PR #79, opened 2013-07-26
Added an argument to addHydrogens() for selecting which Platform to use for minimization. Also marked convertWater() as deprecated.

peastman:


rosemary-cresset:
Hi. I just wanted to make a request for convertWater to be kept in OpenMM rather than deprecated. I have implemented a workflow in my project where the residues that need parameterisation with openmmforcefields are automatically identified with ForceField.getUnmatchedResidues at the beginning. However, if waters are present in the system and we are using a four-point water model I do not want these to be identified, so I first add the dummy atoms to the waters using Modeller.convertWater. I cannot do this with Modeller.addExtraParticles as this throws an error regarding the other unmatched residue templates. Thank you.

peastman:
Could you just remove all the waters from the list returned by `getUnmatchedResidues()`?

```python
unmatched = [res for res in unmatched if res.name != 'HOH']
```

rosemary-cresset:
That solution would probably be fine most of the time, but what I'm trying to do is create a robust workflow that will work for any PDB I throw at it and I want to catch all problems at this point so I can print an intelligible error message for the user rather than a crash in createSystem. Some water molecules might still end up in the list of unmatched residues if they have unusual names. I would also like to check the water molecules themselves with getUnmatchedResidues, so that things like missing H atoms or other problems are caught. Being able to keep using convertWater seems like a much nicer solution.

peastman:
If you're looking for a robust workflow, your current approach won't do it.  Take note of the [warning in the API docs](http://docs.openmm.org/latest/api-python/generated/openmm.app.forcefield.ForceField.html#openmm.app.forcefield.ForceField.getUnmatchedResidues): `getUnmatchedResidues()` is an experimental method.  It isn't robust.

> Some water molecules might still end up in the list of unmatched residues if they have unusual names.

The Topology class requires all residues to follow standard PDB naming conventions.  If a water isn't called HOH, that means your Topology object is invalid.  And take a look at the implementation of `convertWater()`: it identifies water by the residue name.

PR #415, opened 2014-04-22
Added usesPeriodicBoxVectors() methods to Force and System

jchodera:
I've added `usesPeriodicBoxVectors()` methods to both `Force` and `System`.  This method returns `true` if the force (or any force in the system) uses the periodic box vectors in the energy calculation.


swails:
This was implemented by #762 and can be closed.

jchodera:
Thanks, @swails!

Issue #632, opened 2014-10-01
Can I serialize state objects in Python?


kyleabeauchamp:
Not sure if this is unsupported, a bug, or just a silly idea in general:

```
state = simulation.context.getState(getPositions=True, getParameters=True, getEnergy=True)
sstate = cPickle.dumps(state)


---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-eb8b9defca43> in <module>()
----> 1 sstate = cPickle.dumps(state)

/home/kyleb/opt/lib/python2.7/copy_reg.pyc in _reduce_ex(self, proto)
     82             dict = None
     83     else:
---> 84         dict = getstate()
     85     if dict:
     86         return _reconstructor, args, dict

/home/kyleb/opt/lib/python2.7/site-packages/simtk/openmm/openmm.pyc in __getstate__(self)
    953 
    954     def __getstate__(self):
--> 955         serializationString = XmlSerializer.serializeState(self)
    956         return serializationString
```

PR #633, opened 2014-10-01
[WIP] Fixed bug in State serialization in Python where exception would occur.

jchodera:
This addresses #632.  I also added a test to check if serialization is working in Python, which we didn't have before.


jchodera:
This was addressed separately in #632.

Issue #348, opened 2014-03-01
Improve test for anisotropic Monte Carlo barostat


jchodera:
I'll start a PR soon, but this is just an open issue to discuss improvements that can be made to the problematic `Test*MonteCarloAnisotropicBarostat` code.

PR #673, opened 2014-10-26
Eliminated flawed Einstein crystal tests.

jchodera:
This PR eliminates the Einstein crystal tests completely (which had already been disabled some time ago by commenting them out).  These tests were fundamentally flawed in that there is no way to scale the harmonic well centers as the periodic box vectors are scaled.  Issue #348 contains the full discussion.

This issue can be revisited once there is some way to have `CustomExternalForce` energy terms depend on the box vectors, but this capability is not present yet.


jchodera:
Closing this, since I think this was dealt with already.

PR #755, opened 2014-12-16
Correct the mass of hydrogen in AMOEBA2013 force field

leeping:
This reflects a change that Jay made in the TINKER version.


swails:
Bump -- I think this is quite merge-able.  Tests pass and the change is trivial (and obviously correct).

peastman:
We shouldn't just make changes to this file by hand.  It was automatically generated based on a certain set of files from TINKER, which are documented at the top.  If those TINKER files have changed, we should regenerate it properly based on the newest files.

leeping:
Sure, that sounds fine.  I didn't generate these files though, I believe you did it for the version we currently have.  My version of processTinkerForceField.py seems to have issues.

We should regenerate our XML file based on the amoebapro13.prm file contained in the TINKER 7.0.1 distribution.

peastman:
This is fixed by #823.

Issue #180, opened 2013-10-20
System.hasPeriodicBoundaries?


rmcgibbo:
Is there a compelling reason for `System` to not contain a method like `hasPeriodicBoundaries`? I just found myself writing the following code, and wondering if it should be part of the OpenMM API.

```
bool hasPeriodicBoundaries(const OpenMM::System& system) {
  const int numForces = system.getNumForces();
  bool isPeriodic = false;
  for (int i = 0; i < numForces; i++) {
    const OpenMM::NonbondedForce* force  = dynamic_cast<const OpenMM::NonbondedForce *>(&system.getForce(i));
    if (force != NULL) {
      int nbMethod = force->getNonbondedMethod();
      if (nbMethod == OpenMM::NonbondedForce::NoCutoff) {
        isPeriodic = false;
        break;
      } else if (nbMethod == OpenMM::NonbondedForce::CutoffNonPeriodic) {
        isPeriodic = false;
        break;
      } else if (nbMethod == OpenMM::NonbondedForce::CutoffPeriodic) {
        isPeriodic = true;
        break;
      } else if (nbMethod == OpenMM::NonbondedForce::Ewald) {
        isPeriodic = true;
        break;
      } else if (nbMethod == OpenMM::NonbondedForce::PME) {
        isPeriodic = true;
        break;
      }
    }
  }
  return isPeriodic;
}
```

swails:
@rmcgibbo -- this was implemented by #762, so it can be closed.

PR #762, opened 2014-12-30
Implements `usesPeriodicBoundaryConditions` on `Force`s and `System`

swails:
This implements a `usesPeriodicBoundaryConditions` method on all `Force`s as well as `System`.  A few notes:

- The default implementation throws `OpenMMException`
- The `System` returns `true` if at least one force it contains is periodic.  If at least one force is periodic, it will not throw an exception, even if not all forces implement `usesPeriodicBoundaryConditions`.  If no force is periodic and at least one force does not implement `usesPeriodicBoundaryConditions`, an exception is thrown.
- This has been implemented for the main forces, the plugin forces (Drude oscillators and Amoeba plugins), and the thermostat/barostats.
- This interface is tested (only) in the Reference platform, but as it is part of the OpenMM API, that is good enough to ensure coverage.

This addresses #180 and #415, both of which appears to have gone stale.  Since I would like this in the API for my own work, I went ahead and added it.


peastman:
That looks great!

This is totally nitpicking, but could you fix the formatting to match the SimTK style guide?  (I know, I totally need to put together a proper OpenMM style guide.  I'll get to it!)  The two things I notice are that

    bool usesPeriodicBoundaryConditions() const {return false;}

should be

    bool usesPeriodicBoundaryConditions() const {
        return false;
    }

And that

    } else {

should be

    }
    else {

and likewise for some catch blocks.  (Just search for all occurrences of "} " in the diff.)

swails:
Sure thing. I should hopefully get to this a bit later today. Thanks!

swails:
OK -- I think I got all of them in the files that I modified as part of this commit.  Not all of the violations that I fixed were mine (I figured you wouldn't mind a couple extra), but I didn't go through the whole source tree fixing these.

peastman:
Thanks!

Issue #763, opened 2015-01-01
Subtle issue with MonteCarloBarostat random number seed


jchodera:
This may be more of a cautionary tale than an actual issue (unless someone can think of a clever way to address this), but I recently tried to run 100 replicates of an NPT argon fluid using a scheme like this:
```
create System
add barostat to system
for replicate = 1..100
   create Langevin integrator
   create Context
   set initial positions
   simulate
```
This scheme led to an odd synchronization effect in which it appeared that the `MonteCarloBarostat` was using the same sequence of random numbers, inducing synchronization on the density trajectory, similar to what has been previously observed by [Voter](http://dx.doi.org/10.1063/1.1667473) and [Roitberg](http://dx.doi.org/10.1021/ct800573m).

Restructuring this to add a new barostat each replicate eliminated this effect:
```
create System
or replicate = 1..100
   make new deep copy of System
   add barostat to system
   create Langevin integrator
   create Context
   set initial positions
   simulate
```
This is obvious in hindsight, since creating a barostat assigns a new random number seed to it, but I just wonder if there is something we might want to do in order to avoid this kind of (subtle) bug in the future.

rmcgibbo:
Integrators are only permitted to bind to, at maximum, one context. It would be conceivable that a similar restriction could be applied to `Force`s containing prng seeds, but I haven't thought through the implications of this enough to know if it would be desirable.

swails:
> Integrators are only permitted to bind to, at maximum, one context. It would be conceivable that a similar restriction could be applied to Forces containing prng seeds, but I haven't thought through the implications of this enough to know if it would be desirable.

Integrators are fundamentally different here.  The `Integrator` subclass does all of the work for a particular `Context`.  By contrast, `Force`s don't actually _do_ any of their own calculation -- it is the `ForceImpl` objects that get created by the `ContextImpl` that actually do the work and must be uniquely bound to a single `Context`.

A force can create as many `ForceImpl` instances as it wants (and `ForceImpl` cannot modify its parent `Force`, as indicated by the `const Force&` reference in its private variable list).

[...]

Issue #761, opened 2014-12-28
We should report PME grid dimensions for AmoebaMultipoleForce


leeping:
Hi there,

I think it would be very useful if we could query the PME grid dimensions for the `AmoebaMultipoleForce` in a `Context`.  Currently there is no way to get the grid dimensions if the user specifies `ewaldErrorTolerance` - the grid dimensions are calculated automatically when creating the Context, but if I call `simulation.context.getSystem().getForce(force_num).getPmeGridDimensions()` the return value is `(0,0,0)`.

On the CUDA platform, the grid dimensions are automatically set in `AmoebaCudaKernels.cpp` if the user doesn't provide them explicitly.  If the `AmoebaMultipoleForce` argument weren't declared as `const`, then we could call `AmoebaMultipoleForce.setPmeGridDimensions()` and the user could query them from the application layer.  This seems to go against the current design, does anyone have ideas that might work?

Thanks,

- Lee-Ping 

Below is the code snippet for what I'm referring to.

```
CudaCalcAmoebaMultipoleForceKernel::initialize(const System& system, const AmoebaMultipoleForce& force)
        ...
        force.getPmeGridDimensions(pmeGridDimension);
        if (pmeGridDimension[0] == 0 || alpha == 0.0) {
            // LPW This code calculates the grid dimensions automatically
            NonbondedForce nb;
            nb.setEwaldErrorTolerance(force.getEwaldErrorTolerance());
            nb.setCutoffDistance(force.getCutoffDistance());
            NonbondedForceImpl::calcPMEParameters(system, nb, alpha, gridSizeX, gridSizeY, gridSizeZ);
            gridSizeX = findFFTDimension(gridSizeX);
            gridSizeY = findFFTDimension(gridSizeY);
            gridSizeZ = findFFTDimension(gridSizeZ);
            // LPW I wish I could do this
            pmeGridDimension[0] = gridSizeX;
            pmeGridDimension[1] = gridSizeY;
            pmeGridDimension[2] = gridSizeZ;
            force.setPmeGridDimensions(pmeGridDimension);
            // LPW
        } else {
            gridSizeX = pmeGridDimension[0];
            gridSizeY = pmeGridDimension[1];
            gridSizeZ = pmeGridDimension[2];
        }
        defines["EWALD_ALPHA"] = cu.doubleToString(alpha);
        defines["SQRT_PI"] = cu.doubleToString(sqrt(M_PI));
        defines["USE_EWALD"] = "";
[...]

PR #765, opened 2015-01-03
Make it so that random forces and integrators generate unique seeds each time a Context is made

swails:
Any call to "setRandomNumberSeed" prevents this behavior (unless the generator is seeded with 0).

This approach preserves the `const`ness of the `createImpl` methods, but will still generate unique random seeds if numerous contexts are instantiated using the same system by default.  But calls to `setRandomNumberSeed` will preserve the "old" behavior (even though virtually no guarantees are made about random number seeds and their effects on generators by the OpenMM docs).


swails:
This addresses #763

swails:
A few comments about this approach

* If only one `ForceImpl` is created from each `Force`, the behavior is unchanged (i.e., the random number generator is seeded from a call to `osrngseed()`).
* If someone calls `setRandomNumberSeed()` explicitly on any of the stochastic forces, they get the same behavior as they did with previous versions of OpenMM (not that OMM makes any guarantees about the behavior of the generators outside of "different seeds yield different streams", anyway).

It also fixes two instances where users could unwittingly use the same random seed in 2 simulations using the "default" behavior:

* As @jchodera was doing -- if you initialize 2 contexts from the same system, this PR will provide 2 different seeds, while the existing behavior is to assign the same seed to both.
* If you serialize a `System` with a stochastic force, deserializing that XML representation will set the same seed as the original serialized system. With this PR, a different seed will be spit out by `getRandomNumberSeed()` by default.

I actually think this is an worthwhile issue to address explicitly.  I was trained in one of the labs that investigated the phenomena of Langevin synchronization, and luckily that study was being done at the same time one of the other group members unwittingly used the same random seed for numerous simulations in their work.  This led to overly noisy results (since they didn't have nearly as many independent samples as they thought they did), and I think it was probably only caught before publication because someone else in the group was explicitly studying the synchronization artifact.  When the study was redone, the results themselves were not statistically "different", but the result was far less noisy, giving substantially tighter confidence intervals.

I've very often had to think about random generator seeding and stream reproducibility while implementing a variety of methods, but it isn't something I think about when I'm actually running simulations.

The moral there was that using the same seed by accident is historically _very_ easy to do, but detecting that mistake can be very tricky.  I think that anything that can be done to prevent this from occurring _accidentally_ should be done.


What do people think?

peastman:
This seems like a good solution to the problem.  I'd change the implementation, though.  getRandomNumberSeed() should return precisely the value that was set with setRandomNumberSeed().  If it was set to 0, it should return 0.  We can then document that 0 is a special value that means to pick a different seed for every Context.  Instead, we should modify the various platforms to make them interpret 0 differently.  That shouldn't actually be a lot of code change, since most of them just go through a few routines: SimTKOpenMMUtilities::setRandomNumberSeed(), CudaIntegrationUtilities::initRandomNumberGenerator(), etc.

Let's also make the integrators work the same way.  Consistency is good, and the same problem can also come up with them if you serialize an integrator then use that serialized file for multiple simulations.

jchodera:
This sounds like a good solution!

Two additional requests:
* If the user wanted to reuse the same `Context` to run multiple simulations, it would be great to have a public method to cause all integrators/forces with `0` seeds to be reset to some new random numbers.
* It would be great to still be able to query the random seed that was selected.  Perhaps @peastman's suggestion in #761 of adding a `getRandomNumberSeedUsedInContext(context)` method or similar would be a good way to do this.

swails:
@peastman -- I'm a little over my head here.  My understanding of OpenMM internals is definitely improving, and I _think_ I moved this functionality into each platform correctly, but I need smart people to check me here to see that I did it right.

I think the barostats use their own RNGs (instantiating a `OpenMM_SFMT::SFMT` directly rather than using any of the platform-specific generators, which would make sense since all of the decision making moveset generation is done on the CPU and the kernel is just the coordinate scaling).

There seems to be a single (?) RNG stream for each platform that is only (?) utilized by Integrators, and this RNG seeding is done in a platform-specific way.  The barostats get around this single-generator-per-platform limitation by using a CPU (platform-independent) generator based on a separate `SFMT` instance, which doesn't hurt performance since it requires ~0 random numbers.

Last thing -- `Context.h` uses `osrngseed()` in the function prototype for the default seed for random velocity assignment.  I assume that (unlike Python), C++ uses late-binding so that it actually calls `osrngseed()` each time that function is called?  Should I change that to follow the same convention as what I put in everywhere else?

Did I miss anything in my implementation?

swails:
@jchodera -- can we consider your other requests separate from this feature?

jchodera:
> @jchodera -- can we consider your other requests separate from this feature?

Sure.  I'll put them in separate threads.

peastman:
This is the right general idea.  But there are other places where random number generators get initialized.  For example, the CPU platform has its own generator called CpuRandom.  Also, the CUDA and OpenCL implementations of CustomIntegrator provide their own generators for uniform random variables, that they initialize to getRandomNumberSeed()+1 (just to ensure it will be different from the seed used for Gaussian random variables - the exact value doesn't matter).  The best thing to do is search for all calls to methods called getRandomNumberSeed().  That way you can be sure all of them are being handled correctly.

Also, the Doyxgen comments for all the setRandomNumberSeed() methods need to be updated to describe the new behavior.  Just add a second paragraph to each one that reads, "If this is set to 0 (the default), a unique random number seed is automatically generated whenever a Context is created.  This ensure that a different seed is used for every simulation without requiring you to set unique seeds explicitly."

jchodera:
Would it be useful to add a bit more to the manual about this issue and how
the random seeding works?  I'm still happy to do that if @swails (who has
done a fantastic amount of work on this!) prefers this be in a separate PR.

swails:
I'll try to find some time to find the relevant part of the manual to add this to.  In the meantime, @jchodera, you can also submit a PR to the stochastic-forces branch on my repo to have it folded in with this PR if I'm taking too long to get to it.

Or if @peastman wants to merge this, the documentation can be done in a separate PR.  Doesn't matter either way to me.

peastman:
We might as well put everything in one PR, including the documentation.

I noticed you added various references (both in source code comments and the Doxygen comments) to the seed being selected "from the system clock".  That isn't actually correct.  On Mac and Linux it gets read from /dev/urandom, and on Windows it gets generated with CryptGenRandom().  So we should say that a unique seed will be selected, but not say anything about how it gets chosen.  In the current implementation it won't have anything to do with the system clock.

swails:
I was recalling a thread here where we were talking about clock resolution and how the increased resolution would ensure the random seeds would be more likely to be unique.  I think @kyleabeauchamp started the thread, IIRC.  I guess I should've actually looked at `OSRngSeed.h` to see what was being done.

In any case, I've changed the comments and doxygen comments to not reference the source of the unique seed.  I'm not sure where in the manual this should go, though.  It's not actually discussed in the manual as far as I can tell, so there's nothing to _change_.  And I'm not sure where the best place to _add_ it would be.

peastman:
How about in chapter 22?  It's sort of a grab bag of miscellaneous features that only need short discussions.

swails:
OK -- I added a discussion of random numbers and stochastic forces/integrators at the end of Chapter 22 (along with a couple relevant citations).

Comments (and PRs into my branch) welcome.

jchodera:
I really like the discussion you added, @swails!

peastman:
Looks great!  A few minor points.

For class and method names, it's best to use the :class: and :meth: roles.  So instead of \`\`Context\`\`, say :class:\`Context\`.

It's showing the entire bibliography file as being modified.  Did you change the line endings or something like that?

I *think* sphinx requires a blank line before the start of a numbered list.  Did you try generating documentation from this to make sure it came out ok?

swails:
Thanks!

> For class and method names, it's best to use the :class: and :meth: roles. So instead of ``Context``, say :class:`Context`.

Done.

> It's showing the entire bibliography file as being modified. Did you change the line endings or something like that?

I `dos2unix`ed the crap out of that file (the "crap" being the DOS file endings).  In my defense, I did report doing that in the commit log :).

> I think sphinx requires a blank line before the start of a numbered list. Did you try generating documentation from this to make sure it came out ok?

I did.  Turns out I made 2 mistakes.  Sphinx threw a fit about the indentation, but treated it as a newline for some reason.  I changed it to a blank line with no indentation.

If there's anything else this PR needs, let me know.  If not, I say it's ready for merge :).

peastman:
Thanks!  I'll merge it as soon as travis finishes.

PR #774, opened 2015-01-12
Fix typemap for str/bytes -> std::string

rmcgibbo:
In the current code, if you grep the swig-generated `python/src/swig_doxygen/OpenMMSwig.cxx`, you can see that the `%typemap(in) std::string` was actually never being applied at all.


peastman:
The Python tests are segfaulting with this change.

rmcgibbo:
I was having some scoping issues. I think this might be fixed (?). Using this PR, the code for one of the swig wrappers using this typemap is something like (this is the wrapper `Platform::loadPluginsFromDirectory(const std::string&)`)

```
SWIGINTERN PyObject *_wrap_Platform_loadPluginsFromDirectory(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *resultobj = 0;
  std::string *arg1 = 0 ;
  std::string temp21 ;
  PyObject * obj0 = 0 ;
  std::vector< std::string,std::allocator< std::string > > result;

  if (!PyArg_ParseTuple(args,(char *)"O:Platform_loadPluginsFromDirectory",&obj0)) SWIG_fail;
  {
    // if we have a C++ method that takes in a std::string, we're most happy to
    // accept a python bytes object. But if the user passes in a unicode object
    // we'll try to recover by encoding it to UTF-8 bytes
    PyObject* temp = NULL;
    char* c_str = NULL;
    Py_ssize_t len = 0;

    if (PyUnicode_Check(obj0)) {
      temp = PyUnicode_AsUTF8String(obj0);
      if (temp == NULL) {
        SWIG_exception_fail(SWIG_TypeError, "Platform_loadPluginsFromDirectory: 'utf-8' codec can't decode byte");
      }
      PyBytes_AsStringAndSize(temp, &c_str, &len);
      Py_XDECREF(temp);
    } else if (PyBytes_Check(obj0)) {
      PyBytes_AsStringAndSize(obj0, &c_str, &len);
    } else {
      SWIG_exception_fail(SWIG_TypeError, "Platform_loadPluginsFromDirectory: argument must be str or bytes");
    }

    if (c_str == NULL) {
      SWIG_exception_fail(SWIG_TypeError, "Platform_loadPluginsFromDirectory: argument must be str or bytes");
    }

    temp21 = std::string(c_str, len);
    arg1 = &temp21;
  }
  {
    try {
      result = OpenMM::Platform::loadPluginsFromDirectory((std::string const &)*arg1);
    } catch (std::exception &e) {
      PyErr_SetString(PyExc_Exception, const_cast<char*>(e.what()));
      return NULL;
    }
  }
  resultobj = swig::from(static_cast< std::vector<std::string,std::allocator< std::string > > >(result));
  {
    // pass
  }
  return resultobj;
fail:
  {
    // pass
  }
  return NULL;
}

rmcgibbo:
yay.

PR #776, opened 2015-01-12
Fixed bug in reference CustomCompoundBondForce

peastman:
This bug affected CustomCompoundBondForce in the Reference and CPU platforms.  It occurred when the energy expression explicitly depended on the x, y, and z coordinates of an atom.  The coordinates of the wrong atom were being used for those variables.


jchodera:
Many thanks for the descriptive explanation!

PR #777, opened 2015-01-13
Check whether atom number is out of range in 'Add bonds' section

gkiss:
When attempting to load a water-stripped Amber .prmtop file, ''' top.addBond(atoms[bond[0]], atoms[bond[1]])''' aborts with an 'index out of range' error.

This concerns .prmtop files that are edited with the rdparm tool of Amber12 and earlier. I have to check whether this is different in Amber14.

Reason: Stripping of water molecules does not remove the corresponding BondsWithH entries in the .prmtop file. Thus the error when atoms[bond[0]] and atoms[bond[1]] are encountered that belong to the stripped water molecules.

One solution that has worked for me so far is to skip those entries when adding bonds to top.

### DIFF
--- a/wrappers/python/simtk/openmm/app/amberprmtopfile.py
+++ b/wrappers/python/simtk/openmm/app/amberprmtopfile.py
@@ -134,7 +134,8 @@ class AmberPrmtopFile(object):
 
         atoms = list(top.atoms())
         for bond in prmtop.getBondsWithH():
-            top.addBond(atoms[bond[0]], atoms[bond[1]])
+	        if bond[0] < len(list(top.atoms())):
+				top.addBond(atoms[bond[0]], atoms[bond[1]])
         for bond in prmtop.getBondsNoH():
             top.addBond(atoms[bond[0]], atoms[bond[1]])
 


swails:
I would suggest against this.  `rdparm` does not correctly strip Amber topology files, so I wouldn't trust any prmtop that comes from it when you try.  IMO, we _want_ these prmtop files to crash (in fact, rdparm is not even built as part of Amber anymore).  There are other ways to do this correctly.

I suggest looking into ParmEd (http://github.com/swails/ParmEd) -- it may have some functionality that you are looking for (and it is tightly integrated with OpenMM already, capable of instantiating `System` objects from standard Amber prmtops as well as chamber topologies and old-style topologies as well).

gkiss:
Good to know.
Thanks Jason!

wrappers/python/simtk/openmm/app/amberprmtopfile.py
```diff
@@ -134,7 +134,8 @@ def __init__(self, file):
 
         atoms = list(top.atoms())
         for bond in prmtop.getBondsWithH():
-            top.addBond(atoms[bond[0]], atoms[bond[1]])
+	        if bond[0] < len(list(top.atoms())):
+				top.addBond(atoms[bond[0]], atoms[bond[1]])
```
swails: Just a few comments -- the indentation here looks wrong (maybe a mix of tabs and spaces?).  This will break with Python 3, since indentation within a file must be consistent. (It may not work with Python2 either, I'm not sure)

Also, for every bond you loop through, you construct a list from `top.atoms()` and then immediately throw it away.  This is very inefficient.  It would be better to store `len(list(top.atoms))` before the loop and used that cached value.

swails:
I think this can be closed.

PR #780, opened 2015-01-14
Some windows changes

rmcgibbo:
@peastman: does this look the right idea for setting the PATH and plugin dir based on strings in `version.py`? I'm working on another patch for the conda installer that uses the `post-link.(sh|bat)` to modify the `version.py` file inside the package in `conda install` (http://conda.pydata.org/docs/build.html#pre-post-link-unlink-scripts)

### DIFF
--- a/wrappers/python/simtk/openmm/__init__.py
+++ b/wrappers/python/simtk/openmm/__init__.py
@@ -10,11 +10,22 @@ It also tries to load any plugin modules it can find.
 __author__ = "Randall J. Radmer"
 
 import os, os.path
+import sys
+from simtk.openmm import version
+
+if sys.platform == 'win32':
+    _path = os.environ['PATH']
+    os.environ['PATH'] = '%(lib)s;%(lib)s\plugins' % {'lib': version.openmm_library_path}
+
 from simtk.openmm.openmm import *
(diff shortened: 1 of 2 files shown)


peastman:
Does it work?

I think you want to *add* those directories to the path rather than replacing the whole path.  Otherwise, it could fail to find other libraries the plugins need to link to (link CUDA).

rmcgibbo:
> Does it work?

Yes, empirically. 

> I think you want to add those directories to the path rather than replacing the whole path. Otherwise, it could fail to find other libraries the plugins need to link to (link CUDA).

Yeah, that makes sense.

rmcgibbo:
The one wart related to getting this into the cond package is that we either have to do a 6.2.1 "release" that incorporates this PR, or I have to add this PR as a `.patch` file to the `conda-recipes`. Neither of these are optimal.

peastman:
For the moment, let's do it as a patch since that's a lot easier.  There's only one file that actually needs changes.  The typemap wasn't being used, so leaving it in is harmless.

rmcgibbo:
Right, I won't add the patch for the the typemap to the conda recipe, but I figured I should add it here to the PR, since I think the current `typemap.i` that I made in #774 probably isn't right.

PR #783, opened 2015-01-15
Catch nan coordinates before they can produce a segfault

peastman:


platforms/cpu/src/CpuKernels.cpp
```diff
@@ -166,6 +166,12 @@ class CpuCalcForcesAndEnergyKernel::InitForceTask : public ThreadPool::Task {
                 posq[4*i+1] = (float) posData[i][1];
                 posq[4*i+2] = (float) posData[i][2];
             }
+        
+        // Check for invalid positions.
+        
+        for (int i = 4*start; i < 4*end; i++)
+            if (posq[i] != posq[i])
```
swails: Are you sure this won't get optimized away by the compiler?  And what about `inf`?  Does that not cause a segfault?

(I'm just curious here)

peastman: It won't get optimized away.  This is the standard way to check for a nan.  I'm not sure about inf, but in my testing so far just checking for nan has always been sufficient.

rmcgibbo: FWIW, without c99 or c++11, here are the macro-hoops that numpy goes through to define a cross-platform isnan: https://github.com/numpy/numpy/blob/master/numpy/core/include/numpy/npy_math.h#L160-L203

rmcgibbo: With gcc, it's only potentially optimized away if you compile with `-ffast-math`, which is not enabled by any of the `-O` options.

peastman: -ffast-math implies -ffinite-math-only, which tells the compiler it can assume nothing will ever be nan or infinite.  So yes, that would break this!

swails: So let's _don't_ do that? :)

jchodera:
Works like a champ!

For the `LennardJonesGrid` with large initial forces:
```
[LSKI1497:~/code/omnia/openmmtools.jchodera] choderaj% python segfault-test-case.py
initial               : potential = 136626243.414 kcal/mol | force norm = 349722355.930 kcal/mol/nm
iteration     0 /   100 : potential = 94331915978.494 kcal/mol
iteration     1 /   100 : potential = 1045527407773.341 kcal/mol
iteration     2 /   100 : potential = 1262070041743.855 kcal/mol
  File "segfault-test-case.py", line 33, in <module>
    integrator.step(nsteps)
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 11876, in step
    return _openmm.LangevinIntegrator_step(self, *args)
Exception: Particle coordinate is nan
```
For the alanine dipeptide with large timestep:
```
[LSKI1497:~/code/omnia/openmmtools.jchodera] choderaj% python segfault-test-case.py
initial               : potential = -5892.713 kcal/mol | force norm = 116.881 kcal/mol/nm
  File "segfault-test-case.py", line 33, in <module>
    integrator.step(nsteps)
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 11876, in step
    return _openmm.LangevinIntegrator_step(self, *args)
Exception: Particle coordinate is nan
```
Thanks!

PR #789, opened 2015-01-19
Fixgb

swails:
This fixes the Amber custom GB models when you specify a salt concentration _with_ a cutoff.


jchodera:
Is it problematic that OpenMM will also apply the reaction-field correction to the normal Coulomb terms of `NonbondedForce` when using `CutoffNonPeriodic`?

swails:
> Is it problematic that OpenMM will also apply the reaction-field correction to the normal Coulomb terms of NonbondedForce when using CutoffNonPeriodic?

If it _does_ apply the reaction-field correction to the normal Coulomb terms, then yes, it would be problematic.  But it doesn't do that when you use the Amber file parsers.  See https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/amber_file_parser.py#L1077

Note that the problem only occurs if you specify a salt concentration > 0.  Cutoff GB already worked in the absence of salt. (Here "worked" == "gave same answers as Amber".)

Looks like I still have to fix the CHARMM tests, though.

jchodera:
> If it does apply the reaction-field correction to the normal Coulomb terms, then yes, it would be problematic. But it doesn't do that when you use the Amber file parsers.

Ah!  Thanks for the reminder!

peastman:
What is it that changed here?  Where did the old values in the test cases come from?

swails:
> What is it that changed here?

In the cutoff section, `-kappa/solventDielectric` --> `-exp(-kappa*f)/solventDielectric` (have a look at the diff).  The `-exp(-kappa*f)` is the correct term, and is present in the non-cutoff version of the code.  I never run GB with a cutoff, so I must have missed the GB-with-salt-and-cutoff case when I was working on that code.

> Where did the old values in the test cases come from?

Those came from you.  You added them when you optimized the Custom GB models in the Python app layer. (Commit 92a4bbe1e49bdd353ae6f9339ad9aed22a2688b9).

swails:
I only found this because I was trying to duplicate `customgbforces.py` in C++, and all my comparisons matched until I hit GB with a cutoff and salt.

peastman:
Ok.  Could you enter this in the bug tracker so we can include it in the list of fixed bugs for the next release?

swails:
Sure.

swails:
Done in 2069

PR #795, opened 2015-02-02
Add error checking for the input GB radii in the GBn methods.

swails:
The radii must be between 1 and 2 Angstroms, or they fall outside the bounds of the neck lookup
tables.

At least on my Mac, this results in a C++ exception that does not get translated
into a Python exception (i.e., it is not catchable and crashes the application)


peastman:
Looks like some test cases are failing.

swails:
> Looks like some test cases are failing.

I used the wrong index.  I should have applied the WIP tag, apologies.

peastman:
Ok, let us know when it's ready.

swails:
I'm not declaring this 'ready' yet (but I doubt I will need any more changes).  The failure was a stochastic test that succeeded upon re-test.

swails:
OK -- this seems to work correctly.  Good to go.

PR #797, opened 2015-02-06
C++ libraries support triclinic boxes

peastman:
This is the first part of implementing triclinic boxes.  All the C++ libraries (the core library and plugins) now support them.  I haven't yet modified any of the Python classes to take advantage of them.  That will come in a separate PR.

Hopefully I haven't created too many bugs in the process, but there's a ton of new and modified code, so keep an eye out for anything that doesn't look right.  This has really been a slog.


kyleabeauchamp:
Sweet.  I know it's probably too early for benchmarking, but this should have a negligible effect on performance when the number of atoms is kept constant, right?

rmcgibbo:
WOOT! Nice work peter.

jchodera:
Awesome!

swails:
What they said!

peastman:
I haven't benchmarked any systems with triclinic boxes yet.  There will be some overhead, since it does take more flops to apply periodic boundary conditions, but I expect it to be small.  We're able to skip doing it for the majority of interactions.

jlmaccal:
@peastman Is there a standard way to deal with periodic boxes in both Cutstom*Forces and in custom cuda code?

peastman:
That depends on exactly what you mean by "deal with periodic boxes".  But to some extent, yes.  CudaContext defines a set of macros for applying periodic boundary conditions to positions and displacements.  They're defined in CudaContext.cpp starting at line 243.  Take a look at nonbonded.cu for some examples of how they're used.

PR #799, opened 2015-02-06
Fixes vs2013 warnings 4251, 4267 in lepton.

chrisdembia:
The pragma to ignore 4251 already existed in windowsIncludes.h, and it works
when compiling lepton itself. However, when clients use lepton, they get this
warning. I think this is why @sherm1 put the define as shown here:
https://github.com/simbody/simbody/blob/master/SimTKcommon/include/SimTKcommon/internal/common.h#L168
 (whether or not the shared library is being built or used).


chrisdembia:
Thank you!

Issue #800, opened 2015-02-07
AMOEBA stretch-bend: OpenMM does not handle the "general case"


jayponder:
I think I’ve discovered an issue in the OpemMM implementation of AMOEBA’s stretch-bend term. A typical parameter, taken from one of the TINKER parameter files, would be:

strbnd        2    1    3       7.20       7.20

This is a stretch-bend term for the angle consisting of atom classes 2, 1 and 3, connected as the angle 2-1-3. The stretch-bend energy for this angle is a sum of two terms. The first is a force constant
times the 2-1 bond length deviation times the 2-1-3 angle deviation. The second term is a force constant times the 1-3 bond deviation times the 2-1-3 angle deviation (where “deviation” means difference from the ideal/target value for the bond length or angle).

In the above parameter, the first 7.20 is the force constant for the first term, i.e., the one containing the 2-1 bond deviation. The second 7.20 is the force constant for the term involving the 1-3 bond length.
In many case, the two force constants are often the same (and in the AMOEBA Bio09 and Pro13 parameter sets, they are always the same). However, there are many other molecules where the two force constants in a single stretch-bend parameter are different. OpenMM appears to not deal with the case where the two force constants are, in fact, different. It seems to use the first force canstant value (I think..) for both of the terms arising from a single stretch-bend occurrence.

In future AMOEBA parameterizations, we envision that the two force constants in each stretch-bend will be allowed to move to different values. So while this is not significant for users of the current AMOEBA biomolecule parameters, it is for users of the present general organic parameters (AMOEBA09) and will be a frequent/general issue going forward.

jchodera:
From what I can tell, it does appear that the current API in fact only allows *one* force constant to be specified for each stretch-bend term:
https://github.com/pandegroup/openmm/blob/master/plugins/amoeba/openmmapi/include/openmm/AmoebaStretchBendForce.h#L65-L78
```C++
    /**
     * Add a stretch-bend term to the force field.
     *
     * @param particle1     the index of the first particle connected by the stretch-bend
     * @param particle2     the index of the second particle connected by the stretch-bend
     * @param particle3     the index of the third particle connected by the stretch-bend
     * @param lengthAB      the equilibrium length of the stretch-bend in bond ab [particle1, particle2], measured in nm
     * @param lengthCB      the equilibrium length of the stretch-bend in bond cb [particle3, particle2], measured in nm
     * @param angle         the equilibrium angle in radians
     * @param k             the force constant for the stretch-bend
[...]

PR #801, opened 2015-02-08
Generalize Amoeba Stretch-Bend term

swails:
This should address #800.  All of the tests have been updated and continue to pass (including the CUDA test on my machine).  Also, the *old* versions of the test (without the updated energy expression using separate force constants) passed as well, showing that the energies and forces have not *changed* as a result of this PR.

A few notes:

- I tried to maintain as much backwards-compatibility as possible. So the `addStretchBend` and `setStretchBendParameters` functions were supplied with default values for k2 of -1 that results in k1 being used for k2.  However, `getStretchBendParameters` doesn't work this way, and overloading is not an option thanks to the C wrapper generator. This could be worked around if maintaining this backwards-compatibility is important, but it would be clumsy and I would vote against it.
- I have not added any tests with different k's.
- I didn't touch the Python layer at all.  Hopefully, the default argument should keep the app layer working the same way it always has, but that's not fully tested.  This will also necessitate changes to the tinker-to-openmm XML force field writing tool in the pyopenmm repository.  This might be a good time to move it over here, as @jchodera has frequently pushed for.  But since it's *not* here, I'm leaving this be.

I'd like a code review/discussion before this is merged, though.  And if @jayponder has a test case he can either provide or run this PR through, that would be great as well (although it was pretty easy to figure out what changes to make simply by comparing the OMM and Tinker codes...)

### DIFF
--- a/plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
+++ b/plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
@@ -41,7 +41,6 @@ AmoebaStretchBendForce::AmoebaStretchBendForce() {
 
 int AmoebaStretchBendForce::addStretchBend(int particle1, int particle2, int particle3,
                                            double lengthAB,  double lengthCB, double angle, double k1, double k2) {
-    if (k2 == -1.0) k2 = k1;
     stretchBends.push_back(StretchBendInfo(particle1, particle2, particle3, lengthAB,  lengthCB, angle, k1, k2));
     return stretchBends.size()-1;
 }
@@ -67,10 +66,7 @@ void AmoebaStretchBendForce::setStretchBendParameters(int index, int particle1, 
     stretchBends[index].lengthCB   = lengthCB;
     stretchBends[index].angle      = angle;
     stretchBends[index].k1         = k1;
-    if (k2 == -1.0)
-        stretchBends[index].k2     = k1;
-    else
-        stretchBends[index].k2     = k2;
+    stretchBends[index].k2         = k2;
 }
 
 ForceImpl* AmoebaStretchBendForce::createImpl() const {
--- a/plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
+++ b/plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
@@ -41,7 +41,7 @@ AmoebaStretchBendForce::AmoebaStretchBendForce() {
 
 int AmoebaStretchBendForce::addStretchBend(int particle1, int particle2, int particle3,
                                            double lengthAB,  double lengthCB, double angle, double k1, double k2) {
-    if (k2 == -1.0) k2 = k1
+    if (k2 == -1.0) k2 = k1;
     stretchBends.push_back(StretchBendInfo(particle1, particle2, particle3, lengthAB,  lengthCB, angle, k1, k2));
     return stretchBends.size()-1;
 }
--- a/plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
+++ b/plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
@@ -40,31 +40,37 @@ AmoebaStretchBendForce::AmoebaStretchBendForce() {
 }
 
 int AmoebaStretchBendForce::addStretchBend(int particle1, int particle2, int particle3,
-                                           double lengthAB,  double lengthCB, double angle, double k) {
-    stretchBends.push_back(StretchBendInfo(particle1, particle2, particle3, lengthAB,  lengthCB, angle, k));
+                                           double lengthAB,  double lengthCB, double angle, double k1, double k2) {
+    if (k2 == -1.0) k2 = k1
+    stretchBends.push_back(StretchBendInfo(particle1, particle2, particle3, lengthAB,  lengthCB, angle, k1, k2));
     return stretchBends.size()-1;
 }
 
 void AmoebaStretchBendForce::getStretchBendParameters(int index, int& particle1, int& particle2, int& particle3,
-                                                      double& lengthAB, double& lengthCB, double& angle, double& k ) const {
+                                                      double& lengthAB, double& lengthCB, double& angle, double& k1, double& k2 ) const {
     particle1       = stretchBends[index].particle1;
     particle2       = stretchBends[index].particle2;
     particle3       = stretchBends[index].particle3;
     lengthAB        = stretchBends[index].lengthAB;
     lengthCB        = stretchBends[index].lengthCB;
     angle           = stretchBends[index].angle;
-    k               = stretchBends[index].k;
+    k1              = stretchBends[index].k1;
+    k2              = stretchBends[index].k2;
 }
 
-void AmoebaStretchBendForce::setStretchBendParameters(int index, int particle1, int particle2, int particle3, 
-                                                      double lengthAB,  double lengthCB, double angle, double k) {
+void AmoebaStretchBendForce::setStretchBendParameters(int index, int particle1, int particle2, int particle3,
+                                                      double lengthAB,  double lengthCB, double angle, double k1, double k2) {
     stretchBends[index].particle1  = particle1;
     stretchBends[index].particle2  = particle2;
     stretchBends[index].particle3  = particle3;
     stretchBends[index].lengthAB   = lengthAB;
     stretchBends[index].lengthCB   = lengthCB;
     stretchBends[index].angle      = angle;
-    stretchBends[index].k          = k;
+    stretchBends[index].k1         = k1;
+    if (k2 == -1.0)
+        stretchBends[index].k2     = k1;
+    else
+        stretchBends[index].k2     = k2;
 }
 
 ForceImpl* AmoebaStretchBendForce::createImpl() const {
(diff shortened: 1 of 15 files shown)


plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
```diff
@@ -40,31 +40,37 @@ AmoebaStretchBendForce::AmoebaStretchBendForce() {
 }
 
 int AmoebaStretchBendForce::addStretchBend(int particle1, int particle2, int particle3,
-                                           double lengthAB,  double lengthCB, double angle, double k) {
-    stretchBends.push_back(StretchBendInfo(particle1, particle2, particle3, lengthAB,  lengthCB, angle, k));
+                                           double lengthAB,  double lengthCB, double angle, double k1, double k2) {
+    if (k2 == -1.0) k2 = k1;
```
mpharrigan: isn't it dangerous to do floating point equality testing?

rmcgibbo:
> I tried to maintain as much backwards-compatibility as possible. So the addStretchBend and setStretchBendParameters functions were supplied with default values for k2 of -1 that results in k1 being used for k2.

I'm not sure if this is a good idea. I think it would be more clear without a default value, and don't see backwards compatibility as being a real value here. We're already changing the signature, so no reason to do it 'halfway'.

plugins/amoeba/openmmapi/src/AmoebaStretchBendForce.cpp
swails: Sometimes, but not here. (Not if it's an integer, for one thing).

swails:
> I'm not sure if this is a good idea. I think it would be more clear without a default value, and don't see backwards compatibility as being a real value here. We're already changing the signature, so no reason to do it 'halfway'.

Except that I don't know if the Python app layer will work without it.  If it *does* work without the default arguments, then I agree with you (the only places I think using the Amoeba plugin is the app layer and the Tinker/OpenMM plugin, which was hopelessly broken for the latest tinker version, anyway).

rmcgibbo:
If the app layer needs to be changed, then it needs to be changed. No sweat. This is why we don't release the app layer and libopenmm seperately.

swails:
> If the app layer needs to be changed, then it needs to be changed. No sweat. This is why we don't release the app layer and libopenmm seperately.

But there's still a nontrivial amount of work needed to get the app layer *working* (for instance, all of the XML files need to be regenerated after updating the pyopenmm script).  So the process I envision here is (assuming the default arg is needed for the Python app layer), is merge this so it's present and available, then delete the default arg in the PR that updates the app layer.

peastman:
I usually try to maintain API compatibility, but in this case I agree it isn't that important.  There just isn't much code out there using this.

There's no need to regenerate XML files though.  This can all be handled by ForceField.  It will allow the XML file to contain either one or two force constants, then invoke addStretchBend() correctly either way.  We will need to update the script, though, to correctly handle new force fields in the future that use separate force constants.  That script is already in this repository, under devtools/forcefield-scripts.

swails:
> There's no need to regenerate XML files though. This can all be handled by ForceField. It will allow the XML file to contain either one or two force constants, then invoke addStretchBend() correctly either way.

Ah.  If it's that simple, I'll update the PR.

> We will need to update the script, though, to correctly handle new force fields in the future that use separate force constants. That script is already in this repository, under devtools/forcefield-scripts.

I've actually looked at this script before (in the pyopenmm repo), and the change *should* be easy to make.  You want me to try and change that here, or do it in another PR (probably closer to when you'd actually *use* it again)?

swails:
> We will need to update the script, though, to correctly handle new force fields in the future that use separate force constants. That script is already in this repository, under devtools/forcefield-scripts.

Actually I don't think we will.  `ForceField` already assigns `k1` and `k2` to the `AmoebaStretchBendGenerator`... it just uses only `k1` when buliding `AmoebaStretchBendForce`.  So I don't think any additional changes are need in the app layer aside from the extra 10 characters I had to add... sweet.

swails:
OK -- that kludge is gone.  I'm running all of the tests on my machine so that it also gets the CUDA platform.  I'll report back tomorrow with the results (the StretchBend CUDA test passes, but I'd still like to run them all), and if people are satisfied it can be merged.

peastman:
> You want me to try and change that here, or do it in another PR

Feel free if you want to.  It's all really part of the same feature.

jayponder:
There is no need to worry about "backward compatibility" here. TINKER has always (at least for many years) has two force constants associated with each stretch-bend parameter. The TINKER code is not even intended to handle a stretch-bend parameter with only one constant. I'm guessing that someone simply noticed that the AMOEBA protein parameters (2009 and 2013) always have the same value for both force constants, and took a bit of a shortcut. But it's not a "new feature" of AMOEBA for which you want "backward compatibility". As far as I can tell, the current OpenMM XML parameter files translated from TINKER actually have both force constant values present. I'd go so far as to suggest that if someone tries to use a stretch-bend parameter with only one force constant, OpenMM should throw an error and quit... 

I do have a test case (actually lots of them..) and could compare OpenMM directly against TINKER. If someone can suggest the easiest way to apply Jason's patches to my stock OpenMM 6.2.0 downloaded from Stanford, I'm happy to verify the changes.

I'm not sure I followed all of the discussion above, but hopefully these changes propagate to the AMOEBA plugin, so I can make the corresponding changes to the TINKER-OpenMM plugin code. Jason suggests above the TINKER interface is broken, but that's no longer the case :)

swails:
The failure in Travis is the Drude Langevin integrator, which passed when re-run.  On my desktop, `TestOpenCLAndersenThermostatSingle` and `TestOpenCLCustomIntegratorMixed` both failed (but only those failed).  When I reran those two by hand, they each passed.

jchodera:
>  If someone can suggest the easiest way to apply Jason's patches to my stock OpenMM 6.2.0 downloaded from Stanford, I'm happy to verify the changes.

@jayponder : If you want to try the version of the OpenMM current git head that @swails has modified for this pull request, you can grab that using:
```
git clone -b strbnd https://github.com/swails/openmm.git
```
If you're asking how to derive a patch from this pull request, that may be a bit more tricky since it's made against the git head, not the OpenMM 6.2.0 release. The changes in the pull request are available as a patch against the git head through this URL:
https://github.com/pandegroup/openmm/pull/801.patch

That *may* work when applied to your modified 6.2.0 version, but it may be easier to do a `git clone` as above and then compile your plugin separately.

swails:
> There is no need to worry about "backward compatibility" here. TINKER has always (at least for many years) has two force constants associated with each stretch-bend parameter.

I'm convinced :).  I deleted this attempt last night so that setting parameters or adding terms requires 2 force constants in the function call sig.

> The TINKER code is not even intended to handle a stretch-bend parameter with only one constant. I'm guessing that someone simply noticed that the AMOEBA protein parameters (2009 and 2013) always have the same value for both force constants, and took a bit of a shortcut.

They are not the only ones to notice that and take the shortcut :).  The Amber implementation of Amoeba actually does the same thing, as far as I can tell :).  The fix there is a little bit more involved (since `tinker_to_amber` has to be modified as well).

> I do have a test case (actually lots of them..) and could compare OpenMM directly against TINKER. If someone can suggest the easiest way to apply Jason's patches to my stock OpenMM 6.2.0 downloaded from Stanford, I'm happy to verify the changes.

Ah, that would be great!  The *easiest* way would be to clone the OpenMM github repo and pull my PR into a new branch in that repository:

```
git clone https://github.com/pandegroup/openmm.git
git branch strbnd
git checkout strbnd
git pull https://github.com/swails/openmm.git strbnd
```

and then compile that.

If your stock OpenMM 6.2.0 was downloaded from source code, you could also try to apply the patch I created (which are the changes that would be applied to master if this PR was merged), which is available in a gist: https://gist.github.com/swails/61b6f4f1a9058b5ae049.  This will only work if that part of the OMM code hasn't changed since 6.2.0 was released (there is a decent chance that's true, I think).

So download that patch, go to your OpenMM source repository, and run the command:

```bash
patch -p0 -N -n < strbnd.patch
# The -n flag above means "dry run", don't make changes.
# If that command works, then it's safe to run patch for real
patch -p0 -N < strbnd.patch
```

> I'm not sure I followed all of the discussion above, but hopefully these changes propagate to the AMOEBA plugin, so I can make the corresponding changes to the TINKER-OpenMM plugin code. Jason suggests above the TINKER interface is broken, but that's no longer the case :)

Great news!  Where does this new interface live?  I recall toward the end of last summer trying to compile the OpenMM-Tinker plugin with Tinker 7 (? I think? the one where common blocks were replaced with modules), and it was a bit of a bear.  I got it to compile, but it would immediately segfault, so I gave up :).

Thanks!

swails:
Ah!  I didn't see @jchodera's post.  That is certainly one way to do it and probably the easiest (I always pull into a new branch on my repo so I don't have to clone the full history each time).

> The changes in the pull request are available as a patch against the git head through this URL:
> https://github.com/pandegroup/openmm/pull/801.patch

Thanks for this tip (I see it as a ProTip at the bottom here).  This patch is still a git-specific patch, since it has the `a/` and `b/` prefixes at the beginnings of all filenames.  The patch I linked does not, which means that `patch -p0 -N` will not work for the automatic github patch file...

jchodera:
Ah, thanks for catching this, @swails!

peastman:
This looks great.  I'll go ahead and merge it.

PR #803, opened 2015-02-10
Python layer supports triclinic boxes

peastman:
This includes everything except the changes to the CHARMM and Amber readers, which I've left for @swails to do.  The changes to those should be pretty minor.  Creating a System is exactly the same as before, except when you call setDefaultPeriodicBoxVectors(), they don't have to be orthogonal.  When creating a Topology, call setPeriodicBoxVectors() instead of setUnitCellDimensions().  To convert between box vectors and lengths and angles, use the functions in simtk.openmm.app.internal.unitcell.


swails:
> To convert between box vectors and lengths and angles, use the functions in simtk.openmm.app.internal.unitcell.

It would be helpful to have a function like this in the C++ layer as well.  Or maybe have it so that OpenMM takes any *valid* periodic box vectors and then internally transforms them into the reduced form it requires?

jchodera:
> Or maybe have it so that OpenMM takes any valid periodic box vectors and then internally transforms them into the reduced form it requires?

+1 for that!

peastman:
> It would be helpful to have a function like this in the C++ layer as well.

That seems like the sort of thing that belongs more in the Python layer.  The C++ layer is strictly about computation, and having only a single representation of periodic boxes is a lot simpler.  The only reason we need to deal with the length/angles representation at all is that some file formats use it.  Which puts it squarely in the domain of the Python layer.

swails:
> The C++ layer is strictly about computation, and having only a single representation of periodic boxes is a lot simpler.

I wasn't suggesting having multiple representations -- in fact quite the opposite.  IMO, it would be nice if OpenMM had a `UnitCell` class of some sort that would allow C++ coders to define a box however *they* wanted to (with 3 vectors or vector angles and lengths), and that class would convert to the internal representation that OpenMM requires.

> The only reason we need to deal with the length/angles representation at all is that some file formats use it.

I don't think I fully agree here.  IMO, we need to deal with them because there are many ways to represent unit cell vectors "correctly", and different people use different representations for different reasons/applications.  I still think that C++ users will have a similar requirement to Python users in terms of wrangling unit cell vector representations. (I know I am in one of my C++/OpenMM side projects, but I'm using the C++ layer there in much the same way I would the application layer).

> The C++ layer is strictly about computation

There are still convenience functions, though.  Like ``NonbondedForce::createExceptionsFromBonds``.

In my own experience, it would be helpful to have access to a unit cell wrangling class that would take the effort out of making sure I have the appropriate cell representation for OpenMM.  However, the code exists in the Python layer and isn't terribly difficult to duplicate in C++, so it's easy for me to work around.  I just envision many people needing to write the code I'm currently writing to do it in their projects...

peastman:
Ok, I think I see what you mean.

swails:
I wasn't trying to heap work on here.  I'd be happy to work on a POP PR, but I didn't want to put time into it if it didn't have a chance of making it in.

I have some existing code ([header here](https://github.com/swails/omm_cphmd/blob/master/include/amber/unitcell.h), [source here](https://github.com/swails/omm_cphmd/blob/master/src/unitcell.cpp)) that I could migrate to the OMM API (which wouldn't need to be Python-exposed, since you have those Python utils already).

If you want me to go ahead with a PR, let me know.  If you'd rather do it, I'd certainly be fine with that as well :).

peastman:
Let me think about this a little first.  I'm worried it could easily produce a really convoluted API if not done right.

swails:
My basic idea was something like this (I've thought about this a bit before, I'm not just pulling it from the spur-of-the-moment):

```C++

class UnitCell {

public:
    UnitCell(double a, double b, double c, double alpha, double beta, double gamma);
    UnitCell(OpenMM::Vec3 a, OpenMM::Vec3 b, OpenMM::Vec3 c);

    OpenMM::Vec3 getReducedVectorA();
    OpenMM::Vec3 getReducedVectorB();
    OpenMM::Vec3 getReducedVectorC();

private:
    /* implementation details */
};
```

The two constructors will take either box dimensions or "non-reduced" vectors and you could access the reduced vectors through getters.  If there were more ways to represent the unit cell, additional constructors could be provided for them as well.

peastman:
Right, but then what do you do with that object?  What's the rest of the API?  What are the changes to System, State, Context, etc.?

The "reduced vector" terminology is likely to confuse people.  There are infinitely many sets of vectors that all describe the same periodic system, and OpenMM requires you to specify a particular one.  But it's the one that almost everyone would choose as the "obvious" set of vectors, not even thinking about the fact that they could also define it in other ways.  It's also the same definition that Gromacs requires, and I don't think the concept of lattice reduction is ever mentioned in their documentation.

swails:
> Right, but then what do you do with that object?

The way I'm using my equivalent of that class in my project is by embedding it in my `AmberParm` C++ class (and my coordinate readers) and keeping an instance of `UnitCell` as private class member.  Then in the `OpenMM::System AmberParm::createSystem` call, that private variable is used to spit out the 3 vectors that I then pass to `system->setDefaultPeriodicBoxVectors()`.  So I would say it's a utility class that people are free to use if they want to.

I'm sympathetic to arguments in favor of keeping the class list lean and limited to the necessary.  I only brought this up because a *lot* of places specify only PBC dimensions (PDB files, Amber, CHARMM, from the looks of their tutorial files gromacs `.gro` files, ...), so I suspect a lot of people will need to be writing this code to make use of OpenMM in C++. (And I had a personal need for it that I'm currently satisfying with my own code.)

jchodera:
+1 for having these capabilities in the C++ layer.

peastman:
Ok, so System, State, etc. would be unchanged?  This would just be a utility class for doing conversions, similar to what's in unitcell.py?

swails:
> Ok, so System, State, etc. would be unchanged? This would just be a utility class for doing conversions, similar to what's in unitcell.py?

That's what I would do, yea.

peastman:
Ok, that sounds reasonable.

swails:
Do you want to handle this so you can control the API?

peastman:
Sure.  It might take a little while before I get to it though.  Could you open an issue for it?  If the only discussion is on a closed PR, it could easily get forgotten.

swails:
Done in #805.

Thanks!

PR #804, opened 2015-02-10
Replaced TinyXML with irrXML

peastman:
This should use a lot less memory when reading XML files.


mpharrigan:
wooooo

mpharrigan:
maybe don't include the irrxml example files?

peastman:
For 3rd party libraries, I usually include their whole distribution, including READMEs, examples, etc.  It's a pretty tiny library.  Even with all of that, it's well under 200kb of files.

peastman:
I missed checking in one file, which is why the build failed.  It's rerunning now.

PR #807, opened 2015-02-13
Support CUDA runtime compilation

peastman:
CUDA 7.0 introduces a runtime kernel compiler, which eliminates the need to have a separate C compiler installed (very important for Folding@Home users, especially on Windows).  This PR adds support for it.

The relevant code is in a plugin.  That way, if you have an older version of CUDA, it just won't load that plugin and will continue to work as before.  A few years from now once we're no longer concerned with supporting older CUDA versions, we can eliminate the plugin and just do it directly from CudaContext.


jchodera:
Awesome!

jlmaccal:
> Awesome!

+1

Issue #816, opened 2015-02-18
Have travis build openmm-dev conda package and push to binstar on successful commits


jchodera:
This is needed because of problems like https://github.com/pandegroup/pdbfixer/pull/73

jchodera:
We have a workaround right now.

PR #817, opened 2015-02-18
Have travis build and push openmm-dev conda package after success

jchodera:
Attempting to address #816.

Since people were unhappy with the old approach of building the conda package, installing it, then testing it, we now try to build the `openmm-dev` conda package and push to binstar only after travis success.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,10 @@ script:
 after_success:
   # Get libraries necessary for building docs and pushing
   # them to S3
-  - sudo apt-get install python-sphinx python-yaml
-  - sudo pip install sphinxcontrib-bibtex boto
-  - make DoxygenApiDocs
-  - make sphinxhtml
+  - sudo apt-get install -qq python-sphinx python-yaml
+  - sudo pip install --quiet sphinxcontrib-bibtex boto
+  - make --quiet DoxygenApiDocs
+  - make --quiet sphinxhtml
   - python devtools/ci/push-docs-to-s3.py
   # Build conda dev package and push to binstar
   - source devtools/ci/install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ env:
     # encrypted AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY to push docs to s3
     - secure: "VVKz+yOMbKsskR+PfU1HfKBWdGYYrmIXNWQz4nqXCjtg2MRCQmjDulFZaPVDvsBzis9BUhnzAQrBYUrAtN8bZSTYRg7ADFVGdPFicg3Sv0owcghTQwokIvbw3G+HDz/WAnFmqEhqm3t5pNVWNinyHpMM3zYZOVKagyj53cwAM0M="
     - secure: "W2iPU6ooMujfzJNw9ElaEB8Go1rlNFJ5zEldr3FaH7SDRwqtqNOEp9CegCeG/hHtjg1j8TMyytQtvW+OaMKFIbq7Qqu7nIfwIFTV45vBHW6uwT/jAq/J3EgZ8K7JGyysVVHk86D8jT+xu90YVH5Tx/w97luxHOQGfSK8alhCszw="
+    # encrypted BINSTAR_TOKEN for push of dev package to binstar
+    - secure: Qz3pEYXXFnNQ/WK+15ad4cdbLJvzgCIZRwKD9fLiS3CDO2ldAQWxzaz8RQOwqbFtZUWu7lQpr+GukNJz5p0w18QEto+BxLYG9aW5mjoc+F2vCjyWFjkwnJ/Z/3uBKTcr5x9Y7HKaPGivaJ4BNACifjt7cCpeVJzV6u2+bBgSoHc=
 
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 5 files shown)


jchodera:
This will also fix #819, but still debugging the binstar push.

peastman:
Thanks, that will be great.  Let me know when it's ready.

jchodera:
Looks like this works now, though it produces so much output you have to download the travis `log.txt` file.  Let me see if I can suppress some inessential output.

jchodera:
OK, @peastman, I think this works now.  I can't figure out how to make the cmake-driven builds any quieter during the conda `openmm-dev` package build so that we don't have to download the log file to figure out whether the conda package built and uploaded OK---maybe you have ideas that could be applied in the future.

The push to binstar should work after you merge this PR and it builds the commit from the `master` branch to substitute the encrypted environment variables.

peastman:
Shouldn't it only do the conda build after a change has been merged, not for every PR?  Making travis take that much longer to run will be inconvenient, and having such a huge log makes it hard to look at (even for the early part that isn't truncated - the web browser gets really slow).

jchodera:
> Shouldn't it only do the conda build after a change has been merged, not for every PR? Making travis take that much longer to run will be inconvenient, and having such a huge log makes it hard to look at (even for the early part that isn't truncated - the web browser gets really slow).

Good luck debugging that.

jchodera:
The same goes for the code that builds the docs and pushes them to S3, I imagine (which isn't working in the main branch but was fixed in this PR).

peastman:
I'm not sure what you mean.  I assume it only uploads the build after a merge, right?  Otherwise that's a *very* serious security hole, since anyone can send a PR.  Assuming that, what is the difficulty in only doing the build in that case too?

kyleabeauchamp:
Well, one difficulty is that it increases the amount of custom code in the travis script here.  Ideally we'd like to keep things uniform among the different projects to reduce the number of lines of CI code that we maintain.

If the current code is working, maybe we should merge it to move forward on the PDBFixer issues and then plan improvements as a separate PR?

jchodera:
> I'm not sure what you mean. I assume it only uploads the build after a merge, right? Otherwise that's a very serious security hole, since anyone can send a PR. 

The encrypted environment variables are only substituted in travis builds on commits, not PRs, or else anyone could grab your decrypted variables via a PR.

> Assuming that, what is the difficulty in only doing the build in that case too?

There *is* logic to skip the [`after_success.sh`](https://github.com/pandegroup/openmm/blob/master/devtools/ci/after_success.sh) script in pull requests, but it doesn't seem to be working properly.

Even if we can debug that and fix it, we then have to move code around into this script and assume that committing it will actually make it still work the right way.  If it doesn't work, you will need to make more commits to the main branch to get it to work, and you can't figure out if it doesn't work until you commit to the main branch and wait for the build.  I can't do this easily because I don't have access to do that.

Given that, and the fact that it is now 2 am in Berlin, I leave the remainder as an exercise to you.  :)

peastman:
For the moment let's hold off on this then.  It would make travis significantly less useful/convenient for testing, which is its primary purpose.

jchodera:
Can you advise about how to suppress all the cmake output from make during
the conda build? That is what is causing the log file length issue.

jchodera:
Also, @kyleabeauchamp has a good point about consistency among projects. We
had this working before but took it out at your insistence. Keeping a
different version of each tool for openmm is a total pain. What precisely
about the old style of building the conda package and then testing it
didn't you like?

peastman:
> What precisely about the old style of building the conda package and then testing it didn't you like?

Remind we what you're talking about?  It's been a while.

jchodera:
https://github.com/pandegroup/openmm/commit/bcbc107c80563a4c53d3a550a43aaea5de577910

kyleabeauchamp:
Regarding silent mode, can we just redirect output to dev null or is that insufficiently cross platform?
```
make -j4 > /dev/null
make install > /dev/null
```

jchodera:
I think we still want to see build *errors* in case the conda build fails.
My reading suggests it required changes to CMakeLists.txt to get this to
work. @peastman is the only one who really understands the CMake system, so
let's let him fix that if he is insisting on minimizing output from this
stage.

jchodera:
For now, to help you guys move things along, I'll add an `openmm-dev` recipe to our [conda-recipes](https://github.com/omnia-md/conda-recipes) and have Jenkins build and push this.  This is not something we want to continue to maintain in the long run, however, so it would be best if we can work with @peastman to expediently fix this PR so we can drop this workaround.

jchodera:
This Jenkins project will try to build the linux versions when commits are made:
https://jenkins.choderalab.org/job/conda-openmm-dev-linux-vagrant/

peastman:
> What precisely about the old style of building the conda package and then testing it didn't you like?

You posted a pretty good summary of the problems with that approach here: https://github.com/pandegroup/openmm/issues/572

jchodera:
Jenkins is now pushing `omnia-dev` builds for `linux` on commits to `master`:
https://binstar.org/omnia/openmm-dev/files

Let's still try to get this PR sorted out quickly though.

peastman:
Great!  So to get the PDBFixer builds working, I just need to change "openmm" to "openmm-dev" in .travis.yml, yes?

kyleabeauchamp:
Yes

jchodera:
Yup! Let me know if you have issues or need mac builds too.

jchodera:
Any idea what the status of this is?  This PR seems out of date now, but still seems useful, right?

peastman:
It's partly a question of whether we want dev builds getting built by travis or jenkins.  What is the intended use of those builds?  Are they strictly for use by travis scripts that depend on OpenMM?  Or are they supposed to be an easy way for anyone to get the latest development code?  If the latter, they really should include the CUDA and OpenCL platforms, which travis currently doesn't generate.

jchodera:
OK, I'll close this then.

I believe we *do* want to incorporate the fix for tests that fail the first time triggering travis to report a failure even though the tests succeed on re-run, which is not the behavior we want.  I can make a separate PR for that:
https://github.com/pandegroup/openmm/pull/817/files#diff-354f30a63fb0907d4ad57269548329e3L26

peastman:
You already fixed that. :)

jchodera:
Ah, OK!  Thanks!

PR #824, opened 2015-02-20
Fix `unitcell.py`

swails:
This fixes a simple `NameError` in `unitcell.computeLengthsAndAngles` (it's been there since the beginning).  It also adds a set of tests for that module, and adds a function to reduce box vectors if you start with vectors, but not lengths and angles.

Note that because of the reduction, a round-trip between `computePeriodicBoxVectors` -> `computeLengthsAndAngles` does _not_ actually return the same lengths and angles you fed it necessarily.  However, a 2-cycle round-trip works (as you would expect, given that reduced vectors/lengths should be interconvertible without being changed).

### DIFF
--- a/wrappers/python/tests/TestUnitCell.py
+++ b/wrappers/python/tests/TestUnitCell.py
@@ -1,0 +1,64 @@
+import unittest
+import math
+from simtk.unit import *
+from simtk.openmm.app.internal.unitcell import computePeriodicBoxVectors
+from simtk.openmm.app.internal.unitcell import computeLengthsAndAngles
+
+def strip_units(x):
+    if is_quantity(x): return x.value_in_unit_system(md_unit_system)
+    return x
+
+class TestAmberPrmtopFile(unittest.TestCase):
+
+    """Test the AmberPrmtopFile.createSystem() method."""
+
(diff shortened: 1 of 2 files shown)


swails:
The failure was `TestDrudeLangevinIntegrator` which passed upon retesting.  The rest of the C++ and Python test suites passed fine.

wrappers/python/tests/TestUnitCell.py
```diff
@@ -0,0 +1,85 @@
...
+from simtk.openmm.app.internal.unitcell import reducePeriodicBoxVectors
+
+def strip_units(x):
+    if is_quantity(x): return x.value_in_unit_system(md_unit_system)
+    return x
+
+class TestAmberPrmtopFile(unittest.TestCase):
+
+    """Test the AmberPrmtopFile.createSystem() method."""
+
```
peastman: The class name and doc string here are incorrect.  It should be TestUnitCell, not TestAmberPrmtopFile.

swails:
Fixed.  You'll never guess which test file I copied and pasted the template from :).

peastman:
Looks great.  I'll merge it as soon as travis finishes.

PR #825, opened 2015-02-20
Added removeForce() and removeConstraint() to System

peastman:
By popular demand!  I did *not* add a removeParticle(), because that would break all sorts of things.  But removing forces and constraints should be pretty safe, since nothing else in the API references them by index.


jchodera:
Awesome!

PR #827, opened 2015-02-21
Lots of cleanup to reference platform

peastman:
Housecleaning that has needed doing for a long time.  There's still lots to do, of course.


rmcgibbo:
wow this is great. especially the`RealOpenMMVector` to `std::vector<RealOpenMM>`.

PR #839, opened 2015-02-27
PdbStructure handles byte streams correctly in Python 3

peastman:
This is the fix for https://github.com/pandegroup/pdbfixer/issues/81.


rmcgibbo:
lgtm. maybe should add a test that tries to open a pdb with `open(file, mode='rb')`.

peastman:
Good idea.  Added.

PR #844, opened 2015-03-05
Update `Topology` creation from CHARMM parsers and fix Amber prmtop parsing when %COMMENT sections are present

swails:
I still want to turn the prmtop that gave rise to this error into a test case.  Don't merge yet.


swails:
Test case added.  This should be good to go.

peastman:
Did you check in the case case?  I don't see it there.

swails:
> Did you check in the case case? I don't see it there.

Assuming you mean test case, I modified an existing one by adding a `%COMMENT` section.  That is sufficient to guarantee coverage for this fix.

swails:
See 0a0e3ba

peastman:
Ok!

PR #850, opened 2015-03-14
Fixing multiple bugs in dcdfile / dcdreporter.

kyleabeauchamp:
Here is the driver code:

```
from simtk.openmm.app import *
from simtk.openmm import *
from simtk.unit import *
from sys import stdout


temperature = 50*kelvin

prmtop = AmberPrmtopFile('CO.prmtop')
inpcrd = AmberInpcrdFile('CO.inpcrd')
system = prmtop.createSystem(nonbondedMethod=PME, nonbondedCutoff=1*nanometer, constraints=HBonds)
system.addForce(MonteCarloBarostat(1.0 * atmospheres, temperature, 1))

integrator = LangevinIntegrator(temperature, 1/picosecond, 0.0001*picoseconds)
simulation = Simulation(prmtop.topology, system, integrator)
simulation.context.setPositions(inpcrd.positions)
simulation.context.setPeriodicBoxVectors(*inpcrd.boxVectors)


simulation.reporters.append(PDBReporter('output.pdb', 1))
simulation.reporters.append(DCDReporter('output.dcd', 1))
simulation.reporters.append(StateDataReporter(stdout, 1, step=True, potentialEnergy=True, temperature=True, density=True))
simulation.step(10000)
```

Here is the first bug, which is already fixed in this PR:

```
/home/kyleb/opt/lib/python2.7/site-packages/simtk/openmm/app/simulation.pyc in step(self, steps)
    125                 for reporter, next in zip(self.reporters, nextReport):
    126                     if next[0] == nextSteps:
--> 127                         reporter.report(self, state)
    128 

/home/kyleb/opt/lib/python2.7/site-packages/simtk/openmm/app/dcdreporter.pyc in report(self, simulation, state)
     75             self._dcd = DCDFile(self._out, simulation.topology, simulation.integrator.getStepSize(), 0, self._reportInterval)
     76         a,b,c = state.getPeriodicBoxVectors()
---> 77         self._dcd.writeModel(state.getPositions(), mm.Vec3(a[0].value_in_unit(nanometer), b[1].value_in_unit(nanometer), c[2].value_in_unit(nanometer))*nanometer)
     78 
     79     def __del__(self):

/home/kyleb/opt/lib/python2.7/site-packages/simtk/openmm/app/dcdfile.pyc in writeModel(self, positions, unitCellDimensions, periodicBoxVectors)
    113         boxVectors = self._topology.getPeriodicBoxVectors()
    114         if boxVectors is not None:
[...]

PR #851, opened 2015-03-14
Modified travis logic to only fail if retested tests fail.

jchodera:
This modifies the travis logic so that ctests only fail if they fail on retest, rather than triggering a travis failure if they fail the first time.  This was discussed in #850.


jchodera:
Should be good to merge.

PR #856, opened 2015-03-19
Fix HAngles with Amber prmtop files

### LINKED ISSUE #855: Behavior of app.HAngles for amber prmtop
I'm currently using an automated pipeline to simulate various neat liquids parameterized in GAFF.  I'm using the prmtop reader to set up these simulations.  

One of my target molecules is a GAFF-like water model.  Although such a model is not useful for predicting biophysical properties, it's still useful as a unit test for my pipeline code.  

My question is this: I would like to switch my GAFF water model into a rigid model.  From my reading of the OpenMM manual ("The lengths of all bonds are constrained. In addition, all angles of the form H-X-H or H-O-X (where X is an arbitrary atom) are constrained."), this is as simple as switching the constraint flag from HBonds to HAngles.  

As a test of this, consider the script below, which compares the system objects for the two different pipelines.  I would expect the number of angles and constraints to change when the constraint type is switched to HAngles, but no such change occurs.  

```
from simtk.openmm import app
import simtk.openmm as mm
from simtk import unit as u

for constraints in [app.HBonds, app.HAngles]:
    for prmtop_filename in ["./water-box-216.prmtop", "./14314-42-2_1000_303.2.prmtop"]:
        prmtop = app.AmberPrmtopFile(prmtop_filename)
        system = prmtop.createSystem(nonbondedMethod=app.PME, nonbondedCutoff=1.0 * u.nanometers, constraints=constraints)
        print("**********")
        print("filename = %(filename)s constraints = %(constraint_type)s" % dict(filename=prmtop_filename, constraint_type = constraints))
        for k, atom in enumerate(prmtop.topology.atoms()):
            if k < 3:
                print(atom.name, atom.element)        
        output = "particles = %(n_particles)d, bonds = %(n_bonds)d, angles = %(n_angles)d, constraints = %(n_constraints)d" % dict(
            n_particles = system.getNumParticles(),
            n_bonds = system.getForce(0).getNumBonds(),
            n_angles = system.getForce(1).getNumAngles(),
            n_constraints = system.getNumConstraints())
        print(output)
```

```
**********
filename = ./water-box-216.prmtop constraints = HBonds
('O', <Element oxygen>)
('H1', <Element hydrogen>)
('H2', <Element hydrogen>)
particles = 648, bonds = 0, angles = 0, constraints = 648
**********
filename = ./14314-42-2_1000_303.2.prmtop constraints = HBonds
('O1', <Element oxygen>)
('H1', <Element hydrogen>)
('H2', <Element hydrogen>)
particles = 3000, bonds = 0, angles = 1000, constraints = 2000
**********
filename = ./water-box-216.prmtop constraints = HAngles
('O', <Element oxygen>)
('H1', <Element hydrogen>)
('H2', <Element hydrogen>)
particles = 648, bonds = 0, angles = 0, constraints = 648
**********
filename = ./14314-42-2_1000_303.2.prmtop constraints = HAngles
('O1', <Element oxygen>)
('H1', <Element hydrogen>)
('H2', <Element hydrogen>)
particles = 3000, bonds = 0, angles = 1000, constraints = 2000
```

swails:
This PR fixes #855 (and adds the troublesome prmtop file as a test).  A detailed overview of the changes:

- Unrelated; ~20% performance boost in prmtop file reading via a very simple adjustment
- Add required `topology` argument to `readAmberSystem` (backwards-incompatible change) in order to avoid duplicating the atom-element-determination code (much better to keep that in one place).  But `readAmberSystem` is a private function only called in one place -- API stability there is not guaranteed.
- Determine element identity in the H-atom determination more reliably (i.e., use the atomic number if available, or use the first letter of the atom name if not present)
- Add's @kyleabeauchamp's prmtop file that revealed the bug to the test suite


kyleabeauchamp:
Cool.  IMHO your idea of using the elements instead of string searching seems like a good idea here.

peastman:
In the original, numH was calculated as

    numH = len([type for type in (type1, type3) if type.startswith('H')])

It was only looking at the first and third atoms.  You changed that to

    numH = ((atomI.element.atomic_number == 1) + (atomJ.element.atomic_number == 1) +
    (atomK.element.atomic_number == 1))

which looks at all three atoms.  Now, it would be *very* strange for the central atom in an angle to be a hydrogen, so in practice this is unlikely to make a difference.  But if somehow that situation ever came up, the new code would be wrong.

swails:
Turns out the test for Oxygen was wrong, too (should have been `atomJ.element is elem.oxygen` rather than `atomK`.

Fixed both things.

PR #865, opened 2015-04-01
Incorporate gravitational effects in our standard simulations

swails:
One of the greatest shortcomings in modern fixed-charge force fields is the
complete neglect of gravitational forces. These forces are long-ranged, having
the same functional form as electrostatic interactions. Unlike electrostatic
interactions, though, which cancel over long distances due to net neutrality,
gravity is strictly attractive and has been known to stretch over a light-year
(ca. 9.5e24 nm) in our solar system. [1]

Our force fields also neglect earth's gravitational influence on our system.
This commit adds inter-particle gravitational attraction between all massive
particles as well as an elevation-dependent external force due to Earth's
gravity to all standard force field simulations.

This enhancement should resolve the remaining issues with our fixed-charge force
fields, and so should be pushed out as an immediate bugfix update to OpenMM 6.2.

[1] Note that because of the long-range nature of gravity, we need @peastman to
implement a general PME for the CustomNonbondedForce so that the full
gravitational interaction can be accounted for. Until then, a sophisticated
switching function *should* suffice (although a long-range correction is
impossible due to the divergent nature of the potential energy function).

### DIFF
--- a/wrappers/python/simtk/openmm/app/charmmpsffile.py
+++ b/wrappers/python/simtk/openmm/app/charmmpsffile.py
@@ -53,6 +53,7 @@ from simtk.openmm.app.internal.charmm.topologyobjects import (
 from simtk.openmm.app.internal.charmm.exceptions import (
                 CharmmPSFError, MoleculeError, CharmmPSFWarning,
                 MissingParameter, CharmmPsfEOF)
+from simtk.openmm.app.internal.grav import addGravity
 import warnings
 
 TINY = 1e-8
@@ -1454,6 +1455,8 @@ class CharmmPsfFile(object):
 
         # Cache our system for easy access
         self._system = system
+
+        addGravity(system, elevation=0*u.meters)
 
         return system
 
--- a/wrappers/python/simtk/openmm/app/desmonddmsfile.py
+++ b/wrappers/python/simtk/openmm/app/desmonddmsfile.py
@@ -32,9 +32,10 @@ from simtk import openmm as mm
 from simtk.openmm.app import forcefield as ff
 from simtk.openmm.app import Element, Topology, PDBFile
 from simtk.openmm.app.element import hydrogen
+from simtk.openmm.app.internal.grav import addGravity
 from simtk.unit import (nanometer, angstrom, dalton, radian,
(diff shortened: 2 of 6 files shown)


VijayPande:
Nicely done.

peastman:
Looks like you've got a bunch of test failures.  Could you look into them?

I'm also not sure about this bit:

```Python
# Calculate the acceleration; assume Earth's radius is 6.37101e6 meters,
# mass is 5.9736e24 kg, and g is given above
g = G * 5.9736e24*u.kilograms / (6.37101e6*u.meters + elevation)**2
```

I don't think we should be hardcoding values for Earth.  Astrobiology is an important field, and users should be able to do simulations on the moon, Europa, exoplanets, etc.  We can still provide default values that correspond to Earth, but then allow users to override them.

I think it's also important to model changes in gravity with position _within_ the simulation box (getting weaker with increasing height).  On Earth this will generally be negligible, but near a neutron star it could be significant.

Otherwise, this looks great.

swails:
> Looks like you've got a bunch of test failures. Could you look into them?

Small mistakes are common in cutting-edge research like this, but I think I got the last of the major issues.

> I think it's also important to model changes in gravity with position within the simulation box (getting weaker with increasing height). On Earth this will generally be negligible, but near a neutron star it could be significant.

I completely agree -- I'm particularly interested in how a single protein will behave as it approaches the event horizon.  However, gravity this strong often carries with it interesting time distortion.  I figured I could generalize this external force after we coded an integrator that could handle the distortions in spacetime.

peastman:
Yeah, if you want to handle space near a black hole then general relativity will be essential.  But I think that's beyond the scope of this PR.  For a neutron star, though, I think classical gravity should be ok.

rmcgibbo:
I don't think we want a switching function. Do you want to implement Barnes
Hut here, or in the next PR?

jchodera:
We're very interested in getting this PR merged so that we can utilize this functionality in YANK to study the effect of gravitation and weightlessness on protein-ligand binding in order to validate the results of this very important study:
http://www.thermofisher.com/en/about-us/protein-binding-international-space-station.html

PR #868, opened 2015-04-06
Fixed a performance regression on AMD GPUs

peastman:


mpharrigan:
When was this regression introduced?

peastman:
Somewhere between 6.0.1 and 6.1.  By the way, this is a pretty minor issue.  It only slows it down by a few percent at most.

mpharrigan:
Thanks!

PR #870, opened 2015-04-09
Implemented an optimized real-to-complex FFT

peastman:


kyleabeauchamp:
Does the CUDA platform already do something like this?  Just curious.

swails:
NVidia provides cufft, which is what we use.

kyleabeauchamp:
Oh, we have our own FFT code for OpenCL.  Good to know.

PR #871, opened 2015-04-10
Improve CHARMM parsing when resnums have inscodes

swails:
Apparently CHARMM PSF files (particularly those printed by VMD's psfgen) can
have an insertion code tacked on to the end of the residue number, meaning that
casting to an integer will invariably fail.  We can't just ignore the insertion
code, though, since it will differentiate from the residues that came before and
after (which will likely have the same number -- same general idea as insertion
codes from PDB files). So instead we pull the insertion code off of the residue
number and extend the Residue object to accommodate that.

This is the port of the fix from ParmEd: ParmEd/ParmEd#98

I added a test for this case.


swails:
I'm a little short on time to test this locally, so I'm going to rely on Travis for this relatively simple fix (and it's unrelated to any of the GPU platforms, so the Travis tests should ensure correct coverage of the added code)

peastman:
Looks like TestCharmmFiles failed.

swails:
Yea, it was a timeout.  I wonder if I introduced some kind of huge performance regression...  I'll look into it a bit more.

swails:
So I ran it through a profiler and found that `getElementByMass` is unbearably slow.  On my Mac, parsing the PSF took 3 seconds.  Generating the `Topology` took 167 seconds (most of that time spent in `getElementByMass`).  And within that, a lot of time is spent in the `unit` package.

I have two ideas for working around the `unit` overhead here (limiting the number of elements we search through and bypassing units altogether).

As a side-note, I'd never have figured this out without a profiler -- `getElementByMass` is not a place I'd have expected a performance bottleneck (although it makes sense in retrospect, obviously).

swails:
OK.  This PR should finally be good to go if Travis passes.  Everything's working on my computer.

peastman:
Wow, that implementation of getByMass() was... interesting.  If I wrote that, I apologize!

Your new version has a bug, though.  It assumes that atomic masses monotonically increase with atomic number, which isn't always true.  For example, argon (atomic number 18) has a higher mass than potassium (atomic number 19).  You should create a separate list of elements sorted by atomic mass and use that.  I'd also suggest creating a test case that randomly picks masses between 0 and 300, calls getByMass(), then verifies the returned element really is closer than any other element.

swails:
> Wow, that implementation of getByMass() was... interesting. If I wrote that, I apologize!

Actually it was mine (from around when I started contributing) :).  I'd used it before, but never with `unit`s -- normally 100 comparisons aren't much more expensive than 1 to 10 comparisons for basic types.  It was never a bottleneck, so never a target for optimization.  I underestimated the extra work the dimensional analysis introduced.

> Your new version has a bug, though. It assumes that atomic masses monotonically increase with atomic number, which isn't always true.

Oh come on, chemistry.

> You should create a separate list of elements sorted by atomic mass and use that. I'd also suggest creating a test case that randomly picks masses between 0 and 300, calls getByMass(), then verifies the returned element really is closer than any other element.

Yea, this is what I'll do.

swails:
OK.  Bug fixed and test added.  Should be good to go.

peastman:
The failure was TestReferenceDrudeLangevinIntegrator, so nothing related to this.

swails:
Ok. There's actually an option in Travis to trigger a rebuild if you want to.

peastman:
I'm not sure why the OrderedDict?  A simple list would have worked just as well.  But that's probably not important.  Where is the option to rebuild?  I'm not seeing it.

swails:
> I'm not sure why the OrderedDict? A simple list would have worked just as well.

I like `OrderedDict` better than the list of elements (or list of tuples), and all of the other (admittedly private) caches are dicts (which is consistent, at least).  And actually the first thing I was going to do was something like:

```Python
try:
    return Element._elements_by_mass[inputMass]
except KeyError:
    # Now cycle through elements
```

since element lookups are O(1) in dicts.  But I decided this was premature optimization (bypassing dimensional analysis was the big win).  I can change it to a list if you think that'd be a readability improvement in the long run.

> Where is the option to rebuild? I'm not seeing it.

On the Travis webpage for the failed test, there should be an icon in the top-right that looks like the standard "refresh" button (circular arrow).  That's the "re-run" button.  I've had to use it on occasion when Travis kills the my ParmEd tests.

peastman:
> there should be an icon in the top-right that looks like the standard "refresh" button (circular arrow).

Ah, there it is!  That was a little too subtle for me.

peastman:
This time, all tests passed the first time through.  Shrug.  I'll keep an eye on that test.

rmcgibbo:
Just noticed this. Why doesn't it use binary search?

peastman:
It would be slower.  If your system is mostly water, then 2/3 of the atoms will match the very first element, and nearly all of them will be in the first 8.  That makes a linear search from the beginning very fast.

rmcgibbo:
Okay, that's fair.

swails:
> It would be slower. If your system is mostly water, then 2/3 of the atoms will match the very first element, and nearly all of them will be in the first 8. That makes a linear search from the beginning very fast.

That was my rationale.  Even in standard proteins, there are ~1/2 H (succeeds in 2 comparisons), and the majority of the rest are C, N, or O (7, 8, and 9 comparisons).  There few S (18 comparisons).  Ions are also typically small, and transition metals are rare, and typically occur in numbers of less than 5.

If most of the atoms are water, it becomes even better to search from the beginning.

PR #872, opened 2015-04-10
added XML File classes

cxhernandez:
Should make it simpler for users to load XML files.

### DIFF
--- a/wrappers/python/simtk/openmm/app/xmlfile.py
+++ b/wrappers/python/simtk/openmm/app/xmlfile.py
@@ -1,0 +1,28 @@
+from simtk.openmm import XmlSerializer
+from simtk.openmm.openmm import State, System, Integrator
+
+
+class XMLFile(object):
+
+    def __new__(cls, filename):
+        obj = cls._load_xml_(filename)
+        assert isinstance(obj, cls.xmlcls), "File does not contain %s." % cls.xmlcls
+        return obj
+
+    @staticmethod
+    def _load_xml_(filename):
+        with open(filename, 'rb') as f:
+            xmlfile = XmlSerializer.deserialize(f.read())
+        return xmlfile
+
+
+class XMLStateFile(XMLFile):
+    xmlcls = State
+
+
+class XMLSystemFile(XMLFile):
+    xmlcls = System
+
+
+class XMLIntegratorFile(XMLFile):
+    xmlcls = Integrator
--- a/wrappers/python/simtk/openmm/app/xmlfile.py
+++ b/wrappers/python/simtk/openmm/app/xmlfile.py
@@ -1,8 +1,44 @@
+"""
+xmlfile.py: Used for loading XML files.
+
+This is part of the OpenMM molecular simulation toolkit originating from
+Simbios, the NIH National Center for Physics-Based Simulation of
+Biological Structures at Stanford, funded under the NIH Roadmap for
+Medical Research, grant U54 GM072970. See https://simtk.org.
+
+Portions copyright (c) 2015 Stanford University and the Authors.
+Authors: Carlos Xavier Hernandez
+Contributors:
+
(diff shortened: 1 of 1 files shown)


peastman:
What is the intended use of this class?  I don't think I understand it.

cxhernandez:
My understanding of the python API is that if I wanted to load in any sort of XML file to start a simulation, I would have to execute this three times:

``` python
with open(filename, 'rb') as f:
    xmlfile = XmlSerializer.deserialize(f.read())
```

As a beginner, this was not immediately obvious/intuitive to me, and I thought it'd be nice to have functions with the same ease-of-use as PDBFile.

Another suggestion that @mpharrigan had was to define a `load_from_xml` module for the `Simulation` object, which would initialize a simulation from the state, system, and integrator xml files.

mpharrigan:
Yes, I think a class method
```python
simulation = Simulation.from_xml('system.xml', 'integrator.xml', 'state.xml')
```
 would be swell

swails:
Do we still want to make a `from_xml` constructor on `Simulation`?  If so, I'll write up a PR for it if nobody else wants to.  But it seems like we're not moving forward with what's in this PR currently, right?

cxhernandez:
+1 for `from_xml`

I would write the PR but I'm not entirely sure how to gracefully circumvent the `Topology` requirement that the `Simulation` class currently has.

swails:
Ah, turns out to be easy :smiley:.  The `Topology` is no longer needed.  The only thing it's there for is to determine whether or not PBCs are used.  But this is now accessible through `System.usesPeriodicBoundaryConditions()`.

There is still one incredibly obscure corner case where it's better to use `topology.getUnitCellDimensions()` instead of `system.usesPeriodicBoundaryConditions()`, but for 99.9% of cases, the latter is actually better.

That corner case is when someone uses a plugin `Force` they wrote that doesn't implement `usesPeriodicBoundaryConditions`.  If that's the case, and their system is *not* periodic (as defined by some other `Force` in the system), then you get an exception thrown.

I'll put together a PR.

mpharrigan:
Isn't the topology still used for (certain) reporters?

jchodera:
I wonder if we need an XML representation for `Topology` as well.

swails:
> Isn't the topology still used for (certain) reporters?

Sure, but that has nothing to do with `Simulation`

swails:
Have a look at #1370 which implements my ideas.

peastman:
> Sure, but that has nothing to do with Simulation

It has lots to do with Simulation.  If you set `topology` to `None` and then add a PDBReporter, PDBxReporter, or DCDReporter, you'll get an exception.  `None` is not a supported value for that field, and other classes rely on that.

swails:
Oh because those reporters take Simulation as an input parameter...  I forgot about that!

My bad... Yea, I'll need to rethink this a bit.

swails:
A Topology can always be constructed from the System, but it won't have as much information attached to it (but it will have enough for all of the reporters in OpenMM if I recall correctly).

peastman:
Not if you want to write out a valid PDB or PDBx file.  The System has no information about atom or residue names, or how the atoms are grouped into residues.  Or elements, though you could try to guess those from the masses.

swails:
Good point. Ok. I'm done replying about omm details from my phone :). How about I just remove the fromXmlFiles constructor for now and keep the additions of being able to accept XML files in the constructor? Is that a decent addition?

mpharrigan:
> How about I just remove the fromXmlFiles constructor for now and keep the additions of being able to accept XML files in the constructor? Is that a decent addition?

Sounds like a good improvement to me!

peastman:
Sounds reasonable.

cxhernandez:
+1

swails:
#1370 was merged, so this can probably be closed now...

PR #877, opened 2015-04-16
Fixed a bug in OpenCL FFT

peastman:


jchodera:
Any comment on whether this bug was introduced recently or if this was a
long - present bug that may have affected calculations?

peastman:
No, this was just introduced a few days ago.  I've been making a lot of changes to that code.

PR #878, opened 2015-04-16
Enabled OpenCL on OS X

peastman:
As of OS X 10.10.3, Apple has finally fixed their OpenCL bugs!  That means we can finally start supporting it again.


jchodera:
Hooray!

jhprinz:
Cool. That will be very useful here in Amsterdam since most iMacs have AMD chips!

jhprinz:
Does that also mean we could use it already, when compiling OpenMM ourselves?

swails:
> Does that also mean we could use it already, when compiling OpenMM ourselves?

Yes.

jchodera:
The development conda package of OpenMM should also work:
https://binstar.org/omnia/openmm-dev
```
conda install -c https://conda.binstar.org/omnia openmm-dev
```

bb30994:
>swails commented Apr 17, 2015
>>    Does that also mean we could use it already, when compiling OpenMM ourselves?
>Yes.
For stand-alone OpenMM Yes.  For FAH, the servers still won't send you WUs.

Issue #879, opened 2015-04-20
SWIG 3.0.5 broken?


swails:
I'm getting some strange behavior in SWIG 3.0.5 on my Mac (OS X 10.10.3).  In particular, some of the wrapped classes no longer accept ``*args`` as their argument list -- it is set to whatever is in the actual header file.  Some of the custom SWIG code depends on ``*args``, though, like `stripUnits`.  This causes all of the Python CHARMM tests to fail in `System.addParticle` and all Amber tests to fail in `System.addForce`.

In the SWIG-generated `openmm.py` file, this is where the Amber files are failing:

```Python
    def addForce(self, force):
        """
        addForce(System self, Force force) -> int

        Add a Force to the System. The Force should have been created on the heap with the "new" operator. The System takes over ownership of it, and deletes the Force when the System itself is deleted.
           Parameters:
            - force a pointer to the Force object to be added
        """
        if not args[0].thisown:
           s = "the %s object does not own its corresponding OpenMM object" \
              % args[0].__class__.__name__
           raise Exception(s)


        val = _openmm.System_addForce(self, force)
        args[0].thisown=0


        return val
```

This clearly throws a `NameError` when trying to access `thisown`.  Likewise, for the CHARMM files:

```Python
    def addParticle(self, mass):
        """
        addParticle(System self, double mass) -> int

        Add a particle to the System. If the mass is 0, Integrators will ignore the particle and not modify its position or velocity. This is most often used for virtual sites, but can also be used as a   way to prevent a particle from moving.
           Parameters:
            - mass the mass of the particle (in atomic mass units)
        """
        try: args=stripUnits(args)
        except UnboundLocalError: pass


        return _openmm.System_addParticle(self, mass)
```

Since `addParticle` is called with mass being a `unit.Quantity` instance, the `return` line fails with a `TypeError` (not a `double`).

Is anyone else seeing this behavior?  Is it a bad CMake setting?

swails:
So I checked out version 6.1 and compiled that.  I got the same issue, so it's definitely a SWIG 3.0.5 thing.  I don't know how to tell SWIG to go back to using `*args` for all of the wrapped call signatures.

Am I the only one seeing this?

peastman:
I've been going through the SWIG 3 documentation trying to find anything related to this.  At http://www.swig.org/Doc3.0/Python.html#Python_nn67 it says:

[...]

PR #882, opened 2015-04-22
Fix SWIG 3 support for the Python wrappers.

swails:
See swig/swig#387 for a discussion.  It would actually be nice to *not* set the
"compactdefaultargs" feature globally, since that would allow us to actually see
the real API at the Python level the same as you would see by looking at the C++
header files. But this is the only way to support both SWIG 2 and SWIG 3 easily
and avoid a lot of retooling for the SWIG input code generator.

### DIFF
--- a/wrappers/python/src/swig_doxygen/swig_lib/python/features.i
+++ b/wrappers/python/src/swig_doxygen/swig_lib/python/features.i
@@ -6,3 +6,12 @@
 %include pythonprepend.i
 %include pythonappend.i
 %include typemaps.i
+/* SWIG 3.x resolved a bug in which all wrapped C++ functions took *args as its
+ * default argument list. OpenMM then exploited this bug by doing stuff like
+ * passing args to stripUnits (and all added code assumed that the arguments
+ * were in an "args" list). So in order to restore this arguably buggy behavior
+ * from SWIG 2, enable the "compactdefaultargs" feature globally.
+ *
+ * See https://github.com/swig/swig/issues/387
+ */
+%feature("compactdefaultargs");


swails:
This should fix #879 .  I'll check that it does on my machine that has SWIG 3 installed and then report back.

swails:
OK -- fixes the Python wrappers on my Mac with SWIG 3.0.5.

peastman:
Thanks!  I'm glad the fix turned out to be so simple.

swails:
Me too :). Thanks for the quick merge.

rmcgibbo:
I think I know how to fix this actually now, such that we get nice function signatures that don't use the *args.

swails:
AFAICT from the SWIG source code, certain versions of SWIG *always* compressed down to `*args`, and there really wasn't anything you could do about that.  So if we get away from `compactdefaultargs`, we would have to cope with older versions of SWIG 2 always using it.

And then you have [here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/src/swig_doxygen/swig_lib/python/pythoncode.i#L222-L284) that depends on compact default arguments.  So getting away from that means you need to add to the wrapper generator code for each method of each class where an input argument may have units.  Which you'd then have to update every time another force, method, or argument was added.  This would break compatibility with SWIG 2 (that always uses compactdefaultargs by accident).

Actually going to nice function signatures would have two advantages -- nice function signatures and the ability to do better unit checking (so we can check that the right types of units are applied, not just blindly converting to md_unit_system).

swails:
I saw a way to do it myself, but unless you're seeing something different than I do, I didn't think it was actually a net improvement over the current situation when all things are considered...

rmcgibbo:
The argument names are available when writing the swig files, so you can support both compactdefaultargs and the newer behavior without too much fuss, and still using the same principle when stripping the units (not manually listed the expected units of all of the inputs).

rmcgibbo:
I mean, we could also change to listing all of the expected units on the input site if we want, but that's for another day.

Issue #867, opened 2015-04-02
TestCudaLocalEnergyMinimizer mixed fails


swails:
```
swails@batman ~/src/build_openmm_dev $ ./TestCudaLocalEnergyMinimizer single
Done
swails@batman ~/src/build_openmm_dev $ ./TestCudaLocalEnergyMinimizer mixed
exception: Assertion failure at TestCudaLocalEnergyMinimizer.cpp:133
swails@batman ~/src/build_openmm_dev $ ./TestCudaLocalEnergyMinimizer double
Done
```

This is reproducible -- mixed precision always assert-fails at the same line.  If @peastman (or whoever goes about debugging this) can't repro, I'll provide more info.

swails:
This is still happening for me with the latest code.

swails:
There is definitely an issue here -- the `forceNorm` (net normalized force?) in the mixed precision is over 3x larger than that in single precision! (~5.1 for single, >17 for mixed).  Double is, not surprisingly, even lower (ca. 1)

peastman:
I'll try to look into this soon.

swails:
I wasn't trying to be pushy (sorry if I came off that way).  The above was just a mind dump and (some) details of what I did to look into the problem -- this is not a pressing need for me (although I'm curious about the culprit).

peastman:
No problem!  I'm just a little bit buried in a stack of other things, or I would have investigated it already.

swails:
I figured as much

peastman:
I can't reproduce this.  I just tried on two different computers, and it passed repeatedly on both of them.  Does anyone else see this?

kyleabeauchamp:
```
kyleb@jak:~/src/openmm/omm-build$ ./TestCudaLocalEnergyMinimizer single
Done
kyleb@jak:~/src/openmm/omm-build$ ./TestCudaLocalEnergyMinimizer mixed
exception: Assertion failure at TestCudaLocalEnergyMinimizer.cpp:133
kyleb@jak:~/src/openmm/omm-build$ ./TestCudaLocalEnergyMinimizer double
Done
kyleb@jak:~/src/openmm/omm-build$ nvidia-smi 
Mon Apr 27 20:19:16 2015       
+------------------------------------------------------+                       
| NVIDIA-SMI 340.29     Driver Version: 340.29         |                       
|-------------------------------+----------------------+----------------------+
[...]

PR #885, opened 2015-04-28
Fixes the TestCudaLocalEnergyMinimizer

swails:
Not sure why.  Is there a nonbonded "skin" of some sort when building the
pairlist?

### DIFF
--- a/platforms/cuda/tests/TestCudaLocalEnergyMinimizer.cpp
+++ b/platforms/cuda/tests/TestCudaLocalEnergyMinimizer.cpp
@@ -80,7 +80,7 @@ void testLargeSystem() {
     const int numMolecules = 25;
     const int numParticles = numMolecules*2;
     const double cutoff = 2.0;
-    const double boxSize = 4.0;
+    const double boxSize = 5.0;
     const double tolerance = 5;
     System system;
     system.setDefaultPeriodicBoxVectors(Vec3(boxSize, 0, 0), Vec3(0, boxSize, 0), Vec3(0, 0, boxSize));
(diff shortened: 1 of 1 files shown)


swails:
So I bisected the `TestCudaLocalEnergyMinimizer mixed` to f2958043a7a60e19484e3290232dc41a2fadfbf3.  I'm a little wary of a 2A cutoff in a 4A box, since this is the theoretical minimum that will work with MIC.

Does OpenMM use a buffer or "skin" to avoid constant list rebuilds?  Or does it just always search adjacent cells?  If it uses a buffer, then wouldn't the box size need to be strictly larger than twice the cutoff?

peastman:
I think this PR is just trying to make the error go away without fixing the underlying cause.  Yes, it does pad the cutoff when building the neighbor list, but I don't think that should matter.  It still applies the exact cutoff when computing the interaction, so each atom only interacts with the nearest periodic copy of any other atom.  And padding the cutoff will only cause more interactions to get added to the neighbor list (which will then get set to 0 in the interaction kernel).  It wouldn't cause anything to be missed.

swails:
> I think this PR is just trying to make the error go away without fixing the underlying cause.

I suspect you're right.  This PR was intended to be the first step in the bug hunt and the findings of my bisect (although I suppose I could've put the patch in #867).

> Yes, it does pad the cutoff when building the neighbor list, but I don't think that should matter.

I think you're right.  My original line of thinking was MIC only grabs at most 1 image of any atom, but if the skin extends beyond the length of the box then a single particle could have *two* images inside that region, which is where I thought the "missed" interaction was coming from.  But at the end of the day only one image will be included in the nonbonded calc, so it only matters that any atom appears *at least once* in the pairlist.

swails:
This is the wrong approach, and there's already an issue open for this in #867.  No use in keeping this open.

Issue #886, opened 2015-04-30
[FR] Solvate with fixed number of waters


cxhernandez:
I think one thing that's been a big headache for people who are trying to make reproducible clones of simulations has been getting a fixed number of solvent molecules. I think It would be nice to add this functionality into `Modeller`. 

I have a prototype of the code on this [gist](https://gist.github.com/cxhernandez/5afae8cce3dcb034ee95), which is based off of @choderalab's [Ensembler](https://github.com/choderalab/ensembler/blob/aa3cfe690017ad0bb3dce49158ce8e6f31fb0127/ensembler/refinement.py#L565-L683) code.

kyleabeauchamp:
Hmm, possible locations for this might be:

OpenMM
OpenMMTools
Parmed
OpenMolTools

The advantage of doing this downstream of OpenMM is that you could possible have access to extra dependencies...

rmcgibbo:
I've also needed to solvate many copies of the same system (different conformations) with the same number of water molecules, and wasted a couple hours writing a script for it. +1 to putting some version of this functionality into the `Modeller` class.

rmcgibbo:
Neither @cxhernandez's code nor the ensembler code uses anything outside of OpenMM, AFAICT.

rmcgibbo:
IMO, the most logical place for this functionality is as a method on the modeller class, which happens to be in OpenMM.

kyleabeauchamp:
+1

jchodera:
See also:
* https://github.com/pandegroup/openmm/issues/312
* https://github.com/pandegroup/openmm/issues/518

rmcgibbo:
Oh right. If there's a consensus that @danielparton and @cxhernandez's approach is the right one, then that probably solves the biggest hurdle. @cxhernandez: do you want to take a stab at an OpenMM PR?

cxhernandez:
Oh wow, I completely forgot about #312

cxhernandez:
@rmcgibbo: yeah sure

swails:
+1 for `Modeller`

(hey wait guys!  i'm here!)

cxhernandez:
also any suggestions on what this kwarg should be named to keep in line with @peastman's conventions?

peastman:
[...]

PR #888, opened 2015-05-01
Fixed number of waters

cxhernandez:
as discussed in #886. still needs to be tested


cxhernandez:
Now that I think about it maybe it's probably best to just add ions after the fact, instead of deleting excess ones at the end

peastman:
This is a really slow way of implementing this.  Building a large water box can be an expensive operation.  Repeating it several times will make it much worse.  I also see a few other problems:

- If the input Topology doesn't specify a box size, this will crash.

- This deletes crystallographic waters, which is almost never wanted.

- You're clearly trying to produce a densely packed water box, but because you allow it to add too many waters (possibly a lot too many) and then delete the excess ones, it doesn't accomplish that very well.

- The ion concentration comes out wrong.  Neutralizing counterions shouldn't be included when calculating the ionic strength.

- The utility functions you created should be defined locally within the method.  They don't belong as part of the public interface of Modeller.

- The `attempts` parameter really doesn't belong there.  This isn't the sort of problem that you might attempt to solve but fail.

cxhernandez:
Closing this in the wake of your alternative implementation

PR #889, opened 2015-05-01
CPU PME plugin respects OPENMM_CPU_THREADS environment variable

peastman:


jchodera:
So this presumably wasn't being respected before in CPU PME simulations in prior releases?  Or is this just a modification to a post-6.2 commit?

peastman:
The direct space calculation respected it, but the reciprocal space calculation didn't.

Issue #887, opened 2015-05-01
OpenMM failure to load .prmtop file


apeck12:
A prmtop file generated in Amber/CHARMM36 is failing to load in OpenMM. Specifically, it cannot recognize the following line: 
`%FORMAT(i2,a78)`
and returns Nonetype as a result.

The example .prmtop file is available here: https://gist.github.com/apeck12/c32c3760ff1fbae989da

The error message is below:

```
In [1]: from simtk.openmm.app import AmberPrmtopFile 

In [2]: AmberPrmtopFile('apocam_seed0.prmtop')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-1040e2410961> in <module>()
----> 1 AmberPrmtopFile('apocam_seed0.prmtop')

/home/rmcgibbo/miniconda/envs/3.4.2/lib/python3.4/site-packages/simtk/openmm/app/amberprmtopfile.py in __init__(self, file)
     81         # Load the prmtop file
     82 
---> 83         prmtop = amber_file_parser.PrmtopLoader(file)
     84         self._prmtop = prmtop
     85 

/home/rmcgibbo/miniconda/envs/3.4.2/lib/python3.4/site-packages/simtk/openmm/app/internal/amber_file_parser.py in __init__(self, inFilename)
    135                     format = format[index0+1:index1]
    136                     m = FORMAT_RE_PATTERN.search(format)
--> 137                     self._raw_format[self._flags[-1]] = (format, m.group(1), m.group(2), m.group(3), m.group(4))
    138                 elif self._flags \
    139                      and 'TITLE'==self._flags[-1] \

AttributeError: 'NoneType' object has no attribute 'group'
```

Thanks.

swails:
OpenMM does not support chamber-style topology files (and its Fortran Format regex is too simplistic to handle some of the Fortran format specs used in the chamber prmtop file format).

You can try [ParmEd](http://github.com/ParmEd/ParmEd) instead, which has much more complete Amber support (i.e., 10-12 nonbonded support, old-style prmtop file support, and chamber prmtop support).  The beginning of your script should look something like this:

```Python
from chemistry.amber import LoadParm
from simtk.unit import *
from simtk.openmm import *
from simtk.openmm.app import *

# It's best to supply both a prmtop and inpcrd to set the default periodic box vectors
# correctly if there are any
parm = LoadParm('apocam_seed0.prmtop', 'apocam_seed0.inpcrd')

[...]

PR #892, opened 2015-05-02
Better error catch for using chamber prmtops

swails:
Prints out a more helpful error message than a weird regex-based missing
"groups" attribute on NoneType.


swails:
This addresses #887

PR #893, opened 2015-05-04
fix swig build

xantares:


swails:
What version of SWIG does the current version not work for?

xantares:
305

swails:
Weird. 3.0.5 worked for me...

swails:
But I don't know SWIG that well and your changes appear to work on Travis, so thanks!

peastman:
I'm using 3.0.5 and haven't seen any errors with it.  What error was it giving?

xantares:
I've got this this error:
```
./swig_lib/python/extend.i:420: Error: Unknown SWIG preprocessor directive: Look (if this is a block of target language code, delimit it with %{ and %})
```

peastman:
Strange.  No idea why it worked for me and not for you, but if this fixes it, great.

swails:
Maybe @xantares is using a development version of SWIG?  I know that they've been pretty active on fixing Python-related bugs recently, some of which we've been relying on.  So it wouldn't completely surprise me if this is really an issue for the *next* SWIG release (whenever that will be).

xantares:
@swails maybe yes

PR #894, opened 2015-05-04
Added option to add a fixed number of solvent molecules

peastman:


cxhernandez:
This approach looks good!

My only suggestion are perhaps improving the code readability (e.g. there's some rather obfuscated logical statements) and maybe using iterators instead of list construction, in case there are a large number of waters being added to the system.

peastman:
> maybe using iterators instead of list construction

Which lines does this refer to?

cxhernandez:
``` python
+    distToEdge = [min(min(pos-lowerBound), min(upperBound-pos)) for index, pos in addedWaters]
+    sortedIndex = [i[0] for i in sorted(enumerate(distToEdge), key=lambda x: -x[1])]
+    addedWaters = [addedWaters[i] for i in sortedIndex[:numAdded]]
```

Now that I'm looking at it more closely, this is probably not that big of a deal. But I imagine `distToEdge` and `addedWaters` might benefit from this, since they contain floats, right?

swails:
Actually, @cxhernandez raises a good point.  Replace `distToEdge` with a generator and you save memory and execute faster, I think (probably negligibly).  Using the list comp this way forces you to iterate through addedWaters fully once, construct a list of length addedWaters, then iterate through it again.

If you replace the outermost brackets with parentheses, `distToEdge` becomes a generator that only needs to loop through once (and you only use it once).

It's probably not an important optimization, but it's an exceedingly easy one :).

peastman:
Done!  Note, though, that this is a completely negligible optimization.  The time to create that list will be measured in microseconds, and it only happens once.  The total time that will ever be saved by it by all users ever is probably less than it's taken me to type this. :)

swails:
I know :).

cxhernandez:
challenge accepted :stuck_out_tongue:

swails:
haha

```Python
while True:
    # generate modeller
    modeller.addSolvent(numAdded=100000)
```

Come back next week and @peastman will have been wrong!

peastman:
No, then I'll just have wasted even *more* time (yours in this case) by implementing the optimization. :)

PR #896, opened 2015-05-05
do not include reference to source dir

xantares:


peastman:
This is removing potentially useful information.  And the generated file isn't anything that gets distributed.  It only exists inside your build directory.

xantares:
no, it gets installed in /include:
/usr/local/openmm/include/OpenMMSwigHeaders.i

installed files should not reference source dir, please reopen

swails:
> installed files should not reference source dir, please reopen

Installed files should not *depend* on anything in the source directory -- I will agree with you on that (i.e., it should never fail or change behavior based on anything that happens in the source directory after installation).

But this change just affects a comment -- my assertion above does not apply, since comments are not executable code.  What is the motivation besides the statement I quoted above?  If there's none, I agree with @peastman that the information is potential useful and certainly not worth discarding.

xantares:
the motivation is not to get warnings from the package manager with installed files that reference a build dir

swails:
which one?

xantares:
pacman, besides the basename is enough to find the generating script

swails:
ah, Arch linux.  I'll say I'm +0 for this change and step back, since I really don't have a stake in what happens here (the base name *is* enough in this instance to uniquely identify the file that created, and together with the script and printed time stamps will tell you roughly if it's the same version).

On a related note, Gentoo has a phase in their build system that is reserved for modifying the release code of a particular project via patches before building (MacPorts does as well, and I think basically every distro patches the Linux kernel in one way or another).  Is there something analogous in pacman that will let you do this?

xantares:
yes, of course things can be patched afterwards, but it's useless to have the full path from the start

PR #897, opened 2015-05-05
install examples in share

xantares:


peastman:
Why?  What is the point of this, other than making the install directory more complicated?

xantares:
it's because /usr/examples is not a standard linux directory, for non-binaries /usr/share/<pkgname> must be used

peastman:
It doesn't get installed in /usr/examples.  Everything gets installed inside a single folder, which is user configurable.  By default it's /usr/local/openmm, but that can be changed.  For example, if you're on a cluster you probably don't have administrator access, so you need to install it inside your home directory.

xantares:
no, if you write
```
INSTALL(FILES ${EX_ROOT}.c DESTINATION examples)
```
stuff does get installed to /usr/examples provided that you set CMAKE_INSTALL_PREFIX to /usr

peastman:
Why are you setting CMAKE_INSTALL_PREFIX to /usr???

swails:
haha.  I think it was just an example.  It'd be more common to install it to a place like `/usr/local`, `/opt`, or `/opt/local`.  There are also package managers (e.g., portage on Gentoo) that build from source and install to `/usr`, so that's one use-case where OpenMM would install to `/usr`.

Installing to `/usr/local` is not unusual for many libraries and programs, in which case OpenMM's behavior is a little atypical here.

For instance, Gromacs installs their "data" to `$PREFIX/share/gromacs`.  I'd be +1 on this change if it didn't have any negative implications that I don't currently know about.

xantares:
do you know a single linux distro where packages are not supposed to be installed in /usr ?

swails:
Not all package managers install everything to `/usr`.  For example, Gentoo installs the CUDA toolkit to `/opt/cuda`, and then updates the `ld.so.conf` to point to the correct place (so you don't need to augment `LD_LIBRARY_PATH`).  I've never heard of pacman before, though...

peastman:
In any case, OpenMM isn't installed by a package manager.  We support several different installation mechanisms: building from source (which is the only one this PR affects), downloading the precompiled binary and running the install script, or using conda.  For the first two of those, the standard approach is to install everything into a single directory specific to the application, most often inside /usr/local.  Spewing files all over the filesystem is considered very bad form, because there *isn't* a package manager keeping track of them and knowing how to remove all of them again.  And of course conda puts everything inside its own directory structure, so that's a completely different matter.

mpharrigan:
Spewing files all over the filesystem is -- unfortunately -- the standard on unix

http://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

peastman:
Standard for things installed by a package manager, yes.  Not for things installed by their own installers!

xantares:
i dont see why openmm wouldnt be installed by a package manager:
https://aur.archlinux.org/packages/openmm/
it just has to not screw with filesystem standards

jchodera:
I agree that there is benefit to having openmm installable by standard
package managers!

Could a cmake mode or target be added to facilitate this?

mpharrigan:
+1 just moving examples to share/ like it's supposed to be

peastman:
-100 from me. :)  I detest installers that spew things all over the place.  And examples should be a top level folder so it's obvious to users.  Hiding it inside a "share" folder hurts usability for no benefit.

xantares:
you can always provide your own debian mirror or just some .deb, but anyways the debian package manager won't let you install stuff in /usr/examples anyway!

swails:
All package managers are different, so they do not all share the same difficulties.  In particular, Gentoo's package manager would handle OpenMM quite wonderfully (it used to exist by virtue of being an optional Gromacs dependency, but that's since been dropped).

@xantares's efforts here suggest that Arch is another platform where package manager distribution is possible (and indeed, desired).

I don't think the description of "spewing things all over" is representative of what's being proposed here.  While `$OPENMM_PREFIX/examples` may be the first place you'd look, `$OPENMM_PREFIX/share/openmm` may be the first place others would look for non-libs/non-programs.  Neither approach would take longer to acclimate to than the time I spent writing this response ;). A folder being moved from `$OPENMM_PREFIX` to `$OPENMM_PREFIX/share/openmm` doesn't really increase the entropy of our file distribution here.

Again, I'm +0, but I *do* see the benefit to the proposal.

peastman:
You clearly haven't dealt with most of our users. :)  A large fraction of them are biologists with very little computer expertise.  They often are mainly familiar with Windows, and often that's their first experience with OpenMM: installing it on their Windows (or possibly Mac) laptop.  They've never heard of a "share" folder and would have no idea what it meant.  But a folder called "examples" is much more obvious to them.

swails:
I'm persuaded down to 0.  How hard would it be to let `examples` be moved into `share/openmm` via some CMake variable?

leeping:
Hi there,

I think I support Jason's idea of a flexible approach where you could use a CMake variable to move `examples` to `share/openmm`. 

`examples` could still be the default, but this will enable (for example) a package manager to install into `/usr` and the examples into `/usr/share/openmm`.

Having OpenMM be installed by a package manager could be very good in terms of the users we could reach.  I remember Ubuntu's repository had an old version of Gromacs that "just worked" when I was starting to explore molecular dynamics, and once I got more serious I figured out how to build it from source.  I didn't get in touch with the Gromacs community until much later.

Thanks,

- Lee-Ping

swails:
> but this will enable (for example)

I see what you did there :)

peastman:
Could you describe exactly what you would want the alternate layout to be?  #898 calls for also moving the doc folder, so presumably it would be at least the two of those.  Any other changes?

Clearly this option would only be applicable to Linux, since Mac and Windows use completely different organizations.

swails:
@xantares can correct me if I'm wrong, but `examples`, `docs`, and `licenses` are not standard Unix-y folders, so those would all presumably be moved to `${OPENMM_PREFIX}/share/openmm` (so they could use the same variable to control that).

swails:
Everything else can stay as-is

PR #900, opened 2015-05-05
fix mingw build

xantares:


peastman:
I verified this also works under Visual Studio.

PR #902, opened 2015-05-05
[WIP] Implement ternary select(cond,x,y) operation in Lepton

kyleabeauchamp:
Does not work for CUDA yet because I made a bug somewhere, searching now.  It does seem to be OK on reference, however.


peastman:
Let me implement this.  There's a lot of subtlety to it, and it needs to be handled correctly by the various code generators (CUDA, OpenCL, x86).

kyleabeauchamp:
Works for me

PR #903, opened 2015-05-06
Fixed failing test cases.  Increased default minimizer tolerance to 10.

peastman:
See #867.


jchodera:
Should we note something in the manual about how single-precision platforms will have trouble reaching smaller minimization tolerances?

peastman:
First I think we need to study it and find out what the actual limits are for realistic systems.

PR #904, opened 2015-05-06
Implemented select() in custom expressions

peastman:


kyleabeauchamp:
Looks good to me, I was 90% there :)

peastman:
Sorry.  I ended up cheating and just letting the JIT call the standard evaluate() method. :)  Too much trouble for what would probably be negligible performance benefit.

kyleabeauchamp:
Yeah, for my application the `evaluate()` will certainly be dominated by the energy evaluation that just happened.

kyleabeauchamp:
OK, so I integrated the `select()` functionality in my HMC code and things seem to be running, at least on the CUDA and CPU platforms.

kyleabeauchamp:
Silly question: what happens if I add a CustomIntegrator step like the following:

```
u = select(some_variable, 1.0, energy)
```

Is the energy calculated lazily?

peastman:
No, the energy gets computed through a whole series of kernels that get executed before this step begins (in its own kernel).  It has no way to start executing the step, then decide whether the energy will really be needed or not.

PR #908, opened 2015-05-11
Improved accuracy of correction for exclusions with PME

peastman:
See #907


jchodera:
Looks like tests passed!

PR #912, opened 2015-05-12
[WIP] Change erfc approximation of CPU platform to match OpenCL implementation

jchodera:
This change attempts to reduce the error in the erfc approximation utilized by the CPU platform (noted to be significant in #907) to match the OpenCL platform.

Still testing.


platforms/cpu/src/CpuNonbondedForce.cpp
```diff
@@ -473,14 +473,11 @@ void CpuNonbondedForce::getDeltaR(const fvec4& posI, const fvec4& posJ, fvec4& d
...
     // the original source: C. Hastings, Jr., Approximations for Digital Computers (1955).  It has a maximum
-    // error of 3e-7.
+    // absolute error of 1.5e-7.
 
-    float t = 1.0f+(0.0705230784f+(0.0422820123f+(0.0092705272f+(0.0001520143f+(0.0002765672f+0.0000430638f*x)*x)*x)*x)*x)*x;
-    t *= t;
-    t *= t;
-    t *= t;
-    return 1.0f/(t*t);
+    float t = RECIP(1.0f+0.3275911f*x);
```
swails: I don't think `RECIP` is defined on the CPU platform, is it?

jchodera: Whoops!

peastman:
You're missing the exp(-x*x) factor.

Issue #914, opened 2015-05-12
Fixed time simulations


peastman:
A feature that's been requested a few times is the ability to have simulations run for a fixed amount of clock time, then write a checkpoint and exit.  I'd like to get feedback on a few potential designs.

1. A class called SimulationTimeLimiter.  You would add it to a Simulation's list of reporters, and give it either an amount of time to run, or a clock time at which to stop.  When the end time is reached, it would automatically write out a checkpoint (a binary checkpoint file, a serialized State, or both at your choice), then call exit().

2. A method on Simulation to run for a fixed amount of clock time.  You would then be responsible for writing out the checkpoint yourself.

3. Don't add any new features, and just give an example script for doing this.  The script generator could also provide this as an option.

kyleabeauchamp:
So @jchodera and I have used both fixed time blocks and fixed step blocks to converge various properties to the desired accuracy.  Two examples below:

https://github.com/kyleabeauchamp/HMCNotes/blob/master/sandbox/lb_loader.py#L44

https://github.com/choderalab/LiquidBenchmark/blob/master/src/simulation/liquid_tools.py#L165

jchodera:
I actually like suggestion 2, where the `Simulation` class has a method to run for a fixed amount of wall clock time (where it uses a user-settable step block size under the hood), e.g.
```
simulation.runForFixedTime(10 * unit.second, checkIntervalSteps=100)
```
Later, we could also add something like what we have used for `LiquidBenchmark` where the user sets a target statistical uncertainty.

mpharrigan:
Can we do (2) with an auto-checkpoint at the end

```
def step_for(5 * unit.hour):
    while (time_isnt_up()):
        self.step(100)
    self.write_checkpoint()
```

rmcgibbo:
I could imagine wanting to write a script like

```
if not exists('checkpoint.mm')
  [ ... set up simulation like I currently do ... ]
else:
   simulation.loadCheckpoint('checkpoint.mm')

simulation.runFor(12*hours)
simulation.writeCheckpoint('checkpoint.mm')
```

jchodera:
What about adding an option for auto-checkpointing, allowing us to write either
```
simulation.runFor(12*hours)
simulation.writeCheckpoint('checkpoint.mm')
```
or
```
simulation.runFor(12*hours, checkpoint='checkpoint.mm')
```
We could even have it auto-checkpoint at various intervals:
```
[...]

PR #918, opened 2015-05-13
Added checkpointing functions to Simulation

peastman:
This is the first piece of #914.  Comments would be great.

### DIFF
--- a/wrappers/python/simtk/openmm/app/simulation.py
+++ b/wrappers/python/simtk/openmm/app/simulation.py
@@ -6,7 +6,7 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
-Portions copyright (c) 2012 Stanford University and the Authors.
+Portions copyright (c) 2012-2015 Stanford University and the Authors.
 Authors: Peter Eastman
 Contributors:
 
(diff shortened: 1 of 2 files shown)


swails:
Nice additions to the `Simulation` API.

mpharrigan:
Nice! 

Can we have a static method e.g.
```python
simulation = Simulation.fromCheckpoint("checkpoint.mm")
```

kyleabeauchamp:
I agree with the idea of having super convenient ways to load / save things with as few keystrokes as possible.

jchodera:
Looking good!

swails:
@mpharrigan -- `classmethod` you mean, yes? +1 for the alternate XML-based constructor.

mpharrigan:
@swails whatever python likes to call them :)

swails:
Well Python has both.  Alternate constructors are `classmethod`s (they can be implemented as a `staticmethod`, but not in an inheritance-friendly way).

peastman:
It takes a lot more than a checkpoint to create a simulation.  For that, you also need a System, a Topology, and an Integrator.

swails:
> It takes a lot more than a checkpoint to create a simulation. For that, you also need a System, a Topology, and an Integrator.

I know, but all three of those can be exported to XML files, right?  So the `from_xml` constructor could take 3 XML files and an optional checkpoint file to instantiate it, yes?

peastman:
> I know, but all three of those can be exported to XML files, right?

Topology can't be.

swails:
On Thu, May 14, 2015 at 1:57 PM, peastman <notifications@github.com> wrote:

> I know, but all three of those can be exported to XML files, right?
>
> Topology can't be.
> ​
>
> ​
​Bummer.​

PR #919, opened 2015-05-14
Implemented runForClockTime()

### LINKED ISSUE #914: Fixed time simulations

peastman:
This fixes #914.

### DIFF
--- a/wrappers/python/simtk/openmm/app/simulation.py
+++ b/wrappers/python/simtk/openmm/app/simulation.py
@@ -33,6 +33,8 @@ __version__ = "1.0"
 
 import simtk.openmm as mm
 import simtk.unit as unit
+import sys
+from datetime import datetime, timedelta
 
 class Simulation(object):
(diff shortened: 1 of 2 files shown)


mpharrigan:
gromacs runs for 0.99 * number of hours, so it has time to cleanup. Is this something we want here?

PR #921, opened 2015-05-15
Switched to a new Windows pthreads library

peastman:
This one is implemented entirely in source code (mostly one header file), rather than a separate library.  That eliminates all the nightmares I've been going through with all the different types of Windows libraries: 32 vs. 64 bit, dynamic vs. static, /MT vs. /MD.


mpharrigan:
Nice!

PR #922, opened 2015-05-18
Fix Gromacs #define replacement and add support for Gromacs 5

swails:
In my opinion, this is a severe bug.

Gromacs uses the preprocessor to replace previously #define'd tokens in strings.
This is used, for instance, to override certain parameters without having to
introduce new file types.  This is how Gromacs implemented ff99SB-ILDN (i.e., by
simply replacing the chi torsion for isoleucine, leucine, aspartate, and
asparagine with a define token).  So OpenMM *was* missing some in-line
parameters that were defined.

The immediate consequence I see is that whenever someone tried to run
ff99SB-ILDN with OpenMM, what they were *really* getting was ff99SB. Not sure
where else, if anywhere, this kind of thing was happening.

The "Gromacs 5" support is simply looking for the program `gmx` instead of `mdrun` or anything.  All Gromacs programs have been moved to "subcommands" of the `gmx` program.


leeping:
I think it may help to add some tests into OpenMM for setting up / parameterizing the system from different sources.  That is to say, a system like this could have been set up using Gromacs `gro+top` files, AMBER `prmtop`, `pdb+ffxml` and so on; if they're not giving the same energy then something is wrong somewhere.

swails:
> I think it may help to add some tests into OpenMM for setting up / parameterizing the system from different sources. That is to say, a system like this could have been set up using Gromacs gro+top files, AMBER prmtop, pdb+ffxml and so on; if they're not giving the same energy then something is wrong somewhere.

True.  This is the kind of target I was thinking about with the "validation" suite we were talking about earlier.  For this PR, the test I added was strictly to make sure that OpenMM was reading `#define`s the way `grompp` reads them.

peastman:
That looks great.  Thanks!

PR #929, opened 2015-05-19
Moved the MTS and aMD integrators into the main repository

peastman:


jchodera:
@kyleabeauchamp and @peastman: Do you want to discuss the API for the MTS integrator?

swails:
I'd personally prefer to see these integrators imported into the `simtk.openmm` namespace with the rest of the integrators rather than to `simtk.openmm.app`.

jchodera:
What about `simtk.openmm.integrators`?  Or do you want to use them transparently the same way as standard C++ integrators?

swails:
> Or do you want to use them transparently the same way as standard C++ integrators?

This one.  Why are `AMDIntegrator` and `MTSIntegrator` special?  Because they're written in Python?  Why would a user care?

peastman:
I want to keep simtk.openmm as strictly the Python wrapping of the C++ API.  If a class isn't part of the C++ API, it should go somewhere else.

jlmaccal:
Maybe create `smirk.openmm.app.integrators` (or similar) and import them all there? I think it would be strange to have to look for integrators in multiple locations depending on their implementation details.

swails:
Then what about having an `integrators` namespace that imports all of the integrators (both Python and C++)?

I think grouping integrators together is a more sensible grouping than "these are C++ and these are Python", but I think I'm starting to bikeshed here, so I'll just leave it at this :).

peastman:
Personally, I don't like the idea of creating yet another namespace that people have to import.  And I certainly don't want to split the C++ wrapper into multiple modules.  But perhaps we can reach a consensus.  Which of the following three locations do you like best for these two classes?

- simtk.openmm
- simtk.openmm.app
- simtk.openmm.integrators

rmcgibbo:
I vote for `simtk.openmm.app`

swails:
I vote for wherever all of the other integrators are (currently that would be `simtk.openmm`).

But what I was talking about was all of the integrators from C++ be imported into `openmm.integrators` and the 2 you added here be put in that namespace as well (so no breakage of backwards compatibility).  Then you would have something like:

```Python
from simtk.openmm import *
from simtk.unit import *

langevin = integrators.LangevinIntegrator(...)
# or
mts = integrators.MTSIntegrator(...)
# or ...
```

I don't like another namespace either, but I like it slightly better than having "some integrators in one namespace and other integrators in another".  Sure, you can document it, and sure it makes sense to those of us that know the origins of how they're implemented.  But for those users that "don't know don't care", it's more details about the API they have to remember that is avoidable IMO. (And that which they can't remember they have to spend time to go back to the docs and look up).

rmcgibbo:
-1 on adding a new `integrators` namespace.

leeping:
My votes are the same as Robert's.

swails:
To summarize, my ranking is:

1. `simtk.openmm`
2. `simtk.openmm.integrators`
3. `simtk.openmm.app`

I share @rmcgibbo and @leeping's sentiments about namespace proliferation, but think reducing the cognitive load of the API is also important.

leeping:
Would it be bad or confusing to put "copies" of the `simtk.openmm` integrators into the `simtk.openmm.app` namespace?

jchodera:
You can simply import those into the `simtk.openmm.app` namespace as well.  Is that what you mean?

jchodera:
There are advantages to having an `integrators` namespace:
* It makes it easy to test or experiment with all integrators without figuring out how to enumerate all subclasses (and subclasses of subclasses...) of `Integrator` aggregated over all namespaces (which is a pain)
* It keeps all of them in one place for looking at documentation: `help(simtk.openmm.integrators)` makes considerably more sense than scrolling or searching through pages and pages of unrelated crap; the API documentation would be neatly organized by integrator already

By comparison, I don't think segregating logically related things into distinct namespaces is overly burdensome.  It can add just a line (`from simtk.openmm.integrators import *`) to your code.

swails:
> Would it be bad or confusing to put "copies" of the simtk.openmm integrators into the simtk.openmm.app namespace?

To me it would be a little confusing. Someone once explained to me that `simtk.openmm` is the home of the core functionality in OpenMM -- `Force`, `Context`, `Integrator`, etc.  `simtk.openmm.app` is where parts of the application layer live.  This would be what makes OpenMM an *application* as opposed to just a collection of library functions.

So this is where stuff like `Modeller`, `ForceField`, and `Simulation` live.  I think the main goal of our namespaces should be to reduce cognitive load so that people can work quickly and efficiently with OpenMM once they become familiar with its design.  To me, `openmm` and `openmm.app` help compartmentalize the different pieces in a way that makes sense.  I think importing all integrators into `simtk.openmm.app` muddies those waters, but I might be biased by the original explanation I got a year or two ago (by Robert, IIRC).

jlmaccal:
I personally wouldn't find it confusing, but I'm not a huge fan of namespace proliferation -- it's very java-esque.

My preferences would be:

1. Put the new integrators in `simtk.openmm`. As @swails says, this is where all of the low-level building blocks go.
2. Put *all* integrators into `simtk.openmm.integrators` or similar.
3. Put *all* integrators into `smirk.openmm.app`.
4. Put the new python-based integrators in `simtk.openmm.app` and keep the old ones in `simtk.openmm`.

The last option is by far the worst, IMO. It makes no difference to users how the integrators are implemented, which is the primary argument for 4.

leeping:
@swails : Although my vote is the same as Robert's, that's only because I saw moving the new integrators into `simtk.openmm` as breaking some kind of boundary because the former contains SWIG-generated Python wrappers to C++ code and the latter contains hand-written Python code.  I don't know if mixing the two would create any programming or organizational challenges.  That said, I agree with everyone that it would be considerably less confusing if we could get all of the integrators from a single namespace.

I think @jchodera's argument is very reasonable, but I remember learning to use OpenMM myself and I had questions like "Why do I need three import statements?"  "Where do I find something, and why is it there?"  I understood after a while, but I think that having multiple namespaces does make things harder to use for a new user.

swails:
> that's only because I saw moving the new integrators into simtk.openmm as breaking some kind of boundary because the former contains SWIG-generated Python wrappers to C++ code and the latter contains human-written Python code. I don't know if mixing the two would create any programming or organizational challenges.

I don't think it will (we can add custom code to the Swig input to add whatever we want to the generated `openmm.py` file -- we already do this with the `State` class).

> I understood after a while, but I think that having more namespaces does make things harder to use for a new user.

I agree with this, but there is a trade-off.  As time went on, this separation made my life easier (particularly as I began to understand it).  If I was unsure about whether a class existed, or perhaps what it was called, understanding the namespaces gave me a good idea about where to go to look for things.

We don't want a barrier too high to get people started, but at the same time I think we're better off catering more towards the users that will seriously use OpenMM in their daily work (even if they aren't "power" users).

peastman:
> I think the main goal of our namespaces should be to reduce cognitive load so that people can work quickly and efficiently with OpenMM once they become familiar with its design.

Strongly agree, with the addition that we should try to reduce the cognitive load both *while they are learning* and *after they are familiar with it*.

In practical terms for an experienced user, I don't think it makes much difference.  The situations where it actually makes your code simpler to have all integrators in a separate module are few and far between.  Far more often it will just mean an extra import at the top of your program.  And having a single class appear in two different places is definitely confusing and increases cognitive load.

swails:
> Strongly agree, with the addition that we should try to reduce the cognitive load both while they are learning and after they are familiar with it.

My main point was at @leeping who said more than 1 namespace was a complicating factor compared to having everything dumped in the same place.  It might temporarily increase the cognitive load to have 3 imports, but the benefits it has to people that *actually* start using it outweighs the initial confusion (IMO).

> And having a single class appear in two different places is definitely confusing and increases cognitive load.

Except we do this almost everywhere:  `AmberPrmtopFile` is in `simtk.openmm.app` and `simtk.openmm.app.amberprmtopfile`, and that's fine.  We also have `HBonds` as part of `simtk.openmm.app`, `simtk.openmm.app.forcefield`, and `HCT` in `simtk.openmm.app.amberprmtopfile` and `simtk.openmm.app`.  We simply document the `openmm`, `app`, and `unit` packages as the public-facing, documented namespaces and don't worry if people poke around the other files.

It's fine to have `AMDIntegrator` in `simtk.openmm.app.amdintegrator` *and* `simtk.openmm`, IMO (and likewise with the MTS integrator).  I'd personally rather not *also* import it into `simtk.openmm.app`, but that's easy to achieve (i.e., just don't import it in ``app/__init__.py``)

peastman:
I *think* these are people's votes:

Jason: simtk.openmm
Robert: simtk.openmm.app
Justin: simtk.openmm
Lee-Ping: Is your vote for simtk.openmm or simtk.openmm.app?  I couldn't quite tell from your latest post.
Anyone else?

swails:
I think @jchodera voted for simtk.openmm.integrators, and I think @leeping's answer was trying to weigh the technical challenges of putting it in `simtk.openmm` as opposed to `simtk.openmm.app` -- can you comment on that?

(Oh, and you forgot your own vote)

leeping:
Hi Peter, I'll vote for `simtk.openmm`. And yes you should vote. :)

peastman:
For myself, I could go with either simtk.openmm or simtk.openmm.app.  I can see the arguments for either one.  I dislike simtk.openmm.integrators.

It sounds like simtk.openmm is the clear favorite.  I'll move them.

peastman:
Moved.

wrappers/python/simtk/openmm/__init__.py
```diff
@@ -20,6 +20,8 @@
 
 from simtk.openmm.openmm import *
 from simtk.openmm.vec3 import Vec3
+from mtsintegrator import MTSIntegrator
+from amd import AMDIntegrator, AMDForceGroupIntegrator, DualAMDIntegrator
```
swails:
```Python
from simtk.openmm.mtsintegrator import MTSIntegrator
from simtk.openmm.amd import AMDIntegrator, AMDForceGroupIntegrator, DualAMDIntegrator
```

instead?

peastman: The relative import works fine.  I don't know why the two lines above it are using absolute.  There's no particular need for it.

swails: I thought relative imports were removed from Py3?

swails: *implicit* relative imports that is.  `from .mtsintegrator import MTSIntegrator` I think is valid.  Maybe I misunderstood PEP 328.

peastman: We do this in other places and have no trouble with Python 3.  See https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/__init__.py.

swails: Ah, OK, this is something `2to3` fixes (indeed, as-is implicit relative imports raise an exception in Python 3).

There's a little concern that if anybody had downloaded these integrators previously and have them lying around somewhere in `sys.path`, this import will mistakenly pick them up from there rather than this package.

peastman: Fair point.  I just switched them to absolute.

swails: > We do this in other places and have no trouble with Python 3. See https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/__init__.py.

Ah, right you are.

PR #939, opened 2015-05-30
[WIP] Attempting to implement conditional termination

kyleabeauchamp:
This is woefully incomplete at present, but might be useful to get the discussion going.

### DIFF
--- a/openmmapi/include/openmm/CustomIntegrator.h
+++ b/openmmapi/include/openmm/CustomIntegrator.h
@@ -238,7 +238,11 @@ public:
         /**
          * Allow Forces to update the context state.
          */
-        UpdateContextState = 5
+        UpdateContextState = 5,
+        /**
+         * Conditionally terminate the integration step based on an expression.
+         */
+        ConditionalTermination = 6
     };
     /**
      * Create a CustomIntegrator.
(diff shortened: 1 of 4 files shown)


kyleabeauchamp:
Hmm, so a design question that comes up is whether it's possible for this action to "share code and / or kernels" with the `computeGlobal()` pathway.  In practice, I imagine this is quite similar to computeGlobal but with an additional action taken after the kernel is executed.

jchodera:
I wonder if the simplest possible way to implement something that makes use of the `computeGlobal()` code is to explicitly use `addComputeGlobal()` to compute the value of the conditional in an explicitly defined global variable.

That is, instead of
```
integrator.addConditionalTermination(expression)
```
we have the user write
```
integrator.addComputeGlobal(variable, expression)
integrator.addConditionalTermination(variable)
```
That way, you only need to implement a very simple termination kernel.

jchodera:
That may not be as convenient for the user, of course!

kyleabeauchamp:
I was just thinking that as well--that might be the simplest API addition to openmm.

jchodera:
I'm not sure it is any simpler from the API perspective, but it sounds a
hell of a lot easier to implement.

peastman:
I'd actually prefer if we could back off on this.  This isn't something we're ready to start implementing yet.  It needs a lot of design first, and I'm not sure if this is even the right approach to be taking.  For any new feature, step 1 is always to identify the requirements, and we haven't done that yet.  Perhaps you could open a new issue where we could start doing that?

kyleabeauchamp:
OK, so I think the discussion of this is now in #941.  Furthermore, for my immediate needs I may have found a workaround that doesn't require modification of OpenMM.  

If I decide otherwise, I will re-open this or a new pull.

PR #944, opened 2015-06-02
Ported OpenCL FFT to CUDA

peastman:


kyleabeauchamp:
FYI I updated my local installation and will be testing your changes on my GTX 960 CUDA 6.5 linux box.

peastman:
I'd be interested to know whether it's faster with or without the stream.  That's set at line 1706 of CudaKernels.cpp.

kyleabeauchamp:
```
python benchmark.py --platform=CUDA --precision=mixed --test=pme --pme-cutoff=1.0
Platform: CUDA
Precision: mixed

Test: pme (cutoff=1)
Step Size: 2 fs
Integrated 26845 steps in 57.4312 seconds
80.7717 ns/day
```

After my modification:

```
git diff
diff --git a/platforms/cuda/src/CudaKernels.cpp b/platforms/cuda/src/CudaKernels.cpp
index 7e10fc1..6a205be 100644
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -1703,7 +1703,8 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
                 
                 char deviceName[100];
                 cuDeviceGetName(deviceName, 100, cu.getDevice());
-                usePmeStream = (string(deviceName) != "GeForce GTX 980"); // Using a separate stream is slower on GTX 980
+                //usePmeStream = (string(deviceName) != "GeForce GTX 980"); // Using a separate stream is slower on GTX 980
+                usePmeStream = false;
                 if (usePmeStream) {
                     cuStreamCreate(&pmeStream, CU_STREAM_NON_BLOCKING);
                     if (useCudaFFT) {


python benchmark.py --platform=CUDA --precision=mixed --test=pme --pme-cutoff=1.0
Platform: CUDA
Precision: mixed

Test: pme (cutoff=1)
Step Size: 2 fs
Integrated 24266 steps in 59.404 seconds
70.5873 ns/day
```

PR #950, opened 2015-06-04
Minor enhancements to Topology

swails:
- Add an informative ``__repr__`` to Topology
- Add ``__len__`` to Chain and Residue to quickly tell how many residues or atoms are in them, respectively

Now, instead of getting something like this:

```
>>> app.PDBFile('4lzt.pdb').topology
<simtk.openmm.app.topology.Topology object at 0x7fdb419a2050>
```

You get something like this:

```
>>> app.PDBFile('4lzt.pdb').topology
<Topology; 2 chains, 274 residues, 1164 atoms, 1043 bonds>
```

### DIFF
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -59,6 +59,14 @@ class Topology(object):
         self._bonds = []
         self._periodicBoxVectors = None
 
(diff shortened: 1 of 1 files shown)


swails:
These changes are trivial and mainly just useful for interactive Python introspection (which I happen to be doing right now :)).

PR #953, opened 2015-06-05
Make sure RBTorsion parameters are returned with the correct units.

swails:


swails:
BTW -- I need this for ParmEd/ParmEd#170 -- I have a clumsy workaround for the time being that I'd like to get rid of.

jchodera:
Yikes!  This seemed like an important oversight.  Thanks!

PR #954, opened 2015-06-08
Fix clang3.6 tautological-undefined-compare warning.

chrisdembia:
The warnings were:

```
/Users/chris/nmblprojects/psim/pgait/opensim/opensim-core/Vendors/lepton/src/ExpressionTreeNode.cpp:62:85:
warning:
      reference cannot be bound to dereferenced null pointer in well-defined
      C++ code;
      comparison may be assumed to always evaluate to false
      [-Wtautological-undefined-compare]
    ...ExpressionTreeNode& node) : operation(&node.getOperation() == NULL ? NULL : nod...
                                              ^~~~~~~~~~~~~~~~~~~    ~~~~
/Users/chris/nmblprojects/psim/pgait/opensim/opensim-core/Vendors/lepton/include/lepton/ExpressionTreeNode.h:93:22: note:
        'getOperation' returns a reference
    const Operation& getOperation() const;
                     ^
```

```
/Users/chris/nmblprojects/psim/pgait/opensim/opensim-core/Vendors/lepton/src/ParsedExpression.cpp:49:10:
warning:
      reference cannot be bound to dereferenced null pointer in well-defined
      C++ code;
      comparison may be assumed to always evaluate to false
      [-Wtautological-undefined-compare]
    if (&rootNode.getOperation() == NULL)
         ^~~~~~~~~~~~~~~~~~~~~~~    ~~~~
/Users/chris/nmblprojects/psim/pgait/opensim/opensim-core/Vendors/lepton/include/lepton/ExpressionTreeNode.h:93:22: note:
        'getOperation' returns a reference
    const Operation& getOperation() const;
                     ^
```

### DIFF
--- a/libraries/lepton/src/ExpressionTreeNode.cpp
+++ b/libraries/lepton/src/ExpressionTreeNode.cpp
@@ -59,7 +59,7 @@ ExpressionTreeNode::ExpressionTreeNode(Operation* operation) : operation(operati
         throw Exception("wrong number of arguments to function: "+operation->getName());
 }
 
-ExpressionTreeNode::ExpressionTreeNode(const ExpressionTreeNode& node) : operation(&node.getOperation() == NULL ? NULL : node.getOperation().clone()), children(node.getChildren()) {
+ExpressionTreeNode::ExpressionTreeNode(const ExpressionTreeNode& node) : operation(node.getOperation().clone()), children(node.getChildren()) {
 }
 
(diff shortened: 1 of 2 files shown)


swails:
Travis is segfaulting on all of the custom forces, so I don't think this change quite does the trick.

peastman:
clang is wrong in this case.  That check really is required.

PR #959, opened 2015-06-10
Add check for null context inside integrator.step()

### LINKED ISSUE #957: Check context before integrator.step()?
So the following code causes a segmentation fault.  According to the source (https://github.com/pandegroup/openmm/blob/master/openmmapi/src/VerletIntegrator.cpp#L72), shouldn't it be easy enough to check that the context pointer is non-null before the `for` loop?

```
import simtk.openmm as mm
integrator = mm.VerletIntegrator(1.0)
integrator.step(1)
```

kyleabeauchamp:
Fixes #957

### DIFF
--- a/openmmapi/src/BrownianIntegrator.cpp
+++ b/openmmapi/src/BrownianIntegrator.cpp
@@ -72,6 +72,8 @@ double BrownianIntegrator::computeKineticEnergy() {
 }
 
 void BrownianIntegrator::step(int steps) {
+    if (context == NULL)
+        throw OpenMMException("This Integrator is not bound to a context!");  
     for (int i = 0; i < steps; ++i) {
         context->updateContextState();
         context->calcForcesAndEnergy(true, false);
--- a/openmmapi/src/CustomIntegrator.cpp
+++ b/openmmapi/src/CustomIntegrator.cpp
@@ -94,6 +94,8 @@ double CustomIntegrator::computeKineticEnergy() {
 }
 
 void CustomIntegrator::step(int steps) {
+    if (context == NULL)
+        throw OpenMMException("This Integrator is not bound to a context!");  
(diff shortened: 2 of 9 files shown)


kyleabeauchamp:
Test code:

```
import simtk.openmm as mm
import openmmtools

integrator = mm.VerletIntegrator(1.0)
integrator.step(1)

integrator = mm.VariableVerletIntegrator(1.0)
integrator.step(1)

integrator = mm.LangevinIntegrator(1.0, 1.0, 1.0)
integrator.step(1)

integrator = mm.VariableLangevinIntegrator(1.0, 1.0, 1.0)
integrator.step(1)

integrator = mm.VariableVerletIntegrator(1.0)
integrator.step(1)

integrator = openmmtools.integrators.HMCIntegrator()
integrator.step(1)
```

peastman:
Looks good, but you're missing several integrators:

BrownianIntegrator
RPMDIntegrator
DrudeSCFIntegrator
DrudeLangevinIntegrator

jchodera:
What about moving the check to the base `Integrator::step()` and having all the subclasses call this first?

kyleabeauchamp:
If we're going to do that, then we should move *all* the shared code to the base class.  We can do this in a separate PR, as that breaks the current code layout.

kyleabeauchamp:
Added remaining integrators

peastman:
I think the test failure is a stochastic test that just happened to fail twice:

```
72: Test command: /home/travis/build/pandegroup/openmm/TestReferenceDrudeLangevinIntegrator
72: Test timeout computed to be: 1500
72: exception: Assertion failure at TestReferenceDrudeLangevinIntegrator.cpp:174.  Expected 203.333, found 207.612 (This test is stochastic and may occasionally fail)
```

PR #960, opened 2015-06-11
[WIP] Added test script for packaging

jchodera:
Just beginning to add some automation for testing packaging.  Not ready to merge yet.

### DIFF
--- a/devtools/packaging/scripts/osx/test.sh
+++ b/devtools/packaging/scripts/osx/test.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# Test Mac OS X distribution package.
+# Package must have already been downloaded.
+
+# Set relative workspace path.
+export WORKSPACE=`pwd`
+
+# Add conda binaries to path.
+PATH=$WORKSPACE/miniconda/bin:$PATH
+
+# Get Python executable.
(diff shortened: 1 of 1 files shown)


peastman:
What's the status of this PR?  Has it been abandoned, or do you still plan to come back to it?

jchodera:
I'm happy to pick it up again if you think it would be useful to automate testing of the binary package.  If not, we can drop it.

peastman:
I'm not entirely clear on what the intended feature is.  When would this script be run, and on what computer?  Will there be similar tests for other OSes?

jchodera:
The original idea was that this would be run by the Jenkins build system immediately after building the binary package, so that we didn't have to have you open it up and test it manually.

peastman:
If you want to have it do that, I have no objection.  It won't eliminate the need for manual testing, though.  There are a lot of situations where a binary works on the machine it was built on, but fails on other machines (especially if they have a different OS version).

So it's up to you whether you want to finish this or just close it.

swails:
I'm not sure why this branch triggered a build.  But the machine it's running on crashed.

peastman:
Closing abandoned PRs.

PR #961, opened 2015-06-12
Updates to documentation for 6.3

peastman:


jchodera:
I still think it would be very useful to have the "release notes" be part of the distro docs.

Issue #964, opened 2015-06-13
Add TravisCI integration


jcoffland:
TravisCI is free for Open-Source apps.  By adding a ```.travis.yml``` and setting up an account you could have automated Linux builds which could check that multiple build configurations are working.  For example, I cannot build 6.2 with out shared libs enabled.

jchodera:
Travis-CI is already enabled and active for OpenMM.  All PRs are screened
by Travis before merging.  There is even a badge on the README.md that
takes you to the Travis build report.

Is there something else specific to shared libraries here?

swails:
Well you can use multiple environments to test different types of builds (e.g., I test Python 2.7 w/ and w/out numpy, Python 3.3, and Python 3.4 with ParmEd: https://travis-ci.org/ParmEd/ParmEd).

But @jcoffland -- OpenMM *does* already use Travis, but obviously not for static lib builds.

jchodera:
Is the suggestion that we should also be testing static lib builds for
Linux on Travis?  I hadn't realized openmm actually supported static
builds...

swails:
![screen shot 2015-06-13 at 4 10 00 am](https://cloud.githubusercontent.com/assets/779022/8143578/1bb75332-1182-11e5-93a2-1256d5728bda.png)

> I hadn't realized openmm actually supported static
> builds...

curses-cmake says it should, but based on @jcoffland's post it seems not to :).

peastman:
In 6.3 it should work.  6.2 needed changes to make it work, largely due to the pthreads library.

jchodera:
OK, we can have travis attempt to build both static and dynamic versions in a build matrix.  I've opened a PR in #966 to test this.

jchodera:
Make that PR #967, since I screwed up #966.

jchodera:
OK, I've got this working in #967.

swails:
I think this can be closed.

jchodera:
I think only @jcoffland and @peastman can close this.

swails:
Yea, I was just pinging them.

PR #966, opened 2015-06-13
[WIP] Have travis build both static and dynamic libraries.

jchodera:
Working on addressing #964 by having Travis use a build matrix to build both statically-linked and dynamically-linked forms of OpenMM.


mpharrigan:
We shouldn't include that makefile in git if it is autogenerated.

Some paths in your swig file got replaced with jchodera-specific ones

jchodera:
Good catch.  Not sure how those got included.

jchodera:
Argh.  Let me try again.

PR #967, opened 2015-06-13
Add travis build matrix for static and dynamic builds

jchodera:
This attempts to address #964 by having travis do both static and dynamic library builds.

Replacement for PR #966 which I screwed up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ compiler:
   - clang
 
 env:
+  matrix:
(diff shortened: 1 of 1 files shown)


jchodera:
Thanks for pointing out the issues with #966, @mpharrigan!

jchodera:
This seems to work now.

peastman:
Sorry for not merging sooner.  I missed that it wasn't still in progress.

PR #972, opened 2015-06-19
A large number of unit cleanups and fix numpy wart

swails:
The main thing that's done here is that numpy arrays can be given units via the
`*` operator, rather than relying on the Quantity constructor

e.g.,

```Python
>>> import numpy as np
>>> np.zeros(10) * u.angstroms
Quantity(value=array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.]), unit=angstrom)
```

yay!

Still needs a test

### DIFF
--- a/docs-source/usersguide/library.rst
+++ b/docs-source/usersguide/library.rst
@@ -2712,10 +2712,6 @@ multiply operator (‘*’) or the explicit Quantity constructor:
     # or more verbosely
     bond_length = Quantity(value=1.53, unit=nanometer)
 
-When working with Numpy arrays you *must* use the explicit constructor.  You cannot
-multiply them by a unit, because the Numpy array class overloads the multiply operator.
-
-
 Arithmetic with units
 ---------------------
 
--- a/docs-source/usersguide/library.rst
+++ b/docs-source/usersguide/library.rst
@@ -2865,7 +2865,7 @@ in and out.
     [[10.0, 20.0, 30.0], [40.0, 50.0, 60.0]]
 
     >>> import numpy
-    >>> a = Quantity(numpy.array([1,2,3]), centimeter)
+    >>> a = numpy.array([1,2,3]) * centimeter
     >>> print(a)
     [1 2 3] cm
     >>> print(a/millimeter)
--- a/wrappers/python/simtk/unit/basedimension.py
+++ b/wrappers/python/simtk/unit/basedimension.py
@@ -1,6 +1,4 @@
 #!/bin/env python
-
-
 """
 Module simtk.unit.basedimension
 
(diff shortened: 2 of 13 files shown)


peastman:
That looks great!  The manual also needs to be updated.  Search for references to "numpy" in library.rst.

swails:
> That looks great! The manual also needs to be updated. Search for references to "numpy" in library.rst.

Thanks for the pointer on where to look.  I also need to push my test.  The key here is, for anyone that cares, the `__array_priority__` class attribute.  I just set it to 100 for `Unit`, `BaseUnit`, and `ScaledUnit`.  Since this is larger than `np.ndarray`'s priority, control for ufuncs is given up to the unit classes instead.

swails:
Test added and docs updated.  This should be good to merge when Travis finishes (and I'd really like to get this into 6.3 for ParmEd).

peastman:
>  I just set it to 100 for Unit, BaseUnit, and ScaledUnit.

Should Quantity do that too, so you can multiply an array by a Quantity?

peastman:
There's another paragraph in the manual that needs to be changed (or really just removed):

> When working with Numpy arrays you must use the explicit constructor. You cannot multiply them by a unit, because the Numpy array class overloads the multiply operator.

swails:
> Should Quantity do that too, so you can multiply an array by a Quantity?

Yes!  I hadn't even thought about that use-case.

swails:
OK.  Done, with test added.

swails:
Ugh.  The test failure is a result of a bug in numpy that wasn't fixed until version 1.8, and I really don't have time to bisect this down and find a workaround (if there even *is* one).  The tests all pass with numpy versions 1.8 and above.

Any ideas what we should do here?

swails:
BTW -- the Ubuntu 12.04 numpy from aptitude is 1.6.1.

swails:
A couple facts to consider:

- This PR *does* fix unit assigning to numpy arrays via multiplication for all versions of Python.
- This PR *does* fix unit assignment in when multiplying an ndarray by another Quantity that is an ndarray in numpy 1.8+
- This PR keeps the existing buggy behavior in that units are ignored when multiplying a ndarray by a Quantity

So if a user has numpy 1.7 or earlier, this PR adds the convenience of being able to make numpy Quantities via `np.array([1, 2, 3, 4]) * centimeter`, but does not fix the bug described above.  If a user has numpy 1.8 or better, this PR adds the above-mentioned convenience *and* fixes the bug described above.  In both cases, I think this PR is a clear improvement with no caveats.

So I think it's the way to go.  But what should we do about the failing test?  Just run it for compatible versions of numpy?  Take it out altogether?  Switch to using conda's Python in the test suite (and add a Python 3 build env)?

peastman:
> Just run it for compatible versions of numpy?

That sounds like a reasonable solution to me.  Anyone else have opinions on it?

swails:
That was an easy solution, so I went ahead and added the change.  Opinions are still welcome, of course.

swails:
OK, this is ready to go for me.  Nobody else seems to care too much here and this doesn't introduce any new bugs (that I'm aware of).

jchodera:
I care!  This is awesome.  +1 for getting this stuff in ASAP.

peastman:
That's great.  Thanks!

swails:
> I care! This is awesome. +1 for getting this stuff in ASAP.

I more meant that nobody seems to care about how the numpy<1.8 issues were handled for Travis.  It's not a big deal IMO.

PR #982, opened 2015-06-22
Compiling on Windows

rmcgibbo:
1. Compiling with x64 VS2010 compilers, I got an error with two `#defines` already having been defined in `pthreads.h`.
2. Extend #948 to search for other usual suspects paths for OpenCL. This isn't strictly necessary, but it smooths things out a little.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -530,4 +530,11 @@ IF(BUILD_TESTING)
 ENDIF(BUILD_TESTING)
 ADD_SUBDIRECTORY(examples)
 
+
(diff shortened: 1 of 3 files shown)


peastman:
I don't think we should be hardcoding the path "C:/Program Files (x86)/AMD APP SDK/3.0-0-Beta".  I mean, aside from the fact that having to hardcode a different path for every version isn't maintainable, we certainly shouldn't be hardcoding paths to *beta* versions.  Also, what if they installed the 64 bit version, so it's in "C:\Program Files" instead?

rmcgibbo:
Sure, I can get rid of that.

PR #985, opened 2015-06-23
Fix Python 3 support in unit package.

swails:


mpharrigan:
For my edification, why is this change needed? I know they changed `__nonzero__` to __bool__`, but what's the harm in leaving it?

swails:
This failure was what spurred that change: https://travis-ci.org/ParmEd/ParmEd/jobs/67892164#L2691 (ParmEd uses the `simtk.unit` package when available, and its own when not).  Not sure why I thought that deleting `__nonzero__` was the correct fix (I'm not sure why my local test showed that it *worked*, when I don't think it should have).  But when you're doing stuff with `__getattr__` like what the implementation in `Quantity` is doing, it can be tricky to figure out exactly what *should* happen in many cases (particularly when you factor in the fact that `2to3` is changing the OpenMM Python code under the hood).

Thanks for the question though -- made me go back and reevaluate the change, and I think the new one is much better.

PR #995, opened 2015-06-29
Record plugin load failures

rmcgibbo:
When debugging installations of the OpenMM OpenCL/CUDA platforms, the general error message is just 'There is no registered platform named CUDA', without any information about how to debug the problem. Usually it's caused by `LD_LIBRARY_PATH` not being set correctly or similar, and the first thing I always do with people is run `ldd` on the `libOpenMMCUDA.so` library, but that's not something they'd think to do themselves without knowledge of the OpenMM plugin architecture and the linker.

This PR adds a new static method to `Platform`, `getLoadFailures`, that returns a vector/tuple of strings containing the errors produced by the last call to `loadPluginsFromDirectory` (which is called from python during the import process).

So, for example, you can do

```
import simtk.openmm as mm
mm.Platform.getLoadFailures()
```
And get something like
```
('Error loading library /home/rmcgibbo/opt/openmm-6.3.dev/lib/plugins/libOpenMMCUDA.so: libcuda.so.1: cannot open shared object file: No such file or directory',)
```

### DIFF
--- a/olla/include/openmm/Platform.h
+++ b/olla/include/openmm/Platform.h
@@ -174,7 +174,7 @@ public:
     /**
      * Get any failures caused during the last call to loadPluginsFromDirectory
      */
-    static std::vector<std::string> getLoadFailures();
+    static std::vector<std::string> getPluginLoadFailures();
     /**
      * Get the registered Platform with a particular name.  If no Platform with that name has been
      * registered, this throws an exception.
@@ -205,7 +205,7 @@ public:
      * Load multiple dynamic libraries (DLLs) which contain OpenMM plugins from a single directory.
      * This method loops over every file contained in the specified directory and calls loadPluginLibrary()
      * for each one.  If an error occurs while trying to load a particular file, that file is simply
-     * ignored.
+     * ignored. You can retrieve a list of all such errors by calling getPluginLoadFailures().
      *
      * @param directory    the path to the directory containing libraries to load
      * @return the names of all files which were successfully loaded as libraries
@@ -237,7 +237,7 @@ private:
     std::map<std::string, KernelFactory*> kernelFactories;
     std::map<std::string, std::string> defaultProperties;
     static std::vector<Platform*>& getPlatforms();
-    static std::vector<std::string> loadFailures;
+    static std::vector<std::string> pluginLoadFailures;
 };
 
 
--- a/olla/include/openmm/Platform.h
+++ b/olla/include/openmm/Platform.h
@@ -49,13 +49,13 @@ class KernelFactory;
  * More precisely, a Platform object acts as a registry for a set of KernelFactory
  * objects which together implement the kernels.  The Platform class, in turn, provides a
  * static registry of all available Platform objects.
- * 
+ *
  * To get a Platform object, call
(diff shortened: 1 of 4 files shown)


jchodera:
Awesome!  This had been terrible to debug before.

rmcgibbo:
Need to figure out how to make this return type compatible with the C/Fortran wrappers.

rmcgibbo:
@jchodera: even w/ this, it's still nontrivial to debug if you don't know how the linker works...

jchodera:
If there was some way we could have some Python code automatically parse this and give a good guess as to what is wrong, that would be amazing.

peastman:
loadPluginsFromDirectory() also returns a vector\<string>, so you can just copy the code that handles that.  Search for all occurrences of "loadPluginsFromDirectory" in wrappers/generateWrappers.py.

rmcgibbo:
Also, idk what the right name for this method is. Currently `getLoadFailures`. Could also be `getPluginLoadFailures`, or something else.

peastman:
getPluginLoadFailures() sounds a bit better to me.  Also, the documentation for loadPluginsFromDirectory() currently says, "If an error occurs while trying to load a particular file, that file is simply ignored."  Could you add a sentence: "You can retrieve a list of all such errors by calling getPluginLoadFailures()".

Would it be better for the return value to be a map<string, string> mapping filenames to errors?  Right now, there's no way to tell what error goes with what file (unless the filename happens to be included in the error message, which won't always be the case).

Another possibility that occurs to me...  Instead of implementing this the way you have, what about just modifying the Python code that loads the plugins.  Instead of the call to loadPluginsFromDirectory(), have it explicitly loop over files in that directory and call loadPluginLibrary() for each one.  Then it can record any errors, and store a map of them in the module (exactly as it currently stores pluginLoadedLibNames).

mpharrigan:
This might be something nice to have from the cpp library, esp. considering @rmcgibbo already coded it up

rmcgibbo:
I tried to make the C/Fortran wrappers. They build, but I'm not 100% sure it's correct.

rmcgibbo:
If we didn't already have `loadPluginsFromDirectory` in cpp, then doing the whole functionality in python would be fine, but I don't think we're going to undo that, so this just adds the ability to retrieve errors to that existing API.

rmcgibbo:
Ok, name changed.

swails:
@rmcgibbo -- this is great.  This will make everyone's life easier, even those savvy enough to go to `ldd` or `otool`.  :+1:

peastman:
What do you think of my suggestion of making it return a map\<string, string> instead?  (In the C and Fortran wrappers, that's represented by OpenMM_PropertyArray.)

rmcgibbo:
I'm fine either way. What do you want in the strings? Currently, the error message is taken from the `what()` of the exception thrown by `loadOneLibrary`, which is `throw OpenMMException("Error loading library "+file+": "+dlerror());`, so it has both the filename and the dlerror message.

peastman:
Ok, that's fine then.  As long as it's clear what file each error is for.

PR #997, opened 2015-06-29
Sort plugins by length of filename before loading

rmcgibbo:


swails:
Any idea why this works?  And will it *always* work?

swails:
Or will it just work for now because the length of the shared library names just happens to coincide with the dependencies in such a way that it fortuitously comes together?  Or is it somewhere in between (e.g., the way libraries are named, the dependency chain will almost always follow this heuristic pattern)?

rmcgibbo:
I looked at the `ldd` output for a while, and picked a sort order to make it work. The heuristic is that the topological sort of the dependency chain coincides with the sort by length of filename -- basically that long named things (`libOpenMMDrudeOpenCL.so`) might depend on the shorter named things (`libOpenMMOpenCL.so`) but not the other way around.

I believe that this would also be solved on linux by setting the RPATH to '.' on all of the libOpenMM*.so files.

swails:
> basically that long named things (libOpenMMDrudeOpenCL.so) might depend on the shorter named things (libOpenMMOpenCL.so) but not the other way around.

This is what I suspected.  Basically, a plugin contains the name of the plugin *plus* whatever platform that plugin happens to be written for (if the plugin implements its own platform, then the dependency chain probably starts from there for that .so).  According to this convention, the longer names depend on the shorter names, but not the other way around (whenever dependencies *do* in fact exist).

This might affect users writing their own plugins that may not follow the same conventions that @peastman does with the OpenMM-distributed plugins, though, so this might warrant a note somewhere in the dev manual?

rmcgibbo:
Well, nobody has noticed the current situation so far.

rmcgibbo:
Also, at least on linux, I just confirmed that this is also fixed by setting the RPATH on the libraries. e.g.

```
diff --git a/plugins/drude/platforms/opencl/CMakeLists.txt b/plugins/drude/platforms/opencl/CMakeLists.txt
index a5bff30..d14192c 100644
--- a/plugins/drude/platforms/opencl/CMakeLists.txt
+++ b/plugins/drude/platforms/opencl/CMakeLists.txt
@@ -89,7 +89,7 @@ ADD_LIBRARY(${SHARED_TARGET} SHARED ${SOURCE_FILES} ${SOURCE_INCLUDE_FILES} ${AP
 TARGET_LINK_LIBRARIES(${SHARED_TARGET} ${OPENMM_LIBRARY_NAME}  ${OPENCL_LIBRARIES} ${PTHREADS_LIB})
 TARGET_LINK_LIBRARIES(${SHARED_TARGET} ${OPENMM_LIBRARY_NAME}OpenCL)
 TARGET_LINK_LIBRARIES(${SHARED_TARGET} ${SHARED_DRUDE_TARGET})
-SET_TARGET_PROPERTIES(${SHARED_TARGET} PROPERTIES LINK_FLAGS "${EXTRA_COMPILE_FLAGS}" COMPILE_FLAGS "${EXTRA_COMPILE_FLAGS} -DOPENMM_BUILDING_SHARED_LIBRARY")
+SET_TARGET_PROPERTIES(${SHARED_TARGET} PROPERTIES LINK_FLAGS "${EXTRA_COMPILE_FLAGS}" COMPILE_FLAGS "${EXTRA_COMPILE_FLAGS} -DOPENMM_BUILDING_SHARED_LIBRARY" INSTALL_RPATH "$ORIGIN" BUILD_WITH_INSTALL_RPATH TRUE)

 INSTALL(TARGETS ${SHARED_TARGET} DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/plugins)
 # Ensure that links to the main OpenCL library will be resolved.
(END)
```

swails:
> Well, nobody has noticed the current situation so far.

Probably because there are so few real dependencies.  And maybe the *usual* implementation of the directory searching found shorter path names before longer path names?  There are a lot of situations where formally undefined behavior like this works for a long time... until it doesn't.

peastman:
Setting the RPATH definitely seems like the right solution.  We already do something very similar for OS X.  At line 95 of that same file:

```
# Ensure that links to the main OpenCL library will be resolved.
IF (APPLE)
    SET(OPENCL_LIBRARY libOpenMMOpenCL.dylib)
    INSTALL(CODE "EXECUTE_PROCESS(COMMAND install_name_tool -change ${OPENCL_LIBRARY} @loader_path/${OPENCL_LIBRARY} ${CMAKE_INSTALL_PREFIX}/lib/plugins/lib${SHARED_TARGET}.dylib)")
ENDIF (APPLE)
```

rmcgibbo:
This PR seems like it can't hurt though, since the current load order is undefined anyways.

peastman:
But if we set the RPATH correctly, it will have no value.  So it's simply pointless extra code, which *does* hurt.

rmcgibbo:
ok

rmcgibbo:
we might be able to get rid of the apple-specific rpath code in these files that uses install_name_tool too. http://www.kitware.com/blog/home/post/510

swails:
> But if we set the RPATH correctly, it will have no value. So it's simply pointless extra code, which does hurt.

I don't really agree with this.  It adds 7 lines of code that changes certain behavior from potentially undefined to always well-defined.  The issue @rmcgibbo tried to address with this PR indicated that, in general, order *does* matter.  Whether or not that's still the case after the superior RPATH solution is unclear to me -- but that's partially the point; it's unclear.  At least if the order is well-defined, any errant behavior like this will be reproducible for everyone, not just some lucky person on some lucky machine on some lucky day.

That said, this code doesn't do *much*, and alongside the better plugin loading error reporting, if there is any issue like this one in the future, it probably (hopefully?) won't be too hard to track down.

peastman:
It can't merge this due to conflicts.  Could you resolve them in your branch?

rmcgibbo:
Yep, just saw that too. Will do.

rmcgibbo:
Okay, rebased now.

rmcgibbo:
Tests pass.

Issue #996, opened 2015-06-29
Plugin load order undefined


rmcgibbo:
I've observed an issue where the OpenCL Drude plugin fails to load when openmm is initially imported from python. To test it out, try running the script here https://gist.github.com/rmcgibbo/f5ca495d9f121833f1dc, and then try running it again passing the `--reload` flag, and see if you get an error in one case that the platform doesn't support all the required kernels.

The problem is that the load order of the plugins in loadPluginsFromDirectory is undefined, and, on my machine, happens to be (libOpenMMCPU.so, libOpenMMAmoebaReference.so, libOpenMMRPMDReference.so, libOpenMMDrudeOpenCL.so, libOpenMMPME.so, libOpenMMOpenCL.so, libOpenMMRPMDOpenCL.so, libOpenMMDrudeReference.so). For whatever reason, libOpenMMDrudeOpenCL is loaded before libOpenMMOpenCL.

Since libOpenMMDrudeOpenCL refers to symbols in libOpenMMOpenCL, but libOpenMMOpenCL is neither in the LD_LIBRARY_PATH nor currently loaded into the process, the dlopen call fails.

peastman:
How about if you add the plugins directory to your LD_LIBRARY_PATH?

Ideally it shouldn't matter, and the library should just be found by a relative path.  But I'm not certain how to make that work on Linux.

rmcgibbo:
Yeah, that makes it work. But not ideal -- we got away from requiring that with the latest releases, and it's not possible with conda. #997 is kind of a poor heuristic, but it does work.

swails:
Fixed by #997, right?

PR #998, opened 2015-06-30
Add $ORIGIN rpath to plugins

rmcgibbo:
Re #996, #997. This appears to work on linux. I'm not sure if it works on Apple or Windows, or if on apple the extra `install_name_tool` invocations that modify the loader path can be removed with this change.


rmcgibbo:
I'm not sure how to get this to solve the problem on OSX

peastman:
The magic value "$ORIGIN" is completely Linux specific.  This will not have the intended effect on either Mac or Windows, and may break them.

Possibly there's some special CMake incantation that has the desired effect on all platforms, but if so I don't know what it is.  We already have one that works for Mac, and I'm not sure if it even exists on Windows: I *think* DLLs always look in their own directory for dependencies.  I suggest making this change only apply on Linux.

rmcgibbo:
I know that it's linux specific. I've been trying to figure out the mac analog -- no success yet.

rmcgibbo:
The plugins do not correctly load on mac without the sort order fix. e.g. test out this script https://gist.github.com/rmcgibbo/f5ca495d9f121833f1dc

rmcgibbo:
This PR doesn't break the osx build (on my laptop) -- it's just no different that the current master.

rmcgibbo:
#997 solves [this problem](https://gist.github.com/rmcgibbo/f5ca495d9f121833f1dc) on both OSX and linux, whereas this PR only works on LINUX, and there is no feature of the OSX dynamic linker that can be used to solve this problem, I think. My vote is to merge #997.

peastman:
> #997 solves this problem on both OSX and linux, whereas this PR only works on linux, and there is no corresponding feature of the OSX dynamic linker that can be used to solve this problem, I think.

Fair enough.  I think there *is* a feature that can be used to solve it on OS X, but we can save working that out till some time in the future.

PR #1000, opened 2015-06-30
Fixed bug in RPMD contractions

peastman:


swails:
#1000!  But who's counting?

leeping:
I noticed too, that was awesome.

Issue #1003, opened 2015-07-02
OpenMM installation problem


sshukla101:
I have been trying to install openMM in the linux cluster from the source code downloaded from the github. 

I ran 

"./install.sh"

But it is giving the following error:

"/home/sshukla4/project-cse/anaconda/bin/python: can't open file 'setup.py': [Errno 2] No such file or directory

INSTALLATION FAILED"

I have searched other forums for this error. They seem to suggest that setup.py is provided by the package and is not the part of python installation. I am not able to find this file anywhere. 

Any heads-up?

-Saurabh,
UIUC

jchodera:
If you're using Anaconda python, it would be much easier for you to simply use `conda` to install OpenMM:
```
conda config --add channels http://conda.binstar.org/omnia
conda install openmm
```

jchodera:
If you really want to build from the C++ source yourself, the build process (which does not use that `./install` file, which is only used in creating binary packages, I believe) is documented in the [user guide](http://docs.openmm.org/6.2.0/userguide/application.html#installing-openmm).  But I highly recommend you try the `conda` path first---it's really the quickest way to get up and running.

jchodera:
There are also binary packages distributed here:
https://simtk.org/project/xml/downloads.xml?group_id=161

rmcgibbo:
This file should not be in root directory of the source package though: https://github.com/pandegroup/openmm/blob/master/install.sh

rmcgibbo:
That's the installer script we use with the binaries. It should live somewhere in `devtools/` in the source distribution.

sshukla101:
Is installing from conda and installing from the source exactly the same?

PS: Conda seems to have worked. I still need to run some test simulations to check if it is installed correctly.

rmcgibbo:
Installing from conda means that you're using precompiled binaries that we provide. Installing from source means that you're compiling yourself.

jchodera:
> It should live somewhere in devtools/ in the source distribution.

Good call, @rmcgibbo.  Are you going to tackle this in a PR, or should I?

rmcgibbo:
go for it. you know better than I how the non-conda packaging works.

jchodera:
Will do!

PR #1005, opened 2015-07-02
Moved install.sh out of harm's way into devtools/packaging

jchodera:
Also updated binary packaging scripts.

The only thing I'm not sure about is whether `./install.sh` should be packaged with the source distro too.


jchodera:
This addresses an issue brought up in #1003

rmcgibbo:
lgtm

Issue #1007, opened 2015-07-03
Should cmake check the swig version?


rmcgibbo:
Does anyone know what the minimum version  of swig is that we support? I just logged onto a cluster with a really old version in /usr/bin (1.3.40), and got some garbage.

```
$ make PythonInstall
[  0%] Built target ApiWrappers
[ 51%] Built target OpenMM
[ 51%] Installing OpenMM Python binary module...
OPENMM_LIB_PATH = /home/rmcgibbo/opt/openmm-6.3-dev/lib
OPENMM_INCLUDE_PATH = /home/rmcgibbo/projects/openmm/./include/home/rmcgibbo/projects/openmm/./include/openmm/home/rmcgibbo/projects/openmm/./include/openmm/internal/home/rmcgibbo/projects/openmm/openmmapi/include/home/rmcgibbo/projects/openmm/openmmapi/include/openmm/home/rmcgibbo/projects/openmm/openmmapi/include/openmm/internal/home/rmcgibbo/projects/openmm/olla/include/home/rmcgibbo/projects/openmm/olla/include/openmm/home/rmcgibbo/projects/openmm/olla/include/openmm/internal/home/rmcgibbo/projects/openmm/serialization/include/home/rmcgibbo/projects/openmm/serialization/include/openmm/home/rmcgibbo/projects/openmm/serialization/include/openmm/internal/home/rmcgibbo/projects/openmm/plugins/amoeba/openmmapi/include/home/rmcgibbo/projects/openmm/plugins/amoeba/openmmapi/include/openmm/home/rmcgibbo/projects/openmm/plugins/amoeba/openmmapi/include/openmm/internal/home/rmcgibbo/projects/openmm/plugins/rpmd/openmmapi/include/home/rmcgibbo/projects/openmm/plugins/rpmd/openmmapi/include/openmm/home/rmcgibbo/projects/openmm/plugins/rpmd/openmmapi/include/openmm/internal/home/rmcgibbo/projects/openmm/plugins/drude/openmmapi/include/home/rmcgibbo/projects/openmm/plugins/drude/openmmapi/include/openmm/home/rmcgibbo/projects/openmm/plugins/drude/openmmapi/include/openmm/internal
                                  author                                           Randall J. Radmer
                            author_email                                         radmer@stanford.edu
                                cmdclass{'build_py': <class 'distutils.command.build_py.build_py_2to3'>}
                              data_files                                                          []
                             description                Python wrapper for OpenMM (a C++ MD package)
[...]

PR #1010, opened 2015-07-05
Swig cmake version check

rmcgibbo:
re #1007, swig 2.0.4 was the oldest version of swig which produced wrappers that successfully compiled against python 3.4.


jchodera:
+1

This check will save a lot of pain in the future.

rmcgibbo:
we were already doing the appropriate check for python2, so it's not a big change.

swails:
+1

Thanks for figuring this out @rmcgibbo

Issue #1004, opened 2015-07-02
Some CPU platform test failures


rmcgibbo:
Running on sherlock (CentOS 6.6, Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz) with a couple different compilers. All tests with 42c7781a84e2f828595bdf3c3d732bfb45532849. 

1) clang version 3.4 (tags/RELEASE_34/final). No failures :white_check_mark: 
2) gcc (GCC) 4.8.1 20130715 (Red Hat 4.8.1-4) TestCpuCustomGBForce fails
```
exception: Assertion failure at TestCpuCustomGBForce.cpp:150.   Expected [-117.436, 78.1114, -10.5441], found [-128.571, 44.9566, -8.68634]
```
3) gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-11) TestCpuCustomGBForce fails
```
exception: Assertion failure at TestCpuCustomGBForce.cpp:150.   Expected [-117.436, 78.1114, -10.5441], found [-128.571, 44.9566, -8.68634]
```
4) icc (ICC) 15.0.0 20140723, although perhaps this isn't formally supported
```
exception: Assertion failure at TestCpuNeighborList.cpp:118
exception: Assertion failure at TestCpuGBSAOBCForce.cpp:226.  Expected 0, found 8.18476
exception: Assertion failure at TestCpuCustomNonbondedForce.cpp:843.   Expected [0, 0, 0], found [-94.193, -205.722, -94.83]
exception: Assertion failure at TestCpuCustomManyParticleForce.cpp:614.  Expected 3.3147e+07, found 3.29341e+07
exception: Assertion failure at TestCpuNonbondedForce.cpp:472.   Expected [-0.622213, -1.35938, -3.25274], found [-0.292664, -2.93669, -2.84411]
```

------

Running on vspm9 (Ubuntu 15.03, Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz) with ca200ea3eee2f4f0139afc057d7d8bf51315acd1

1) gcc (Ubuntu 4.9.2-10ubuntu13) 4.9.2, TestCpuCustomGBForce fails
```
exception: Assertion failure at TestCpuCustomGBForce.cpp:150.   Expected [-117.436, 78.1114, -10.5441], found [-128.571, 44.9566, -8.68634]
```
[...]

PR #1014, opened 2015-07-06
[WIP] Bug in CPU neighborlist?

rmcgibbo:
This change triggers a test failure on my machine -- want to run on travis.

### DIFF
--- a/platforms/cpu/tests/TestCpuNeighborList.cpp
+++ b/platforms/cpu/tests/TestCpuNeighborList.cpp
@@ -40,11 +40,11 @@
 #include "CpuPlatform.h"
 #include "sfmt/SFMT.h"
 #include <iostream>
+#include <cstdio>
 #include <set>
 #include <utility>
 #include <vector>
 #include <algorithm>
-#include <cstdio>
 
 using namespace OpenMM;
 using namespace std;
@@ -87,7 +87,6 @@ void testNeighborList(bool periodic, bool triclinic) {
             exclusions[i-j].insert(i);
         }
     }
-
     ThreadPool threads;
     CpuNeighborList neighborList(blockSize);
     neighborList.computeNeighborList(numParticles, positions, exclusions, boxVectors, periodic, cutoff, threads);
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Reverting 95423468ab fixes the test failure, so I think there's a bug here.

jchodera:
Did the test from this PR get incorporated somewhere?  Seems like it would be useful to retain the test that caught this bug.

swails:
@rmcgibbo found this just running the existing test suite with a different compiler (icc).

swails:
See #1004

jchodera:
Wasn't this a PR containing test changes that revealed the error using `gcc`?  He demonstrated that the test fails on travis too (which uses `gcc`):
https://github.com/pandegroup/openmm/pull/1014/files

I know he was able to get things to fail with `icc`, but if there is a test that reveals the issue with `gcc` too, that may be worth including...

swails:
No, the PR just changed 1 file, no tests (it reverted a single commit).  @rmcgibbo posted test failures from Intel and one or two from GCC from the existing test suite.  The test failures seem to be intermittent and hardware-dependent, which is why Travis didn't catch it but @rmcgibbo did.

jchodera:
Sorry, I'm still confused; maybe we're mixing up issues?

If you click on the "files" tab on this PR, or you click on [this link which takes you to the same place](https://github.com/pandegroup/openmm/pull/1014/files), you will see:

Showing  1 changed file  with 14 additions and 5 deletions.
`platforms/cpu/tests/TestCpuNeighborList.cpp`

which clearly shows edits made to `TestCpuNeighborList`, which when run on travis (using `clang`) show that this test *fails*:
```
37: Test command: /home/travis/build/pandegroup/openmm/TestCpuNeighborList "single"
37: Test timeout computed to be: 1500
37: exception: Assertion failure at TestCpuNeighborList.cpp:127
 36/108 Test  #37: TestCpuNeighborList ...........................***Failed    0.10 sec
```

swails:
Oh, my mistake.  I misread the filename.  I'll wait for @rmcgibbo to comment.

PR #1015, opened 2015-07-06
Reverted an incorrect optimization

peastman:
Fixes #1014.

### DIFF
--- a/platforms/cpu/src/CpuNeighborList.cpp
+++ b/platforms/cpu/src/CpuNeighborList.cpp
@@ -233,34 +233,24 @@ public:
                 
                 float minx = centerPos[0];
                 float maxx = centerPos[0];
-                float offset[3] = {-xoffset, -yoffset+voxelSizeY*y+(usePeriodic ? 0.0f : miny), voxelSizeZ*z+(usePeriodic ? 0.0f : minz)};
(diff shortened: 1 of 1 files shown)


swails:
Any ideas what was wrong?  Or is this a fix-for-release with intent to go back to it later?

rmcgibbo:
Also fixes the gcc failures in`TestCpuCustomGBForce` (#1015)

peastman:
Yes, though it's a little complicated to describe.  This bit of code is identifying neighbors of a block of atoms, and is working out what range of atoms in a particular list it needs to search through.  In doing that, it applies a correction for atoms that are in different voxels.  That correction gets skipped if the atoms are both in the same voxel.  I previously vectorized this so it would do the calculation for four atoms at a time.  But it didn't take into account the fact that some of those atoms might be in the same voxel as the target atom and others not, so some would require the correction and others not.

swails:
Thanks for the explanation

rmcgibbo:
(I just went over and asked @peastman. In particular, he pointed out the lines line `if (y != atomVoxelIndex[k].y)`, which is only checking the `k`th atom, but since we're looping by 4 atoms at a time, that check is just short of assuming that the remaining 3 atoms in the block are in the same voxel as the 1st, which isn't always true.

PR #1019, opened 2015-07-07
ENH: TestVectorize8

rmcgibbo:
I don't think we have a test that checks the basic properties of fvec8 and ivec8 yet. I'm adapting TestVectorize.cpp, which checks fvec8/ivec4.


rmcgibbo:
As I suspected with clang 3.6.0
```
$ ./TestVectorize8
exception: Assertion failure at TestVectorize8.cpp:106.   Expected (1.5,3,4.5,6,7.5,9,10.5,12), found (1.5,3,4.5,6,8.40779e-45,0,0,0)
```

peastman:
Note that I intentionally did *not* implement operator[]() for fvec8 and ivec8, because it would necessarily be less efficient than calling lowerVec() or upperVec(), and hence would lead people into unwittingly writing inefficient code.

rmcgibbo:
I'm only using it to make the assert more convenient -- it's not really necessary.

peastman:
Understood!

rmcgibbo:
Would it be bad form to make some slow convenience functions only conditionally compiled in debug mode or something? operator[] (and also a print() method) are kind of useful for debugging sometimes.

peastman:
Yes, that would be very bad form. :)  Besides, test cases need to work when compiled in release mode!  Just call lowerVec() or upperVec(), then get the appropriate element from that.

rmcgibbo:
:) okay.

rmcgibbo:
Okay. I think this is ready. @peastman: do you know why `bool isVec8Supported()` was [defined](https://github.com/pandegroup/openmm/blob/master/platforms/cpu/src/CpuNonbondedForceVec8.cpp#L52) in `CpuNonbondedForceVec8.cpp` instead of in  `vectorize8.h`? Can I move it over? Otherwise I need to copy-paste it into this file.

peastman:
Moving isVec8Supported() to the header would be reasonable, but let's wait till after 6.3 to do that.  It feels like a slightly dangerous change to make this late.

What's up with the commented out blocks that it says require AVX2?  OpenMM doesn't use any AVX2 instructions anywhere.

rmcgibbo:
It's just a placeholder for tests that we would add if we supported those operations on ivec8 (which we don't).

rmcgibbo:
Okay, I removed the commented out blocks.

PR #1020, opened 2015-07-07
Fixed a change to AmoebaStretchBendForceProxy that broke compatibility with old files

peastman:
This fixes an error that was introduced in https://github.com/pandegroup/openmm/commit/94ca75c6d65404cf9239e6c33ecdb22d657af5e9


swails:
In my defense, I asked about maintaining backwards compatibility IIRC :)

peastman:
Sorry!  Backwards compatibility with code: less important.  Backwards compatibility with serialized files: more important.  It's a lot easier from someone to update the former than the latter.

swails:
Good point.

PR #1021, opened 2015-07-07
Disable AVX CPU NonbondedForce on Clang 3.6.0 due to compiler bugs

rmcgibbo:
I tested clang 3.5.0 and the [nightly build](http://llvm.org/apt/) of clang 3.7.0 (in development), and they both work.

@peastman: should we be doing these compiler checks (this code is adjacent to a similar msvc check that disables AVX) in CMake, or here?

### DIFF
--- a/platforms/cpu/src/CpuNonbondedForceVec8.cpp
+++ b/platforms/cpu/src/CpuNonbondedForceVec8.cpp
@@ -37,6 +37,12 @@ using namespace OpenMM;
     #undef __AVX__
 #endif
 
+#if defined (__clang__) && (__clang_major__==3) && (__clang_minor__==6)
+  // Workaround for a compiler bug in Clang 3.6.0
+  #undef __AVX__
+#endif
+
+
 #ifndef __AVX__
 bool isVec8Supported() {
     return false;


peastman:
Have you tested 3.6.1?  It was released in late May, and the bug report you linked before had comments posted in early May saying it was fixed internally.  So it might well be working now.

I'm also not sure what's the best way to handle this.  In some ways, I'd rather it just refuse to compile instead of silently generating slow code.

rmcgibbo:
We could switch the `#undef __AVX__` to `#error`. It would be better if that kind of thing came up in CMake though, not as a compiler error.

rmcgibbo:
Just tested. `TestVectorize8` also fails with 3.6.1

peastman:
I'm fine with making it a CMake error rather than a compiler error.

rmcgibbo:
> I'm fine with making it a CMake error rather than a compiler error.

Done.

peastman:
> Done.

I don't think you checked it in.

rmcgibbo:
It's in #1022

peastman:
Oh, I see.  Different PR.  This can be closed then?

PR #1025, opened 2015-07-08
MAINT: Docstrings

rmcgibbo:
1. Update the module docstring in `__init__.py`
2. Update the package docstring in `setup.py`
3. Modify the `swigInputBuilder.py` to strip some of the HTML tags from the C++ docstrings when generating the python docstrings, so that they're a little bit easier to read in the console when using the python docstring introspection facilities. The changes to the generated file OpenMM_docstrings.i, relative to the current version are here: https://gist.github.com/rmcgibbo/556464bd6ccfb74bf7f3. (This is the file that swig reads the python docstrings from.)


peastman:
The build failed with errors in OpenMMSwigHeaders.i.

peastman:
I'm a little nervous about your changes to docstring generation.  Have you gone through the generated Doxygen docs for the Python API to make sure everything is still correct?  There's a lot of machinery that goes into generating those, including some custom scripts, and it would be really easy to break something.

rmcgibbo:
We can wait until after 6.3. Still probably needs some tweaks.

rmcgibbo:
here are the doxygen docs, as built with 804cbb2 http://stanford.edu/~rmcgibbo/api-python//annotated.html

peastman:
Is this ready, or is it still waiting for tweaks?

Also, it can't be merged due to conflicts.  Probably just needs to have the latest code from master merged into the branch.

rmcgibbo:
Okay, I think this is ready to merge.

PR #1027, opened 2015-07-08
WIP: Nicer swig wrappers

rmcgibbo:
Re #882, #879, this now generates nicer swig wrappers that have the proper function signatures when using a sufficiently new version of swig (definitely 3.0.6, not 3.0.2, not sure exactly where in between).

Previously, all of the python wrappers formally had a signature like `openmm.LangevinIntegrator.setTemperature(self, *args)`. Using new swig, with this PR the signature is `LangevinIntegrator.setTemperature(self, temp)`. This allows you to use keyword arguments for example, for most methods.

Still some bugs to be worked out, I suspect. Try it out and let me know if it works for you.

### DIFF
--- a/wrappers/python/CMakeLists.txt
+++ b/wrappers/python/CMakeLists.txt
@@ -163,7 +163,6 @@ set(SWIG_INPUT_FILES2
     "${SWIG_OPENMM_DIR}/swig_lib/python/header.i"
     "${SWIG_OPENMM_DIR}/swig_lib/python/pythoncode.i"
     "${SWIG_OPENMM_DIR}/swig_lib/python/typemaps.i"
-    "${SWIG_OPENMM_DIR}/swig_lib/python/pythonprepend_all.i"
 )
 
 # Create input files for swig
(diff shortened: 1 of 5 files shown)


peastman:
Cool!  But *definitely* something to wait until after 6.3 to merge.

rmcgibbo:
Yep :) (which is hopefully imminent, right?)

peastman:
I hope so!  I'm still corresponding with Nvidia to try to track down the Windows performance problem.

rmcgibbo:
TODO notes for me:

 - [ ] Unit stripping in constructors
 - [ ] Overloaded methods are not treated correctly (`system.getForce`)
 - [ ] Methods that use pass-by-reference to return values like `System.getConstraintParameters` not handled correctly when stripping units.

rmcgibbo:
Actually, I think it will be more robust to use typemaps to apply the unit conversion.

peastman:
Looks like there are some test failures.

rmcgibbo:
Incorporating this into #1031

PR #1028, opened 2015-07-09
Spherical harmonics

andysim:
Take II.  Tests pass on my machine, after fixing a small PBC-related error.  This *should* give the spherical harmonic direct space terms for the reference platform's PME implementation.

### DIFF
--- a/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp
+++ b/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp
@@ -364,9 +364,15 @@ void AmoebaReferenceMultipoleForce::checkChiralCenterAtParticle(MultipoleParticl
     RealOpenMM volume = deltaC.dot(deltaAD);
 
     if (volume < 0.0) {
+#if SPHERICAL_MULTIPOLES
+        particleI.sphericalDipole[2]     *= -1.0;   // y
+        particleI.sphericalQuadrupole[2] *= -1.0;   // yz
+        particleI.sphericalQuadrupole[4] *= -1.0;   // xy
+#else
         particleI.dipole[1]         *= -1.0; // pole(3,i)
         particleI.quadrupole[QXY]   *= -1.0; // pole(6,i)  && pole(8,i)
         particleI.quadrupole[QYZ]   *= -1.0; // pole(10,i) && pole(12,i)
+#endif
     }
 }
 
@@ -554,21 +560,20 @@ void AmoebaReferenceMultipoleForce::applyRotationMatrixToParticle(      Multipol
 
 
 #if SPHERICAL_MULTIPOLES
-void AmoebaReferenceMultipoleForce::formQIRotationMatrix(const MultipoleParticleData& particleI,
-                                                         const MultipoleParticleData& particleJ,
+void AmoebaReferenceMultipoleForce::formQIRotationMatrix(const RealVec& iPosition,
+                                                         const RealVec& jPosition,
                                                          const RealVec &deltaR,
                                                          RealOpenMM r,
                                                          RealOpenMM (&rotationMatrix)[3][3]) const
 {
     RealVec vectorZ = (deltaR)/r;
     RealVec vectorX(vectorZ);
-    if ((particleI.position[1] != particleJ.position[1]) || (particleI.position[2] != particleJ.position[2])){
+    if ((iPosition[1] != jPosition[1]) || (iPosition[2] != jPosition[2])){
         vectorX[0] += 1.0;
     }else{
         vectorX[1] += 1.0;
     }
     RealVec vectorY;
-
 
     RealOpenMM dot      = vectorZ.dot(vectorX);
     vectorX            -= vectorZ*dot;
@@ -587,6 +592,8 @@ void AmoebaReferenceMultipoleForce::formQIRotationMatrix(const MultipoleParticle
     rotationMatrix[2][1] = vectorY[0];
     rotationMatrix[2][2] = vectorY[1];
 }
+
+
 
 
 void AmoebaReferenceMultipoleForce::buildSphericalQuadrupoleRotationMatrix(const RealOpenMM (&D1)[3][3], RealOpenMM (&D2)[5][5]) const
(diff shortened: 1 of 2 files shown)


swails:
Cool!

leeping:
That's awesome.  Let me know if there's anything I can do to help test this.

andysim:
Thanks guys.  The code automagically converts the standard input to spherical harmonics behind the scenes and, currently, keeps both Cartesian and spherical harmonic moments around.  This isn't a problem in terms of storage, but it's a little messy, so I'll work on the reciprocal and self terms over the next couple of weeks.  I'll also add the non-PME version.

I probably won't be coding this up in CUDA, but I'll have a crack at a new OpenCL implementation that will be easy to port back to CUDA.  If anybody want to shoehorn the spherical harmonics from the reference code into CUDA, be my guest.  If you take a look at the code, you should see that most of it is primed for vectorization; perhaps a CPU implementation using AVX instructions will be worthwhile also.

peastman:
That's great!  I can do the CUDA version.  But not for a little while yet. :)

If it's OK with you, I'm not going to merge this PR.  Instead I'll merge it by hand, clean out all the now-obsolete code, and send a new PR with the complete feature (including CUDA implementation) once it's all ready.

andysim:
That's completely fine with me.  I was just sending the PR so that you can have immediate access to the code because I think it should be quite helpful for the CUDA implementation.  Merging by hand will be extremely easy; I deliberately housed all new code in preprocessor macros to help you find what changed, and will continue to do this as I add the spherical harmonics in more places.  I can easily make this a spherical harmonic-only code if we determine that this is indeed faster (I haven't done any timing yet at all, but it should be) although the generalized Kirkwood code will have to Cartesians.  I haven't looked into the theory for that.

I will continue to push any updates to my public fork of the repo and will let you know when I have what I think is a complete version.  Perhaps it's easier if you follow changes in my repo, rather than me bombarding you with pull requests?  Either way, I'll drop you an email whenever I get to a big milestone.

swails:
I think PRs are a good way to go, personally.  It would make it easy for me to keep up with developments.

> If it's OK with you, I'm not going to merge this PR. Instead I'll merge it by hand, clean out all the now-obsolete code, and send a new PR with the complete feature (including CUDA implementation) once it's all ready

@peastman -- can you start from @andysim's branch so that his commits are still in the history?  That way they'll appear on his Github profile if/when a future employer is trying to gauge his activity here.  (Always a nice thing for postdocs and grad students.)

peastman:
Where else do you still plan to add spherical harmonic code?  I don't see much reason to modify the reciprocal space code.  You said it would have no performance benefit, so it would just risk creating bugs in code that already works.  We wouldn't even be able to simplify our data structures to get rid of the Cartesian multipoles, because we'll still need them for GK.

andysim:
I originally planned to eliminate the Cartesians completely, but decided against it when I saw the GK code and realized it's a lot of work.  The reciprocal space terms won't speed up through the addition of spherical harmonics, but it will allow us to only rotate the sphericals to the lab frame for each call, instead of having to rotate both Cartesians and spherical harmonics.  This probably isn't a major bottleneck; it's done in O(N) effort.  I'm happy to leave this alone if you'd prefer to keep the existing code.

I was also planning to implement the non-PME version of the spherical harmonic code.  It's mostly a cut and paste exercise, but I think it's worth having in the reference platform if you plan to add the non-PME version to the CUDA platform.

I only just saw the last part of your message, @swails; thanks for the consideration.  I think the more we can keep common ancestry, the easier it will be for us to merge codebases.  To that end, perhaps the easiest place to start this merge is to pull in the diff, and just delete the preprocessor macros (and the #else clause in the actual electrostatic code).  I'm happy to follow whatever protocol you decide is best; my aim is simply to make the code faster, not impose my design decisions on you.

peastman:
I'll be ready to start working on this soon.  Is this basically in good shape and ready for me to base the CUDA version on it?  I notice that TestReferenceAmoebaGeneralizedKirkwoodForce is failing.  Is that caused by a partial reimplementation of the GK calculation that should be reverted, or an interaction with something else?

andysim:
That's my bad.  I didn't realize that subsequent pushes to my repo would be included in the original PR.  I was messing around with the rec and self terms, to eliminate all instances of the cartesian moments and pushed to my GH repo to make sure it's backed up somewhere.  The GK code seems to have been broken in the process, but it should be an easy fix.  I'll look into this when I get a chance later this week, and will fix the problem.  I've been working with Jay and I think we have an even better approach in the works, so it may be worth waiting to see if that pans out, before implementing a production level QI code.

The GK code needing Cartesian moments is quite unfortunate, because it's annoying to have to keep track of both those and spherical moments.  I'm considering making the function that rotates the moments to the lab frame virtual, so that the GK can rotate Cartesians, while the regular code only rotates sphericals.

peastman:
Ok, I'll wait then.  Thanks!  What's the even better approach?

andysim:
The new approach will still use the QI/spherical approach to describe the permanent moments, but will avoid some FFTs and will make better use of intermediates.  *If* it works...

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp
```diff
@@ -347,9 +364,15 @@ void AmoebaReferenceMultipoleForce::checkChiralCenterAtParticle(MultipoleParticl
...
     if (volume < 0.0) {
+#if SPHERICAL_MULTIPOLES
+        particleI.sphericalDipole[2]     *= -1.0;   // y
+        particleI.sphericalQuadrupole[2] *= -1.0;   // yz
+        particleI.sphericalQuadrupole[4] *= -1.0;   // xy
+#else
         particleI.dipole[1]         *= -1.0; // pole(3,i)
         particleI.quadrupole[QXY]   *= -1.0; // pole(6,i)  && pole(8,i)
         particleI.quadrupole[QYZ]   *= -1.0; // pole(10,i) && pole(12,i)
+#endif
```
peastman: @andysim should it be executing both blocks here, not just the first one?  The cartesian multipoles are still used for some things (PME, GK), so shouldn't those get reversed as well?

andysim: Yes it should.  My bad.  I was hoping to get rid of the Cartesians entirely when I wrote this, but the pesky GK code got in the way.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp
```diff
@@ -907,6 +1181,396 @@ RealOpenMM AmoebaReferenceMultipoleForce::calculateElectrostaticPairIxn(const Mu
...
+    RealOpenMM rInv = 1.0 / r;
+
+    // The rInvVec array is defined such that the ith element is R^-i, with the
+    // dieleectric constant folded in, to avoid conversions later.
+    rInvVec[1] = prefac * rInv;
+    for(int i = 2; i < 7; ++i)
+        rInvVec[i] = rInvVec[i-1] * rInv;
+
+    RealOpenMM mScale = scalingFactors[M_SCALE];
+    RealOpenMM pScale = scalingFactors[D_SCALE];
```
peastman: This is clearly wrong here.  You have D_SCALE and P_SCALE swapped.  Am I correct that all "d" and "p" suffixes throughout the routine should be reversed?

peastman: No, it's something more complicated than that.  I'm very confused by what's going on here, but clearly there is *something* wrong.

andysim: When I first coded it up, I had all of the dscale and scales switched in the energy and derivative intermediates in that chunk of code.  For all of the tests in the reference multipole force test case, the d and p groups are identical so they always pass.  The villin headpiece in the Amoeba GB test case has different d and p groups, so that test case originally failed.  I realized the mistake and fixed it (in 92b0297844df6e7ed1ffc3e86a110d0a9a0b7cc5) by just redefining the variable names.  I meant to go back through and properly change the code to make it more readable, but forgot to do so.  Sorry for the confusion.

peastman: If I swap dScale and pScale as shown in that commit, then TestReferenceAmoebaGeneralizedKirkwoodForce fails.  That's how I originally found this: my CUDA version didn't have the scales swapped, and it was failing.

peastman: Oh, I see.  That commit is where you swapped them.  But it's not just a question of the labels being swapped.  *Some* of the d's and p's are clearly correct.  For example:

            valIP += inducedDipoleRotationMatrix[ii][jj] * _inducedDipolePolar[iIndex][jj];
            valID += inducedDipoleRotationMatrix[ii][jj] * _inducedDipole[iIndex][jj];

There are some places where the labels d and p refer to which induced dipoles they work with.  And there are others where they refer to which scale factor they use.  Yet they then seem to be used consistently.  For example,

    Vji[0]  += ePermCoef*qiQI[1] + eUIndCoef*qiUindI[0] + eUInpCoef*qiUinpI[0];

eUIndCoef is calculated with dScale, so it should really be p.  qiUindI is calculated based on the non-polar induced dipole, so that one really is correct as d.  Yet the d terms are always used together, and the p terms are always used together.

andysim: It's only the eCoef and dCoef terms (which are really the T^n and dT^n/dR terms) that are incorrect.  The dipoles are just fine; I just accidentally applied the wrong T tensor coefficients in my first pass.

peastman: Could you clarify what you mean?  I tried swapping the definitions of eCoef and dCoef, but that didn't fix it.  What change do I need to make?

andysim: The code in the repo is working and passes all tests. The only weird part is that the dScale variable is set to PSCALE, and vice versa. To make the code more readable you may want to redefine dScale and pScale to have the proper meanings, but change all instances of pScale usage to dSclae, and dScale to pScale. This is trivial to do in context-aware editors like the excellent QT Creator. I will gladly do it tomorrow if you like. I've been traveling for over 24 hours now, so I don't trust myself to do it right now!

peastman: Right, I got that.  But if I do that, the code doesn't make sense.  D_SCALE is *supposed* to be the scale factor applied to `_inducedDipole`, and P_SCALE is *supposed* to be the scale factor applied to `_inducedDipolePolar`.  If I make the change you suggest, then that will be reversed.

andysim: Not quite.  D_SCALE is the scale factor used to generate the field that defines inducedDipole, as P_SCALE is used to generate the field for inducedDipolePolar; this means that the scale factors used to generate the permanent fields are in the order we'd expect.  However the energy expression (see eqn. A4 in dx.doi.org/10.1021/ct200304d) is

    U  =  -1/2 Ep µd  =  -1/2 Ed µp

so the p scale factors should be applied to µd, and d scale factors to µp.  I forgot this when coding it up the first time, and will surely forget it again next time I write this sort of code.  The p and d subscripts make things quite tricky.

Issue #974, opened 2015-06-19
Allow different force groups for nonbonded forces in GPU platforms


swails:
As a proposed enhancement, it would be great to allow different nonbonded forces to be assigned to different force groups, which is currently possible only in the CPU and Reference platforms.

This feature is going to be necessary to introduce some of the resonance-free, isokinetic multiple time stepping integrators that Mark Tuckerman's group has been working on.  (@peastman -- you can add the enhancement label to this if we want to start using labels for our issues)

peastman:
I have an idea in mind for how to implement this and #975.  It will involve relaxing but not completely eliminating the various restrictions on nonbonded forces in the CUDA and OpenCL platforms.  Would the following be acceptable for everyone?

1. Nonbonded forces will no longer be required to all be in the same force group.

2. All nonbonded forces *in a particular force group* must have the same cutoff, but different force groups can have different cutoffs.

3. All nonbonded forces will still be required to have identical exclusions, just as they do now.

swails:
+1.  Completely covers my use-cases.

For the third point, I believe that a clever use of `CustomBondForce` will be sufficient to completely work around any real restriction that institutes, if someone *needs* different exclusions.

swails:
cc #975.  @peastman mentioned a fix to #975 in his second point.  So refer to that thread for the second part of this request.

jchodera:
This sounds great!  Would definitely use this feature.

Lnaden:
+1 on this. 

One question though, does the identical exclusions requirement mean that all molecules with exclusions must appear in all `NonbondedForce` objects, or simply that if a molecule with exclusions appears in `NonbondedForce` 1, it must have the same exclusions (and atoms) if it also appears in `NonbondedForce` 2?

peastman:
It means that all nonbonded forces (NonbondedForce, CustomNonbondedForce, etc.) must have precisely identical lists of exclusions.  If the interaction between two atoms is excluded in one, it must be excluded in all of them.  That's because it only stores one list of exclusions that needs to be shared by all of them.

jchodera:
What about interaction groups?  Can these continue to be different in different forces?

peastman:
Interaction groups are strictly internal to a CustomForce object.   If you have two of those, there's no requirement the interaction groups be the same.  The restrictions only apply to forces computed by CudaNonbondedUtilities and OpenCLNonbondedUtilities.

peastman:
This is implemented by #1030.

PR #1030, opened 2015-07-09
Different force groups can have different cutoffs

peastman:
This is the implementation of #974.  I won't merge this until after 6.3, but you can try it out now if you want.


swails:
Awesome!  I'll play around with implementing the isokinetic integrator with this when I get a chance (probably won't be until after 6.3, anyway).  I'll let you know how it goes (and contribute the integrator back if I get it cleaned up enough to do that).

swails:
w00t!  Nice work @peastman!

PR #1031, opened 2015-07-10
MAINT: Nicer swig wrappers

rmcgibbo:
Related to #1027, https://github.com/pandegroup/openmm/pull/1027#issuecomment-119789257. This removes the python `stripUnits` function that was applied indiscriminately to the arguments to every method, and moves this functionality into the swig typemaps. Only OpenMM C++ functions that take `double`, `Vec3`, and `std::vector<Vec3>` get the units stripped off, and it happens inside the C++ wrapper functions.

edit:
Also, Re #882, #879, this now generates nicer swig python wrappers that have the proper function signatures when using a sufficiently new version of swig (definitely 3.0.6, not 3.0.2, not sure exactly where in between). Previously, all of the python wrappers formally had a signature like `openmm.LangevinIntegrator.setTemperature(self, *args)`. Using new swig, with this PR the signature is `LangevinIntegrator.setTemperature(self, temp)`. This allows you to use keyword arguments for example, for most methods.


jchodera:
Cool!

Do the typemaps do checking for unit compatibility (eg distance, time, etc)?

rmcgibbo:
> Do the typemaps do checking for unit compatibility (eg distance, time, etc)?

No. They behave in the same way as the current ones.

jchodera:
Ah, OK.  We *could* enforce automatic unit checking if there was interest:
https://github.com/pandegroup/openmm/issues/204#issuecomment-27707770

rmcgibbo:
It is possible. It's a little trickier than just that, because you have to integrate it with the swig build system.

rmcgibbo:
The first step, if you want to take a stab at it, is to write down the expected base dimension for the arguments for every function in a format similar to https://github.com/pandegroup/openmm/blob/master/wrappers/python/src/swig_doxygen/swigInputConfig.py#L148-L429

jchodera:
> The first step, if you want to take a stab at it, is to write down the expected base dimension for the arguments for every function in a format similar to

Why can't we just use that `UNITS` list?

rmcgibbo:
That only lists the units for the return values, it doesn't list the units for the arguments.

jchodera:
Ah, of course.

peastman:
What's the reason for stripping units in C rather than Python?  It certainly makes the code a lot more complicated.

> Only OpenMM C++ functions that take double, Vec3, and std::vector<Vec3> get the units stripped off

You also need to do vector\<double>.  Possibly some other types - I'm not sure.

rmcgibbo:
> What's the reason for stripping units in C rather than Python? It certainly makes the code a lot more complicated.

Because we don't reliably know the name of the arguments in python.

rmcgibbo:
This is really the kind of thing that the whole swig typemap system is designed to handle.

swails:
> What's the reason for stripping units in C rather than Python? It certainly makes the code a lot more complicated.

To expand on @rmcgibbo's answer, this is because some versions of SWIG have a bug that *force* `compactdefaultargs`.  And this was actually the default for most of OpenMM's development history since only recently did we see fatal problems arising from SWIG 3 that was a result of this bug being fixed.

So *not* supporting `compactdefaultargs` isn't really an option.  I hadn't even thought of trying to implement this at the C level... clever @rmcgibbo.  We will probably need quite a few tests to make sure that units are handled everywhere they used to be correctly handled, though.

peastman:
This still needs to have support for vector\<double> added, right?

rmcgibbo:
Yeah.

rmcgibbo:
Okay. This should be clear, assuming that the tests pass. It's definitely possible that there are some bugs with this though, possibly either with methods that crash in the swig wrappers or leak memory. The probability of any silent-but-incorrect-results type bugs seems very low.

rmcgibbo:
Test failure looks legit.

swails:
Definitely is legit.  Something, somewhere is either stripping units prematurely or, more likely, not adding them when they should be.

I actually think our test coverage on this stuff is woefully inadequate.  What do you say I work on some comprehensive unit tests based off master that try and improve (or, hopefully, even complete) the test coverage completely on unit handling within forces (i.e., make sure all of them are handled correctly and that the return value has the appropriate units attached).

I could file this PR either to master for @peastman to merge or to this PR.

swails:
FTR, I'm a big +1 on this, but I would like to add my tests before this gets merged if @rmcgibbo and @peastman are OK with that.  If so, I can get those done by Monday afternoon.

peastman:
That would be great.

rmcgibbo:
@swails do you have more tests? Otherwise I think this is ready to go.

swails:
Ack.  I forgot!  Sorry!

I can add the tests later if you don't want to wait on me.  I dropped the ball here, sorry about that.  I'll add the tests in to a later PR if that's best for people.

rmcgibbo:
My pref. would be to merge. If there are bugs to be found, i suspect they're shallowest with everyone who's using `openmm-dev` or the master branch using it.

rmcgibbo:
(After merging, it would be a good idea to make sure that @jchodera's openmm-dev build system is using a latest enough version of swig to use this feature.)

jchodera:
> (After merging, it would be a good idea to make sure that @jchodera's openmm-dev build system is using a latest enough version of swig to use this feature.)

Whoops, I just saw this!

Looking into upgrading swig now.

jchodera:
The `osx` builder uses swig 3.0.2
```
mski1620:~ jenkins-slave$ swig -version
SWIG Version 3.0.2
```
and the `linux` VM should be using swig 3.0.2 as well, installed via `conda`:
```
[vagrant@localhost ~]$ swig -version
SWIG Version 3.0.2
```
though it may not have been using this version initially.

jchodera:
I see the problem.  The linux VM was originally using `swig` installed within Miniconda, but my Jenkins scripts were blowing this away and reinstalling Miniconda without reinstalling conda `swig`.   The Mac was similarly using the system `swig`, which was thankfully more recent.

`swig 3.0.2` is now available via binstar:
https://binstar.org/anaconda/swig

We should just be using that.  I'll submit a PR to update the `openmm` and `openmm-dev` recipes to specify it as a `build` prerequisite.

jchodera:
I see you already have it specified in the `openmm` recipe:
```
    - swig   # [not win]                                                                                                                                                                       
```

jchodera:
It looks like I don't need to do any upgrading of `swig` on the build boxes then!

rmcgibbo:
I actually think swig 3.0.6 is required to fully use this feature.

jchodera:
D'oh.

jchodera:
OK, the most recent binstar version is 3.0.2, but the [`conda/conda-recipes/swig` recipe](https://github.com/conda/conda-recipes/tree/master/swig) is for 3.0.6.  Can we just pull that into `omnia-/conda-recipes/` and things will automagically work?

You might want to explicitly add `swig >=3.0.6` to the `openmm` and `openmm-dev` recipes then?

PR #1037, opened 2015-07-14
Fixed error in constructing CCMA matrix

### LINKED ISSUE #1035: NaNs with 6.3, not 6.2
Using the following (state.xml, integrator.xml, system.xml) triplet: http://stanford.edu/~rmcgibbo/wu.tar.bz2, I get an explosion with OpenMM 6.3, but not with 6.2. (These xml files come from FAHBench.)

```
from os.path import join
from simtk.openmm import *
from simtk import unit as units
from six.moves.urllib.request import urlopen
import tarfile
import tempfile
import shutil
print(Platform.getOpenMMVersion(), version.version)


tempdir = tempfile.mkdtemp()
with urlopen('http://stanford.edu/~rmcgibbo/wu.tar.bz2') as f:
    with open(join(tempdir, 'wu.tar.bz2'), 'wb') as fout:
        fout.write(f.read())

    tf = tarfile.open(join(tempdir, 'wu.tar.bz2'),  mode='r:bz2')
    system = XmlSerializer.deserialize(tf.extractfile('dhfr/system.xml').read().decode('utf-8'))
    integrator = XmlSerializer.deserialize(tf.extractfile('dhfr/integrator.xml').read().decode('utf-8'))
    state = XmlSerializer.deserialize(tf.extractfile('dhfr/state.xml').read().decode('utf-8'))
shutil.rmtree(tempdir)

context = Context(system, integrator)
context.setState(state)

print(context.getPlatform().getName())
for key in context.getPlatform().getPropertyNames():
    print(key, context.getPlatform().getPropertyValue(context, key))

print(context.getState(getEnergy=True).getPotentialEnergy())
integrator.step(1)
positions = context.getState(getPositions=True).getPositions(asNumpy=True)._value
print(positions)
```

With 6.2,  I have no problem
```
6.2 6.2.0.dev-bf8de1f
OpenCL
OpenCLDeviceIndex 0
OpenCLDeviceName Tahiti
OpenCLPlatformIndex 0
OpenCLPlatformName AMD Accelerated Parallel Processing
OpenCLPrecision single
OpenCLUseCpuPme false
-337078.90986203356 kJ/mol
[[ 3.86168408  2.00907898  3.94630194]
 [ 3.84794426  2.07484078  4.02171803]
 [ 3.95497465  2.01952648  3.9090364 ]
 ..., 
 [ 5.14553118  5.82013035  5.09981012]
 [ 5.08554411  5.84606075  5.16974926]
 [ 5.14642143  5.72447443  5.10319376]]
```

But with 6.3, I get NaNs
```
$ python test-nan.py
6.3 6.3.0.dev-5246ed6
OpenCL
OpenCLDeviceIndex 0
OpenCLDeviceName Tahiti
OpenCLPlatformIndex 0
OpenCLPlatformName AMD Accelerated Parallel Processing
OpenCLPrecision single
OpenCLUseCpuPme false
-337078.9317334164 kJ/mol
[[        nan         nan         nan]
 [        nan         nan         nan]
 [        nan         nan         nan]
 ..., 
 [ 5.14553118  5.82013035  5.09981012]
 [ 5.08554411  5.84606075  5.16974926]
 [ 5.14642143  5.72447443  5.10319376]]
```

Anyone have any idea what's going on, or how to debug this? The initial energy is basically identical -- it all happens during the first `step()`.

peastman:
Fixes #1035.

### DIFF
--- a/platforms/reference/src/SimTKReference/ReferenceCCMAAlgorithm.cpp
+++ b/platforms/reference/src/SimTKReference/ReferenceCCMAAlgorithm.cpp
@@ -33,24 +33,22 @@
 #include "openmm/Vec3.h"
 #include "openmm/internal/ThreadPool.h"
 #include <map>
+#include <utility>
 
-using std::map;
-using std::pair;
-using std::vector;
-using std::set;
 using namespace OpenMM;
+using namespace std;
 
 // This class extracts columns from the inverse matrix one at a time.  It is done in parallel,
 // since this can be very slow.
 
 class ExtractMatrixTask : public ThreadPool::Task {
 public:
-    ExtractMatrixTask(int numConstraints, vector<vector<pair<int, RealOpenMM> > >& matrix, const vector<RealOpenMM>& distance, RealOpenMM elementCutoff,
+    ExtractMatrixTask(int numConstraints, vector<vector<pair<int, RealOpenMM> > >& transposedMatrix, const vector<RealOpenMM>& distance, RealOpenMM elementCutoff,
                       const int* qRowStart, const int* qColIndex, const int* rRowStart, const int* rColIndex, const double* qValue, const double* rValue) :
-                numConstraints(numConstraints), matrix(matrix), distance(distance), elementCutoff(elementCutoff), qRowStart(qRowStart), qColIndex(qColIndex),
+                numConstraints(numConstraints), transposedMatrix(transposedMatrix), distance(distance), elementCutoff(elementCutoff), qRowStart(qRowStart), qColIndex(qColIndex),
                 rRowStart(rRowStart), rColIndex(rColIndex), qValue(qValue), rValue(rValue) {
     }
-    
+
     void execute(ThreadPool& pool, int threadIndex) {
         vector<double> rhs(numConstraints);
         for (int i = threadIndex; i < numConstraints; i += pool.getNumThreads()) {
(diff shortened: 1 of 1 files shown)


rmcgibbo:
What's the test case we should have written to find this earlier. I'm kind of surprised it hid this long.

platforms/reference/src/SimTKReference/ReferenceCCMAAlgorithm.cpp
```diff
@@ -33,24 +33,22 @@
 #include "openmm/Vec3.h"
 #include "openmm/internal/ThreadPool.h"
 #include <map>
+#include <utility>
```
rmcgibbo: is this used?

peastman:
> What's the test case we should have written to find this earlier?

I'm not sure.  We do have lots of tests that involve CCMA.  Possibly they didn't have enough variation between the distances for the different constraints?

platforms/reference/src/SimTKReference/ReferenceCCMAAlgorithm.cpp
peastman: Yes.  That's where make_pair() is defined.

rmcgibbo: right.

rmcgibbo:
Do you know what the relationship between the inverse constraint matrix from before this PR to after it is?  It seems like it would have just been all messed up, which makes me surprised that it seemed to work at all.

peastman:
Not that messed up.  Each element depends most strongly on the angle between the two constrained directions, and that's symmetric.  But it's multiplied by distance1/distance2.  If both constraints have the same distance, the symmetric elements will be exactly equal.  If they don't, they'll be slightly different.  But since most bond lengths are in a fairly narrow range, they usually won't be *that* different.

Also, if your matrix is wrong, that can affect the radius of convergence and the number of iterations needed to converge.  But as long as it converges, that doesn't affect the accuracy of the final result.

rmcgibbo:
Sounds like I should go back and read the paper.

rmcgibbo:
Right. Okay, I see how this works. It's actually very tricky to build a test that catches this because of the very fact that makes the CCMA derivation work (that you can use an approximate inverse jacobian and still get a convergent algorithm).

Do you want me to make a test with the positions and constraints from the state.xml that triggered this?

rmcgibbo:
It will be a little large, unfortunately.

rmcgibbo:
Anyways, +1 to merge this. It fixes that particular state.xml and it's correct.

peastman:
I don't want a test that big, both because of the size and because of how long it would take to run.  But if you can put together something smaller, that would be great.  A case where the distances for different constraints vary widely would be most sensitive to it, I think.

rmcgibbo:
I tried, but I couldn't come up with anything.

PR #1044, opened 2015-07-20
Build without doxygen

mpharrigan:
Cmake would refuse to continue without doxygen.

### DIFF
--- a/docs-source/CMakeLists.txt
+++ b/docs-source/CMakeLists.txt
@@ -3,8 +3,7 @@
 #
 
 find_package(Doxygen QUIET)
-mark_as_advanced(CLEAR DOXYGEN_EXECUTABLE)
-IF(DOXYGEN_EXECUTABLE)
+IF(DOXYGEN_FOUND)
     # Generate C++ API documentation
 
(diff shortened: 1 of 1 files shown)


rmcgibbo:
doxygen is required for building the swig wrappers tho.

mpharrigan:
If you don't want swig wrappers

rmcgibbo:
what happens w/ this PR if you don't have doxygen but enable the
OPENMM_BUILD_PYTHON_WRAPPERS option?

mpharrigan:
You're right. That's the source of the error:
```
  Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
  C:/Program Files (x86)/CMake/share/cmake-3.2/Modules/FindPackageHandleStandardArgs.cmake:374 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files (x86)/CMake/share/cmake-3.2/Modules/FindDoxygen.cmake:107 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  wrappers/python/CMakeLists.txt:122 (find_package)
```

The proposed change in this PR does nothing

PR #1056, opened 2015-07-29
Optimizations to custom integrator

peastman:
These are fairly minor optimizations that slightly reduce the cost of per-dof computations.


jchodera:
Hooray!

peastman:
Don't expect anything huge from this, but it does make a difference.  I'm looking to see what else can be done.

PR #1059, opened 2015-07-29
Added badges for Jenkins openmm-dev conda package builds.

jchodera:
This adds badges for the Jenkins builds of the `openmm-dev` conda package so we can tell when it breaks.


peastman:
Yay!

jchodera:
I can make it prettier if you don't like the formatting.

peastman:
As you like.  It would be good to put a label in front of the travis badge, since it's now unclear what that refers to.  Also, if we're going to have several lines of badges, let's move them down a bit so the "Introduction" is at the top.

PR #1061, opened 2015-07-29
Badging update

jchodera:
* labeled Travis badge
* moved all badges to below Introduction
* added binstar badges


jchodera:
Preview [here](https://github.com/jchodera/openmm/blob/add-jenkins-badges/README.md).

peastman:
Looks good.

Issue #1058, opened 2015-07-29
Jenkins builds of openmm-dev dying


jchodera:
They seem to be dying with this weird error:
https://jenkins.choderalab.org/job/conda-openmm-dev-linux-vagrant-2/61/console
```
[ 83%] Building CXX object platforms/cuda/tests/CMakeFiles/TestCudaEwald.dir/TestCudaEwald.cpp.o
/home/vagrant/miniconda/conda-bld/work/platforms/cuda/tests/TestCudaCustomIntegrator.cpp:838:9: error: call to 'abs' is ambiguous
        ASSERT_EQUAL_TOL(i+1, integrator.getGlobalVariable(0), 1e-5);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/vagrant/miniconda/conda-bld/work/openmmapi/include/openmm/internal/AssertionUtilities.h:55:66: note: expanded from macro 'ASSERT_EQUAL_TOL'
#define ASSERT_EQUAL_TOL(expected, found, tol) {double _scale_ = std::abs(expected) > 1.0 ? std::abs(expected) : 1.0; if (!(std::abs((expected)-(found))/_scale_ <= (tol))) {std::stringstream details; details << "Expected "<<(expected)<<", found "<<(found); throwException(__FILE__, __LINE__, details.str());}};
                                                                 ^~~~~~~~
/usr/bin/../lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/cmath:94:3: note: candidate function
  abs(double __x)
  ^
/usr/bin/../lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/cmath:98:3: note: candidate function
  abs(float __x)
  ^
/usr/bin/../lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/cmath:102:3: note: candidate function
  abs(long double __x)
  ^
/home/vagrant/miniconda/conda-bld/work/platforms/cuda/tests/TestCudaCustomIntegrator.cpp:838:9: error: call to 'abs' is ambiguous
        ASSERT_EQUAL_TOL(i+1, integrator.getGlobalVariable(0), 1e-5);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1062, opened 2015-07-30
Fix issue with test type casting.

jchodera:
This should address #1058


jchodera:
I unfortunately don't have an easy way to test if these eliminate the jenkins build problem, @peastman.  If these changes are correct, maybe it's easiest to merge these and see if the automated builds resolve on their own?

PR #1069, opened 2015-08-03
Install SFMT headers

### LINKED ISSUE #1066: SFMT headers aren't installed
When developing a plugin, it's convenient to use the OpenMM_SFMT RNG, but unfortunately the header files are not installed in internal/ or anything.

rmcgibbo:
Resolves #1066


peastman:
The failure seems to be stochastic.  TestReferenceRpmd failed twice in static mode, but passed the first time in dynamic mode.  I don't think it's related to this.

PR #1070, opened 2015-08-03
Added a more useful string representation of atoms, residues, and chains

peastman:
This is mainly to help with debugging.


peastman:
For some reason travis never ran on this.  But I'm pretty sure it hasn't broken anything.

PR #1078, opened 2015-08-07
add RPMDMonteCarloBarostat header to OpenMM.h

zonca:
Trying to compile the [mbpol plugin](https://github.com/paesanilab/mbpol_openmm_plugin) with OpenMM 6.3, I noticed that the `RPMDMonteCarloBarostat` header is not included in `OpenMM.h`.
However, it is added in the `swig` headers, so when compiling the Python wrapper for the plugin, it fails with the error:

    MBPolPluginWrapper.cpp:18178:42: error: ‘RPMDMonteCarloBarostat’ is not a member of ‘OpenMM’

Adding this fixes the compilation issue on my machine, so I wonder if this issue is more general.

I thought I tested compilation on the 6.3 beta release, but I guess I overlooked something in the configuration.

### DIFF
--- a/openmmapi/include/OpenMM.h
+++ b/openmmapi/include/OpenMM.h
@@ -57,6 +57,7 @@
 #include "openmm/MonteCarloAnisotropicBarostat.h"
 #include "openmm/MonteCarloBarostat.h"
 #include "openmm/MonteCarloMembraneBarostat.h"
+#include "openmm/RPMDMonteCarloBarostat.h"
 #include "openmm/NonbondedForce.h"
 #include "openmm/Context.h"
 #include "openmm/OpenMMException.h"


peastman:
OpenMM.h only includes classes that are defined in the main library.  Headers for plugins, such as RPMD, need to be included separately.

We should add a top level OpenMMRPMD.h header though.  I didn't create one before, because the RPMD plugin only defined a single public class.  Now that it has two of them, there should be a header that includes both at once, similar to the OpenMMAmoeba.h and OpenMMDrude.h headers.

zonca:
Thanks, do you recommend a workaround for compiling plugins with 6.3?

rmcgibbo:
Are you talking about compiling the plugin's swig wrappers? Just include the extra headers, e.g. https://github.com/peastman/openmmexampleplugin/blob/master/python/exampleplugin.i#L21-L24

zonca:
exactly!
this solved my issue...and tracking `exampleplugin` is a good idea for the future!
thank you very much

PR #1080, opened 2015-08-07
Fixed errors executing force field scripts on Python 3

peastman:
There was a change in behavior between Python 2 and 3 that caused some scripts embedded in force field XML files to stop working: in particular, the one used for the CHARMM polarizable force field.  This change fixes the problem.  See https://simtk.org/forums/viewtopic.php?f=161&t=5888 for more details.


rmcgibbo:
It would be nice to slowly switch the python code over to being written directly in 2/3 compatible syntax, and not use 2to3.

peastman:
Sounds reasonable.  That wouldn't have affected this, though.  The problem was a change in behavior, not syntax.

Issue #1074, opened 2015-08-05
Add PDBx/mmCIF reporter


swails:
An enhancement stub to remind me to do this

cc #1072

swails:
Added by #1081

PR #1081, opened 2015-08-08
Add PDBxReporter class

### LINKED ISSUE #1072: "Fixed width overflow detected" in PDBFile.writeFile with keepIds=True
We're trying to preserve original residue numberings in models generated by `simtk.openmm.app.Modeller` from constructs that match standard Uniprot numberings, but contain lots of solvent molecules. 

Writing out PDB files after the `Modeller` stage using 
```python2.7
        app.PDBFile.writeFile(fixer.topology, fixer.positions, outfile, keepIds=True)  
```
gives the error
```
Fixed width overflow detected
```
presumably because too many waters are included.

Complete input files here:
https://gist.github.com/jchodera/d4e967c29fb124d76d44

Example invocation (requires `pdbfixer-dev`):
```
python setup_mutations_ensembler.py --source_pdbfile='MTOR_HUMAN_D0-selected.pdb' --output_directory='MTOR_HUMAN_D0-mutations'
```

Is there some way we could have `writeFile` preserve numbering for non-water residues but do *something* to just not die on writing water residues (where we essentially *never* care about numbering being preserved)?

swails:
Add a `PDBxReporter` class that writes a PDBx/mmCIF trajectory just like the PDB reporter does.  The mmCIF writer implemented here is my own code and unlike the RCSB module does not require storing everything in memory before writing the file, so it can be written continuously and does not have to store stuff in memory.

Also adds two tests and makes sure that the CIF files generated can actually be read back in by `PDBxFile` correctly.

There are also a couple other fixes, like changing some implicit relative imports to explicit relative imports (part of the "single code-base for Py2 and Py3 without using `2to3`" -- still a lot to do on that regard, though).

The last change I made here was to have PDBReporter and PDBxReporter automatically call `flush()` on any file object opened for that reporter following any frame writes as long as that file object supports flushing.

This closes #1072 and #1074

### DIFF
--- a/wrappers/python/simtk/openmm/app/pdbxfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbxfile.py
@@ -240,12 +240,12 @@ class PDBxFile(object):
         if vectors is not None:
             a, b, c, alpha, beta, gamma = computeLengthsAndAngles(vectors)
             RAD_TO_DEG = 180/math.pi
-            print('_cell.length_a     %10.4f' % a_length*10, file=file)
-            print('_cell.length_b     %10.4f' % b_length*10, file=file)
-            print('_cell.length_c     %10.4f' % c_length*10, file=file)
-            print('_cell.angle_alpha  %10.4f' % alpha*RAD_TO_DEG, file=file)
-            print('_cell.angle_beta   %10.4f' % beta*RAD_TO_DEG, file=file)
-            print('_cell.angle_gamma  %10.4f' % gamma*RAD_TO_DEG, file=file)
+            print('_cell.length_a     %10.4f' % (a*10), file=file)
+            print('_cell.length_b     %10.4f' % (b*10), file=file)
+            print('_cell.length_c     %10.4f' % (c*10), file=file)
+            print('_cell.angle_alpha  %10.4f' % (alpha*RAD_TO_DEG), file=file)
+            print('_cell.angle_beta   %10.4f' % (beta*RAD_TO_DEG), file=file)
+            print('_cell.angle_gamma  %10.4f' % (gamma*RAD_TO_DEG), file=file)
             print('##', file=file)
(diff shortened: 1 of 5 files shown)


wrappers/python/simtk/openmm/app/pdbxfile.py
```diff
@@ -201,3 +203,125 @@ def getPositions(self, asNumpy=False, frame=0):
...
+        """
+        if entry is not None:
+            print('data_%s' % entry, file=file)
+        else:
+            print('data_cell', file=file)
+        print("# Created with OpenMM %s, %s" % (Platform.getOpenMMVersion(), str(date.today())), file=file)
+        print('#', file=file)
+        vectors = topology.getPeriodicBoxVectors()
+        if vectors is not None:
+            (a, b, c) = vectors.value_in_unit(angstroms)
```
peastman: Why not just call computeLengthsAndAngles() here?

swails: I didn't know about it.  I adapted `PDBFile` for this, and `computeLengthsAndAngles` wasn't used there, either.

I can change both if you want.

peastman: That would be great.  I must have missed replacing the duplicate code there.

peastman:
This looks great!  One very minor point: there are several places where the docstrings still refer to "PDB files".

swails:
OK, `PDBFile` and `PDBxFile` have now both been updated to use `computeLengthsAndAngles`.  Since I was modifying `pdbfile.py`, anyway, I went ahead and made the changes necessary to support both Python 2 and Python 3 in the same code base (using the print function and absolute imports).

Assuming this passes the tests and you don't see anything else, this should be ready to merge.

peastman:
Thank you!

PR #1083, opened 2015-08-11
Single py2/3 codebase

rmcgibbo:
Not sure if this is sufficient or not.


rmcgibbo:
(Besides the changes to `setup.py`, I just ran python-moderize http://python-modernize.readthedocs.org/en/latest/ with the --no-six option)

peastman:
Unless there's a good reason, I'd prefer not to make any changes to internal/pdbx.  That's the PDBx library from RCSB, and I'd like to keep it as close as possible to the original, in case we need to upgrade to a newer version at some point.  The changes to it mostly look stylistic, not functional.

What were the changes to pdbstructure.py?  Github is showing the entire content of the file as having been replaced.

swails:
> Unless there's a good reason, I'd prefer not to make any changes to internal/pdbx.

It's impossible to support Python 2 and Python 3 in the same code base that way (you need 2to3).  So the good reason is to get rid of 2to3.  Is that a good enough reason?  (For instance, that code makes extensive use of `dict.has_key`, which was removed in Python 3 in favor of `__contains__`).

I already went through this code and made the requisite changes needed to run under Python2 and Python3 in the same code base, and there are definitely changes that need to be made for that.  I don't think these changes will make it harder to update in the future if need be -- these changes will just be blown away.

swails:
> Not sure if this is sufficient or not.

Can you test under python 3 as well inside Travis?  If we're going to support Python 2 and Python 3 without `2to3`, then it'll be important to make sure we don't break Python 3 support with any of our changes.  It's far more likely *to* break Python 3 if we stop using 2to3.

I can add to this PR for that if you want.

rmcgibbo:
> What were the changes to pdbstructure.py? Github is showing the entire content of the file as having been replaced.

It looks the that file had windows line endings or something which the tool fixed. Run `git diff` with `-w` to ignore that. I pasted the diff here: https://gist.github.com/rmcgibbo/cc2d66eb3aa3513d0f24

rmcgibbo:
I had to open https://github.com/travis-ci/apt-package-whitelist/issues/770, https://github.com/travis-ci/apt-package-whitelist/issues/769, and  https://github.com/travis-ci/apt-package-whitelist/issues/768 to get the required python3 packages, unless there's a different way of doing it.

peastman:
I don't think that getting rid of 2to3 is itself a goal.  It works nicely and doesn't cause any problems.  For general code cleanliness, I agree that it's nice if as much as possible uses the newer syntax, etc.  But for 3rd party libraries that we don't intend to modify and will rarely have reason to look at, that's a minor point, whereas diverging from the original is a significant concern.

Perhaps what you want to do is get RCSB to update their library?

mpharrigan:
To ignore whitespace changes you can add `?w=true` to the end of the github url

https://github.com/pandegroup/openmm/pull/1083/files?w=true

rmcgibbo:
Either way the PDBx thing doesn't matter. If we want to use their library, but it doesn't support all the platforms that we want to support (py3), then we have to modify it somehow. Either that happens in the source code, or it happens at install time. But the code that the user is executing and that shows up in the traceback when there's a bug is the transformed code, one way or the other.

If someone wants to file a PR to the RCSB, great. Or if someone wants to graft in a new copy of their code into openmm, just run `pip install modernize; python-modernize --no-six *.py` -- that's all I did.

mpharrigan:
+1 removing 2to3 as a goal unto itself. Rewriting code at compile time obscures what is being executed from the user and removes control from the developer.

I think 2to3 is not widely used by the python community. It doesn't receive much support. A unified codebase (esp since >3.3 and >2.6) is the go-to

peastman:
> Either that happens in the source code, or it happens at install time. But the code that the user is executing and that shows up in the traceback when there's a bug is the transformed code, one way or the other.

The point is that in the future, we may want to upgrade to a newer version of the library.  When that happens, the fewer changes that need to be backport, the better.

rmcgibbo:
Since it's mechanical, I don't think we should waste too much time worrying about it.

swails:
> I think 2to3 is not widely used by the python community.

What makes you say this?  It's incorporated directly into `distutils` and is one of the recommended approaches for supporting both Python 2 and Python 3 in the same codebase.  Also, I believe 2to3 works on the swig-generated `openmm.py`, in which case, generating a single codebase compatible with both Python 2 and Python 3 is going to be a little harder (unless it already "just works").

The approach I had been taking was to make these changes incrementally.  e.g., Convert implicit relative imports to absolute (or explicit relative) imports.  Use `__future__.print_function`.  That kind of thing, within code that I was actively modifying at the time.  Since I'm adding tests at the same time, I could be more confident about making sure I don't break anything given our less-than-complete test coverage in the app layer and lack of regular Python 3 testing.

This seems like a sledgehammer type of approach, and I'm worried that it will introduce problems at first for a benefit that appears from this thread to be debatable.

rmcgibbo:
"Sledgehammer"? This PR is a couple of single line changes to `print()`, `raise`, and `has_key`.

swails:
The dos ending conversion admittedly makes the diff look more intimidating.  But it's doing a bit more than those 3 things.  Implicit imports, `xyz.next()` -> `next(xyz)`, `range` -> `list(range)`, `zip` -> `list(zip)`, ...

What you're changing here is not just what's showing up in the diff -- it's also changing the `2to3` code base that is autogenerated in master.  The way we have things now, we effectively have two Python code bases and only one of them is tracked.  It's the changes to the "invisible" code base that concerns me most, not the ones we can see in the diff.

Nothing about Python 2 -> Python 3 is a huge difference in and of itself, but it's a lot of little changes.  It looks like you're beginning to add Python 3 testing to Travis, which will alleviate a lot of my concerns.

I'm +0 on making elimination of `2to3` a goal in and of itself (i.e., it would be nice to have, but not something I think will be a real functional or maintenance win in the near-term; I'd be happy to hear counterexamples).

That said, the `pdbx` library is small, lightweight, hasn't changed in 3+ years, and the changes to it here are small.  So I don't think changes to that are as concerning as other changes to things like asmjit or other dependencies we ship.

peastman:
This is failing on Python 3.

rmcgibbo:
Got python3 tests running on travis-ci, but this has be stumped:

```
$ nosetests -vv
nose.config: INFO: Ignoring files matching ['^\\.', '^_', '^setup\\.py$']
  File "/home/travis/virtualenv/python3.4.2/bin/nosetests", line 11, in <module>
    sys.exit(run_exit())
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/core.py", line 121, in __init__
    **extra_args)
  File "/opt/python/3.4.2/lib/python3.4/unittest/main.py", line 93, in __init__
    self.runTests()
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/core.py", line 207, in runTests
    result = self.testRunner.run(self.test)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/core.py", line 62, in run
    test(result)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/suite.py", line 178, in __call__
    return self.run(*arg, **kw)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/suite.py", line 225, in run
    test(orig)
  File "/opt/python/3.4.2/lib/python3.4/unittest/suite.py", line 87, in __call__
    return self.run(*args, **kwds)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/suite.py", line 73, in run
    for test in self._tests:
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/suite.py", line 100, in _get_tests
    for test in self.test_generator:
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/loader.py", line 184, in loadTestsFromDir
    entry_path, discovered=True)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/loader.py", line 429, in loadTestsFromName
    discovered=discovered)
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/nose/loader.py", line 329, in loadTestsFromModule
    elif isfunction(test) and self.selector.wantFunction(test):
  File "/opt/python/3.4.2/lib/python3.4/inspect.py", line 174, in isfunction
    return isinstance(object, types.FunctionType)
NameError: Unknown C global variable
```

rmcgibbo:
Okay, it seems like swig 2.0.4 produced incorrect code with python3.

rmcgibbo:
@peastman is it necessary that [two](https://github.com/pandegroup/openmm/blob/4b46bee6561ad97ab8be5361c5dac546c8431987/wrappers/python/tests/TestAmberPrmtopFile.py#L291) implicit solvent [tests](https://github.com/pandegroup/openmm/blob/50442b0a2f072ed95181993b2d667d6b9c1cde22/wrappers/python/tests/TestCharmmFiles.py#L134) run on the CPU platform? Looks like it's not available on Appveyor, which is the only failure.

peastman:
In principle there's no reason they couldn't use the reference platform instead, but how long do they take to run?  I wouldn't want to make the tests significantly slower on all other platforms.

And seriously, it doesn't support the CPU platform?  What kind of CPUs are they using?  It should work on just about anything made in the last five years!

rmcgibbo:
> And seriously, it doesn't support the CPU platform? What kind of CPUs are they using? It should work on just about anything made in the last five years!

I'm not sure how to get the equivalent of /proc/cpuinfo from the command line on windows.

On my laptop, when I change TestAmberPrmtopFile.test_ImplicitSolventForces to use the Reference platform, the execution time drops from ~1.5 seconds to ~1.2 seconds.

peastman:
Ok, that's fine then.

rmcgibbo:
Kk. Tests pass on py2 and py3 (this PR adds py3 tests to travis). This should be ready to review/merge.

peastman:
I'm still -1 on the changes to pdbx.

jchodera:
What if the `pdbx` changes were contributed back upstream to the wwpdb?

rmcgibbo:
@peastman, we could make cmake run 2to3 on specifically those files when it copies them into the staging directory. at least that way it would be restricted to that subset of enumerated files.

mpharrigan:
The pdbx code is going to be changed when running on py3. Either an automated tool is going to do it at install time or @rmcgibbo is going to do it here. The latter feels safer to me

swails:
> The pdbx code is going to be changed when running on py3. Either an automated tool is going to do it at install time or @rmcgibbo is going to do it here.

True.  But in one case we are maintaining the same *exact* code as the RCSB releases and in another we are maintaining something slightly different.  (For ParmEd I chose "something slightly different".)

> The latter feels safer to me

It's really no safer, though.  In fact, @rmcgibbo has stated that it's functionally equivalent:

> Since it's mechanical, I don't think we should waste too much time worrying about it.

Code changes -- any of them -- are only as safe as the test coverage is complete.  I'd rather see these changes accepted and merged because it feels better to me, but I can't really claim to have a good *real* reason.

rmcgibbo:
peter, what's your opinion on https://github.com/pandegroup/openmm/pull/1083#issuecomment-135899201

peastman:
Since other people prefer this approach, I guess I can go with it.  I'm not thrilled with it, but my opinion isn't strong enough to argue about it.

PR #1086, opened 2015-08-12
Fix two tests for python3

rmcgibbo:


peastman:
It's been 23 minutes since you posted this.  Travis finished in under 8 minutes.  And AppVeyor still hasn't even started running.  I don't think we want to have to wait that long for every PR.  Maybe this can be used as a secondary test: we merge if travis passes, but then get notified if AppVeyor fails?

swails:
+1

rmcgibbo:
That sounds good. Do you want an email or something for failures?

jchodera:
What if we foot the $19/mo bill for super fast builds?

rmcgibbo:
Naw. It was going super fast earlier for me today. I think it's just stuck or something.

PR #1087, opened 2015-08-12
Show fewer warnings on Visual Studio

peastman:

PR #1088, opened 2015-08-12
Update appveyor.yml

rmcgibbo:

PR #1090, opened 2015-08-13
Eliminate extra warning about overriding the warning level

rmcgibbo:
cc #1087, https://github.com/pandegroup/openmm/pull/1087#issuecomment-130478720


peastman:
We have about 800 copies of this error message:

```
C:\projects\openmm\tests\TestVectorize8.cpp(180) : warning C4305: 'argument' : truncation from 'double' to 'float' 
```

I should look into how to make that go away!

By the way, it again had a stochastic test fail, didn't rerun it, and then didn't run any python tests.

rmcgibbo:
maybe i can add /wd4305

rmcgibbo:
> By the way, it again had a stochastic test fail, didn't rerun it, and then didn't run any python tests.

Should be fixed by #1088

PR #1096, opened 2015-08-13
Ensure neighbor list is large enough on first step

peastman:
This is a workaround for simulating large systems on Windows.  If any kernel takes longer than 2 seconds, the OS resets the driver.  That was sometimes happening because there wouldn't initially be enough space for the neighbor list, so it would fall back to the O(n^2) algorithm for the first step.


swails:
FTFY
            -- Windows

PR #1098, opened 2015-08-14
Appveyor faster

rmcgibbo:


rmcgibbo:
Also, we could switch the optimization options to something lower like "/Od"

peastman:
Or we could pay $19/month to get faster CPUs, multiple cores, and possibly a machine that is capable of running the CPU platform tests.

> Also, we could switch the optimization options to something lower like "/Od"

Tests should be run with the same compile options used for release builds.  Anyway, that would be as likely to make it slower as faster.  Builds would take less time, but tests would take more.

rmcgibbo:
> Or we could pay $19/month to get faster CPUs.

Go for it.

jchodera:
I took a look at the pricing again and I think faster builds start at $59/mo:
> "Basic" and "Free" plans run builds on "Azure" environment where each build worker has 1 CPU core and 1.75 GB of RAM ("Small" instance). "Pro" and "Premium" plans run builds on dedicated hardware with faster CPUs, SSD drives, 1 GB network uplink and use Hyper-V for creating isolated build workers.

http://www.appveyor.com/pricing

Would this still be worth it?

swails:
Well right now the build goes over 40 minutes, so it doesn't even finish...

peastman:
That starts sounding like less of a deal.  It doesn't take many months at that rate before it's less expensive to just buy our own build machine.

rmcgibbo:
Maybe it wouldn't be _as useful_ without using the same optimization options, but at least it does finish with `/Od /Ob0`. But we don't need the end all and be all of testing, given that we were doing just fine before we had this set up at all.

swails:
> Maybe it wouldn't be as useful without using the same optimization options, but at least it does finish with /Od /Ob0. But we don't need the end all and be all of testing, given that we were doing just fine before we had this set up at all.

+1

PR #1101, opened 2015-08-21
Implemented spherical harmonics for multipoles

peastman:
When using direct polarization, this produces an overall speedup of about 35%.  With mutual polarization, the speedup is only about 5%.  Once we get extrapolated polarization, that should be somewhere in between.  This also led to a significant overall simplification of the CUDA AMOEBA code.


peastman:
The failure is an unrelated stochastic test (TestReferenceDrudeLangevinIntegrator).

Issue #1109, opened 2015-08-26
Index not rendering correctly in OpenMM 6.3 Users Guide


jchodera:
Not sure we ever looked at this before, but it's mostly empty:
http://docs.openmm.org/6.3.0/userguide/genindex.html

jchodera:
Whoops, I'm wrong about the search box!

rmcgibbo:
Looks like that page hasn't really had any content in the past either: http://docs.openmm.org/6.2.0/userguide/genindex.html, http://docs.openmm.org/6.1.0/userguide/genindex.html

peastman:
The index doesn't get populated automatically.  You have to tell it what to include there.  And given that you can search the documentation, I don't see much value in having an index at all.  Maybe there's a way we can tell it to remove it.

jchodera:
Good point, @peastman.

For completeness, here is the documentation on creating index markdown entries:
http://sphinx-doc.org/markup/misc.html#index-generating-markup

swails:
I believe Sphinx auto generates that page. It gets filled when it's used to generate API docs (since I think apidoc populates it by default). 

But we just use Sphinx to build the manuals, so that doesn't happen. I'm not sure how to suppress the generation of that page, but it is probably easy to google.

PR #1110, opened 2015-08-26
Turn off HTML index generation in Sphinx.

jchodera:
Addresses #1109 (I think!).


jchodera:
Seems to work on my machine.  No more index in the HTML version!

peastman:
Looks good.  Turn it off for the Developer Guide too?

jchodera:
Good catch!

PR #1116, opened 2015-08-31
Fix OpenCL Nonbonded Force on certain CPU devices

rmcgibbo:
Fixes https://github.com/pandegroup/openmm/issues/1079#issuecomment-136193271

### DIFF
--- a/platforms/opencl/src/kernels/nonbonded_cpu.cl
+++ b/platforms/opencl/src/kernels/nonbonded_cpu.cl
@@ -237,9 +237,9 @@ __kernel void computeNonbonded(
         if (numTiles <= maxTiles) {
             x = tiles[pos];
             real4 blockSizeX = blockSize[x];
-            singlePeriodicCopy = (0.5f*periodicBoxSize.x-blockSizeX.x >= CUTOFF &&
-                                  0.5f*periodicBoxSize.y-blockSizeX.y >= CUTOFF &&
-                                  0.5f*periodicBoxSize.z-blockSizeX.z >= CUTOFF);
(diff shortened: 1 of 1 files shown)


rmcgibbo:
Note that this compilation fix leads to an error in `TestOpenCLNonbondedForce` in the dispersion correction.

```
vagrant@precise64:~/openmm/build$ ./TestOpenCLNonbondedForce 
exception: Assertion failure at TestOpenCLNonbondedForce.cpp:721.  Expected -1304.2, found -1304
```

The test machine is an Ubuntu 12.04 VM running inside a Mac Mini. The VM has the AMD APP SDK, and the OpenCL device is the CPU (OpenCL info: https://gist.github.com/rmcgibbo/7a20862e5f34dc3bc806)

rmcgibbo:
Actually, there are quite a bit more OpenCL test errors.

swails:
Maybe this isn't the right fix?  I *think* `MAX_CUTOFF` and `CUTOFF` can be different on account of different force groups can have different nonbonded cutoffs for their nonbonded forces.

rmcgibbo:
No, it's the same change as https://github.com/pandegroup/openmm/commit/8ae2eba560313a211dab115a3ef73807ff911177, which is a fix for the previous PR that let different nonbonded force kernels have different cutoffs.

swails:
Ah, OK.

peastman:
I'm going to merge this since, whatever the other problems are caused by, this fix is clearly needed.

jchodera:
Good catch!

PR #1120, opened 2015-09-03
Created getPMEParametersInContext()

### LINKED ISSUE #1111: Extend NonbondedForce API to calculate, retrieve PME parameters
(Originally discussed in #1108)

It would be useful to expose to users (in both C++ and Python) ways to
* retrieve PME parameters in use in a given `Context`, perhaps via a `NonbondedForce.getPMEParametersInContext(context)` method
* compute the PME parameters that would be used in a specific platform, perhaps via `NonbondedForce.calcPMEParameters(platform)` method

peastman:
Fixes #1111.


peastman:
For some reason travis never ran on this.  I don't know why.  It passed on AppVeyor, though, so I'll go ahead and merge it.

jchodera:
Awesome!

PR #1123, opened 2015-09-04
Legalese at top of PDF, bottom of HTML

### LINKED ISSUE #1047: Reformat "landing" page of sphinx docs
We might consider reformatting the sphinx docs in such a way that the landing page doesn't hit you with a big block of legal text. It's a little offputting.

(orignally at https://github.com/pandegroup/openmm/issues/992#issuecomment-124269718)

swails:
Depending on whether we want to build HTML docs or a PDF, we want the legalese
either at the end or the beginning of the index page. There is no way that *I*
know of to do this with a single index.rst file, so have conf.py automatically
build the index.rst file, and inspect the command-line to see if it's building
an HTML or a PDF.  Then construct the index.rst file in the desired order for
the type of document that's being generated.

Fixes #1047


swails:
Superseded by #1124

PR #1125, opened 2015-09-04
Throw an exception if an illegal device or platform index is specified

### LINKED ISSUE #1034: Strict mode for device id
If the given device id isn't legit, openmm will silently select a device of its choosing. Can we add an option to raise an exception instead. 

I've run into this problem running simulations on multi-gpu clusters where I thought I was spreading jobs across gpus, but since I was outside the range of the number of devices, it was all getting assigned to gpu 0.

It's a pretty surprising behaviour from a moral/philosophical standpoint as well.

peastman:
Fixes #1034.


mpharrigan:
Thanks @peastman !

PR #1128, opened 2015-09-09
Reenabled cuFFT on CUDA 7.5 or later

peastman:
Now that there's an official CUDA release with the bug fixed, we can start using it again.  In my tests, this gives about a 10% speedup on the DHFR benchmark.


jchodera:
Woohoo!

PR #1129, opened 2015-09-10
Use MTSIntegrator for AMOEBA benchmarks

peastman:
Previously we used a 1 fs time step for the AMOEBA benchmarks, but really it's only the bonded forces that need to be evaluated that often.  Now that MTSIntegrator is a standard feature, it makes sense to use it to compute the nonbonded forces half as often.  This leads to a nearly 2x speedup.


jchodera:
Nice!

Issue #1054, opened 2015-07-27
Update ffxml files for AmberTools15


jchodera:
Tagging @rafwiewiora here.  This issue will discuss our conversion and validation of all forcefields distributed with AmberTools15.

@swails: Do you already have some validation scripts to use ParmEd to compare energies between OpenMM-generated energies and Amber-generated energies?

swails:
> @swails: Do you already have some validation scripts to use ParmEd to compare energies between OpenMM-generated energies and Amber-generated energies?

ParmEd makes it easy to compute energies and forces with both OpenMM and Amber (through the sander API), if that's what you mean.

jchodera:
> ParmEd makes it easy to compute energies and forces with both OpenMM and Amber (through the sander API), if that's what you mean.

That's exactly it!  We were just wondering if you had example scripts for this (e.g. the scripts you used to test/validate the GB parameters recently).

swails:
I'll try to dig them up.  Probably won't be until tomorrow though.

I'm not sure how I missed this message in the first place.  Can you open an issue on ParmEd for this with a description of what you're trying to do (for example, what kinds of files you're starting with)?

jchodera:
@rafwiewiora: Can you tackle that?

rafwiewiora:
Done!

rafwiewiora:
I want to touch base on the issues with `sigma = 0` - see #732 and #979 - @jchodera has had expressions blow up because of `sigma = 0`. 

Now, @jchodera thought in #979 that the presence of `epsilon = 0 sigma = 0` atoms in the force fields was caused by [these lines in the process script] (https://github.com/pandegroup/openmm/blob/master/devtools/forcefield-scripts/processAmberForceField.py#L407-L409). This should be changed to `sigma = 1`, happy to do the PR, but I just realized the real culprit is something else. Hydrogens of `class="HO"`, at least in the amber99 fields actually have `epsilon = 0 sigma = 0` parameters. The appropriate line in parm99.dat is 
`  HO          0.0000  0.0000             OPLS Jorgensen, JACS,110,(1988),1657`

It doesn't matter what sigma is when `epsilon = 0`, but clearly we shouldn't have `sigma = 0` 's, so how about adding:
````python
if sigma == 0 and epsilon == 0:
    sigma = 1
````
 to the processAmberForceField.py?

swails:
[...]

PR #1133, opened 2015-09-11
Resolve sigma = 0 problems from processAmberForceField

rafwiewiora:
As discussed in #1054. In case of unknown vdW parameters `sigma, epsilon = 1, 0` will now be assigned; in case Amber forcefield specifies `sigma, epsilon = 0, 0` this will be changed to `1, 0`.


peastman:
Looks good.  Now we need to regenerate the force fields.  I can do that next week.

rafwiewiora:
@peastman: I'm working on converting all the Amber forcefields to ffxml right now, so happy to regenerate the existing ones too.

jchodera:
@rafwoewiora is also using the latest AmberTools and automating the conversion of current and old forcefields.  He will open a PR soon for this.

This will also be a good time to add some additional information about the provenance of those files to the ffxml format.

PR #1134, opened 2015-09-14
Add ffxml-converted and validated Amber15 forcefields

rafwiewiora:
See issue #1054. 

This will create and validate ffxml equivalents of all forcefields shipping with Amber15. 

Starting with uploading the original Amber15 files into `devtools/amber/original-parameters`:

`/cmd` contains leaprc's for all the forcefields - I only for now excluded the GLYCAM ffs - is this right? Would you like to exclude any more of these / include the GLYCAMs?

`/lib, /parm, /prep` contain the original files (.lib, .dat, frcmod., .in) - only the ones that are called by the leaprc's included in /cmd. 

Workflow:
- [x] Upload original Amber files
- [ ] Convert to ffxml
- [ ] Validate against the Amber originals
- [ ] Include provenance information in the ffxml's


jchodera:
> /cmd contains leaprc's for all the forcefields - I only for now excluded the GLYCAM ffs - is this right? Would you like to exclude any more of these / include the GLYCAMs?

I think the GLYCAM forcefields could be of interest to people.  Is there any reason to selectively exclude these?

Let's add a `README.md` that notes that these forcefield parameters were imported from AmberTools 15.

I wonder if we want some way to keep these forcefield parameters more easily current with AmberTools. These files also appear in `ambermini`, which we are trying to keep current.  We could in principle use a subrepository mechanism to import [`share/amber/dat/` from `ambermini`](https://github.com/choderalab/ambermini/tree/master/share/amber/dat) if @peastman thinks this is a better idea.

We'll want to make sure to automate the conversion of both current forcefields and those in `oldff/` for backwards-compatibility.

We had also talked about modifying the conversion script to add more provenance data in an XML tag at the [beginning of the ffxml file](https://github.com/pandegroup/openmm/blob/master/devtools/forcefield-scripts/processAmberForceField.py#L302).  Perhaps we could add something like
```
<provenance>
   <conversion-script filename="processAmberForceField.py" version="1.0.5"/>
   <conversion-date>Mon Sep 14 09:49:07 CEST 2015</conversion-date>
   <source-distribution name="AmberTools" version="15.0.0"/>
   <source-files>
    <source-file filename="all_amino94.lib" md5="2dde4b366104b248789852ae0a7a73d4"/>
    ...
   </source-files>
</provenance>
```

rafwiewiora:
> I think the GLYCAM forcefields could be of interest to people. Is there any reason to selectively exclude these?

Just because they use the .prep files - I haven't seen this before so need to get familiar and the conversion script will need a new section for these files I think. So I'm gonna do the conversion of the other ones first, and then go back to GLYCAMs. Sounds good?

Updating from ambermini sounds good to me, waiting to see what @peastman thinks. 

Thanks for the provenance example!

peastman:
Let's *NOT* duplicate the entire Amber data directory!  This information doesn't belong in the OpenMM repository.  It's already managed and version controlled by Amber.  All we need to do is cite which Amber version we used the files from.

Also, we explicitly do *not* want to include all force fields from Amber.  One of the guiding principles of OpenMM is to only include features that we actually think users should use, and the decision to add a new force field is something that happens after serious discussion.  Absolutely no one should be using, say, AMBER94 anymore, so adding it to OpenMM would accomplish nothing except encourage people to make decisions.

jchodera:
An important component of this update will be traceability and provenance.  Having the forcefield files there helps with that, but we can use another scheme if it can be easily automated.  The ambermini scheme is easiest, but it may also be reasonably easy to require the person running the conversion script unpacks or downloads the appropriate version of AmberTools and its patches (but it may be harder to get the version info).

Old forcefields are critical to reproducibility.  Even Amber acknowledges this.  They still distribute parm94.  This is because you still need to compare results to published values when replicating work or validating against old work.  The science has to be reproducible. 

They do make it harder to use by pretending "old/" to these forcefield names so users cannot accidentally use a forcefield that is considered deprecated.  We can easily do the same, but we cannot shirk our responsibility of enabling reproducible science by omitting those forcefields.

jchodera:
@peastman: Any suggestions on the provenance XML block?  I bet that can be improved in multiple ways.

peastman:
> Old forcefields are critical to reproducibility.

Yes, but that only means OpenMM needs to keep providing all force fields *it* has provided in the past.  It doesn't mean it needs to include all force fields that *any* MD application has ever included!  And that's part of why we need to be very cautious about adding new force fields: once added, we can never remove them.

> Any suggestions on the provenance XML block?

I assume you wouldn't be asking that unless you had something in mind? :)  The AMBER force fields don't currently include it, because they were generated before we started doing that.  But is there other information you think should be added beyond what's in the other force fields?  For example, here's the relevant section from the AMOEBA2013 force field:

```
 <Info>
  <Source>amoebapro13.prm</Source>
  <DateGenerated>2015-02-19</DateGenerated>
  <Reference>Yue Shi, Zhen Xia, Jiajing Zhang, Robert Best, Chuanjie Wu, Jay W. Ponder, and Pengyu Ren. Polarizable Atomic Multipole-Based AMOEBA Force Field for Proteins. Journal of Chemical Theory and Computation, 9(9):4046–4063, 2013.</Reference>
 </Info>
```

The one obvious thing I see that could be added is which version of TINKER that file was taken from.

jchodera:
@peastman, can you specifically take a look at the provenance XML block I suggested earlier in this thread?  I think you may have just missed it.

peastman:
We already have a standardized format for that information.  We can add more information to it, but let's not throw it out and design something new from scratch.

jchodera:
I think it is very useful to support the same forcefields Amber provides, provided we can validate their consistency between openmm and amber.  ParmEd makes this validation easy now. 

We can survey what people think, but I think the overhead for supporting all the amber forcefields is minimal once we automate conversion.  So why not include them all if this adds value?

peastman:
Because it is an actively harmful thing to do.  Including a force field that no one should use, simply because it was introduced in an ancient version of Amber, will harm our users.

swails:
I'm with @peastman on this one -- using an old, bad FF is almost always a mistake.  In the very few instances where a power user wants it, they can use the XML converters themselves.  The FFs we distribute should be chosen with care and never should we find ourselves saying "no! never use that FF!" the first time we introduce the XML files to the app layer.

As for versions/provenance, AmberTools has a way of extracting version information:

```
swails@batman ~ $ $AMBERHOME/update_amber --version
Version is reported as <version>.<patches applied>

	AmberTools version 15.03
	     Amber version 14.13

```

That said, the file name itself is provenance enough, IMO.  `leaprc.ff99SB` sources the same files in every version.

jchodera:
> We already have a standardized format for that information.

Can you remind me where? I'm not finding it in either the converted forcefield files or the forcefield conversion scripts.

> Because it is an actively harmful thing to do. Including a force field that no one should use, simply because it was introduced in an ancient version of Amber, will harm our users.

We simply *cannot* omit forcefields with enormous bodies of literature behind them.  We *must* let users run an alanine dipeptide simulation with parm96 if they want to because this is still a standard test system for the construction of Markov state models.  We *must* let users compute the hydration free energy of SPCE water in SPCE because this is a standard reference number for validating transfer free energies.  We *must* let people run simulations with TIP3P because there are thousands and thousands of publications in the literature using this forcefield, and someone (like my lab) may need to run a reference calculation.  These are all outdated forcefields, but suggesting we cannot provide them (with appropriate warnings) is like suggesting we should discard all prior literature using them---it is simply impossible.

I'm all for warning users appropriately.  If they have to type `forcefield='deprecated-for-historical-use-only/myforcefield.xml'`, there's no way they can somehow be confused about what they're doing.  But expecting normal users to be able to figure out how to convert forcefields automatically is unreasonable.  Those conversion scripts are not even documented in the slightest at this point, nor are they fully compatible with all of the forcefield files yet---they actually give incorrect results on some.  I would *much* rather we fix that on our end then end up just driving users to another code.

> In the very few instances where a power user wants it, they can use the XML converters themselves.

This is absolutely false.  My lab has been working to convert CHARMM and AMBER forcefields and is having a great deal of trouble.  We have to fix and document these scripts first, and verify they produce correct results on those files.  At that point, it is far better for us to include them with appropriate warnings than to expect the user to convert the forcefield themselves.  Besides, what is the difference between forcing the user to prepend a warning message and having them run a script first?  The difference here is just that *we have to make sure the scripts work properly and are documented first* when we convert things ourselves, which is something we have not done yet, but are working on.

jchodera:
> As for versions/provenance, AmberTools has a way of extracting version information:

Thanks!

> That said, the file name itself is provenance enough, IMO. leaprc.ff99SB sources the same files in every version.

But `gaff.dat` has changed a few times over the years, as a simple example.  Users sometimes also hand-edit these files.  I think the `md5` hash is useful just in case.

swails:
> But gaff.dat has changed a few times over the years, as a simple example. Users sometimes also hand-edit these files. I think the md5 hash is useful just in case.

gaff is the only force field that has not really followed strict version control standards, unfortunately.  The others should be fine.

jchodera:
I think the point is that there is at least one important counterexample; there may be more, I honestly don't know.  

Can we be sure it will never happen to other files?  No, of course not.

Does it cost us anything to include this extra information and be a bit paranoid about provenance? No. And it won't break anything to add this information because it's XML.

Could it be problematic  in the future if we don't include provenance information? Yes.

Have we been burned by doing a bad job about provenance information in the past?  Yes, just recall the issue with the GBSA parameter origin.

swails:
@jchodera -- I'm going to respond to the general gist of your post rather than in-line responses, since it's fairly lengthy.  First, the ff2xml scripts are broken if they don't create correct XML files by simply running them on existing Amber files.  If they are broken and undocumented, the correct fix for *that* is to fix and document them (and won't it be the case that once we have "working" XML files for old FFs, the ff2xml scripts will have been fixed?).  Distributing FFs actively discouraged for production use should be an orthogonal decision (and we are benefitting from your lab's struggle to use old force fields since the ff2xml scripts are getting fixed!).  In the meantime, there are workarounds.

I have been answering Amber emails prolifically for ~6 years or so (~1000 responses/year over that time), and the number of users that use some crappy FF like FF96/FF99 simply because it's there and it "worked" (i.e., didn't crash) *far* outstripped the number of people that wanted an old FF for some reason (the latter whom tended to just have simple, technical questions).  And that's only the subset of people whose question indirectly included information about which force field they were using (since they never ask directly about FF choice). While including Amber 94/96/99 in OpenMM would be convenient to some, I suspect it would be inconvenient to many more people (many of whom may not even be aware they are being inconvenienced until they try to publish).

These "obsolete" force fields can always be released as a separate download (e.g., [here](http://wiki.simtk.org/openmm/VirtualRepository)) rather than in OpenMM proper once they've been validated.

swails:
> I think the point is that there is at least one important counterexample; there may be more, I honestly don't know.
> 
> Can we be sure it will never happen to other files? No, of course not.
> 
> Does it cost us anything to include this extra information and be a bit paranoid about provenance? No. And it won't break anything to add this information because it's XML.
> 
> Could it be problematic in the future if we don't include provenance information? Yes.
> 
> Have we been burned by doing a bad job about provenance information in the past? Yes, just recall the issue with the GBSA parameter origin.

I'm all for adding more provenance -- I was just saying why I don't think it's necessary to go hunting back through Amber releases of the past 20 years to make sure ff94 has never changed ;).

peastman:
> I'm not finding it in either the converted forcefield files or the forcefield conversion scripts.

A quick search for "\<Info>" tags shows that it appears in the following files: amoeba2009.xml, amoeba2009_gk.xml, amoeba2013.xml, amoeba2013_gk.xml, charmm_polar.xml, tip3pfb.xml, and tip4pfb.xml.  Basically, every force field file that's been added or regenerated within the last three years.

>  We must let users run an alanine dipeptide simulation with parm96 if they want to because this is still a standard test system for the construction of Markov state models.

And guess what?  We do!  Look in the data directory and you'll find amber96.xml is already there.

>  We must let users compute the hydration free energy of SPCE water in SPCE because this is a standard reference number for validating transfer free energies.

And sure enough, there's spce.xml.  What's your point?

>  We must let people run simulations with TIP3P because there are thousands and thousands of publications in the literature using this forcefield

And we do.  So...?

I'm really confused about what your point is.  You listed off a bunch of force fields that we already support, because at some point in the past we discussed whether to support them and decided we should.  Somehow you jumped from that to, "We should support every single force field that Amber has ever supported, even thoroughly outdated ones like parm94, just because, with no discussion about whether it's a good idea."  How do you get to that conclusion?

jchodera:
> A quick search for "<Info>" tags shows that it appears in the following files: amoeba2009.xml, amoeba2009_gk.xml, amoeba2013.xml, amoeba2013_gk.xml, charmm_polar.xml, tip3pfb.xml, and tip4pfb.xml. Basically, every force field file that's been added or regenerated within the last three years.

Great!  Here's the format:
```
 <Info>
  <Source>amoebapro13.prm</Source>
  <DateGenerated>2015-02-19</DateGenerated>
  <Reference>Yue Shi, Zhen Xia, Jiajing Zhang, Robert Best, Chuanjie Wu, Jay W. Ponder, and Pengyu Ren. Polarizable Atomic Multipole-Based AMOEBA Force Field for Proteins. Journal of Chemical Theory and Computation, 9(9):4046–4063, 2013.</Reference>
 </Info>
```
Let's think about how we can extend this to be more useful:
* Extend the `Source` to allow for multiple source filenames (since AMBER forcefields are scattered in multiple files)
* Add the `md5` hash of the file it was generated from if applicable
* Add the `SourcePackage` name (e.g. `AmberTools`) if relevant
* Add the `SourceVersion` name (e.g. 15) if relevant

> And guess what? We do! Look in the data directory and you'll find amber96.xml is already there.

These were illustrative examples motivating the need for deprecated forcefields, which you were insisting should not be provided (but some still are).

> I'm really confused about what your point is. You listed off a bunch of force fields that we already support, because at some point in the past we discussed whether to support them and decided we should. Somehow you jumped from that to, "We should support every single force field that Amber has ever supported, even thoroughly outdated ones like parm94, just because, with no discussion about whether it's a good idea." How do you get to that conclusion?

If we are converting forcefields from the curated repository available in AmberTools, we should make all of those available.  There are reasons they are there.  I don't think we need to make additional judgements as to whether something should be provided or not, though we can certainly make it harder to do things that are bad ideas.

jchodera:
Any further thoughts here on extending the provenance XML block?

peastman:
> Extend the Source to allow for multiple source filenames (since AMBER forcefields are scattered in multiple files)
> Add the md5 hash of the file it was generated from if applicable
> Add the SourcePackage name (e.g. AmberTools) if relevant
> Add the SourceVersion name (e.g. 15) if relevant

Those all sound reasonable.  How about having multiple `<Source>` tags, one for each file.  That tag can then optionally include `md5hash`, `sourcePackage`, and `sourcePackageVersion` attributes.

jchodera:
> How about having multiple `<Source>` tags, one for each file. That tag can then optionally include `md5hash`, `sourcePackage`, and `sourcePackageVersion` attributes.

I like it!  @rafwiewiora: Can you modify the AMBER forcefield conversion script to include this provenance data for each of the files processed?  The `<Info>` block would look something like this for AMBER:
```
 <Info>
  <Source md5hash="fd8622e8e96b18c1f36851a3115db9f6" sourcePackage="AmberTools" sourcePackageVersion="13">leaprc.ff10</Source>
  <Source md5hash="dac593da1c0afc852956bbc39bdf53fc" sourcePackage="AmberTools" sourcePackageVersion="13">amino10.lib</Source>
  ...
  <DateGenerated>2015-02-19</DateGenerated>
  <Reference>reference goes here</Reference>
 </Info>
```
@swails: Any suggestions on how to automatically get the `<Reference>` for each AMBER forcefield?  The AmberTools manual lists some references we can code by hand, but some of them (e.g. `ff10`) are only defined in this weird table format.

We could add a `<Notes>` tag can hold the comments block from the `leaprc` file, e.g.:
```
#
# ----- leaprc for loading the ff10 force field
# ----- NOTE: this is designed for PDB format 3!
#       ff10 =  ff99SB for proteins; ff99bsc0 for DNA; ff99sbsc_chiOL3 for RNA
#       phosphoaa parms from N. Homeyer, A.H.C. Horn, H. Lanig, H. Sticht
#           J. Mol. Model. 2006, 12, 281-289.  OP vdW parameters modified
#           by T. Steinbrecher and J. Latzer.
#
```
That might help?

swails:
Sorry for the late response

> @swails: Any suggestions on how to automatically get the <Reference> for each AMBER forcefield?

I can't think of a way to do this 'automatically'.  In fact, I don't even know where I would look to find all of the references (I'd probably start with the files themselves and then move to the manual, although not all FFs actually have a citation).

jchodera:
What if we take an approach that centers on processing each of the `.leaprc` files available in `AmberTools`?  We can compile a master XML or YAML file that manually specifies which `.leaprc` files to process and additional XML provenance blocks (such as literature references) to be added.

We can also pull in the comments block from the `leaprc` into a `<Notes>` section in the `ffxml` file.  

We would then just need to curate this master forcefields file to decide which forcefields to include and which references go along with them.

peastman:
I think it's totally fine to fill in the references by hand.

jchodera:
Great.  Let's decide to do this with either a YAML or XML "master index file" of forcefields that @rafwiewiora will initially curate (maybe with help from @swails) from the AmberTools manual.

This will list the `leaprc` files we will process to generate `ffxml` files.  I think the current conversion script doesn't process `leaprc` files, but instead processes `lib', 'off', 'prep', and `parm` files supplied via the command line, which will no longer be how we specify files.  Instead, we would add a parsing block that will parse the `leaprc` files for the forcefield files to process, since these index all the files that go with each class of forcefield.

Does that sound reasonable?

jchodera:
Cross-referencing discussion in #1200 where we are discussing whether ParmEd may be the best route to convert forcefield definitions.

rafwiewiora:
Closing this as not useful anymore / ffxmls converted with the ParmEd approach will be PR'ed in a new one.

PR #1076, opened 2015-08-05
Add a set of unit tests for unit handling in the core API.

swails:
This is the start of what I promised in #1031 

So far it just has the basic classes, but I still want to add:

- [x] `CMAPForce`
- [x] custom forces
- [x] integrators
- [x] Amoeba forces

PR #1137, opened 2015-09-18
Created periodicdistance() function for CustomExternalForce

### LINKED ISSUE #988: CustomExternalForce and periodic boundary conditions
I was having a very difficult time using `CustomExternalForce` to impose positional restraints for an Amoeba simulation recently.  What I tried basically was this:

```Python
force = mm.CustomExternalForce('k*((x-x0)^2+(y-y0)^2+(z-z0)^2)")
force.addGlobalParameter('k', 10*u.kilocalories_per_mole/u.angstrom**2)
force.addPerParticleParameter('x0')
force.addPerParticleParameter('y0')
force.addPerParticleParameter('z0')
force.setForceGroup(1)
for atom in pdb.atoms:
    if atom.residue.name in AMINO_ACIDS:
        force.addParticle(atom.idx, [atom.xx/10, atom.xy/10, atom.xz/10])
system.addForce(force)
# Create the context
context.setPositions(pdb.positions)
```

The simulation would almost immediately blow up.  I dug some more into what was happening and found that the particles seem to be getting reimaged before being applied to the `CustomExternalForce`.  In particular, I found the following:

```Python
context.getState(getPositions=True).getPositions()
context.getState(getPositions=True, enforcePeriodicBox=True).getPositions()
```

These two functions return different coordinates, since some of the coordinates are translated by a box length.  Based on the coordinates obtained in the first line, the CustomExternalForce should return a net force very close to 0 (since those were the coordinates used both to seed the context as well as to set the CustomExternalForce per-particle parameters).  But the forces are very much *not* zero for many particles.  The forces in fact match (almost exactly) what I calculate by hand using the positions I get from the second line of the above example.

Which leads me to the conclusion that the `CustomExternalForce` is using the imaged coordinates in its energy expression.

Is this intended?  Is this only true for Amoeba?  I don't recall having this problem before with `NonbondedForce`, but this system is rather atypical (crystal simulation with *far* more protein atoms than solvent).

peastman:
Fixes #988.

### DIFF
--- a/openmmapi/include/openmm/CustomExternalForce.h
+++ b/openmmapi/include/openmm/CustomExternalForce.h
@@ -67,6 +67,14 @@ namespace OpenMM {
  * force->addPerParticleParameter("z0");
  * </pre></tt>
  *
+ * Special care is needed in systems that use periodic boundary conditions.  In that case, each particle really represents
+ * an infinite set of particles repeating through space.  The variables x, y, and z contain the coordinates of one of those
+ * periodic copies, but there is no guarantee about which.  It might even change from one time step to the next.  You can handle
+ * this situation by using the function periodicdistance(x1, y1, z1, x2, y2, z2), which returns the minimum distance between
+ * periodic copies of the points (x1, y1, z1) and (x2, y2, z2).  For example, the force given above would be rewritten as
+ *
+ * <tt>CustomExternalForce* force = new CustomExternalForce("k*periodicdistance(x, y, z, x0, y0, z0)^2");</tt>
+ *
  * Expressions may involve the operators + (add), - (subtract), * (multiply), / (divide), and ^ (power), and the following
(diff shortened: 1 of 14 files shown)


swails:
A few things I thought about here while working on #1076:

1. What happens when someone uses the `periodicdistance` function on a system where no periodic box vectors are defined?
2. What *should* `CustomExternalForce.usesPeriodicBoundaryConditions()` return when `periodicdistance` is in the energy expression?

What I would naively expect for 1 is a thrown exception.  Likewise, for 2 I would expect `usesPeriodicBoundaryConditions()` to return `true` if `periodicdistance` is in the energy expression and false otherwise (but I don't see this implemented anywhere).

peastman:
> What should CustomExternalForce.usesPeriodicBoundaryConditions() return when periodicdistance is in the energy expression?

Good point.  Yes, it should return true in that case.  I'll fix it.

> What happens when someone uses the periodicdistance function on a system where no periodic box vectors are defined?

There are always periodic box vectors defined.  `getDefaultPeriodicBoxVectors()` always returns something.  I think the default value is something fairly arbitrary, like a 2 nm cube,

PR #1138, opened 2015-09-19
Typemaps

rmcgibbo:
This add typemaps for `vector<vector<double> >` and `vector<vector<vector<double> > >`.

@swails, I think this solves the issue in https://github.com/pandegroup/openmm/pull/1076#issuecomment-141569078


swails:
+1 on the added typemaps, I actually need to solve this in `%pythonprepend` anyway, since the TorsionTorsion angles need to be in degrees, not radians -- I've already made the necessary fix in `pythoncode.i` in #1076

PR #1140, opened 2015-09-21
Changes to generating nonbonded kernels

peastman:
This includes two related changes.  First, I made it so that in mixed precision mode, the nonbonded kernel accumulates energy in double precision.  This leads to more accurate and more consistent energies.

Unfortunately, it also hurts performance.  The difference isn't huge, but it isn't negligible either, especially on consumer level GPUs with bad double precision performance.  So I also modified how it generates the nonbonded kernel.  It now can compile separate versions of the kernel for computing just forces, just energies, or both.  Since most of the time it's only computing forces, that avoids the added cost except when you actually query the energy.  It even helps in single precision mode, though in that case the difference is tiny.

So the upshot: energies in mixed precision are now more accurate, and simulations are very very slightly faster.

### DIFF
--- a/platforms/cuda/include/CudaNonbondedUtilities.h
+++ b/platforms/cuda/include/CudaNonbondedUtilities.h
@@ -138,8 +138,12 @@ public:
     void prepareInteractions(int forceGroups);
     /**
      * Compute the nonbonded interactions.
+     * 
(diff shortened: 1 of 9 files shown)


jchodera:
Awesome!  This is going to be tremendously helpful for alchemical free energy calculations and anything involving reweighting.

PR #1142, opened 2015-09-23
Refactored test cases

### LINKED ISSUE #651: Is there some strategy that could reduce code duplication in unit tests?
Currently, all unit tests are implemented separately for each `platform`.  Most, but all, of this code is an exact duplication for each `platform`, with only a few lines changed to invoke the appropriate `Platform` class.

Is there some way that the common code could be extracted into a common `unit-testing/` directory and compiled separately for each platform, with only the platform-specific tests remaining within each `platform/` directory?  This might significantly aid code maintenance down the road.

Also, it would be useful to annotate each unit test with a human-readable description of what exactly is tested in each test component.  For example, many of the tests currently have unilluminating names like `testTemperature()` that essentially require reading the whole function in detail to comprehend what is being tested and how.  A simple statement like

> Ensure average of instantaneous kinetic temperature for 8 charged LJ particles matches control temperature when an Andersen thermostat is used over 150 ps simulation.

would significantly speed comprehension.

peastman:
Fixes #651.  This was a bit of a slog, but hopefully worth it.  156 files changed, the total size of the source tree reduced by well over a megabyte.  I haven't yet dealt with the tests for the plugins, but those can wait a little while.


jchodera:
> total size of the source tree reduced by well over a megabyte

Whoa!

peastman:
We have a *lot* of test cases.  And when there's three or four copies of each one...  There were also a few data files I was able to delete extra copies of.

PR #1153, opened 2015-09-25
General perturbation theory code for induced dipoles

andysim:
Here is a rough implementation of arbitrary-order PT approach to induced dipoles, with arbitrary-coefficients for each order of dipoles.  The code works with and without PME.  There are a few things that still need to be done, but I wanted to give you access to it ASAP.  TODO:-

- [x] Implement Cartesian direct space dipole-dipole code, because the QI formulation is probably not ideal for such low angular momentum.
- [x] Implement unit tests.
- [x] Hook into https://simtk.org/svn/pyopenmm/trunk/test/system-tests to get some finite difference tests; I've also implemented the current code in TINKER and verified it by finite differences.
- [ ] Serialize the OPT coefficients, and add this to the serialization test.

To activate the code, simply set the polarization types and coefficients:

    amoebaMultipoleForce.setNonbondedMethod(AmoebaMultipoleForce::OPT);
    std::vector<RealOpenMM> coefs;
    coefs.push_back(0.0);  // The mu_0 coefficient
    coefs.push_back(-0.3); // The mu_1 coefficient
    coefs.push_back(0.0);  // The mu_2 coefficient
    coefs.push_back(1.3);  // The mu_3 coefficient
    amoebaMultipoleForce.setOPTCoefficients(coefs);

This example is for the ExPT coefficients and I'm still working to come up with better values; I'll let you know as soon as I have some sensible defaults.  Because they're input parameters, the code can be developed before this issue is resolved.

I also only just noticed that the QI code I committed has a small mistake that will cause the energy (but not forces) to be incorrect when different µp and µd scale factors are present and PME is used.  I use a more symmetric energy expression than what was implemented before; either is appropriate, but they must be consistent across real, reciprocal and self terms.  The fix is implemented in e6dbc86; sorry for not catching the problem sooner.


peastman:
Looks great!  Let me know when you have everything done, and I'll start working on the CUDA implementation, Python support, etc.

How is this intended to be used in practice?  Are users supposed to actually supply their extrapolation coefficients?  Or will there be one "official" set of coefficients that get used by default (and it would be very rare for anyone to change them)?  Or should the coefficients be specified by the force field?

I wonder if we can come up with a better name for the NonbondedMethod that "OPT".  It's a not very enlightening name, and will likely confuse people.  In my original version I called it "Extrapolated" which seems clearer and more informative.

peastman:
I separated out the bug fix and created a PR with just that (#1174), since users are running into it.

andysim:
Sorry for the slow reply.  The method is intended to be a drop-in replacement for the iterative solver and there will hopefully be a single set of universal (i.e. for all molecules) coefficients that we're working to determine now.  Our initial tests are promising, but we need more diversity in our test set to be able to say anything definitive.  Ultimately the user should be able to just ask for the PT method and not have to worry about coefficients; it would be nice for the more intrepid users to have the option of specifying coefficients though.

The first attempt was extrapolating the series to infinity (ExPT), but now we're optimizing (in a least squares sense) the coefficients so internally we're using the moniker optimized perturbation theory (OPT).  I have no problem with using any other nomenclature in OpenMM so feel free to change any of the naming.

peastman:
Ok, thanks!  So:

- There will be a default set of coefficients.
- Users can change them if they want, but hopefully that will be a rare thing.

We'll need to decide whether this should become the default method, or whether users should have to explicitly select it.  We should probably put off that decision until the coefficients are finalized and we see how well they work.

peastman:
@andysim just wondering what the status of this is.  You listed several things that still needed to be done.  Are you working on those?

andysim:
AFAIK the commits that I pushed are just fine, and they pass the tests that I created.  It looks like one of the Travis tests has a setup problem.  The AppVeyor problem is less apparent (perhaps a timeout?), but none of the test cases seem to be broken.  Either way, I think the code is good and stable at this point, so I'd appreciate your help to figuring out the validation issues when you get a chance, @peastman.

The remaining task on the TODO is to serialize the PT coefficients, which should be just a trivial modification of plugins/amoeba/serialization/src/AmoebaMultipoleForceProxy.cpp.  I'm just not sure how you handle the version number of the save file.  Should it be bumped from 2 to 3 when a new member is added?

peastman:
It looks like travis passed on all but one machine, and that one failed trying to install OpenCL, so pretty clearly nothing related to this.  And we've been having trouble with AppVeyor lately.  It always says the build failed, but when you look up the log, you find that all tests passed.  We still need to debug what's going on with that, but again it's unrelated to this PR.

Any change to the serialized representation of a class requires incrementing the version number.  For an example of how to handle this, take a look at GBSAOBCForceProxy.cpp.  In `serialize()` it sets the version to 2.  Then in `deserialize()` it accepts either 1 or 2:

```C++
    if (version < 1 || version > 2)
        throw OpenMMException("Unsupported version number");
```

For the piece of information that was added in version 2, it only tries to load it if the version indicates it's there:

```C++
        if (version > 1)
            force->setSurfaceAreaEnergy(node.getDoubleProperty("surfaceAreaEnergy"));
```

jchodera:
>  and that one failed trying to install OpenCL

xref https://github.com/pandegroup/openmm/pull/1286

rmcgibbo:
I changed the appveyor python tests to use py.test instead of nosetests, and https://ci.appveyor.com/project/peastman/openmm/build/1.0.395 it appears to be working?

peastman:
I'm going to go ahead and get started on this.  I'll take this branch as the starting point, then revise the API, add the CUDA version, update the documentation, etc.  When that's all done, I'll send a new PR containing the complete feature.

peastman:
I notice you only implemented this for PME, not GK.  Is there any reason this wouldn't work for that as well?  In `convergeInduceDipolesByOPT()` you hardcoded the assumption that `updateInducedDipoleField` would be of length 2 (for GK it's 4).  And you modified `AmoebaReferencePmeMultipoleForce::calculateElectrostatic()`, but not `AmoebaReferenceGeneralizedKirkwoodMultipoleForce::calculateElectrostatic()`.

andysim:
The method should work just fine for the GK approach.  I didn't implement it, primarily because I don't have any familiarity with the GK code, so I can't be as confident in the correctness of any implementation changes.  The stuff I put into the regular PME version has been implemented in CHARMM and TINKER, using quite a different algorithm, so I'm very confident it's correct.

I didn't realize the GK code computes 4 different fields.  I wondered why the list of UpdateInducedDipoleFieldStruct structs was used, rather than just writing the two terms out, and I unpacked them to reflect the µ_d µ_p terms explicitly.  I think the best way to proceed is to undo this unpacking, by consolidating the _ptDipoleFieldGradientP and _ptDipoleFieldGradientD arrays into a single, higher-dimensional structure.  By rewriting the code as a loop over the (two) different fields, as it used to be, the method should be easier to extend to the case of 4 different fields.  AFAIK, Jay has added an ExPT-like solver to his TINKER version of the GK code, so it should be possible.

Thanks for the advice with the tests.  The serialization code will be trivial to add, but I just wanted to make sure I follow the proper protocols in doing so.

peastman:
Ok, thanks!  Could you take a look at `AmoebaReferenceGeneralizedKirkwoodMultipoleForce::calculateElectrostatic()` and see if you can tell how it should be updated?  Presumably by adding a block similar to the one in `AmoebaReferencePmeMultipoleForce::calculateElectrostatic()`.  But since there are four fields instead of two (and to be honest, I don't entirely understand why!), it presumably will be somewhat different.

I can take care of the serialization code.  I've already branched from your code and modified it significantly, so at this point it's just as easy for me to add the serialization of that one field as to merge changes by hand.

andysim:
I'm trying to digest what's happening.  It looks like there are just two more combinations of fields that each generate a set of dipoles: with and without the GK field present, in addition to the P/D subscripts.  The fields and matching dipole vectors are all pushed into the updatedipoles structs [here](https://github.com/andysim/openmm/blob/2913b6866545103cc9cfcf3d9211547752f4ac21/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L2585-L2588).  Therefore, it looks like [this code](https://github.com/andysim/openmm/blob/2913b6866545103cc9cfcf3d9211547752f4ac21/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L962-L1039) needs to be modified by restoring the loop over the vector of fields (instead of unpacking d and p terms), before calling it from a brach point [here](https://github.com/andysim/openmm/blob/2913b6866545103cc9cfcf3d9211547752f4ac21/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L2590).  After that, the rest of the code should work right out of the box, I think.  I have no idea how the dipole response force terms enter into it at this point; i'd need time to figure out the GK theory to be able to properly address those.

peastman:
I found several places where it checks to see if mutual polarization is enabled, and adds in extra terms to the interaction if it is.  Should those also be enabled for extrapolated polarization?  The specific places are in `AmoebaReferenceMultipoleForce::calculateElectrostaticPairIxn()`, `AmoebaReferenceGeneralizedKirkwoodMultipoleForce::calculateKirkwoodPairIxn()`, and `AmoebaReferencePmeMultipoleForce::computeReciprocalSpaceInducedDipoleForceAndEnergy()`.

andysim:
That's a slightly tricky one to answer.  For AmoebaReferenceMultipoleForce::calculateElectrostaticPairIxn() and AmoebaReferencePmeMultipoleForce::computeReciprocalSpaceInducedDipoleForceAndEnergy() the answer is no; in those cases the PT energy and forces are correct in the patches provided by this PR.  The PT forces are just the same as the "direct" forces, with some dipole response terms that are added in [here](https://github.com/andysim/openmm/blob/genpt/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L1849-L1878) and [here](https://github.com/andysim/openmm/blob/genpt/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L6803-L6834).  These dipole reponse forces are computed using dipole field gradients that are computed and cached during the generation of the induced dipoles, so they're basically free (at a cost of 6*Natoms memory).

The GK code is not quite as straightforward; the algorithm here is the same as Jay's TINKER version.  He always computes the µ-µ terms, and then recomputes them and backs them out [if they're not needed](https://github.com/andysim/openmm/blob/genpt/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L4478-L4489).  The easiest way to put the PT scheme into the GK routine is to also do this removal of the mutual terms for the case of PT being used.  Then you'd need to add the cacheing of the µ field gradients to AmoebaReferenceGeneralizedKirkwoodMultipoleForce::calculateInducedDipolePairIxns(), exactly like [this](https://github.com/andysim/openmm/blob/genpt/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L829-L874).  Once those terms are cached, you can simply add some code to the GK computeElectrostatic(), exactly analogous to the regular reference and PME versions, linked above.  As I mentioned in my previous comment, this should be done by replacing the _ptDipoleFieldGradientP and _ptDipoleFieldGradientD arrays with a single list of lists to contain them; handling the four fields needed in the GK code will be a lot easier this way.

andysim:
I should add that there are two if statements that should be activated to remove the mutual terms in the GK electrostatic code: the block I linked to and another one [here](https://github.com/andysim/openmm/blob/genpt/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L4743-L4752).  The way it's set up, the dipole terms with an S subscript interact only with each other, and those without an S subscript only interact with each other.  Therefore, when adding in code like [this](https://github.com/andysim/openmm/blob/genpt/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L1849-L1878) to process the dipole response forces, you just need to cut and paste the contents of the inner loop to add in corresponding terms for the dipoles with an S subscript.  At least that's my understanding from looking through the code; I still haven't had time to go and read the GK paper in detail.  Thankfully the forces are very easy to verify with finite differences.

PR #1157, opened 2015-09-28
optionally pass a PSF to charmmparameterset from which atomtypes will be taken

mj-harvey:
For many of our models we have charmm par files but no corresponding rtf file. Without the latter the CharmmParameterSet reader fails, complaining about undefined atom types.
This patch adds the ability to extract those missing atom types from a PSF file.

Two new named arguments are added to readParameterFile
 * psf=None     - either a CharmmPsfFile object or PSF filename. If set, the atom types are extracted from this file
 * permissive=False - if True, accept a parameter file that contains parameters for undefined atom types (in this case nonbonded terms for atom types not used in the PSF)

### DIFF
--- a/wrappers/python/simtk/openmm/app/charmmparameterset.py
+++ b/wrappers/python/simtk/openmm/app/charmmparameterset.py
@@ -175,7 +175,33 @@ class CharmmParameterSet(object):
                 inst.readStreamFile(sfile)
         return inst
 
-    def readParameterFile(self, pfile):
+    def _read_atom_types_from_psf( self, psf ):
+       if( isinstance( psf, str ) ):
+           data = CharmmPsfFile( psf )
+       else:
+           data = psf
+       idx=1000
+       for a in data.atom_list:
+            name = a.attype
+            mass = a.mass
+            # Figure element out from the mass
+            masselem = Element.getByMass(mass)
+            if masselem is None:
+                atomic_number = 0 # Extra point or something
+            else:
+                atomic_number = masselem.atomic_number
+
+            if not name in self.atom_types_str:
+                atype = AtomType(name=name, number=idx, mass=mass,
+                                 atomic_number=atomic_number)
+                self.atom_types_str[atype.name] = atype
+                self.atom_types_int[atype.number] = atype
+                self.atom_types_tuple[(atype.name, atype.number)] = atype
+                idx=idx+1
+            else:
+                pass
+  
+    def readParameterFile(self, pfile, psf=None, permissive=False ):
         """
         Reads all of the parameters from a parameter file. Versions 36 and
         later of the CHARMM force field files have an ATOMS section defining
@@ -184,12 +210,20 @@ class CharmmParameterSet(object):
 
         Parameters:
             - pfile (str) : Name of the CHARMM PARameter file to read
+            - psf         : Optional CharmPsfFile object or PSF filename
+                            from which atom types will be read
+            - permissive  : Boolean - accept a parameter file that contains
+                            references to undefined atom types 
 
         Notes: The atom types must all be loaded by the end of this routine.
         Either supply a PAR file with atom definitions in them or read in a
         RTF/TOP file first. Failure to do so will result in a raised
         RuntimeError.
         """
+        # Read in the atom types in the pdb
+        if(psf):
+            self._read_atom_types_from_psf( psf )
+
         conv = CharmmParameterSet._convert
         if isinstance(pfile, str):
             own_handle = True
@@ -481,12 +515,16 @@ class CharmmParameterSet(object):
         # instances. In order for this to work, all keys in nonbonded_types
         # must be in the self.atom_types_str dict. Raise a RuntimeError if this
         # is not satisfied
-        try:
-            for key in nonbonded_types:
+        for key in nonbonded_types:
+            try:
                 self.atom_types_str[key].set_lj_params(*nonbonded_types[key])
-        except KeyError:
-            raise RuntimeError('Atom type %s not present in AtomType list' %
+            except KeyError:
+                if( permissive ):
+                    print( " Atom type " + key + " unknown" )
+                else:
+                    raise RuntimeError('Atom type %s not present in AtomType list' %
                                key)
+
         if parameterset is not None: self.parametersets.append(parameterset)
         if own_handle: f.close()
 


wrappers/python/simtk/openmm/app/charmmparameterset.py
```diff
@@ -481,12 +515,16 @@ def readParameterFile(self, pfile):
...
-        try:
-            for key in nonbonded_types:
+        for key in nonbonded_types:
+            try:
                 self.atom_types_str[key].set_lj_params(*nonbonded_types[key])
-        except KeyError:
-            raise RuntimeError('Atom type %s not present in AtomType list' %
+            except KeyError:
+                if( permissive ):
+                    print( " Atom type " + key + " unknown" )
```
swails: Turn this into a `warnings.warn` to avoid printing from the API.  This allows it to be intercepted via loggers and otherwise handled/silenced (and sends it to stderr instead of stdout).

wrappers/python/simtk/openmm/app/charmmparameterset.py
```diff
@@ -175,7 +175,33 @@ def loadSet(cls, tfile=None, pfile=None, sfiles=[]):
...
+
+            if not name in self.atom_types_str:
+                atype = AtomType(name=name, number=idx, mass=mass,
+                                 atomic_number=atomic_number)
+                self.atom_types_str[atype.name] = atype
+                self.atom_types_int[atype.number] = atype
+                self.atom_types_tuple[(atype.name, atype.number)] = atype
+                idx=idx+1
+            else:
+                pass
```
swails: `else: pass` is a no-op here.

wrappers/python/simtk/openmm/app/charmmparameterset.py
```diff
@@ -175,7 +175,33 @@ def loadSet(cls, tfile=None, pfile=None, sfiles=[]):
                 inst.readStreamFile(sfile)
         return inst
 
-    def readParameterFile(self, pfile):
+    def _read_atom_types_from_psf( self, psf ):
+       if( isinstance( psf, str ) ):
+           data = CharmmPsfFile( psf )
+       else:
+           data = psf
+       idx=1000
```
swails: Are you sure this doesn't collide with other atom type indices?  My particular worry is from CHARMM CGenFF.  Although I've been told CHARMM is basically doing away with atom type indices and moving to the XPLOR-style PSF files (so indices won't matter much).

Perhaps `idx` can be assigned from `max(self.atom_types_int.keys())` (and adding 1000 to that, maybe), and just make sure that the PSF is the last file parsed when given in a list of files?

swails:
Since CHARMM 36, the parameter files contain the `MASS` section, so RTF files are entirely unnecessary here.  For anything older (or that otherwise does *not* have the atom type definitions), then I would expect to see the error you reported.

I don't think this approach will work for CHARMM-style PSF files, since there is no integer<->string type mapping that tells you what atom types in the PSF correspond to the type names listed in the parameter files.  And in versions of CHARMM that only use atom type names (i.e., XPLOR-format), the parameter files should already have the MASS section present (so this PR wouldn't be necessary).

Can you give more details, and perhaps an example (that can be turned into a unit test) of where this is needed and how you generated these files?  This code is originally taken from ParmEd, so I'd like to backport improvements there when applicable.

Thanks!

mj-harvey:
Hi Jason,

We've models built with VMD in conjunction with CHARMM 22 and 27. The PSFs are in XPLOR format (with string atomtypes) and the parameter files we have are par_all27_prot_lipid.prm and par_all22_prot.inp  derivaties, in all cases without MASS entries. Both ACEMD and NAMD will start up from these files alone, and I want to be able to replicate that behaviour with OPENMM.

 http://www.ks.uiuc.edu/Research/namd/tutorial/NCSA2001/dhfr/ is a good example

swails:
Gotcha.  I can see the appeal here.  I actually think I'd opt for something slightly different.  I don't really like the idea of `CharmmParameterSet` taking a PSF file as input if that can be avoided.  I think it would be better to have `CharmmParameterSet` generate new atom types when it encounters nonbonded parameters for types it doesn't know about rather than *just* define those types specified within a PSF file (after all -- you should be able to use the same CharmmParameterSet for *many* different PSF files in the same script).

So there can be a "strict" and "permissive" mode where Lennard-Jones types for unknown atom types either throw an exception (as it does now) or generate a new type implicitly, respectively.  Given the potential danger of implicit type generation, I'd say this should be something you have to opt-in to (that danger being that you load RTF files that define an atom type *after* reading a parameter file that defines the parameters, and the RTF file will silently wipe out those Lennard-Jones parameters).

I can put together an alternative PR that does this, or if you want to modify yours that would work also.  Just let me know.

mj-harvey:
Hi Jason,

I thought about that, but there's no robust way to guess the mass based on
the atom type string alone, so the AtomType object initialisation may be
wrong.
Perhaps that doesn't matter - will OpenMM override the masses (and charges)
found in the RTF parameter file with those defined in the PSF?

On 29 September 2015 at 18:20, Jason Swails <notifications@github.com>
wrote:

> Gotcha. I can see the appeal here. I actually think I'd opt for something
> slightly different. I don't really like the idea of CharmmParameterSet
> taking a PSF file as input if that can be avoided. I think it would be
> better to have CharmmParameterSet generate new atom types when it
> encounters nonbonded parameters for types it doesn't know about rather than
> *just* define those types specified within a PSF file (after all -- you
> should be able to use the same CharmmParameterSet for *many* different
> PSF files in the same script).
>
> So there can be a "strict" and "permissive" mode where Lennard-Jones types
> for unknown atom types either throw an exception (as it does now) or
> generate a new type implicitly, respectively. Given the potential danger of
> implicit type generation, I'd say this should be something you have to
> opt-in to (that danger being that you load RTF files that define an atom
> type *after* reading a parameter file that defines the parameters, and
> the RTF file will silently wipe out those Lennard-Jones parameters).
>
> I can put together an alternative PR that does this, or if you want to
> modify yours that would work also. Just let me know.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pandegroup/openmm/pull/1157#issuecomment-144109478>.
>



-- 
M J Harvey
CTO
Acellera Ltd.
London, Barcelona.
www.acellera.com | m.j.harvey@acellera.com
Follow us on: twitter <http://www.twitter.com/acellera>, linkedin
<http://www.linkedin.com/company/acellera>, youtube
<http://youtube.com/acelleracom>.
Skypename: acellera

swails:
> Perhaps that doesn't matter - will OpenMM override the masses (and charges)
> found in the RTF parameter file with those defined in the PSF?

Exactly.  The masses from the PSF are the ones that are used.

mj-harvey:
Ok, very simple then. I'll work up a new patch with just the "permissive" option.

mj-harvey:
Closing. New pull rq to follow

PR #1165, opened 2015-10-02
Further improved energy accuracy in mixed precision mode

peastman:


jchodera:
Cool!  Can you give us any additional insight into what was improved?

peastman:
There are three levels of accumulation:

1. Each thread in each kernel has a variable for accumulating the energies of all interactions it computes.
2. There's an array in global memory where threads add in their contributions to the energy.  That's how it accumulates energies across kernels.
3. All the elements of that array need to be summed to get the final energy.

3 has always been done in double precision.  My previous change was that for the nonbonded kernel, the variables in 1 became double precision.  And this change makes the array in 2 double precision.

AppVeyor failed with the error, "jom: Can't connect to job server: Cannot determine jobserver name."  I assume this means they're having some sort of problem on their end?

jchodera:
Thanks for the explanation!

Not sure what happened with AppVeyor.  It does sound like a problem on their end.

PR #1171, opened 2015-10-06
Hard wall constraints for DrudeLangevinIntegrator

peastman:


peastman:
AppVeyor got the error again about "jom: Can't connect to job server: Cannot determine jobserver name."  Anyone have an idea what that means?

PR #1172, opened 2015-10-06
Replace JOM with NMake in Appveyor build

rmcgibbo:


rmcgibbo:
it looks like it works okay with nmake. i think the appveyor build machines are single core, so it might not really help anyways to use jom.

peastman:
Looks like it passed.  Should I go ahead and merge this?

rmcgibbo:
Yes, I think so.

PR #1173, opened 2015-10-06
Add permissive flag to permit undefined atomtypes when reading charmm parameter files

mj-harvey:
Hi Jason,

This is the re-worked patch to allow undefined atomtypes when reading Charmm param files, added the named argument "permissive" to the constructor and readParameterFile(). Default is False, giving current behaviour. If set to True, any undefined atom type used by the nonbonded params will be implicitly created, rather than raise an exception.

Matt


peastman:
Test cases are failing:

```
E     File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/simtk/openmm/app/charmmparameterset.py", line 103
E       def __init__(self, *args, permissive=False):
E                                          ^
E   SyntaxError: invalid syntax
```

I think `*args` is required to be the very last entry in the argument list.

mpharrigan:
In python 2 :)

https://www.python.org/dev/peps/pep-3102/

peastman:
It's also failing in Python 3, but with a different error message:

```
>       for par in pars: self.readParameterFile(par, permissive=self.permissive )
E       AttributeError: 'CharmmParameterSet' object has no attribute 'permissive'
```

wrappers/python/simtk/openmm/app/charmmparameterset.py
```diff
@@ -113,7 +113,7 @@ def __init__(self, *args):
         self.cmap_types = dict()
         self.nbfix_types = dict()
         self.parametersets = []
-
+
```
swails: Extra space can be deleted

wrappers/python/simtk/openmm/app/charmmparameterset.py
```diff
@@ -100,7 +100,7 @@ def _convert(data, type, msg=''):
         except ValueError:
             raise CharmmFileError('Could not convert %s to %s' % (msg, type))
 
-    def __init__(self, *args):
+    def __init__(self, *args ):
```
swails: trailing space can be deleted

swails: The constructor should also take `permissive`.  You can do this via `**kwargs` and then popping off the `permissive` keyword (and then to avoid silently passing errors, throw a `TypeError` if the `kwargs` dict is not empty after you pop off `permissive`).

wrappers/python/simtk/openmm/app/charmmparameterset.py
```diff
@@ -477,6 +481,23 @@ def readParameterFile(self, pfile):
...
             self.cmap_types[current_cmap] = ty
+
+        # If in permissive mode create an atomtype for every type used in  
+        # the nonbonded parameters. This is a work-around for when all that's
+        # available is a CHARMM22 inp file, which has no ATOM/MASS fields
+
+        if permissive:
+            try:
+               idx = max(self.atom_types_int.keys())+1000
+            except:
```
swails: Add an explicit exception type here.  Bare excepts like this hide bugs.

swails:
I made some fairly minor comments.  Aside from those, looks good!  I think this PR also needs a test case (hopefully you have one handy) -- but this would ensure that this functionality is not broken in the future.

mj-harvey:
Test case added to TestCharmmFiles

swails:
Looks like there's a missing file that's causing the tests to fail: https://travis-ci.org/pandegroup/openmm/jobs/84059277#L5171

mj-harvey:
Don't understand what's wrong there - the file it's complaining about is in the commit 
eb77eb6  wrappers/python/tests/systems/par_all22_prot_with_mass.inp

swails:
You didn't tell CMake to copy that file over to the build directory.  Look at https://github.com/mj-harvey/openmm/blob/charmm-permissive/wrappers/python/CMakeLists.txt#L51-L71

You need to add the following line to that block:

```
        "${CMAKE_CURRENT_SOURCE_DIR}/${SUBDIR}/par_all*"
```

or something to that effect.

swails:
+1 from me on this.

PR #1174, opened 2015-10-07
Fixed a bug in AmoebaMultipoleForce

peastman:
This fixes the bug mentioned in #1153.


swails:
Thank you!  I was getting so baffled by this!  I was within an hour of filing an issue (FTR, I was seeing an error in energies, but not forces as @andysim's email said would happen).

I'll see if this fixes my problems.

andysim:
Yeah, sorry for the hassle.  The problem will only occur in systems where PME is used, and  µ_d !=  µ_p.  There is no coverage in the test cases for this right now, so perhaps it'd be a good idea to take one of the tests (villin headpiece, I think?) from the generalized Kirkwood AMOEBA tests.  The ammonia tests in the existing unit test pass because µ_d == µ_p for that case.

swails:
OK, this fixed the energy differences (as you probably knew it did/would).

Thanks!

PR #1175, opened 2015-10-07
Removed some unnecessary double precision

peastman:


peastman:
The test failure was stochastic.

PR #1177, opened 2015-10-08
Amber switching

mj-harvey:
Add the switchdistance= named argument to the AmberPrmtopFile constructor.
Semantics as for that parameter in CharmmPsfFile

### DIFF
--- a/wrappers/python/simtk/openmm/app/amberprmtopfile.py
+++ b/wrappers/python/simtk/openmm/app/amberprmtopfile.py
@@ -150,7 +150,7 @@ class AmberPrmtopFile(object):
                      implicitSolventSaltConc=0.0*(unit.moles/unit.liter),
                      implicitSolventKappa=None, temperature=298.15*unit.kelvin,
                      soluteDielectric=1.0, solventDielectric=78.5,
-                     removeCMMotion=True, hydrogenMass=None, ewaldErrorTolerance=0.0005):
+                     removeCMMotion=True, hydrogenMass=None, ewaldErrorTolerance=0.0005, switchDist=0.0*unit.nanometer):
         """Construct an OpenMM System representing the topology described by this prmtop file.
 
         Parameters:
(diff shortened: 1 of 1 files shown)


wrappers/python/simtk/openmm/app/amberprmtopfile.py
```diff
@@ -248,4 +254,16 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
                 force.setEwaldErrorTolerance(ewaldErrorTolerance)
         if removeCMMotion:
             sys.addForce(mm.CMMotionRemover())
+
+        if switchDistance and nonbondedMethod is not ff.NoCutoff:
+            # make sure it's legal
+            if switchDistance >= nonbondedCutoff:
```
swails: This comparison fails if `switchDistance` is a unit and `nonbondedCutoff` is not (or vice versa) (and it fails in a weird way, depending on which is a unit and which is not).  Probably best to convert both to nanometer scalars and compare the numbers that way:

```Python
if unit.is_quantity(switchDistance):
    switchDistance = switchDistance.value_in_unit(unit.nanometer)
if unit.is_quantity(nonbondedCutoff):
    nonbondedCutoff = nonbondedCutoff.value_in_unit(unit.nanometer)
```

wrappers/python/simtk/openmm/app/amberprmtopfile.py
```diff
@@ -248,4 +254,16 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
...
+                raise ValueError('switchDistance is too large compared '
+                                     'to the cutoff!')
+                if abs(switchDistance) != switchDistance:
```
swails: If you do the above, this can be simplified to `switchDistance < 0`, which will be easier to see intent immediately.

mj-harvey: Can do. That was a copy and paste job from the CharmmPsfFile code

swails: Ah, thanks for the pointer.  I'll clean up that code a bit (it was some of the first OpenMM code I wrote... :)

mj-harvey: No apologies necessary, I've been writing python for all of a week myself!

peastman:
I just noticed this PR was never merged.  Is it all ready at this point?

swails:
Not quite.  It needs the same changes as I did in #1180.  Or you can just merge this and I'll do that in a separate PR.

peastman:
No hurry.  I just wanted to check.

PR #1178, opened 2015-10-08
Added checks for nan coordinates in CUDA and OpenCL

peastman:
The check is done during molecule reordering, which means it has two limitations.  First, it can take up to 100 time steps for the nan to be detected.  Second, the check isn't done if the simulation is run with no cutoff.  But it's still a big improvement, and doing the check there means that it adds negligible overhead.


jchodera:
Thanks!

PR #1180, opened 2015-10-09
Add code supporting input with and without units for CharmmPsfFile.

swails:
Add relevant tests (that used to fail). This also fixes a subtle bug where
switching functions were not applied when NBFIXes were specified.


peastman:
AppVeyor has come up with a new way to fail:

```
doxygen.portable not installed. An error occurred during installation:
 The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.
```

I'm not very impressed with it...

swails:
I think you're being too hard on AppVeyor -- trusting relationships take time.

PR #1182, opened 2015-10-09
[WIP] Per-atom parameters can be taken from residue templates

### LINKED ISSUE #1168: Extending ffxml to separate charge and Lennard-Jones parameter assignment
We have been thinking about ways we can make the `ffxml` format for `simtk.openmm.app.ForceField` a bit friendlier both for users (who often want to "mix and match", such as combining AMBER forcefields with standard or multisite ion parameters) and for tools that need to use small molecules (such as YANK).  

While the `type` and `class` mechansim of the `ffxml` format is very powerful, one of the main sources of complexity right now is the definition of a new atom `type` for each atom appearing in each residue template.  Forcefields like AMBER define a very limited set of atom types (e.g. [`parm96.dat` defines 54 atom types](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/parm96.dat#L2-L55)), and even fewer Lennard-Jones parameters ([`parm96.dat` defines 35 LJ types](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/parm96.dat#L2-L55)).

If we had a mechanism by which the `NonbondedForce` section of the `ffxml` file could also specify a `charge` attribute for each object, and use this to assign charges, the complexity of the resulting `ffxml` files could be considerably simplified, and we could additionally mix-and-match things like the [`gaff.xml`](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/gaff.dat) small molecule forcefield that specifies parameters for small molecules because no name collisions would occur (provided we also permit parameters to be specified in a separate `ffxml` file from residue templates).

This change would allow us to extend the `<Atom>` element tag to permit an additional `charge attribute.  For alanine, for example, this would change from
```XML
  <Residue name="ALA">
   <Atom name="N" type="0"/>
   <Atom name="H" type="1"/>
```
to
```XML
  <Residue name="ALA">
   <Atom name="N" type="0" charge="-0.41570"/>
   <Atom name="H" type="1" charge="0.27190"/>
```
The atom types and classes would be redefined to match the AMBER atom type names and LJ classes, changing from:
```XML
 <AtomTypes>
  <Type name="0" class="N" element="N" mass="14.00672"/>
  <Type name="1" class="H" element="H" mass="1.007947"/>
  <Type name="2" class="CT" element="C" mass="12.01078"/>
  <Type name="3" class="H1" element="H" mass="1.007947"/>
```
to
```XML
 <AtomTypes>
  <Type name="N" class="N" element="N" mass="14.00672"/>
  <Type name="H" class="H" element="H" mass="1.007947"/>
  <Type name="CT" class="C" element="C" mass="12.01078"/>
  <Type name="H1" class="H1" element="H" mass="1.007947"/>
```
The `<NonbondedForce>` section of the `ffxml` would then be simplified by utilizing the smaller number of LJ atom classes, and either omit the charge altogether:
```XML
<NonbondedForce coulomb14scale="0.833333" lj14scale="0.5">
  <Atom class="N" sigma="0.324999852378" epsilon="0.71128"/>
  <Atom class="H" sigma="0.106907846177" epsilon="0.0656888"/>
  <Atom class="C" sigma="0.339966950842" epsilon="0.4577296"/>
  <Atom class="H1" sigma="0.247135304412" epsilon="0.0656888"/>
```
or specify a keyword that says the charge should be obtained from the `charge` attribute of the `<Atom>` tag:
```XML
<NonbondedForce coulomb14scale="0.833333" lj14scale="0.5">
  <Atom class="N" charge="atom@charge" sigma="0.324999852378" epsilon="0.71128"/>
  <Atom class="H" charge="atom@charge" sigma="0.106907846177" epsilon="0.0656888"/>
  <Atom class="C" charge="atom@charge" sigma="0.339966950842" epsilon="0.4577296"/>
  <Atom class="H1" charge="atom@charge" sigma="0.247135304412" epsilon="0.0656888"/>
```
Other syntax could be used to denote the attribute of interest, such as XPath syntax (is this `Atom/@charge`?) that could let us extend the `<Atom>` tag in the future with additional atom-specific information (such as AMOEBA multipole parameters).

Questions for the implementation:
* Can this simply be a backwards-compatible extension to the `ffxml` format, rather than a new non-backwards-compatible revision?  
* Even if this is not a revision, should we start adding an attribute that specifies the version of the `ffxml` spec we are using, such as
```XML
<ForceField version="1.0">
```
in order to better handle revisions in the future?
* Should we modify `NonbondedGenerator` to handle these extensions, or is this better handled by adding a custom `<Script>` section to the `<NonbondedForce>` sections of the `ffxml` files?  If a `Script`, how would we access the `<Atom>` `charge` attribute?
 * Can we easily change the behavior to allow residue templates to be specified in a different file from the parameters to facilitate mix-and-match use without causing problems?  We would still want an `Exception` to be thrown on atom type or class name collisions.
* Are there other opportunities for generalization we can exploit along the same lines?

peastman:
This fixes #1168.

I still need to write the documentation for this.  Also, there are a couple of force types that aren't yet supported.  AmoebaMultipoleForce is really complicated, and I'm not sure how this should work for it.  For example, it has two different sets of tags for per-atom properties, each with its own attributes: `<Multipole>` and `<Polarize>`.  And then there's AmoebaGeneralizedKirkwoodForce.  I have no idea what the heck is going on with that code!  Its `addParticle()` method takes three arguments: `charge`, `radius`, and `scalingFactor`.  The XML format, on the other hand, only has two attributes: `charge` and `shct`.  But it doesn't actually use them!  The generator completely ignores the attributes, instead getting the charge from the AmoebaMultipoleForce, selecting the radius based on the element, and hardcoding the scaling factor as 0.69.  So until I can figure out what's going on with that, I'm not going to touch it.

Anyway, if you want to start trying this out, it should be functional.


jchodera:
Do you want to tackle the changes to `<Bond>` to permit `fromName` and `toName` in a separate PR?  I also realize that the same extension should apply to `<ExternalBond>` as well.

peastman:
Yes, that's really a separate feature, so I'll put it in a separate PR.

jchodera:
OK, let me know if I should put it in a separate issue to make tracking easier.

jchodera:
Also see #1183 that relates to the feature implemented here.  I *think* that the proposal there fits well with this scheme (provided we *also* define dummy atom types), but there could be utility in reversing our decision as to whether `<Atom>` elements can have attributes that override default charges (to specify some atoms as dummy atoms in #1183).

jchodera:
I've moved the `<Bond>` and `<ExternalBond>` issue to #1187.

Is this ready to be merged?

peastman:
No, I still need to do the documentation for it.  But I'm on vacation right now, so that might not happen until I get back.

jchodera:
Thanks!

Issue #1191, opened 2015-10-18
Energy is NaN error while using periodicdistance() function


dr-nate:
Hey all,

I've been trying to use the new periodicdistance() function to restrain an atom for NPT equilibration, but at the first energy check it throws an error:

```
#"Progress (%)"	"Step"	"Time (ps)"	"Potential Energy (kJ/mole)"	"Kinetic Energy (kJ/mole)"	"Total Energy (kJ/mole)"	"Temperature (K)"	"Box Volume (nm^3)"	"Speed (ns/day)"	"Time Remaining"
Traceback (most recent call last):
  File "openmm-equil.py", line 131, in <module>
    simulation.step(numsteps)
  File "/Users/stanlen1/anaconda/lib/python3.4/site-packages/simtk/openmm/app/simulation.py", line 96, in step
    self._simulate(endStep=self.currentStep+steps)
  File "/Users/stanlen1/anaconda/lib/python3.4/site-packages/simtk/openmm/app/simulation.py", line 173, in _simulate
    reporter.report(self, state)
  File "/Users/stanlen1/anaconda/lib/python3.4/site-packages/simtk/openmm/app/statedatareporter.py", line 162, in report
    self._checkForErrors(simulation, state)
  File "/Users/stanlen1/anaconda/lib/python3.4/site-packages/simtk/openmm/app/statedatareporter.py", line 309, in _checkForErrors
    raise ValueError('Energy is NaN')
ValueError: Energy is NaN
```
Looking deeper into the issue by writing out parameters for that atom every frame, it seems that the position and force on the particle (a single atom) are NaN from the first step of the simulation. Here is the code for applying the restraint:

``` 
# Add positional restraints to ion
force = mm.CustomExternalForce("k*periodicdistance(x, y, z, x0, y0, z0)^2")
force.addGlobalParameter("k", 5.0*u.kilocalories_per_mole/u.angstroms**2)
force.addPerParticleParameter("x0")
force.addPerParticleParameter("y0")
force.addPerParticleParameter("z0")
for i, atom_crd in enumerate(parm.positions):
    if parm.atoms[i].name in ('K+'):
        print(atom_crd.value_in_unit(u.nanometers))
        force.addParticle(i, atom_crd.value_in_unit(u.nanometers))
system.addForce(force)
```
If I swap out that first line with the one below, the simulation can usually complete without issue. 

```
[...]

PR #1192, opened 2015-10-19
Fix gradient of periodicdistance at 0

rmcgibbo:
cc #1191

### DIFF
--- a/platforms/cuda/src/CudaExpressionUtilities.cpp
+++ b/platforms/cuda/src/CudaExpressionUtilities.cpp
@@ -109,7 +109,8 @@ void CudaExpressionUtilities::processExpression(stringstream& out, const Express
                 }
                 out << ");\n";
                 out << "APPLY_PERIODIC_TO_DELTA(periodicDistance_delta)\n";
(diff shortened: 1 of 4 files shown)


peastman:
I don't think that's actually guaranteed to work.  Depending on the implementation, inf*0 could be nan.  Instead of

```
periodicDistance_delta.x*periodicDistance_rinv*(periodicDistance_r2>0)
```

write

```
(periodicDistance_r2 > 0 ? periodicDistance_delta.x*periodicDistance_rinv : 0)
```

Could you give the test case a meaningful name?  Something like "testZeroPeriodicDistance".

rmcgibbo:
Okay, done.

rmcgibbo:
For CUDA and OpenCL, the test I added didn't actually fail with the old code -- it was giving a very very large force, but not NaN.

peastman:
There was a travis failure that looks like a networking error or problem on Github:

```
Cloning into 'pandegroup/openmm'...
warning: Could not find remote branch add-test-for-1191 to clone.
fatal: Remote branch add-test-for-1191 not found in upstream origin
Unexpected end of command stream
```

Since only one of the four failed, I'm not going to worry about it.

PR #1199, opened 2015-10-21
Modify Context.getState to accept a set of indices or bitmask

rmcgibbo:


swails:
:+1:

rmcgibbo:
Actually I don't think the sum( is correct if the set has repeats. It should be reduce(operator.bisewiseor

swails:
I like this a lot.  I use force groups fairly extensively (since it really helps to be able to compare energy components when doing validation).

peastman:
This doesn't look right to me.  You shouldn't need a special case for -1.  And if you give it groups=7 (111, so include groups 0, 1, and 2), won't this give you group 7 instead?  It should be checking to see whether the argument is a number or set, then handling each one correctly.

Also, the same change should be made to RPMDIntegrator, since it also has a getState() method.

swails:
He's only special-casing -1 when it's part of a collection.  In this case you *do* need to special-case it, because

```Python
>>> 1 << -1
ValueError: negative shift count
>>> 
```

If `groups` is just -1, then it's handled in the exception clause (group_mask is -1).  If you give it an integer, then the group mask *is* that integer.  If you give it a collection of groups, then the group mask is constructed by collecting the bits of all the groups you want to add up.

So `context.getState(groups=7)` will pull out groups 0, 1, and 2.  On the flipside, `context.getState(groups=[7])` will pull out just group 7 (`1<<7`).  This doesn't break any backwards compatibility (that is tested for in the test case @rmcgibbo added), and adds a nice, intuitive convenience to the Python app layer IMO.

rmcgibbo:
Perhaps the code is too clever in exception handling. I wanted to make sure it handles duck typed sets properly.

swails:
> Perhaps the code is too clever in exception handling. I wanted to make sure it handles duck typed sets properly.

I think it's fine.  It converts all iterables to sets, which covers listcomps, generator comps, lists, and tuples.  I can imagine all 3 of those being common input with the same intent.

The way I would expect (and want) this API to work is to interpret iterables as a list of groups to get the force for and interpret non-iterables the way everything used to be interpreted (which is exactly what it does).  This discussion suggests that the code slice that is responsible for this probably warrants an explanatory comment, but I like how it behaves as-is.

Note that this just changes the behavior of `getState(groups=<some iterable>)` from throwing a TypeError to actually doing something intuitive :). :+1:

swails:
> I can imagine all 3 of those being common input with the same intent.

There are 3 types of people in this world: those that can count, and me.

`s/3/4/`

wrappers/python/src/swig_doxygen/swig_lib/python/extend.i
```diff
@@ -55,26 +55,32 @@
...
-        if enforcePeriodicBox: enforcePeriodic=1
-        else: enforcePeriodic=0
+        import functools, operator
+
+        getP, getV, getF, getE, getPa, enforcePeriodic = map(bool,
+            (getPositions, getVelocities, getForces, getEnergy, getParameters,
+             enforcePeriodicBox))
+
+        try:
+            groups = set(groups)
```
rmcgibbo: @swails instead of using the `TypeError` thrown here if `groups` is an integer to trigger the control flow, I'll move it out into a more documented check. Obviously I'll still keep the duck typing.

wrappers/python/src/swig_doxygen/swig_lib/python/extend.i
```diff
@@ -55,26 +55,28 @@
...
+            (getPositions, getVelocities, getForces, getEnergy, getParameters,
+             enforcePeriodicBox))
+
+         try:
+             # is the input integer-like?
+             groups_mask = int(groups)
+         except TypeError:
+             # nope, okay, then it should be an iterable
+             groups_mask = functools.reduce(operator.or_,
+                     ((1<<x) & 0xffffffff for x in set(groups)))
```
swails: Did you mean to get rid of -1 support in the force set?

rmcgibbo: yeah. i don't think it serves much purpose if you're explicitly specifying the set anyways.

swails: OK then :)

rmcgibbo: if the set includes -1, no matter the remaining content, then the call is just equivalent to using the default argument.

swails: You mean in principle, or now?  Because if -1 is in the group set now, it'll throw a ValueError when it tries to do `1 << -1`.

I have no problems with -1 in the group set throwing an exception rather than rendering every other element in the set irrelevant, I'm just trying to ascertain intent here.

rmcgibbo: I mean that adding support for `-1`-handling in the set doesn't add any new functionality.

Also, it seems contrary to the purpose of passing in a set from Python, which is to avoid dealing with bit-twiddling and the binary representation of integers. To understand why -1 means "all forcegroups", you need to know how two's complement works.

swails: > I mean that adding support for -1-handling in the set doesn't add any new functionality.

Got it.

> Also, it seems contrary to the purpose of passing in a set from Python, which is to avoid dealing with bit-twiddling and the binary representation of integers. To understand why -1 means "all forcegroups", you need to know how two's complement works.

Agreed.  Hence why I said I have no problems with it :).  I was just remarking because it was a change in functionality from one commit to the next.

peastman:
Much clearer now.  And I agree, -1 doesn't belong in the set.

It still needs the corresponding changes to RPMDIntegrator.  It's in the same file.

rmcgibbo:
The changes at lines 212-228 are the RPMD integrator.

peastman:
I've been having second thoughts about accepting arbitrary iterables.  I'm worried users (especially ones with limited Python experience) will type `groups=(3)`, expecting to get group 3, not knowing or remembering that to get a one element tuple you have to write `(3,)`.  If we only accept sets so you have to write `groups={3}` instead, that eliminates a bug opportunity without making it significantly less convenient to use.

jchodera:
I am not sure this is a real issue. Beginners to Python are far more likely
to be familiar with lists than tuples, and [3] is a perfectly valid
iterable.  I think the (3) vs (3,) thing is something that the
documentation could warn against.

Alternatively, what about using different arguments for bitmask vs set?

peastman:
A lot of tutorials make pretty heavy use of tuples.  The point is that someone *will* make this mistake sooner or later.  And it's a high level bug opportunity: it silently produces wrong results, and could be really hard to track down.

Compare that to what you lose: at the absolute worst, you'll have to wrap the argument in `set(...)`, so five extra characters to type.  That's not much of a downside, and usually it's not even as much as that, because creating a set right from the beginning is just as easy as creating a list or tuple.

rmcgibbo:
-1. The majority of python APIs are duck-typed like this.

jchodera:
I'm with @rmcgibbo. I think we want to do a complete review of the Python API at some point and see what opportunities exist to make it more pythonic.  We want to make it *more * intuitive for Python users, not less.

peastman:
The majority of Python APIs do not take *either* iterables *or* ints, and interpret them in very different ways.  The more I think about this, the more convinced I am that allowing arbitrary iterables would be a very very bad decision.  It's a very easy mistake to make, people *will* make it, and it *will* lead to errors.  And we can easily prevent that with negligible loss in power and convenience, just by requiring it to be a set.

rmcgibbo:
Ideally, `groups` should not accept an integer -- it should only accept an iterable. For backward compatibility, IMO, support for accepting an int should be maintained but deprecated.

swails:
> The majority of Python APIs do not take either iterables or ints

A sizeable number do, though.  The ones that immediately come to mind are array indexing, particularly in numpy:

```Python
>>> import numpy as np
>>> arr = np.random.rand(100)
>>> arr[(1)]
0.24045127255333898
>>> arr[((1,2))]
IndexError: too many indices for array
>>> arr[((1,2),)]
array([ 0.24045127,  0.88458535])
```

And duck typing reigns supreme.  It takes really any indexable iterator.  Indexing via integers returns an integer, indexing via a tuple returns, also, a number or an ndarray, and indexing via a tuple of tuples yields an ndarray.

I tend to lean toward the duck typing stance myself, although mistaken uses of numpy array indexing is more likely to yield some kind of type or value error than passing an integer instead of a tuple so I can understand the concerns here.

But if Python-OpenMM users are also very likely to be numpy users, I don't foresee the same severity of issues in duck typing.

peastman:
I'm still very strongly opposed to this.  To understand why, take a look at this article: http://www.drdobbs.com/jvm/bug-opportunity-levels/222200364.  It's a really useful guide to avoiding bugs, written by a really amazing programmer. :)  Everyone should read it.  What you're proposing is that we introduce a level 3 bug opportunity for negligible benefit, just because it's a "more pythonic" thing to do.

rmcgibbo:
lol.

rmcgibbo:
I relent :)

swails:
As I do *not* feel strongly either way, so do I. :)

PR #1204, opened 2015-10-24
Add OpenCL on TravisCI

rmcgibbo:
With the OpenCL tests running on Travis, the test suite takes longer than 50 minutes (the timeout), so I changed the `ctest` invocation to run the tests in a random order with a ~~30~~ 40 minute timeout. Across multiple invocations, hopefully this will get reasonable coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: python
-python:
-  - "2.7_with_system_site_packages"
-  - "3.4"
 
-sudo: false
 addons:
   apt:
     sources:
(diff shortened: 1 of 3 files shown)


peastman:
What OpenCL implementation is it using?

It looks like there were a number of failures in OpenCL test cases.  But since you told it failures were OK, you'll never know that unless you actually pull up the build log for that VM.  Which kind of limits the value of this.

rmcgibbo:
Yes, obviously it has limited value, but it seems like a small improvement over the current situation. Presumably it's using whatever CPU OpenCL is provided by `fglrx`. If you want to turn off the `allowed_failures` -- go for it. None of this is set in stone.

jchodera:
What if we install both CUDA and the AMD APP SDK using a scheme similar to [this](https://github.com/FoldingAtHome/docker-core-linux-build-environment/blob/master/Dockerfile#L11-L22)?  That way, we can at least make sure the CUDA and OpenCL platforms *compile* correctly, even if we can't run the CUDA or OpenCL tests.

rmcgibbo:
The OpenCL is compiling and running just fine with this PR. That could work for CUDA -- I'm not sure. I recommend you open another PR if you want to try.

peastman:
Are you sure it's using fglrx?  It may be using the Mesa OpenCL implementation, which is still buggy and incomplete.

> The OpenCL is compiling and running just fine with this PR.

It's not running fine.  There were four failures in OpenCL tests, and that's not even having run the full test suite.

rmcgibbo:
I don't have any secret information beyond what you have, Peter. I assume that `sudo apt-get install -qq fglrx=2:8.960-0ubuntu1 opencl-headers`, installs `fglrx, but I may be wrong. I copied that line from the Travis-CI configuration from boost::compute (https://github.com/boostorg/compute), which they appear to be using succesfully.

rmcgibbo:
Do you know if there's something similar to the CUDA `deviceQuery` program that's included with the CUDA samples for OpenCL? We could run that at the top of the test script. From looking through the boost::compute issues, I see a link to https://travis-ci.org/ddemidov/vexcl/jobs/18265418#L280-L300 which might be useful to add.

rmcgibbo:
I added https://github.com/rmcgibbo/openmm/blob/cl/platforms/opencl/tests/TestOpenCLDeviceQuery.cpp that just prints to stdout, but that might not be the best way to organize it.

rmcgibbo:
```
OpenCL devices:
OpenCL Platform 0, Device 0: "AMD Opteron(tm) Processor 6376                 "
    CL_PLATFORM_NAME                 = Advanced Micro Devices, Inc.
    CL_DEVICE_VENDOR                 = AuthenticAMD
    CL_DEVICE_MAX_COMPUTE_UNITS      = 32
    CL_DEVICE_GLOBAL_MEM_SIZE        = 3221225472
    CL_DEVICE_MAX_CLOCK_FREQUENCY    = 2300
```

rmcgibbo:
grr. git is weird.

Issue #1203, opened 2015-10-23
Test suite fails when GPUs are in compute exclusive mode


rmcgibbo:
@dr-nate and I have both lost a couple hours independently trying to debug apparent test failures on two different clusters that, I'm pretty sure, stem from the same issue. The issue is an apparent failure in many of many of the tests. The observed behavior is

```
[rmcgibbo@xs-0061 ~/projects/openmm/build]$ ./TestCudaHarmonicBondForce 
exception: Error initializing Context: CUDA_ERROR_INVALID_DEVICE (101) at /home/stanford/pande/users/rmcgibbo/projects/openmm/platforms/cuda/src/CudaContext.cpp:167
```

AFACT this occurs on any machine where the GPUs are in compute exclusive mode. You can reproduce the same failure mode with [this](https://gist.github.com/rmcgibbo/81688b8f7e683c1c6db9#file-main-cpp) very simple CUDA program. Running under strace (`strace -f -e clone ...`) you see that the creation of a new CUDA context internally involves the creation of a new child process with a different PID, which means that you can't create two separate CUDA contexts inside of a single test case if the GPUs are configured in exclusive process. This pattern is used a lot in the tests.

I'm not exactly sure if there's a solution. I guess perhaps we could query the device mode? Or perhaps when `CudaContext.cpp:167` throws `CUDA_ERROR_INVALID_DEVICE`, the code should loop back and try to use a different `deviceIndex` -- this at least makes things work if you have >=2 GPUs.

jchodera:
We've run into this too when running production codes, and it took a while to debug what was going on.  This would be great to document in a few places (or maybe we need an FAQ?) to help save other users time.

Our batch queue system (Torque/Moab) allows the user to specify what mode the GPU is to be reset into on submission time.  We set the default to keep the GPUs in shared mode, which works just fine for us.  Is this impossible on your clusters?

> you can't create two separate CUDA contexts inside of a single test case if the GPUs are configured in exclusive process. This pattern is used a lot in the tests.

Could we just fix the tests?  Or recommend the tests be run in shared mode?

rmcgibbo:
It's not possible to change the compute mode without root. On some of our setups, it also needs to be changed cluster-wide, and cannot be changed on a per-GPU or per node-basis, according to the system administrators that I asked.

jchodera:
What queuing system is used? Any queueing system that supports GPUs allows
a prerun script to be run sudo that can control the GPU setting.

[...]

PR #1205, opened 2015-10-24
Ensure context can be created when selecting CUDA device

rmcgibbo:
cc #1203, the idea is to check for failures in `cuCtxCreate` and try the next device if it fails, instead of throwing an exception.

### DIFF
--- a/platforms/cuda/src/CudaContext.cpp
+++ b/platforms/cuda/src/CudaContext.cpp
@@ -1408,7 +1408,7 @@ vector<int> CudaContext::getDevicePrecedence() {
         if (major == 1 && minor < 2)
             continue;
 
-        if ((useDoublePrecision || useMixedPrecision) && (major*0.1*minor < 1.3))
+        if ((useDoublePrecision || useMixedPrecision) && (major+0.1*minor < 1.3))
             continue;
 
         CHECK_RESULT(cuDeviceGetAttribute(&clock, CU_DEVICE_ATTRIBUTE_CLOCK_RATE, thisDevice));
--- a/platforms/cuda/src/CudaContext.cpp
+++ b/platforms/cuda/src/CudaContext.cpp
@@ -122,49 +122,48 @@ CudaContext::CudaContext(const System& system, int deviceIndex, bool useBlocking
     CHECK_RESULT(cuDeviceGetCount(&numDevices));
     if (deviceIndex < -1 || deviceIndex >= numDevices)
         throw OpenMMException("Illegal value for CudaDeviceIndex: "+intToString(deviceIndex));
-    if (deviceIndex == -1) {
-        // Try to figure out which device is the fastest.
 
-        int bestSpeed = -1;
-        int bestCompute = -1;
-        for (int i = 0; i < numDevices; i++) {
-            CHECK_RESULT(cuDeviceGet(&device, i));
-            int major, minor, clock, multiprocessors;
-            CHECK_RESULT(cuDeviceComputeCapability(&major, &minor, device));
-            if (major == 1 && minor < 2)
-                continue; // 1.0 and 1.1 are not supported
-            CHECK_RESULT(cuDeviceGetAttribute(&clock, CU_DEVICE_ATTRIBUTE_CLOCK_RATE, device));
-            CHECK_RESULT(cuDeviceGetAttribute(&multiprocessors, CU_DEVICE_ATTRIBUTE_MULTIPROCESSOR_COUNT, device));
-            int speed = clock*multiprocessors;
-            if (major > bestCompute || (major == bestCompute && speed > bestSpeed)) {
-                deviceIndex = i;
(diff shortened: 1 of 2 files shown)


jchodera:
Since the default behavior without any specified device index is to try to find the fastest device of all available devices anyway, the failthrough to find an *available* device sounds reasonable.  But it's probably worth discussing whether this behavior might be dangerous by causing surprising, unexpected, or difficult to control behavior.

For example, we want to make sure that if you're trying to benchmark a or request specific GPU by either setting `CUDA_VISIBLE_DEVICES` or using the `cudaDeviceIndex` property, you still get an exception if that doesn't work.  It looks like the code changes to preserve this.

I'm also actually a little confused why requesting a specific precision model doesn't cause an exception, but instead causes a silent fail-through---that certainly doesn't seem like something we want unless the user somehow says that this is OK:
```
        if ((useDoublePrecision || useMixedPrecision) && computeCapability < 1.3)
            continue;
            // throw OpenMMException("This device does not support double precision");
```

mpharrigan:
cc #1034 ?

rmcgibbo:
The `CUDA_VISIBLE_DEVICES` works as expected without any special code, because it changes the number of devices reported by the CUDA API and the indexing. I didn't change the behavior if `CudaDeviceId` is specified, so I don't see anything "surprising, unexpected, or difficult to control behavior."

peastman:
Looks good.  What aspects are still WIP?

jchodera:
You commented out that exception that is thrown if the platform does not
support the requested precision model. Isn't that a sudden change in
behavior?

rmcgibbo:
It just changes the exception message from "This GPU doesn't support double precision" to "No compatible CUDA device is available".

rmcgibbo:
Originally I wanted this to fix #1203, but there are some other blockers that could be addressed independently. Should I port this to OpenCL too? Otherwise, I'll remove the WIP tag.

rmcgibbo:
I can also clean it up a bit -- there's a bit of redundancy between the checks done in `getDevicePrecedence` and the calling loop.

rmcgibbo:
@peastman: I tried to do this for OpenCL too, but I'm not super confident with all of the moving parts here, so I might have gotten something wrong: https://github.com/rmcgibbo/openmm/compare/which-device...rmcgibbo:which-device-cl2

peastman:
This is probably a lot less important for OpenCL than for CUDA.  The main place this situation is going to come up is clusters with multiple GPUs per node, and those pretty universally have Nvidia GPUs.  So let's leave OpenCL alone for the present and first see how this works on CUDA, and whether anyone runs into problems with it.

jchodera:
> It just changes the exception message from "This GPU doesn't support double precision" to "No compatible CUDA device is available".

Ah, OK!

rmcgibbo:
> This is probably a lot less important for OpenCL than for CUDA.

Got it. I changed the title, and I think it's ready to merge unless you see any bugs or anything that needs to be changed.

platforms/cuda/src/CudaContext.cpp
```diff
@@ -1393,3 +1392,41 @@ void CudaContext::WorkThread::flush() {
...
+
+    CHECK_RESULT(cuDeviceGetCount(&numDevices));
+    for (int i = 0; i < numDevices; i++) {
+        CHECK_RESULT(cuDeviceGet(&thisDevice, i));
+        int major, minor, clock, multiprocessors, speed;
+        CHECK_RESULT(cuDeviceComputeCapability(&major, &minor, thisDevice));
+        if (major == 1 && minor < 2)
+            continue;
+
+        if ((useDoublePrecision || useMixedPrecision) && (major*0.1*minor < 1.3))
```
peastman: I think that's supposed to be major+0.1*minor, not major*0.1*minor.

platforms/cuda/src/CudaContext.cpp
```diff
@@ -122,49 +122,48 @@ CudaContext::CudaContext(const System& system, int deviceIndex, bool useBlocking
...
+            this->deviceIndex = deviceIndex;
+            break;
         }
     }
-    if (deviceIndex == -1)
-        throw OpenMMException("No compatible CUDA device is available");
-    CHECK_RESULT(cuDeviceGet(&device, deviceIndex));
-    this->deviceIndex = deviceIndex;
+    if (this->deviceIndex == -1)
+        if (devicePrecedence.size() == 1)
```
peastman: The check should be `deviceIndex == -1`, not `devicePrecedence.size() == 1` (except that you've overwrtten the value originally passed for `deviceIndex`, which isn't a good idea--use a new variable at line 135 instead).  If `devicePrecedence.size() == 1`, that often is because there's only one device installed, not because they requested a specific device.

peastman:
Also note that recent CUDA releases only support devices with compute capability 2.0 and up, all of which support double precision.

PR #1209, opened 2015-10-26
Fix thread synchronization in customCentroidBond.cl

rmcgibbo:
This fixes the `TestOpenCLCustomCentroidBondForce` failures on the CPU OpenCL device I reported in https://github.com/pandegroup/openmm/pull/1208#issuecomment-150995439

Based on my undersanding of the [spec](https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/barrier.html), and [some forum answers](https://software.intel.com/en-us/forums/opencl/topic/413956), the SYNC_WARPS barrier can't be in a conditional that's hit by only a subset of the threads.


jchodera:
Good catch!

PR #1210, opened 2015-10-26
Added getNumAtoms, getNumResidues, getNumChains to simtk.openmm.app.Topology

jchodera:
This addresses #1207 by adding the following to `simtk.openmm.app.Topology`:
* `getNumAtoms()`
* `getNumResidues()`
* `getNumChains()`

I've also added a corresponding test.

### DIFF
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -70,24 +70,18 @@ class Topology(object):
 
     def getNumAtoms(self):
         """Return the number of atoms in the Topology.
-
-        Returns: the number of atoms in the Topology
         """
-        natom = sum(1 for a in self.atoms())
+        natom = self.__numAtoms
         return natom
 
     def getNumResidues(self):
         """Return the number of residues in the Topology.
-
-        Returns: the number of residues in the Topology
         """
-        nres = sum(1 for r in self.residues())
+        nres = self.__numResidues
         return nres
 
     def getNumChains(self):
         """Return the number of chains in the Topology.
-
-        Returns: the number of chains in the Topology
         """
         nchain = len(self._chains)
         return nchain
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -89,7 +89,7 @@ class Topology(object):
 
         Returns: the number of chains in the Topology
         """
-        nchain = sum(1 for c in self.chains())
+        nchain = len(self._chains)
         return nchain
 
     def addChain(self, id=None):
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -62,8 +62,8 @@ class Topology(object):
 
     def __repr__(self):
         nchains = len(self._chains)
-        nres = sum(1 for r in self.residues())
-        natom = sum(1 for a in self.atoms())
+        nres = self._numResidues
+        natom = self._numAtoms
         nbond = len(self._bonds)
         return '<%s; %d chains, %d residues, %d atoms, %d bonds>' % (
                 type(self).__name__, nchains, nres, natom, nbond)
@@ -71,13 +71,13 @@ class Topology(object):
     def getNumAtoms(self):
         """Return the number of atoms in the Topology.
         """
-        natom = self.__numAtoms
+        natom = self._numAtoms
         return natom
 
     def getNumResidues(self):
         """Return the number of residues in the Topology.
         """
-        nres = self.__numResidues
+        nres = self._numResidues
         return nres
 
     def getNumChains(self):
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -68,6 +68,30 @@ class Topology(object):
         return '<%s; %d chains, %d residues, %d atoms, %d bonds>' % (
                 type(self).__name__, nchains, nres, natom, nbond)
 
+    def getNumAtoms(self):
+        """Return the number of atoms in the Topology.
+
+        Returns: the number of atoms in the Topology
+        """
+        natom = sum(1 for a in self.atoms())
+        return natom
+
+    def getNumResidues(self):
+        """Return the number of residues in the Topology.
+
+        Returns: the number of residues in the Topology
+        """
+        nres = sum(1 for r in self.residues())
+        return nres
+
+    def getNumChains(self):
+        """Return the number of chains in the Topology.
+
+        Returns: the number of chains in the Topology
+        """
+        nchain = sum(1 for c in self.chains())
+        return nchain
+
     def addChain(self, id=None):
         """Create a new Chain and add it to the Topology.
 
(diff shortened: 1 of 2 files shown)


jchodera:
Should be good to merge once appveyor completes.

wrappers/python/simtk/openmm/app/topology.py
```diff
@@ -68,6 +68,30 @@ def __repr__(self):
...
+        nres = sum(1 for r in self.residues())
+        return nres
+
+    def getNumChains(self):
+        """Return the number of chains in the Topology.
+
+        Returns: the number of chains in the Topology
+        """
+        nchain = sum(1 for c in self.chains())
+        return nchain
```
swails: Maybe change this to `return len(self._chains)`?

jchodera: Sure.

peastman:
There's a much more efficient implementation: `Topology` already has `_numAtoms` and `_numResidues` fields you can return directly.

It's just a minor stylistic point, but I'm not fond of this sort of repetitively redundant redundancy:

>        """Return the number of atoms in the Topology.
>
>        Returns: the number of atoms in the Topology

In cases like this, I think it's sufficient to say it only once. :)

jchodera:
> There's a much more efficient implementation: Topology already has _numAtoms and _numResidues fields you can return directly.

Then why did you write [this](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/topology.py#L65-L66) in the `__repr__()` implementation?
```python
        nres = sum(1 for r in self.residues())
        natom = sum(1 for a in self.atoms())
```

> It's just a minor stylistic point, but I'm not fond of this sort of repetitively redundant redundancy:
> In cases like this, I think it's sufficient to say it only once. :)

Would you like to write a standards document describing in detail how OpenMM `app` layer docstrings should be written?  :)

I did my best to follow whatever nonstandard scheme you had adopted for docstrings.  At this point, you're asking me to read your mind.

I believe all our other `omnia` projects have adopted the [`numpy` docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt).  It actually physically pains me to depart from that.

jchodera:
Translation:
1. Do you want me to change the `__repr__()` implementation to?
2. I'll change the docstring, but will grumble about it.

swails:
> Then why did you write this in the __repr__() implementation?
> 
```
>         nres = sum(1 for r in self.residues())
>         natom = sum(1 for a in self.atoms())
```

All you can blame @peastman for on that one is merging my PR...  I did that because I didn't know about `_numAtoms` or `_numResidues` attributes, most likely.

> Would you like to write a standards document describing in detail how OpenMM app layer docstrings should be written? :)

Unlike every other omnia package, OpenMM docs are doxygen-generated rather than Sphinx-generated.  I believe it's this distinction that necessitates differently-formatted docstrings (although that may not be the *only* one).

> I did my best to follow whatever nonstandard scheme you had adopted for docstrings. At this point, you're asking me to read your mind.

I think that's unfair.  He pointed out a minor preference with an explanation -- not something one would do if they expected their mind read (nor was it phrased in a way that I interpreted as you should have known better).

I'm not sure how much a standards doc on this would do.  Most people (or maybe just me?) wouldn't really read it, but try and follow existing style when contributing.  If we got something wrong, it'd come up in code review (as it has now), and I'd change it; sans grumbles.

> 1. Do you want me to change the __repr__() implementation too?

Yes, please.

> I believe all our other omnia projects have adopted the numpy docstring standard. It actually physically pains me to depart from that.

My 1yo son shared his viral infection with me, which included in its symptoms full body muscle aches.  I found that over-the-counter ibuprofin worked wonders (at least for awhile).

wrappers/python/simtk/openmm/app/topology.py
```diff
@@ -68,6 +68,24 @@ def __repr__(self):
         return '<%s; %d chains, %d residues, %d atoms, %d bonds>' % (
                 type(self).__name__, nchains, nres, natom, nbond)
 
+    def getNumAtoms(self):
+        """Return the number of atoms in the Topology.
+        """
+        natom = self.__numAtoms
```
swails: `_numAtoms`

wrappers/python/simtk/openmm/app/topology.py
```diff
@@ -68,6 +68,24 @@ def __repr__(self):
...
+        return natom
+
+    def getNumResidues(self):
+        """Return the number of residues in the Topology.
+        """
+        nres = self.__numResidues
```
swails: `_numResidues`

jchodera:
> Unlike every other omnia package, OpenMM docs are doxygen-generated rather than Sphinx-generated. I believe it's this distinction that necessitates differently-formatted docstrings (although that may not be the only one).

That's true of the OpenMM API layer, which is distinct from the `app` layer.  We could adopt a different scheme for the `app` layer.  In fact, we do---the formats are different (especially regarding function signatures and return values).

The autogenerated OpenMM API Python docs are actually problematic because they are inaccurate.  They reflect the C++ API, and the reader must mentally translate them into Python by omitting or translating arguments as appropriate.  This can be super confusing to new users (and *was* at the recent OpenMM Workshop in Jülich).  We might consider changing this too.

> My 1yo son shared his viral infection with me, which included in its symptoms full body muscle aches. I found that over-the-counter ibuprofin worked wonders (at least for awhile).

You can take up to two grams of ibuprofen a day!  We're actually trying to use this fact that it binds HSA to saturate binding sites that would otherwise soak up imaging molecules.  

Feel better!

wrappers/python/simtk/openmm/app/topology.py
jchodera: Fixed!  Thanks.

wrappers/python/simtk/openmm/app/topology.py
jchodera: Fixed!

jchodera:
Travis passes!

swails:
> That's true of the OpenMM API layer, which is distinct from the app layer. We could adopt a different scheme for the app layer. In fact, we do---the formats are different (especially regarding function signatures and return values).

I believe that the SWIG-generated wrapper Python script is automatically run through some kind of scrubber to prettify doxygen output.  I know we had a discussion in the past about switching to Sphinx for API docs and it was decided against.

But Sphinx *definitely* processes the app-layer docstrings -- [you can see some doxygen-generated stuff here](http://docs.openmm.org/6.3.0/api-python/classsimtk_1_1openmm_1_1app_1_1amberprmtopfile_1_1AmberPrmtopFile.html#a04c2b6fd2a8d956129b2be4517534cc7) and [the docstring that generates it here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/amberprmtopfile.py#L154-L177).

> The autogenerated OpenMM API Python docs are actually problematic because they are inaccurate. They reflect the C++ API, and the reader must mentally translate them into Python by omitting or translating arguments as appropriate. This can be super confusing to new users (and was at the recent OpenMM Workshop in Jülich). We might consider changing this too.

True, and this warrants an extra issue IMO (I'm never 100% certain that I'm getting the call signature right before I test it just by looking at the docs, but .  But this is only in the API layer).  There is no C++-like pass-by-reference functionality for immutables in Python for obvious reasons, so where C++ uses that idiom, the docs for Python are wrong.

I think it still stands that switching to numpydoc conventions isn't likely, since it won't work nicely with Doxygen (if I'm mistaken, please correct me).

Issue #1190, opened 2015-10-15
copy.deepcopy of Force objects does not give same type of Force


jchodera:
Somehow `copy.deepcopy` was broken recently, and now gives generic `Force` objects instead of the correct subclasses:
```python
#!/usr/bin/env python                                                                                                                                                                          
from simtk import openmm                                                                                                                                                                       
import copy                                                                                                                                                                                    
force = openmm.NonbondedForce()                                                                                                                                                                
force_copy = copy.deepcopy(force)                                                                                                                                                              
print force                                                                                                                                                                                    
print force_copy
```
gives
```
<simtk.openmm.openmm.NonbondedForce; proxy of <Swig Object of type 'OpenMM::NonbondedForce *' at 0x7f81ada05cf0> >
<simtk.openmm.openmm.Force; proxy of <Swig Object of type 'OpenMM::Force *' at 0x7f80fdc1fc60> >
```
instead of both objects being of type `NonbondedForce`.

jchodera:
The [fast copy implementation](https://github.com/pandegroup/openmm/blob/master/wrappers/python/src/swig_doxygen/swig_lib/python/extend.i#L458) from #1055 may be the culprit.  Here's [the diff](https://github.com/pandegroup/openmm/commit/a707b2b41a60914e2214839a08bbfa86bdb68745) for that commit.

We should also add some python tests that ensure copied `Force` subclasses end up as the same subclass.

jchodera:
Actually, where *are* the Python tests for the Python API?

rmcgibbo:
https://github.com/pandegroup/openmm/tree/master/wrappers/python/tests

jchodera:
Would it be valuable for me to add a Python test in a PR even if I don't
know how to fix the actual bug?  Or should I just wait for @peastman?

jchodera:
Alternatively, I could add a test and revert the change that broke things.

peastman:
Let's not revert it yet.  Hopefully we can find a way to make it work without losing the benefits.

But if we can't find a way to do that, it's probably OK to remove the fast copy implementation for Force and Integrator.  I suspected it's a lot more common to clone Systems than Forces, so that's the most important one.  And Integrators are very cheap to clone.

peastman:
This is fixed by #1219.

PR #1211, opened 2015-10-26
Fix force deepcopy bug

jchodera:
Intended to address #1190 

Adding a Python test to catch the deepcopy bug as a first step.

### DIFF
--- a/wrappers/python/tests/TestPickle.py
+++ b/wrappers/python/tests/TestPickle.py
@@ -9,7 +9,7 @@ import copy
 import pickle
 
 class TestPickle(unittest.TestCase):
-    """Pickling / deepcopy of OpenMM state and integrator objects."""
+    """Pickling / deepcopy of OpenMM objects."""
 
     def setUp(self):
         """Set up the tests by loading the input pdb files and force field
@@ -26,9 +26,8 @@ class TestPickle(unittest.TestCase):
         self.pdb2 = PDBFile('systems/alanine-dipeptide-implicit.pdb')
         self.forcefield2 = ForceField('amber99sb.xml', 'amber99_obc.xml')
 
-
-    def test_deepcopy(self):
-        """Test that serialization/deserialization works (via deepcopy)."""
+    def test_system_integrator_deepcopy(self):
+        """Test that serialization/deserialization of system and integrator works (via deepcopy)."""
 
         system = self.forcefield1.createSystem(self.pdb1.topology)
         integrator = VerletIntegrator(2*femtosecond)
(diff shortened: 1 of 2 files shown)


jchodera:
Should be good to merge once appveyor completes.

peastman:
It looks like you've just removed the faster copy implementation?  Like I said in #1190, I'd like to see if I can find a way to make that work without losing the benefits.

jchodera:
#1055 essentially broke *all* of our code that depends on the [`alchemy`](http://github.com/choderalab/alchemy) module, including `yank` and a bunch of other projects.  We can't use or test anything that uses `openmm-dev` as a result.

I'd love to keep the benefits of a faster `deepcopy`, but can we merge this PR (which is the minimal reversion since it just reverts the `Force` part, but not the `System` part)---which also adds a helpful test!---ASAP and then test other fixes in a separate PR?  It's just copying-and-pasting like four lines into a new PR attempt, whereas everything is broken right now for us.

jchodera:
For comparison, it's much harder for me to monkey-patch `alchemy` to work around this bug, since I have to copy *every* force type via its API calls instead of just doing `copy.deepcopy`.  That's hundreds or thousands of lines of code.

wrappers/python/tests/TestPickle.py
```diff
@@ -39,16 +45,23 @@ def test_deepcopy(self):
...
-
         del context, integrator
 
+        # Check deep copy of each force.
+        forces = [ system.getForce(index) for index in range(system.getNumForces()) ]
+        for force in forces:
+            force_copy = copy.deepcopy(force)
+            # Check class name is same.
+            self.assertEqual(force.__class__.__name__, force_copy.__class__.__name__)
+            # TODO: Check to make sure all force data properly copied?
```
peastman: You can do that by serializing each of them and comparing the XML.  That would be a good test to do for the system, state, and integrator as well (above).

    self.assertEqual(XMLSerializer.serialize(force), XMLSerializer.serialize(force_copy))

peastman:
Fair enough.  In that case, though, we should revert the code for Integrator as well, since it has exactly the same problem.

wrappers/python/tests/TestPickle.py
jchodera: Great idea!

jchodera:
Will fix the integrator issue and improve the tests!

wrappers/python/tests/TestPickle.py
jchodera: Done!

peastman:
Looks good.  Once travis finishes, I think this is ready to merge?

jchodera:
Should be!

peastman:
Looks like the test is failing.

jchodera:
Huh!  We've come across an unexpected issue in which `HarmonicBondForce` is refusing to be pickled:
https://travis-ci.org/pandegroup/openmm/jobs/87535600#L4732-L4771
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <Swig Object of type 'OpenMM::HarmonicBondForce *' at 0x26489990>
proto = 0
    def _reduce_ex(self, proto):
        assert proto < 2
        for base in self.__class__.__mro__:
            if hasattr(base, '__flags__') and not base.__flags__ & _HEAPTYPE:
                break
        else:
            base = object # not really reachable
        if base is object:
            state = None
        else:
            if base is self.__class__:
>               raise TypeError, "can't pickle %s objects" % base.__name__
E               TypeError: can't pickle SwigPyObject objects
```

jchodera:
We had apparently forgotten to add `__getstate__` and `__setstate__` methods for `Force` objects.

jchodera:
OK, travis passes!

PR #1212, opened 2015-10-26
Enable testing of OpenCL platform on Travis-CI

rmcgibbo:
This patch creates a new build in the Travis-CI matrix that compiles and
runs the OpenCL tests using the CPU OpenCL device (AMD drivers). Because
not all of the tests can complete in the allotted 50 minutes, the ctest
invocation is changed to run them in a random order.

This also adds a new executable, TestOpenCLDeviceQuery, that prints
out information about all of the OpenCL devices on the system.

(fixed/rebased version of #1204)


swails:
Can it not run in the allotted time because the VM running the OpenCL tests have to be run on the legacy hardware until the fglrx drivers are whitelisted on the container architecture?

rmcgibbo:
I'm not sure there's a big speed difference between the two Travis architectures, once the VM/container is booted (the containers are definitely faster to boot). The test suite itself just takes longer than 50 minutes on the hardware.

peastman:
I wonder if a better solution would be to set OPENMM_BUILD_OPENCL_DOUBLE_PRECISION_TESTS to OFF?  That will take much less time, so it will complete and you don't need to tell it to allow failures.

rmcgibbo:
Okay, I'll try that. Although I did find a double-specific OpenCL failure in https://github.com/pandegroup/openmm/pull/1208#issuecomment-150995439

rmcgibbo:
And/or, I could turn off the reference/cpu tests in this branch of the build matrix -- i'm sure they're covered just fine in the other branches of the build matrix.

swails:
That last idea would get my vote.

swails:
Also--the container architecture cut the existing build/test times in half when it was implemented, so I suspect it would help here as well if it were possible to get the opencl stuff into the Docker instances.

rmcgibbo:
There's currently no cmake flag `OPENMM_BUILD_CPU_TESTS` or `OPENMM_BUILD_REFERENCE_TESTS` to disable building those tests -- can I add one?

rmcgibbo:
> Also--the container architecture cut the existing build/test times in half when it was implemented, so I suspect it would help here as well if it were possible to get the opencl stuff into the Docker instances.

I suspect based on https://github.com/boostorg/compute/pull/483 that it's not going to happen.

peastman:
Good idea.  You can turn off OPENMM_BUILD_CPU_LIB.  You can also use "ctest -R TestOpenCL.*" to only run the tests that start with "TestOpenCL".

rmcgibbo:
Oh, I should have thought of using `ctest -R`. Instead I just disabled all of these tests from being built, but that required some fiddling with the cmake scripts. I've been changing the travis script around to try to cut out as many superfluous things as possible for speed.

peastman:
The reason there's no option to disable the Reference and CPU tests is that, generally speaking, there's no reason to ever not build them.  If you're building the platform, you might as well build the tests as well.  The OpenCL and CUDA options were created for cases where you're on a machine that can't run those tests (e.g. no GPU), but you still want to build the platforms.

rmcgibbo:
Well, my purpose is to get the build as fast as possible.

rmcgibbo:
I could move those options to the advanced tab.

peastman:
> I could move those options to the advanced tab.

That sounds like a good compromise.  You can optimize this, but it doesn't overload the user with quite as many options that they really shouldn't mess with.

rmcgibbo:
Okay, I think this is slimmed down as much as possible, and gets about a 30-minute build/test for all of the single precision OpenCL platform.

@peastman, would you prefer it like this, or with the double and mixed precision tests mixed in and the random order / timeout?

peastman:
The important thing is to remove the `allow_failures`.  I assume the timeout is what made that necessary?  Realistically, we're not going to pull up the build log for every PR.  So if test failures don't get reported, we won't know they're happening.

swails:
What about adding another build to the build matrix and do single/mixed in one and double in the other to speed up the build a bit?

rmcgibbo:
I talked with peter, and I think we can go with the random order. In the latest build of this PR, the single precision tests take 12 minutes (+ ~6 minutes for everything else in the script, including rerunning the failing tests). Adding the double and mixed precision tests and setting a ~30 minute timeout should mean that each test will still get run pretty frequently, and we'll have better coverage.

rmcgibbo:
Okay, this is ready to merge, perhaps modulo fixing the OpenCL CPU tests failures which I think are genuine bugs in the OpenCL `CustomGBForce` and a potential bug, or maybe just too strict of a tolerance in `TestOpenCLLocalEnergyMinimizer`.

peastman:
I'll look into those failures tomorrow.  Interestingly, all tests passed this time.  I suppose that means it just didn't get to the ones that fail?

rmcgibbo:
I have a good minimal test case of the GB error now.

rmcgibbo:
here's the test case. https://gist.github.com/rmcgibbo/6f40a5f2979b5db5cd9b. I tracked it down a little to the `atomIndices` array getting corrupted around https://github.com/pandegroup/openmm/blob/master/platforms/opencl/src/kernels/customGBEnergyN2_cpu.cl#L346, but not the root cause

peastman:
I found the problem in CustomGBForce.  I can't reproduce the test failure for LocalEnergyMinimizer.  Perhaps I can look at it on your computer?

peastman:
The problems should be fixed by #1217.

peastman:
The fix is merged, and I've triggered travis to rerun.  Let's see what it does.

peastman:
Still some failures, which I'm looking into now.

rmcgibbo:
@peastman here's my best test case for the remaining cpu customgbforce error with cutoffperiodic. https://gist.github.com/anonymous/ea8327b01e617dc19ffb.

peastman:
Tests passed!  Which means I think we can merge this now.  Everyone ok with that?

swails:
Nice work @rmcgibbo and @peastman!

rmcgibbo:
+1

rmcgibbo:
oops.

peastman:
Very oops.

peastman:
You can usually undo deleting a branch.

rmcgibbo:
> You can usually undo deleting a branch.

Right. Okay.

Issue #1214, opened 2015-10-27
Weird Python issue with System.getDefaultPeriodicBoxVectors()


jchodera:
Here's a weird issue that just ate a couple hours of my life:
```python
from simtk import unit, openmm                                                                                                                                                                 
for cycle in range(100):                                                                                                                                                                       
    print('cycle %d' % cycle)                                                                                                                                                                  
    system = openmm.System()                                                                                                                                                                   
    box_vectors = system.getDefaultPeriodicBoxVectors()                                                                                                                                        
 ```
output:
```
Suspended
[LSKI1497:choderalab/openmmtools/openmmtools.jchodera] choderaj% python openmm-boxvectors-error.py
cycle 0
cycle 1
cycle 2
cycle 3
Traceback (most recent call last):
  File "openmm-boxvectors-error.py", line 6, in <module>
    system = openmm.System()
AttributeError: 'module' object has no attribute 'System'
Segmentation fault
```

**Comment by jchodera** [comment]
Commenting out the `box_vectors = system.getDefaultPeriodicBoxVectors()` call causes it to run through all cycles to completion without incident.

**Comment by swails** [comment]
I get the same thing running it in a Python interpreter, although the segfault waited until I shut down Python:

```
>>> from simtk import unit, openmm                                                                                                                                                                 
>>> for cycle in range(100):                                                                                                                                                                       
...     print('cycle %d' % cycle)                                                                                                                                                                  
...     system = openmm.System()                                                                                                                                                                   
...     box_vectors = system.getDefaultPeriodicBoxVectors()
... 
cycle 0
cycle 1
cycle 2
cycle 3
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
AttributeError: 'module' object has no attribute 'System'
>>> 
Segmentation fault
```

Is it just `getDefaultPeriodicBoxVectors()`?  Or have you not tried other System methods?

**Comment by jchodera** [comment]
I haven't tried anything else yet, but found this while trying to track
down a segfault that occurs when trying to iterate through
openmm.testsystems (which may be unrelated if that segfault only happens on
interpreter exit).

**Comment by rmcgibbo** [comment]
I can reproduce it too with HEAD, but not with 6.3.

```
(lldb) bt
* thread #1: tid = 0xe6db7a, 0x0000000100129dd0 libpython3.4m.dylib`collect + 352, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x18)
[...]

PR #1216, opened 2015-10-27
Fix segfault in swig wrappers

rmcgibbo:
The fix for #1214 is on line 472, because PyImport_AddModule gives a borrowed reference. The other decrefs are just fixing leaked references.

### DIFF
--- a/wrappers/python/src/swig_doxygen/swig_lib/python/typemaps.i
+++ b/wrappers/python/src/swig_doxygen/swig_lib/python/typemaps.i
@@ -1,10 +1,44 @@
+%fragment("Vec3_to_PyVec3", "header") {
+/**
+ * Convert an OpenMM::Vec3 into a Python simtk.openmm.Vec3 object
+ *
+ * Returns a new reference.
+ */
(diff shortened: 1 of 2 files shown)


rmcgibbo:
I suspect it would speed things up slightly to have these imports done only once, statically, as is done by the unit stripping code at the top of the file.

jchodera:
Would it be worthwhile to add a corresponding test?

jchodera:
Maybe at least this one?
```python
from simtk import unit, openmm                                                                                                                                                                 
for i in range(100):                                                                                                                                                                       
    system = openmm.System()                                                                                                                                                                   
    box_vectors = system.getDefaultPeriodicBoxVectors()     
```

peastman:
> I suspect it would speed things up slightly to have these imports done only once

Feel free to change that too if you want.

rmcgibbo:
> Feel free to change that too if you want.

Will do.

rmcgibbo:
Okay, I cleaned this up a little bit and factored some copy-paste code out into a function.

jchodera:
Hooray!  Thanks!

PR #1219, opened 2015-10-28
Reinstate fast copy methods

rmcgibbo:
Proper fix for #1190, https://github.com/pandegroup/openmm/issues/1190#issuecomment-148428653, partially reverts #1211.


jchodera:
Cool!

What about expanding the tests a bit to cover both `Integrator`s and `Force`s?

peastman:
Nice!

PR #1221, opened 2015-10-29
Check for index validity when setting interaction groups

### LINKED ISSUE #1220: Segfault with Custom*Force terms, even on Reference platform
This simple script
```python
from simtk import openmm                                                                                                                                                                                                                                                      
from simtk.openmm import XmlSerializer                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                              
system = XmlSerializer.deserialize(open('system.xml','r').read())                                                                                                                                                                                                             
integrator = XmlSerializer.deserialize(open('integrator.xml','r').read())                                                                                                                                                                                                     
state = XmlSerializer.deserialize(open('state.xml','r').read())                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                              
print "Creating Context..."                                                                                                                                                                                                                                                   
platform = openmm.Platform.getPlatformByName('Reference')                                                                                                                                                                                                                     
print "Setting positions..."                                                                                                                                                                                                                                                  
print "Taking a step..."                                                                                                                                                                                                                                                      
integrator.step(1)                                                                                                                                                                                                                                                            
print "Computing energies..."                                                                                                                                                                                                                                                 
state = context.getState(getEnergy=True)                                                                                                                                                                                                                                      
print state.getPotentialEnergy()                                                                                                                                                                                                                                              
del context, integrator      
```
yields a `segfault` for HEAD with `Reference` or `CPU` platforms:
```
[LSKI1497:~] choderaj% python openmm-segfault.py
Creating Context...
Setting positions...
Taking a step...
Segmentation fault
```
Corresponding XML files are [here](https://jenkins.choderalab.org/userContent/openmm-segfault-2015-10-28.tgz).

rmcgibbo:
Fixes #1220

### DIFF
--- a/openmmapi/src/CustomNonbondedForce.cpp
+++ b/openmmapi/src/CustomNonbondedForce.cpp
@@ -262,6 +262,10 @@ void CustomNonbondedForce::setFunctionParameters(int index, const std::string& n
 }
 
 int CustomNonbondedForce::addInteractionGroup(const std::set<int>& set1, const std::set<int>& set2) {
+    for (set<int>::iterator it = set1.begin(); it != set1.end(); ++it)
+        ASSERT_VALID_INDEX(*it, particles);
+    for (set<int>::iterator it = set2.begin(); it != set2.end(); ++it)
+        ASSERT_VALID_INDEX(*it, particles);
     interactionGroups.push_back(InteractionGroupInfo(set1, set2));
     return interactionGroups.size()-1;
 }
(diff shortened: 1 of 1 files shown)


jchodera:
I don't think this test should happen inside `CustomNonbondedForce.addInteractionGroup()`.  To my knowledge, none of the other `Force` objects validate their arguments at force population time---we are free to first create the `Force`, add it to `System`, set various properties, and then add the particle terms in any order we want. Usually, such `Exception`s are thrown when we create a `Context`---aren't they?

If we do keep this, we should add an explicit statement to the docstrings that note that a particular order is required in making calls to these functions.

(I note this because this change broke all our code---even those cases where we were correctly setting the interaction groups.)

jchodera:
What if we move this to [`CustomNonbondedForceImpl::initialize()`](https://github.com/pandegroup/openmm/blob/master/openmmapi/src/CustomNonbondedForceImpl.cpp#L57-L115) where all the other validity checks occur?

rmcgibbo:
Sure.

Issue #587, opened 2014-08-16
CMAKE_OSX_DEPLOYMENT_TARGET warnings for 10.9


jchodera:
I'm still getting this error on all of my OS X 10.9 machines:

```
(2.7)bash-3.2$ cmake .
CMake Warning at /opt/local/share/cmake-3.0/Modules/Platform/Darwin.cmake:143 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.7' but the matching SDK does not exist
  at:

   "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.7.sdk"

  Instead using SDK:

   "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.9.sdk"

  matching the host OS X version.
Call Stack (most recent call first):
  /opt/local/share/cmake-3.0/Modules/CMakeSystemSpecificInformation.cmake:36 (include)
  CMakeLists.txt:20 (PROJECT)


CMake Error at /opt/local/share/cmake-3.0/Modules/Platform/Darwin.cmake:213 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.7' but CMAKE_OSX_SYSROOT:

   ""

  is not set to a MacOSX SDK with a recognized version.  Either set
  CMAKE_OSX_SYSROOT to a valid SDK or set CMAKE_OSX_DEPLOYMENT_TARGET to
  empty.
Call Stack (most recent call first):
  /opt/local/share/cmake-3.0/Modules/CMakeSystemSpecificInformation.cmake:36 (include)
  CMakeLists.txt:20 (PROJECT)


-- Configuring incomplete, errors occurred!
See also "/Users/Shared/Jenkins/Home/workspace/openmm-conda-dev-mac/openmm/CMakeFiles/CMakeOutput.log".
```

Is anyone else getting this too?

jchodera:
Should we update the `CMAKE_OSX_DEPLOYMENT_TARGET` at some point?

peastman:
Good question.  I guess the arguments are 1) not many people are likely to still be using an OS that old, vs. 2) why break it if there's no reason to?

What would we gain by changing it?

rmcgibbo:
[...]

PR #1225, opened 2015-10-29
[CMake] set CMAKE_OSX_SYSROOT

rmcgibbo:
With OS X, and CMake >(?) 3.0, (I'm not sure the precise version of CMake. I'm using 3.2.3) I always get issues like this. Also #587
```
$ cmake ..
CMake Error at /Users/rmcgibbo/miniconda/envs/3.4.2/share/cmake-3.2/Modules/Platform/Darwin.cmake:76 (message):
  CMAKE_OSX_DEPLOYMENT_TARGET is '10.7' but CMAKE_OSX_SYSROOT:

   ""

  is not set to a MacOSX SDK with a recognized version.  Either set
  CMAKE_OSX_SYSROOT to a valid SDK or set CMAKE_OSX_DEPLOYMENT_TARGET to
  empty.
  /Users/rmcgibbo/miniconda/envs/3.4.2/share/cmake-3.2/Modules/CMakeSystemSpecificInformation.cmake:36 (include)
  CMakeLists.txt:21 (PROJECT)


-- Configuring incomplete, errors occurred!
See also "/Users/rmcgibbo/projects/openmm/build3/CMakeFiles/CMakeOutput.log".
```

This cleans that up by setting the `CMAKE_OSX_SYSROOT` variable from the output of `xcrun --show-sdk-path`


jchodera:
Thank your finally fixing this!

PR #1227, opened 2015-10-30
Add OS X and CUDA builds to Travis-CI config

rmcgibbo:
For the CUDA build, the CUDA platform is compiled, but the
tests are not run (no GPU). For OS X, the OpenCL platform
is also compiled, but not tested (we skip Apple CPU OpenCL).


peastman:
Nice!

PR #1230, opened 2015-10-31
CustomNonbodedForce interaction group validation now occurs during kernel initialization

jchodera:
Addresses modifications to #1221 discussed therein.

PR #1231, opened 2015-11-01
Fix travis build?

rmcgibbo:


rmcgibbo:
cc #1230

rmcgibbo:
It looks like #1225 killed the OS X build.

Issue #1229, opened 2015-10-31
`Simulation.minimizeEnergy` not working for me


swails:
This is a preliminary report, but I've been finding that applying the OpenMM minimizer to my system is actually making things *worse* than they were before.

Not only is it not resolving bad contacts, it is actually *creating* some where none were before.  Since it doesn't print a progress log, it's impossible for me to track what's happening at each iteration without hacking the LBFGS code to do so.

Here is my general workflow -- I create and solvate a system using tleap for use with the Amoeba force field (I have a fix in the works to let `Modeller` work with a `ForceField` that doesn't assign a `NonbondedForce`, but it doesn't currently work).  I then take this PDB file, generate a `System` from that, and then minimize it using the following script:

```Python
#!/usr/bin/env python
from __future__ import print_function, division

import parmed as pmd

import simtk.openmm as mm
import simtk.openmm.app as app

print('Loading the system and PDB file...')
system = pmd.load_file('system.xml')
parm = pmd.load_file('2igd_solvated.pdb')

print('Creating the Simulation...')
sim = app.Simulation(parm.topology, system, mm.VerletIntegrator(0.001),
                     platform=mm.Platform.getPlatformByName('CUDA'),
                     platformProperties=dict(CudaPrecision='mixed'))

print('Minimizing...')
sim.context.setPositions(parm.positions)
sim.minimizeEnergy(1.0)

# Now write a serialized state that has coordinates
print('Finished. Writing serialized XML restart file...')
with open('2igd_solvated.min.xml', 'w') as f:
    f.write(
            mm.XmlSerializer.serialize(
                sim.context.getState(getPositions=True, getVelocities=True,
                                     getForces=True, getEnergy=True,
                                     enforcePeriodicBox=True)
            )
    )

print('Done!')
```

I check my structure for bad contacts using the `checkoverlaps` command in cpptraj which scans through all interatomic distances and reports any that fall below 0.8 A.  If I use ParmEd to take the minimized state XML file and create a new PDB (since cpptraj can't read serialized states), I can feed it through and get the close contacts.  The list I get following minimization is *longer* than the list I get for the original PDB.

Sure enough, as soon as I try to run a simulation with this system, it runs off to NaNdyland and never comes back :(.

The simulation gets farther when I don't minimize at all (but this is rarely an option).  The next thing I tried was hijacking the `scipy` LBFGS optimizer and using that instead via this script:

```Python
#!/usr/bin/env python

import parmed as pmd
[...]

PR #1234, opened 2015-11-01
getState enforcePeriodicBox with triclinic unit cells

rmcgibbo:
cc #1229.


tests/TestEnforcePeriodicBox.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "openmm/NonbondedForce.h"
+#include "openmm/Platform.h"
+#include "openmm/VerletIntegrator.h"
+#include "sfmt/SFMT.h"
+#include <iostream>
+
+using namespace OpenMM;
+using namespace std;
+
+void testTrunactedOctahedron() {
```
swails: Truncated

swails:
This fixes the problem, btw.  Thanks!

rmcgibbo:
I think it would be still worthwhile to think about how to make the optimizer more robust, but I guess that requires a different test case.

swails:
Here's the error in one of the Travis builds: https://travis-ci.org/pandegroup/openmm/jobs/88685253#L1291-L1293

I'm pretty sure this is a race condition due to some dependencies that are not correctly set.  Here's what I've observed on my own machines:

- `make install` always works.
- `make -j# install` sometimes fails on one thread with an error like that one, and that failure brings down the whole install.  However, subsequent `make install` or `make -j# install` commands always complete just fine.

So we can do one of two things here:

1. Hunt down where the missing dependency is coming from and fix it
2. Change `make -j2 install` to `make -j2 install || make -j2 install`

The latter is easier and should be just as robust.

rmcgibbo:
No, the error is fixed by #1232.

swails:
I've been getting the strange parallel-make errors basically since I started compiling OpenMM.  Will #1232 fix those?

rmcgibbo:
That error applies regardless of things being in parallel. There's an extra newline in the `CMAKE_OSX_SYSROOT`, which just kills the whole construction of the input arguments to the compiler, I think. Because the variables are cached in CMake, you have to recreate the build directory to see it though.

rmcgibbo:
It seems like all of the RPMD tests are failing.

rmcgibbo:
perhaps https://github.com/pandegroup/openmm/blob/master/plugins/rpmd/openmmapi/src/RPMDIntegrator.cpp#L132-L145 also needs to be updated.

peastman:
Exactly what I was just about to post.

rmcgibbo:
Okay, let's see if that fixes it.

peastman:
Looks good!  The one problem was TestOpenCLRPMDDouble, which timed out.  Possibly that test takes too long to run on travis?

rmcgibbo:
Yeah, it timed out the first time it was run, which is expected since we need to enforce the time limit on ctest, but then it was rerun and timed out again with `No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.` It makes sense that it would be one of the slower tests.

PR #1236, opened 2015-11-03
Added Context::getParameters()

### LINKED ISSUE #1206: Add methods for discovering parameters available in Context?
While `Context` has `getParameter(name)` and `setParameter(name, value)`, there are no other functions for inquiring or discovering which context parameters are available.

What do you think about extending the API to include something like this:
```python
nparameters = context.getNumParameters()
name = context.getParameterName(index)
```
so that we can autodiscover parameters like
```python
context_parameters = [ context.getParameterName(index) for index in range(context.getNumParameters()) ]
```
?

peastman:
Fixes #1206.


jchodera:
Thanks!

PR #1228, opened 2015-10-30
Better docstrings

rmcgibbo:
Example:
```
In [4]: mm.HarmonicBondForce.getBondParameters?
Signature: mm.HarmonicBondForce.getBondParameters(self, *args)
Docstring:
getBondParameters(self, index)

Get the force field parameters for a bond term.

Parameters
----------
index : int
    the index of the bond for which to get parameters

Returns
-------
particle1 : int
    the index of the first particle connected by the bond
particle2 : int
    the index of the second particle connected by the bond
length : double
    the equilibrium length of the bond, measured in nm
k : double
    the harmonic force constant for the bond, measured in kJ/mol/nm^2
```

~~I also found https://github.com/m7thon/doxy2swig that does a pretty good job, so I might fork that and improve it a little, and then we could just delete all of this code.~~

PR #1238, opened 2015-11-03
New Sphinx API Docs

rmcgibbo:
What do you all think of ~~http://stanford.edu/~rmcgibbo/python-api/~~ http://openmm-python-api.s3-website-us-east-1.amazonaws.com/_build/index.html? (depends on #1228)


kyleabeauchamp:
+1, I find it more usable than the doxygen version.

swails:
A couple observations:

- It looks like compactdefaultargs was on... at least for a lot of the classes (I didn't check carefully enough to see if the arg list was *always* `*args` for the SWIG-generated interface or not).  Is there any way to get named arguments for the wrapper classes?
- Can you do the same thing for the C++ API docs?

I like these docs more than what Doxygen generates for sure, but I'd personally rather have consistency between the two sets of docs than one pretty and one not, I think.  I consult both docs, but Python more than C++.  If every time I visited the C++ docs it not only had a different theme but also a different layout from what I was used to, it would make using the C++ docs far more irritating.

dr-nate:
Wow. That is much better, IMO.

jchodera:
Love it!

mpharrigan:
I disagree with @swails a little. IMO c++ developers will be comfortable with the ugly but usable doxygen docs and python people / end users will find the nicely themed version more pleasing

peastman:
To be fair, Doxygen is capable of doing pretty much all the same things you've done here.  It has lots of features for categorizing, customizing, etc.  Just because we've never used them, that isn't Doxygen's fault!

Also, a large part of the reason these docs are cleaner is that you've replaced automatically generated indices with handwritten indices.  But that's something we should be very cautious about doing, since it creates all sorts of new opportunities for the docs to get out of sync with the code.  Perhaps there's some way the indices could be automatically generated while still being better organized?

rmcgibbo:
Which indices? Most of it is generated with the templates. We can
introspect to find all of the classes.

IMO, the biggest difference is the CSS, not the organization. And that's
the part I don't know how to do with doxygen.

peastman:
app.rst.jinja2 and library.rst.jinja2.

For a description of options to customize the appearance, see http://www.stack.nl/~dimitri/doxygen/manual/customize.html.  This page has links to various themes other people have created: https://stackoverflow.com/questions/165451/suggest-a-good-doxygen-stylesheet.

rmcgibbo:
If you wanted all the classes just listed alphabetically, that could be
done. I thought it was more useful to separate out a small number of key
classes, but autogenerate the rest with the template engine where the
inheritance tree was useful.

It's not a big deal either way.

peastman:
I wonder if we could autogenerate the list but still break them into categories the way you did.  For example, all subclasses of Force go in one category, all subclasses of Integrator go in another category, and everything else goes in a third category.

Doxygen also has a feature for manually specifying categories like this: http://www.stack.nl/~dimitri/doxygen/manual/grouping.html

rmcgibbo:
> For example, all subclasses of Force go in one category, all subclasses of Integrator go in another category, and everything else goes in a third category.

I think that's already how it works, modulo separating out `Context`, `System` and `Platform` as "core objects".

peastman:
No, there isn't currently any kind of clean "front page" with a nicely categorized list of classes, like the one you created.  But there could be.

rmcgibbo:
Oh, you mean with doxygen?

peastman:
Right.

peastman:
Or even with Sphinx.  My point is that you created handwritten front pages to serve as an index.  And they'll get out of date the moment someone adds a new class and doesn't remember (or know) they need to be added there.

rmcgibbo:
I'm confused with exactly what you're saying. Most of it is based on the inheritance tree. Especially the forces and integrators. e.g. https://github.com/pandegroup/openmm/pull/1238/files#diff-eac865fb938da96253704b0e12bb1594R22

mpharrigan:
> getStepSize((self) -> double)

Is this really how we want to present the method signatures?

mpharrigan:
And I agree the `*args` on each class is unhelpful

rmcgibbo:
The `*args` is hard to avoid, because it comes from swig dealing with the fact that method overloading in C++ works differently than in Python. It only happens for classes with multiple constructors.

mpharrigan:
The methods summary table "squishes" the right column due to the large number of arguments. Can we remove arguments from those, and have people click through to get the full signature. 

This will also fix the strageness of the methods that specify a return type. They're rendered nicely in the full docs, but badly in the summary table

rmcgibbo:
How about this? http://openmm-python-api.s3-website-us-east-1.amazonaws.com/_build/index.html

rmcgibbo:
With 4d55c9e, _every_ class defined in simtk.openmm.openmm and simtk.openmm.app should be listed. Some of them are explicitly categorized, either by name, or who they inherit from, and the others are just put into "extras". So even if nobody changes the file again after adding a new class, things shouldn't get completely dropped.

mpharrigan:
The logo doesn't scale in the sphinx_rtd_theme version

rmcgibbo:
Everyone's a critic :)

rmcgibbo:
How should the logo look?

peastman:
Well, not the way it currently does: taking up most of the height of the window, but only seeing a sliver of its left edge.

rmcgibbo:
? It looks fine on my machine.

peastman:
On my computer it's messed up in Firefox, but looks OK in Safari.

rmcgibbo:
Any better now?

peastman:
Yes, it's back to the proper size now!

peastman:
There seem to be problems in how the text gets processed.  Take a look at http://openmm-python-api.s3-website-us-east-1.amazonaws.com/_build/generated/simtk.openmm.app.modeller.Modeller.html#simtk.openmm.app.modeller.Modeller.addSolvent/  The numbered lists are messed up.

rmcgibbo:
Yeah, definitely not perfect. The docstring markup was probably correct for doxygen but not sphinx/rst.

rmcgibbo:
I'm not sure if I got all of the CMake stuff right, but I think it looks pretty reasonable. It does depend on #1228, and probably requires some cosmetic improvements to some of the docstrings, especially those that use lists or html.

jchodera:
This is looking great!

mpharrigan:
For improvement:
- The blue background in the logo area. We changed this for mdtraj/msmbuilder
- summary tables now have a horizontal scrollbar. Is this due to the change of napoleon? I tried using napoleon for msmbuilder docs and it had a hard time dealing with a lot of the more advanced markup we use there, like unordered lists in parameter descriptions. I also recall it adding horizonal scrollbars to the summary tables

rmcgibbo:
The horizontal scrollbar come from sphinx_rtd_theme, I think, not napoleon. It keeps it from being too squished on the right.

mpharrigan:
c.f. http://msmbuilder.org/3.3.0/featurization.html

rmcgibbo:
Do you want me to try using the latest msmb_theme?

rmcgibbo:
(Also, @mpharrigan do you we want to update msmb_theme to be based off the latest version of sphinx_rtd_theme?)

mpharrigan:
The latest msmb_theme (the one i've been messing with) inherits from whatever version you've got. I've been testing it with the latest (0.1.9)

rmcgibbo:
Right, okay.

mpharrigan:
Oh, I thought I hadn't messed with the table linebreaking behavior, but now that I look, it *is* msmb_theme that's doing it. Sorry

rmcgibbo:
Here are a couple different themes for the layout and CSS:

- [alabaster](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/alabaster/index.html)
- [sphinx_rtd_theme](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/_build/index.html)
- [msmb_theme](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/msmb_theme/index.html)
- [nature](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/nature/index.html)

peastman:
I really like alabaster.  Nature is ok, but not as nice.  sphinx_rtd_theme is kind of ugly, and msmb_theme is really ugly (sorry!).

jchodera:
I don't like the fact that Alabaster serif and sans serif fonts (yuck!).  Is there a way to fix that?

peastman:
> I don't like the fact that Alabaster serif and sans serif fonts (yuck!)

It doesn't bother me that much, since they're being used for different purposes.  But I assume it would be easy to modify the style sheet to use a fixed width serif font (e.g. courier) for class names?

rmcgibbo:
From the other thread

> Could you be more specific? For example,

> > Simplicity, navigability.

> In what ways do you want it to be simpler and more navigable?

> > Lack of extraneous or irrelevant information.

> What is the specific information you think should be removed?

> And so on. I'm trying to get some specific, objective goals and criteria. I could say, "I find the Doxygen docs more attractive and professional looking than the Sphinx ones," (which is true), but unless I give specifics, that doesn't give any guidance on how to make them better.

It's hard to articulate precisely since I'm not a web designer, but I'll try a couple things that I think go to simplicity and navigability.

 - If you misclick on the doxygen pages, there are tons of pages of really odd, confusing, and extraneous information
     - http://docs.openmm.org/6.3.0/api-python/functions.html.
     - http://docs.openmm.org/6.3.0/api-python/namespacemembers.html

- When you go to the main page, literally nothing shows up: http://docs.openmm.org/6.3.0/api-python/index.html.
- The page is not mobile responsive.
- The expandable tabs on the left side are extremely deep, and most of the levels of hierarchy have no relevant semantic meaning, and it's really difficult to navigate to the page you want.
- There are big sections of many of the pages that are totally meaningless or provide no information
    * http://docs.openmm.org/6.3.0/api-python/classsimtk_1_1openmm_1_1openmm_1_1GBVIForce.html#a9a4139e80f5ae78c0b0c7a093bf5b9da

- And complete pages that have no information
   * http://docs.openmm.org/6.3.0/api-python/namespacesimtk_1_1openmm_1_1app_1_1modeller.html#details 

The summary is that I think I'm on solid ground when I say the doxygen docs are "hard to use".

peastman:
For the manuals, we currently use the agogo theme.  It might be worth trying that.

rmcgibbo:
- [agogo](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/agogo/index.html)

peastman:
Thanks!  That looks OK, but I like alabaster better.  I should try that for the manuals.

jhprinz:
First. I would also go with alabaster.

Another option maybe for the future: What about bootstrap compatible themes like this: [sphinx-bootstrap-theme](https://ryan-roemer.github.io/sphinx-bootstrap-theme/)

The rest coming is related but more for the future...

I would like to have the docs website being [bootstrap](http://getbootstrap.com) compatible which would allow to use lots of (free) themes, being responsive and other nice features like 
- menubars, 
- flexible navigation
- sidebar that moves with scrolling of the docs
- sticky navbar or footer
with minimal effort. Ipython/Jupyter notebooks use bootstrap.

Would be nice to have a bundle of sphinx tools that solve this problem (maybe for all omnia projects alike). This would require/contain robust solutions for
- a tweak of napolean/numpydoc to produce bootstrap compatible HTML -> easier styling using CSS. I especially dislike the way the descriptions of parameters/attributes are displayed. Although the changes you made in the showed docs are much better than what I am used to.

- a bootstrap compatible sphinx theme

- a better ipynb style that is compatible with bootstrap

    I made some tweaks that make it more compatible with `rtd` but that can definitely be improved

- someone who makes/tweaks a nice bootstrap theme like alabaster, etc...

- a sphinx directives to include ipython notebooks (including reruns if desired)

    ```
    .. notebook:: examples/ipython/alanine.ipynb
    ```
- a sphinx directive to include markdown (like include the README.md)

    ```
    .. markdown:: docs/README.md
    ```

At the end changing the style of your docs would boil down to only change a `.css` file. Most of this stuff exists and _just_ needs to be put together. I can offer to put some time into a packge like `omnia-sphinx-tools` provided there is some interest.

jchodera:
Wow, I had no idea you could have sphinx generate bootstrap-compatible pages.  That definitely looks very nice and modern.

mpharrigan:
The alabaster theme is responsive

I don't see how using bootstrap makes the css easier to change. You can change the css for alabaster just as easily

peastman:
@jchodera you aren't allowed to like the bootstrap theme.  It uses a serif font for class and method names, but a sans-serif font for everything else.  You find that yucky. :)

Bootstrap is OK, but I like alabaster better.  The bootstrap theme makes it look like twitter.  Alabaster makes it look like a great work of literature.

jhprinz:
Maybe I should add that I the bootstrap theme is okay, but not super. The main point about using bootstrap is that it is an agreement on additional HTML tags and attributes while providing an easy way to keep the website responsive like "this block should appear on desktop size windows, but not on mobile sized ones" or how to size blocks while staying responsive.

You can of course do all of that with other themes, but bootstrap is an intermediate layer that is widely accepted and used by web designers. I assume because it simplifies things.

The default theme of bootstrap is okay but not great, but that is just the default setting. While there are tons of free templates you can also buy professional ones for a few bucks. That look very different.

The sphinx-bootstrap-theme is one approach to use the new html tags and navbars to enhance the sphinx documentation, but that is just one approach and it ships with a set of 20 templates that change
the look to some extend. These themes can be looked at under [bootswatch](http://bootswatch.com). Do support this it specific to this theme.

Another example is `builder.openmm.org`. The tabbing, the navbar on the top, all these things are components that are parts of bootstrap. How they look like is part of the concrete css.

Once you agree on certain naming of classes and objects it is easier to change the css. Hope this clarifies the idea. It is more a longterm goal...

jchodera:
> @jchodera you aren't allowed to like the bootstrap theme. It uses a serif font for class and method names, but a sans-serif font for everything else. You find that yucky. :)

Are we looking at the same thing?  I was looking at these linked examples:
* http://ryan-roemer.github.io/sphinx-bootstrap-theme/
* http://ryan-roemer.github.io/django-cloud-browser/

> The bootstrap theme makes it look like twitter.

Which is, I hate to say it, what "modern" web things look like.

peastman:
> Are we looking at the same thing? I was looking at these linked examples:

Right, the first one.  Click on the "Code Documentation" link.

peastman:
Does anyone else have opinions on specific ways you would like the documentation to be improved?

jchodera:
> Right, the first one. Click on the "Code Documentation" link.

None of my browsers show any serif fonts there---it's all sans serif.

I specifically like the following properties of some of the new proposals:
* Clean, uncluttered look with sans serif fonts (no mixing of serif and sans serif!)
* "Modern" web UI look, specifically avoiding the look of tools that have been around for a long time and subconsciously cue the user that the code might be a "dusty deck"
* The separation into "Library Layer" and "Application Layer" (probably in that order), rather than [dumping everything into one place](http://docs.openmm.org/6.3.0/api-python/annotated.html) which just causes confusion
* Clear indication when something is linked (underlining or boldface with hover-above underlining)
* No [gradient shading](http://docs.openmm.org/6.3.0/api-python/classsimtk_1_1openmm_1_1openmm_1_1GBVIForce.html#a9a4139e80f5ae78c0b0c7a093bf5b9da)!
* Removing the `self` argument from the call signature might be good
* Fixed-width fonts for function names
* Documentation of individual parameters, as [in here](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/nature/generated/simtk.openmm.app.amberinpcrdfile.AmberInpcrdFile.html#simtk.openmm.app.amberinpcrdfile.AmberInpcrdFile)
* A search box!

A lot of the new styles satisfy all these criteria, and all look great and are very functional to me.

peastman:
Here's how it looks on my screen.  Note the serif font for the class name.

untitled

> Clean, uncluttered look with sans serif fonts...
> "Modern" web UI look...
> No gradient shading!

All of those are personal aesthetic preferences.  What I conclude from them is that your personal preferences are very different from mine. :)  But let's try to stick to objective things.

> Clear indication when something is linked (underlining or boldface with hover-above underlining)

It's already extremely clear in the existing docs: boldface and in a different color.

> Documentation of individual parameters, as in here 

The existing docs already have documentation of individual parameters.

> A search box!

The existing docs have a search box:

untitled2

rmcgibbo:
> All of those are personal aesthetic preferences. What I conclude from them is that your personal preferences are very different from mine. :) But let's try to stick to objective things.

-1

Personal aesthetic preferences are important.

rmcgibbo:
Discounting what you might pejoratively call "aesthetic preference" is essentially equivalent to discounting design. And design is really closely connected with UX and ease of use, especially for new users.

jchodera:
> Here's how it looks on my screen. Note the serif font for the class name.

I am fine with mixing fixed-width (terminal) fonts with sans-serif fonts, as I noted above.

> And design is really closely connected with UX and ease of use, especially for new users.

Agreed.  I think @peastman complained earlier "The bootstrap theme makes it look like twitter", but twitter is very usable, and people are very used to using twitter and the umpteen-other bootstrap-powered responsive websites now, so that is a very "usable" interface.

peastman:
No, ease of use is objective.  There's a whole field of study around that.  Aesthetics is subjective.  It's important to distinguish between the two.

For example, I object to the lime green text used for links in the bootstrap theme because it's hard to read.  I object even more to the medium gray text on slightly lighter gray background that it uses for some of its code samples, because those are *really* hard to read.  Those are objective things.  I also have some personal aesthetic preferences: white backgrounds, prefer gradients over large blocks of solid color, serif fonts.  But those are purely subjective.

> I am fine with mixing fixed-width (terminal) fonts with sans-serif fonts, as I noted above.

I'm not talking about it being fixed width.  I'm talking about it being a serif font.  See the little bits sticking out from the tops and bottoms of all the letters in the class name "MyAnimal"?  Those are serifs!  Those two questions are independent.  For example, compare Inconsolata (https://web.archive.org/web/20150612082006/http://a.pomf.se/a80ds.png), which is a fixed width sans-serif font, and UM Typewriter (https://web.archive.org/web/20150610010407/http://a.pomf.se/j36dx.png), which is a fixed width serif font.

rmcgibbo:
I guess my point is that there seems to be wide agreement from everyone involved in this conversation except for you that the current docs are not easy to use, and that the sphinx docs improve on the ease of use. We've now turned to bike-shedding about the particular theme and the font, but don't miss the forest for the trees.

swails:
> The separation into "Library Layer" and "Application Layer" (probably in that order), rather than dumping everything into one place which just causes confusion

This is being unfair.  The organization is undoubtedly better, but to say that the current situation *only* causes confusion is easily disproven -- I'll simply cite myself as a counterexample.

The organization (app layer vs. Forces vs. Integrators vs. other) on its own is well worth the changes being discussed here, but I still maintain a desire to see the C++ and Python docs have at least the same layout or organization even if the themes/styling are different.  I'm really not liking the attitude of "who gives a crap about the C++ docs, let's focus everything on Python" and the vibe of "a choice between the Sphinx docs that are awesome and the Doxygen docs that sucksobad".  I *like* the doxygen docs (but sorry, @peastman, I *do* like the Sphinx docs better, too).

IMO this is a decision between good and better -- I've found the docs to be remarkably accurate and helpful -- moreso than many other packages.

peastman:
I think you misunderstand me.  I absolutely agree that the current docs could be improved in a lot of ways.  The question now is figure out how to improve them.  That's why I asked people to describe specific, objective ways they would like them to be improved.  Once we have that, we can move on to the next question, which is to evaluate possible tools and technologies to use.

For example, most of the items on your list are related to the organization rather than the appearance: provide a meaningful front page, replace the expandable tree with a more useful index, eliminate the pages that don't contain useful content.  Each of those can be done with Doxygen.  That isn't to say we have to stick with Doxygen: we absolutely can and should consider other tools as well.  But I don't feel like we've really done that analysis.  It feels more like you created a prototype based on Sphinx, people agreed that it looks nice, and so instantly we're talking about switching without considering whether switching to Sphinx is actually an essential part of the changes, and whether it might also have negative consequences.

rmcgibbo:
IMO, I created not only a prototype, but a full solution that improves the status quo. If someone want's to do the leg work to come up with _another_ solution that improves the status quo, that would be great too. I don't know what more we can do to "analyze" the situation. Punting this off into endless debate over fonts seems like a waste of time to me.

swails:
Here are some comments I have.

- The upper-left section of the doxygen docs feels like a cluster to me.  So many buttons, some contain useless information, at least half lead to blank or useless pages (main page, related pages, classes->class members).  I basically ignore its existence because it's easier for me (and the inheritance diagram is easy enough to remember).  The Sphinx docs clean that up significantly and present a view in which a much higher percentage of what you see is important.  But I can easily filter out the upper-left section and make efficient use of what we have, so this is minor.

screen shot 2015-11-09 at 5 54 00 pm

- The table width in Sphinx is capped so that lines can't run out to the edge of the browser.  This makes it easier to read long lines (for me and I believe studies have been done that show this is generally true).  Doxygen takes up all available space and so on a widescreen monitor can run out to *incredibly* long line lengths.  You can shrink the browser window for doxygen, but I use tabs so this can get annoying for other websites.

- The Search bar in Doxygen is much better.  As you type, you get real-time suggestions.  In the Sphinx themes I didn't see that -- you need to type, press enter, and look.  Doxygen's search engine is much more efficient on a standard computer (i.e., not mobile).  This is important IMO.

- The organization in the Sphinx docs is much better.

- I want the C++ docs to get *some* of this love, too.

cing:
I'm just a wandering OpenMM user and I've been lurking this discussion a bit. Lots of positive changes in my opinion. Any reason why the Doxygen or Sphinx docs don't include links to source code? MDAnalysis docs (Sphinx) include them and I find them pretty useful ([example](http://pythonhosted.org/MDAnalysis/documentation_pages/analysis/psa.html)). Sync issues don't seem like a major problem to me, even when linking to the master branch of OpenMM on Github (which could be slightly newer). Might be alone on this one though...

rmcgibbo:
For the App layer, we could do links to the source code with the built in sphinx support. But for the library, it's a little harder since the python code is just a wrapper.

peastman:
The C++ docs do include links to the source files, but only the headers.  Should the API docs contain a complete extra copy for the full source code?

peastman:
Let me offer a list of specific objectives for the docs.  This is my attempt to combine everyone's comments while avoiding subjective things we aren't likely to get agreement one.

1. The docs should have a meaningful "front page" that organizes everything.  Right now, the front page is just blank.

2. There currently is a lot of extraneous clutter: buttons that aren't useful, pages that don't contain useful information, etc.  We should eliminate that.

3. The tree on the left doesn't organize the classes in a very useful way.

4. All of the above apply to both the C++ and Python docs.  They both need improvement.

I think everyone agrees with all of the above?  In addition, there are several things that have been mentioned but that we don't have agreement on:

1. Should the Python docs omit the "self" argument from methods?  This was mentioned once, but I don't think anyone else commented one way or the other.

2. A couple of people mentioned making the page responsive for mobile.  Do you really read API docs on your phone?  I don't recall ever doing that myself.  Is this an actual issue for you, or just more of an "in an ideal world" thing?

3. Different opinions have been expressed about how similar the C++ and Python docs should be to each other.  I think most people agree they should at least have similar organization.  Whether it's important for them to look the same seems to be an open question.

4. Everyone wants the docs to "look nice", but there's not a clear agreement on what that means, or even on whether the current docs "look nice".

Finally, I'll offer one point of my own that hasn't been discussed much but I think is important: the whole system needs to be as easily maintainable as possible.  A system that is simpler and takes less work to maintain is much preferred over one that's more complicated and takes more work to maintain.

jchodera:
> Do you really read API docs on your phone?

I've done this multiple times when helping students.

rmcgibbo:
Yes, I agree with all of those things. I would add that the overall focus should be on **ease-of-use**. Most of the specific objectives mentioned above in the first list are specific elements that advance goal.

peastman:
If we're mostly agreed on those points, then let's start talking about tools.  I think there are three main options:

1. Use Doxygen for both C++ and Python.
2. Use Doxygen for C++ and Sphinx for Python.
3. Use Sphinx for both C++ and Python.

Unless, of course, someone has yet another tool we should consider while we're at it?

Does anyone know how practical option 3 is?  From what I've read, Sphinx can't parse C++ files, but you can use Doxygen to parse the code and then Breathe to convert it to a format Sphinx can read.  Any idea how well that works?

peastman:
Any comments on the three options listed above?  I was expecting everyone would have lots of say about them!

peastman:
As much as I like listening to the sound of crickets, I'm starting to feel a little lonely here!  Has everyone disappeared?  Or do you just have no opinion at all on what tools we should use?

I want to try to keep this moving forward.  I think at this point we need to do some research.  If we go with either option 1 or 2, that means trying to make some of these improvements in Doxygen.  So we need to figure out what's possible and how to do it.  On the other hand, if we want to go with option 3, we need to determine whether Breathe+Sphinx really works well enough to make that practical.  That's two different questions that need research.

Does anyone want to take on either of those?  I can do them if no one else does, but it might be a while before I'm able to get to them.

swails:
The most important thing IMO is the improved organization and added simplicity.  That said, I do like the Sphinx-generated docs better minus the search bar, and more of us know how to use it than Doxygen (I think -- please correct me if I'm wrong).  So my top choice would be 3, assuming that Breathe (or similar) works well.  My next would probably be both doxygen if the key improvements could be made (organization and simplification).  Last choice would be sphinx-Python/doxygen-C++.

I *think*.

mpharrigan:
I think @rmcgibbo said it eloquently when he pointed out that he already has made sphinx work in a way that all of us agree is an improvement

swails:
> I think @rmcgibbo said it eloquently when he pointed out that he already has made sphinx work in a way that all of us agree is an improvement

You dodged the question ;).  That said, I don't know that the agreement is universal.  For instance, the search bar in Doxygen is *way* better when on a keyboard IMO.

The organization and simplicity is a big improvement, but that can also be done with Doxygen.

peastman:
@mpharrigan since you like the Sphinx docs much better, I take it you're volunteering to try to get that working for C++? :)

rmcgibbo:
I think we should work incrementally. There's no need to couple every change to both docs systems into 1 changeset.

My vote is for option 2. We can then consider option 3 asynchronously.

rmcgibbo:
Also, obviously this is speculative, but it might be pretty doable to plug in a better client-side live search functionality into sphinx, eg. with lunr.js.

swails:
> Also, obviously this is speculative, but it might be pretty doable to plug in a better client-side live search functionality into sphinx, eg. with lunr.js.

:+1: for that

jchodera:
> My vote is for option 2. We can then consider option 3 asynchronously.

+1

peastman:
We can implement it asynchronously.  But we need to at least be confident about which of the three options we're going with, and that that option will be technically practical.

rmcgibbo:
> But we need to at least be confident about which of the three options we're going with, and that that option will be technically practical.

I'm not sure why this need be the case, if the plan is (say option 4) is "change python API docs to Sphinx, and continue to brainstorm about C++ docs"

peastman:
If we end up sticking with Doxygen for C++, and if it turns out we can make all the same improvements with Doxygen that you've made with Sphinx, then option 1 would be much better than option 2.  It would have all the same benefits, while having much better consistency between C++ and Python and being easier to maintain.

Your option 4 is basically, "Let's rule out option 1, and decide later between 2 and 3."  I don't think we have enough information yet to do that.

rmcgibbo:
@swails: I overhauled the search. Try it out [here](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/_build/index.html). The code is https://github.com/rmcgibbo/sphinxcontrib-lunrsearch

swails:
Search bar is awesome. Works really well on mobile, too. Very responsive and as-you-type results. 

Nice work!

swails:
Also thanks for the source. I'll probably use it for the ParmEd docs (it would probably be good for mdtraj and other omnia projects too!)

jchodera:
New search bar is awesome.

I still don't like the serif text font---do other people feel the same way, or is it just me?

I think changing the font just requires experimenting with [these lines in `theme.conf`](https://github.com/bitprophet/alabaster/blob/master/alabaster/theme.conf#L63-L64).

swails:
I don't mind the serif fonts.

rmcgibbo:
I'm sure there are still quite a few bugs, so feel free to file issues on the repo. I noticed, for example, that it doesn't behave too well with long class names that go over the width, and stuff like that.

peastman:
I like the serif font.  I think it's really just a matter of personal preference.  That's why there are so many fonts in the world of both kinds.

If no one else is planning to, I'll try to find time this week to investigate using Sphinx for C++.  No promises though, since I also have some other things I need to do.

peastman:
Here's a first attempt at using Breathe to produce some C++ documentation:

https://web.stanford.edu/~peastman/breathe

Some problems:

1. It doesn't include any kind of table indexing the class members.  It just dumps the full documentation for each member, one after another, which makes it really hard to use.  I haven't been able to find any way of fixing this.

2. All the code samples are missing.  I haven't investigated this yet.

3. It also doesn't generate any sort of an index to the classes.  But since Robert's Sphinx documentation generates its own index, we can probably do something similar.

rmcgibbo:
That looks pretty good, actually. The remaining stuff is mostly template / theming, I assume. What is the source code / config file required to generate this?

rmcgibbo:
Neat!

dr-nate:
Looks good. I'm already mainly using the new sphinx docs and they've been pretty handy.

rmcgibbo:
> Looks good. I'm already mainly using the new sphinx docs and they've been pretty handy.

I'm curious: which ones? There are like 8 different copies with different themes and such.

dr-nate:
The sphinx/alabaster one with the fancy search. https://github.com/pandegroup/openmm/pull/1238#issuecomment-156788981 Suuper helpful.

rmcgibbo:
Cool. Let me know if you find any problems, esp. with the search.

rmcgibbo:
I tried to read some of the breathe source to figure out how to change the templating. Bad decision. Too many `DoxygenToRstRendererFactoryCreatorConstructor`s. It's like a java nightmare.

peastman:
I take it they're into Using Design Patterns (TM)? :)

I'm trying to debug the missing code samples.  Any idea how I can inspect the rst output that Breathe is generating?  As far as I can tell it never writes it to disk, just creates it in memory.

peastman:
It looks like the easiest solution may be to just reformat the examples in the source code.  Currently they're wrapped in html `<tt><pre>` tags.  But Doxygen also supports markdown syntax, so if we just remove those tags and instead indent them four spaces, it works correctly.

peastman:
Here's a progress update on this.  @rmcgibbo was sufficiently unimpressed with Breathe that he decided to write his own version from scratch.  See https://github.com/rmcgibbo/sphinxcontrib-autodoc_doxygen.  We've both been testing it, and it's now producing pretty nice results.  See https://rawgit.com/rmcgibbo/sphinxcontrib-autodoc_doxygen/gh-pages/generated/OpenMM::NonbondedForce.html for an example.  So I think we can move forward on using Sphinx for both C++ and Python.

Here are the next steps:

1. Implement the C++ API docs.
2. Do a thorough review of both sets of docs, to make sure no errors have been introduced.

mpharrigan:
Cool :sunglasses:

jchodera:
This is really nice, and even works decently well on mobile (though some table widths could be optimized for that).

swails:
@rmcgibbo -- this is excellent work.  Everything I liked better about doxygen (Py/C++ consistency and search bar) you have incorporated into an aesthetically pleasing set of docs, and everything I disliked about the doxygen docs you got rid of.

I actually thought the previous docs walked the border between fine and good.  But these docs are great.  Thanks for all the hard work.

rmcgibbo:
Okay, new C++ docs using sphinxcontrib-autodoc_doxygen are here: http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/. Still rough around the edges.

rmcgibbo:
Obvious errors:
 - Some links between objects documented on different pages to not work properly e.g. [here](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/generated/OpenMM::System.html#_CPPv2NK6OpenMM6System14getVirtualSiteEi)
 - The static methods in [XmlSerializer](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/generated/OpenMM::XmlSerializer.html#_CPPv2N6OpenMM13XmlSerializerE) do not render.

swails:
State appears twice here http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/core.html

peastman:
That's looking really nice!  Here are a few errors I noticed:

Look at the fourth paragraph of http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/generated/OpenMM.CustomNonbondedForce.html.  The code sample there is messed up.

Deprecation warnings aren't being shown correctly.  For example, see http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/generated/OpenMM.CustomNonbondedForce.html#_CPPv2N6OpenMM20CustomNonbondedForce11addFunctionERKNSt6stringERKNSt6vectorIdEEdd.  It does show the text from the warning, but without setting it off in any way or showing the word "deprecated".

Types of function arguments that are templates are shown incorrectly.  For example, see the `values` argument of http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/generated/OpenMM.CustomIntegrator.html#_CPPv2N6OpenMM16CustomIntegrator17setPerDofVariableEiRKNSt6vectorI4Vec3EE.  It's written as "(const std::vector\< )".  I don't think it's actually necessary to list types after the argument names.  That information is already given right above in the method signature.

rmcgibbo:
Okay, those are all fixed now.

peastman:
The deprecated methods are being marked as "Note" rather than "Deprecated".  Are you using the `deprecated` directive for them?

http://sphinx-doc.org/markup/para.html#directive-deprecated

rmcgibbo:
I tried that, but the problem is that the sphinx deprecative directive has one _required_ argument, which is the version number when the deprecation went into effect, and gets rendered as "Deprecated since version <version>", and the doxygen source doesn't contain that info.

peastman:
How about `admonition:: deprecated`?

rmcgibbo:
Good suggestion. That works better. Uploaded now.

rmcgibbo:
Remaining work:
 - [ ] Search box doesn't work on C++ docs
 - [ ] Update user guide on how to build the API docs
 - [ ] Update travis-ci to build the API docs
 - [ ] Fix up doxygen invocations in CMakeLists
 - [ ] Check everything for correctness
 - [ ] Add some narrative text where appropriate and finalize the organization / layout

rmcgibbo:
Okay, I think this is basically finished. Obviously there's still room for improvement, but overall it does the trick, I think.

- [Python API Docs](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/index.html)
- [C++ API Docs](http://openmm-python-api.s3-website-us-east-1.amazonaws.com/api-c++/index.html)

peastman:
In the C++ docs, when a method takes a single argument, there's no bullet mark in front of it.  For example, if the argument is called "index", it just says, "Parameters index".  That looks kind of strange.  Also, the first parameter is always listed on the same line as the word "Parameters".  That's different from the Python docs, where it goes on a different line.  I like the latter approach better.

When a return value is a reference, the & is missing.  Look, for example, at System::getForce().

In the description of State on the front page, it says, "As discussed earlier..."  Actually, it wasn't.

rmcgibbo:
Looks like lost my previous comment. Most of these are stylistic and could be fixed later though.

- OutOfPlaneSite's detaileddescription has some stuff (`<sub>`) that doesn't work.
- Maybe show the header file required for each C++ class? Doxygen does this and it's kind of nice.
- If we're going to customize the theme, I like having some horiztonal lines running between functions. 
- The method signatures don't look that great, mostly because there is excess use of italics, and I think there should be more monospaced.
- Pass-by-reference method parameters that are used for output (`@param[out]` vs `@param` in doxygen comments) aren't demarcated separately. This could be done pretty simply just by adding the string [out] or something to the parameter name / description.

rmcgibbo:
Okay, all those issues are fixed (except the aspirational ones about improving the theme or syntax highlighting that I don't consider blcokers)

peastman:
This is looking great.  I think we may finally be just about ready to merge it!  Anything else we should wait on?

rmcgibbo:
I think we're ready to go. Only thing would be if you want me to rebase the commits? Once it gets merged in, the docs should get pushed to docs.openmm.org/development after each build, so we can improve them incrementally if we see any problems.

peastman:
> Only thing would be if you want me to rebase the commits?

Totally up to you.

rmcgibbo:
Okay, let's merge then.

swails:
Sweet!  Awesome work, @rmcgibbo!

PR #1176, opened 2015-10-07
Add support of a 4-dummy atom tetrahedral zinc ion model

rafwiewiora:
See #1032.

### Changes:

`modeller.py`: new `VirtualSite` type added to `addExtraParticles`, the rest of the changes to do with bonding issues. Starting PDB contains CONECT records between the ligating residues (cysteine and histidine) and zinc. These have to be removed in the final topology, as the bonding between the amino acids and metal is only electrostatic in this model, however the information must also be used to place the dummy atoms along the zinc - ligand vector. 

`bondedToAtom` and `bondedToAtomNoEP` will now not contain bonds between the ligating residues and metals (for now only zinc). This enables both the residues and the metal to be correctly matched with their forcefield templates. 

`delBonds` contains information on the bonds along which dummy atoms were added, those bonds are not copied from the original topology to the new topology. `newVirtualBonds` contains information on the new bonds required to be added between the central zinc and the dummy atoms and between dummy atoms themselves. The bonds between zinc and dummy atoms are added in the loop that iterates over sites in the residue and calculates positions of dummy atoms. Subsequently, we iterate over the dummy atoms in those bonds and add bonds between each of them. 

`templateOutsideLigands` collects information on which atoms are bound to the central zinc (which is the only atom that will be found in `matchingAtoms`). dummyTetrZinc will only go ahead if each zinc has 4 ligands. The virtual site comes with two attributes: `atom1`, which specifies the central zinc atom and `weight1`, which specifies the length of the Zn - Dummy bond in nanometers (0.09 nm). The central zinc atom --> ligating atom vector is normalized and the dummy atoms placed 0.09 nm away from the zinc, along the vector. 

`forcefield.py`: add the new `VirtualSite`. 

`dummyTetrZinc.xml`: ffxml containing the ZN residue (tetrahedron, central Zn and 4DA's) and the HIN residue - negatively charged histidine. Negatively charged cysteine (CYM) is already present in the OpenMM FFs. Parameters from http://www.mayo.edu/research/labs/computer-aided-molecular-design/projects/zinc-protein-simulations-using-cationic-dummy-atom-cada-approach and Amber14's parm10.dat / frcmod.ff14SB. - this to be added as provenance info to the ffxml when I get the provenance stuff done for other FFs as well, which is next on the list!

`hydrogens.xml`: add HIN as variant of HIS. CYM does not need to be added as variant of CYS as it's identical hydrogen-wise to CYX. 

[...]

PR #1240, opened 2015-11-03
addExtraParticles() can use bonds to select positions

peastman:
This is one piece of #1176.  It allows addExtraParticles() to work for multi-site ion models where several extra particles are connected to the atom (and to each other) by harmonic bonds.  @rafwiewiora and @jchodera could you take a look at this and see if it looks OK for your purposes?

### DIFF
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -1029,11 +1029,11 @@ class Modeller(object):
                             if atom1 in newExtraPoints:
                                 a1 = newExtraPoints[atom1]
                             else:
-                                a1 = matchingAtoms[atom1]
+                                a1 = newAtoms[matchingAtoms[atom1]]
                             if atom2 in newExtraPoints:
                                 a2 = newExtraPoints[atom2]
                             else:
-                                a2 = matchingAtoms[atom2]
+                                a2 = newAtoms[matchingAtoms[atom2]]
                             newTopology.addBond(a1, a2)
                             
         for bond in self.topology.bonds():
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -35,7 +35,7 @@ __author__ = "Peter Eastman"
 __version__ = "1.0"
 
 from simtk.openmm.app import Topology, PDBFile, ForceField
-from simtk.openmm.app.forcefield import HAngles, _createResidueSignature, _matchResidue, DrudeGenerator
+from simtk.openmm.app.forcefield import HAngles, AllBonds, _createResidueSignature, _matchResidue, DrudeGenerator
 from simtk.openmm.app.topology import Residue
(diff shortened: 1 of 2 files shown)


rafwiewiora:
Awesome, thanks @peastman! I'm going to have a more detailed look ASAP.

jchodera:
Seems like it might work.

peastman:
Shall I go ahead and merge this?

jchodera:
@rafwiewiora: Does this meet your needs?

rafwiewiora:
Sorry, too many talks this week - let me check this out now.

rafwiewiora:
@peastman: looks like there's a problem with the addition of bonds to `newExtraPoints` and hence `system = forcefield.createSystem(newTopology, constraints=AllBonds)` can't be created (ZN residues do not match templates on bonds). 

Looking at the code it seems to me this is the issue: [here] (https://github.com/peastman/openmm/blob/extraparticles/wrappers/python/simtk/openmm/app/modeller.py#L1032) the central ZN's will be in `matchingAtoms`, but `matchingAtoms` is a template atom --> old topology dictionary, so you actually create a new topology extra point to old topology central Zn atom bond, right?

Now you can't use `newAtoms` either, because that's an old topology atom --> new topology atom dictionary. What you're missing and need here is a template atom --> new atom dictionary. Perhaps `newAtoms` could become that by changing: 

```python
for atom in residue.atoms():
    if atom.element is not None:
        newAtoms[atom] = newTopology.addAtom(atom.name, atom.element, newResidue)
        newPositions.append(deepcopy(self.positions[atom.index]))
```

to:

```python
for atom in template.atoms:
    if atom.element is not None:
        newAtoms[atom] = newTopology.addAtom(atom.name, atom.element, newResidue)
        newPositions.append(deepcopy(matchingAtoms[atom].index))
```

rafwiewiora:
Also, thinking about this, it seems to me there's a certain redundancy here - you only add to `newTopology` those atoms from the old topology for which `atom.element is not None`, why not all `matchingAtoms`? If I load a PDB in which the dummy atoms already exist, this will recreate the `element == None` dummies and then have to SHAKE the positions - why not use the existing information already?

peastman:
Do you have a test case that is failing?  The one I added to TestModeller is passing.

peastman:
Ok, I figured out how to reproduce it.  I updated the test case and fixed the problem.

rafwiewiora:
Awesome! Works like a charm! Here's applied to the 2-Zn cluster in 4H12:
untitled

Would you expect that any problems might arise with the SHAKE positioning, i.e. would this benefit from testing on a larger number of systems? Or is it a works on one, will work on all kind of thing?

rafwiewiora:
Good to merge.

peastman:
SHAKE is a pretty robust algorithm.  And in this case it doesn't need to produce perfect positions, just ones that are good enough not to cause problems for energy minimization.  So it will probably be fine, but keep an eye out for any problems.

PR #1242, opened 2015-11-04
Fix volume calculation in StateDataReporter for triclinic unit cells.

swails:


peastman:
This isn't necessary.  The calculation was already correct.  See http://docs.openmm.org/6.3.0/userguide/theory.html#periodic-boundary-conditions.

swails:
Oh yea, *reduced* vectors... :)

PR #1245, opened 2015-11-05
Implemented CompoundIntegrator

### LINKED ISSUE #1196: CompoundIntegrator
I am in the process of implementing a non-equilibrium candidate Monte Carlo procedure where I drive (perturb) internal degrees of freedom using some Monte Carlo move, and then propagate the remaining degrees of freedom using Velocity Verlet. After some discussions with John and Peter it seems these two steps are best coded in my own Integrator plugin - to allow specifying mutually exclusive sets of degrees of freedom. This should be no problem.

However, after a number of steps of this specialized integrator, I would like to run a number of steps with a generalized MD integrator on the entire system. For this purpose it would be extremely useful to have a CompoundIntegrator, which alternates between running two integrators, running a specified number of steps of each. For my purposes, no communication is needed between the integrators.

Would such an integrator be difficult to implement? - and would anyone else find it useful?

peastman:
Fixes #1196

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -1125,6 +1125,13 @@ perform integration on a modified potential energy surface to allow much faster
 sampling of conformations.  For details on how to use them, consult the API
 documentation.
 
+Compound Integrator
+-------------------
+
+The :class:`CompoundIntegrator` class is useful for cases where you want to use
+multiple integration algorithms within a single simulation.  It allows you to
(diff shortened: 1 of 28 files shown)


jchodera:
Wow!  Awesome!

rmcgibbo:
I think all of the Appveyor tests have been failing since this, with the same issue noted https://github.com/pandegroup/openmm/pull/1244#issuecomment-153943359. I kind of doubt that it's related to this code at all, but I'm not sure what the cause is.

PR #1248, opened 2015-11-05
Fix a handful of compiler warnings

swails:
These are worth fixing because they're in header files that are always included
(rather than just hitting them at OpenMM compile-time). The SWIG interface now
builds without warnings on my machine.

The `AmoebaVdwForce` diff looks bigger than it really is since it was all dossed up and I smashed it with dos2unix.  Look at 197d776 for the substantive differences here.


mpharrigan:
Plugging my favorite github secret: append `?w=true` to web url to hide whitespace changes from the diff

swails:
You've posted that before, but I can never remember...

One of these days it'll stick :)

peastman:
If only they would add a "hide whitespace changes" button to the web page, then we wouldn't need to remember it...

PR #1249, opened 2015-11-05
Fixed a race condition in energy computation

peastman:
This fixes an error that very occassionally caused energies to be slightly off.  When using a separate stream for PME, it was possible for two different kernels to try to update energies at the same time, causing some of the contributions to be lost.  This mainly happens with CUDA 7.5.  There's no reason it couldn't happen with earlier releases, but something in that update seems to have made it more likely to happen.

### DIFF
--- a/platforms/cuda/include/CudaKernels.h
+++ b/platforms/cuda/include/CudaKernels.h
@@ -592,7 +592,7 @@ class CudaCalcNonbondedForceKernel : public CalcNonbondedForceKernel {
 public:
     CudaCalcNonbondedForceKernel(std::string name, const Platform& platform, CudaContext& cu, const System& system) : CalcNonbondedForceKernel(name, platform),
             cu(cu), hasInitializedFFT(false), sigmaEpsilon(NULL), exceptionParams(NULL), cosSinSums(NULL), directPmeGrid(NULL), reciprocalPmeGrid(NULL),
-            pmeBsplineModuliX(NULL), pmeBsplineModuliY(NULL), pmeBsplineModuliZ(NULL),  pmeAtomRange(NULL), pmeAtomGridIndex(NULL), sort(NULL), fft(NULL), pmeio(NULL) {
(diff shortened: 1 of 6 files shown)


jchodera:
Can you give us an indication of how long this bug has been in there and what magnitude of energy errors we are likely to have seen?  This may have affected some published results.

peastman:
That's actually a kind of complicated question to answer.  The separate stream for PME was introduced in 6.2.  But due to many CUDA bugs we've had to work around, it's been disabled on and off in a variety of situations.  In 6.2 it was only enabled if your GPU had compute capability < 5.0, your CUDA version was at least 6.0, and the system size was <130,000 atoms.  In 6.3 that switched to being enabled on all cards *except* GTX 980.

And just because the bug was present in the code, that doesn't mean it would manifest itself.  The person who reported this saw it appear when he upgraded to CUDA 7.5.  That doesn't mean it *couldn't* appear on earlier versions, but in the case of his system it didn't.  Using his system, I was also able to reproduce it with CUDA 7.5 (which is how I figured out what the problem was).  But only with the CUDA platform, not the OpenCL platform.  Which again doesn't mean that it *couldn't* happen with OpenCL, but at least in that case, it didn't.

The magnitude of the energy errors was small in relative terms (around 1%), but of course that could still be large in absolute terms.  Also, it only happened occasionally.  To reproduce the problem, I used a script that calculated the energy 10,000 times in a row, then looked for the difference between the minimum and maximum values.

Finally, note that this *only* affected energies.  Forces were always correct.

jchodera:
Thanks for the detailed explanation!

I think it's important to track critical bugfixes like this that could have resulted in erroneous data being used in publications.  Perhaps we should have a category "critical bugs" where we try our best to estimate the scope and magnitude of the problem, and then compile these somewhere, just so that researchers can check this log to see if they were affected?

peastman:
Sounds like a good idea.

jchodera:
Does it make sense to backport this bugfix to OpenMM 6.3?  I can't tell for certain, but it seems likely that this issue was responsible for a lot of the Bad State errors we had with the Folding@home core 21 v0.0.12 on projects that used PME.

PR #1250, opened 2015-11-06
ForceField will not create duplicate constraints

### LINKED ISSUE #1188: ForceField can add multiple copies of same bond, which causes constraint issue
We have a simple small molecule system where integration explodes if `HBond` constraints are used, but things work fine if no constraints are used.

The system was parameterized with `ForceField` otherwise identically in every way.

Here is some simple test code:
```python
from simtk import openmm
import numpy
import simtk.unit as u
from simtk.openmm import XmlSerializer

postfix = 'None' # this is fine                                                                                                                                                                                                                                                                                                                                           
postfix = 'HBonds' # this breaks                                                                                                                                                                                                                                                                                                                                          

state = XmlSerializer.deserialize(open('state-%s.xml' % postfix, 'r').read())
system = XmlSerializer.deserialize(open('system-%s.xml' % postfix, 'r').read())
integrator = XmlSerializer.deserialize(open('integrator-%s.xml' % postfix, 'r').read())

context.setVelocities(state.getVelocities())

print "Simulating..."

for i in range(1000):
    integrator.step(1)
    positions = context.getState(getPositions=True).getPositions(asNumpy=True)
    if numpy.isnan(positions / u.nanometers).any():
        print("NaN positions detected on step "+str(i))
        exit(1)

print("Complete!")
```

Changing the constraint tolerance does not change the behavior.

Serialized XML input files (for convenience) are all contained in [this gist](https://gist.github.com/jchodera/ea2287a0fdf7f8e5f51f).

A mol2 file of the molecule and the `ffxml` file the `System` was built from are also included.

peastman:
This fixes #1188.  If there are multiple constraints for a pair of atoms, but they all have the same length, it simply removes the duplicates.  If they have different lengths, it throws an exception.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -270,6 +270,17 @@ class ForceField(object):
             self.impropers = []
             self.atomBonds = []
             self.isAngleConstrained = []
+            self.constraints = {}
+
+        def addConstraint(self, system, atom1, atom2, distance):
+            """Add a constraint to the system, avoiding duplicate constraints."""
+            key = (min(atom1, atom2), max(atom1, atom2))
+            if key in self.constraints:
+                if self.constraints(key) != distance:
+                    raise ValueError('Two constraints were specified between atoms %d and %d with different distances' % (atom1, atom2))
+            else:
+                self.constraints[key] = distance
+                system.addConstraint(atom1, atom2, distance)
 
     class _TemplateData:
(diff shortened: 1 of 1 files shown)


jchodera:
Why would we not want to throw an exception if multiple constraints of the same length are imposed, out of curiosity?  Is this about being permissive when possible?

peastman:
We don't forbid having multiple bonds, so it doesn't necessarily indicate an error.

PR #1256, opened 2015-11-11
Changes necessary for VS2015

rmcgibbo:
I encountered a couple build failures with Visual Studio 2015 fixed here.
 - The `WIN32` macro doesn't seem to be defined (I think `_WIN32` is the canonical spelling).
 - VS2015 now has a `struct timespec`.


libraries/pthreads/include/pthread.h
```diff
@@ -121,6 +121,9 @@
 #endif
 
 
+#if _MSC_VER >= 1900
+#include <ctime>
+#else
 /* Windows doesn't have this, so declare it ourselves. */
```
mpharrigan: update this comment?

mpharrigan: oh it was already merged nevermind

PR #1258, opened 2015-11-11
Residue templates can specify atoms by name instead of index

### LINKED ISSUE #1187: Add fromName and toName attributes to ffxml parsing for Bond and ExternalBond
(Continued from #1168)

We want to extend the `ffxml` format to add `fromName` and `toName` attributes to `<Bond>` and `<ExternalBond>` elements:
```XML
   <Bond fromName="O" toName="H1"/>
   <Bond fromName="O" toName="H2"/>
   ...
   <ExternalBond fromName="N">
```

peastman:
Fixes #1187.  I also converted the XML files that were written by hand rather than generated from something else.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -1889,12 +1889,12 @@ The residue template definitions look like this:
       <Atom name="HH33" type="710"/>
       <Atom name="C" type="712"/>
       <Atom name="O" type="713"/>
-      <Bond from="0" to="1"/>
-      <Bond from="1" to="2"/>
-      <Bond from="1" to="3"/>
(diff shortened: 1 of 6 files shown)


jchodera:
The `VirtualSite` docs don't mention the deprecated interface retained for backward compatibility, while the `Bond` docs do.

jchodera:
Do the `ExternalBond` docs need to be updated too?

jchodera:
Looks like this can be merged since everything passes but a test timeout.

jchodera:
Oh, except the `ExternalBond` docs still need updating.

jchodera:
Thanks!

PR #1260, opened 2015-11-12
Deploy development docs to S3

rmcgibbo:
RE https://github.com/pandegroup/openmm/pull/1228#issuecomment-156209569, this should deploy the development docs to S3 at http://docs.openmm.org/development, which could aid with debugging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ matrix:
            CUDA=false
            CC=clang
            CXX=clang++
+           DOCS_DEPLOY=true
            CMAKE_FLAGS=""
 
     - sudo: false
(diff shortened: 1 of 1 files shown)


peastman:
I hope you're not doing anything insecure by hardcoding the access keys?

rmcgibbo:
They're encrypted. http://docs.travis-ci.com/user/environment-variables/#Encrypting-Variables-Using-a-Public-Key

peastman:
Ok, just making sure!  Seeing a field called `secret_access_key` hardcoded in the script made me a little nervous.

mpharrigan:
but it's followed by `secure:` :)

rmcgibbo:
Kk. I think this is correct now. We'll only see the site once it runs in `master` after the merge.

rmcgibbo:
The docs for the master branch can now be found here.

 - http://docs.openmm.org/development/api-python/annotated.html
 - http://docs.openmm.org/development/api-c++/annotated.html

PR #1262, opened 2015-11-13
AMD docstring fix

rmcgibbo:


peastman:
There are three different integrators in that file, each with a `getEffectiveEnergy()` method.  You changed one of them, but not the other two.  Also, perhaps list the return type as "energy" rather than "value"?  It will have units on it.

rmcgibbo:
I changed the two that have the return value separated out. The other one just has a single-line unformatted docstring.

PR #1264, opened 2015-11-17
Fix #1213 addInteractionGroup should accept `set` values as parameters

### LINKED ISSUE #1213: CustomNonbondedForce.addInteractionGroup should accept arbitrary iterables
`CustomNonbondedForce.addInteractionGroup` only seems to accept `list` arguments, but not `set`:
```
>>> from simtk import openmm
>>> force = openmm.CustomNonbondedForce('r')
>>> force.addInteractionGroup(set([0,1,2]), set([1,2,3]))
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 7859, in addInteractionGroup
    return _openmm.CustomNonbondedForce_addInteractionGroup(self, *args)
TypeError: in method 'CustomNonbondedForce_addInteractionGroup', argument 2 of type 'std::set< int,std::less< int >,std::allocator< int > > const &'
>>> force.addInteractionGroup([0,1,2], [1,2,3])
0
```

philiptzou:
apparently SWIG typemaps maps Python's list but not set with std::set. Instead, Python's `set` is mapped with `std::unordered_set`.

The original code of `swigInputBuilder.py` is kindly in a mess. Too much lines in a single method `writeMethods` which causes debug becomes difficult. Also there are a lot of warnings about [PEP8](https://www.python.org/dev/peps/pep-0008/).

A refactoring in the future may be a good idea. :)

### DIFF
--- a/wrappers/python/src/swig_doxygen/swigInputBuilder.py
+++ b/wrappers/python/src/swig_doxygen/swigInputBuilder.py
@@ -378,7 +378,6 @@ class SwigInputBuilder:
             self.fOut.write("\n%s};\n" % INDENT)
         if len(enumNodes)>0: self.fOut.write("\n")
 
-
     def writeMethods(self, classNode):
         methodList=getClassMethodList(classNode, self.skipMethods)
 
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Can we add a test for this?

philiptzou:
Actually I am thinking about to add a test... Will do that later.

rmcgibbo:
Sounds good. Especially because this `swigInputBuilder.py` code is a kludgey mess (I'm probably the last persion to mess around in that mud pit and am partly to blame), tests seem particularly critical.

philiptzou:
@rmcgibbo Well, so I assume a refactoring pull-request is also welcomed, right? I might do that.

rmcgibbo:
I would be +1 for refactoring, yes, as long as we're careful to ensure correctness. The code may be kludgey, but it's also been extensively deployed over many years.

PR #1270, opened 2015-11-19
Force usePmeQueue = false to eliminate potential race condition on NVIDIA hardware

jchodera:
Suggested by @peastman as a monkey-patch for the issue in https://github.com/FoldingAtHome/openmm-core/pull/90


jchodera:
Whoops, wrong merge direction.

PR #1272, opened 2015-11-20
Amber switching

swails:
Closes #1177 

Still needs a test.  I did this because the code bases were beginning to diverge and if we waited longer we may have to just redo it :).  I preserved @mj-harvey's commits so he will be represented in the contributor list.

I originally rebased this on master, but the end result was history pollution that was not worth it IMO. (I don't like rebasing if the rebase is not clean.)  So we get an extra merge commit here.


swails:
Tests added.  Good to merge.

wrappers/python/simtk/openmm/app/amberprmtopfile.py
```diff
@@ -192,6 +202,11 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
             total mass the same.
         ewaldErrorTolerance : float=0.0005
             The error tolerance to use if nonbondedMethod is Ewald or PME.
+        switchDistance : float=0*nanometers
+            A positive value turns on a potential energy switching function for
+            Lennard-Jones interactions. If the switchDistance is 0 or evaluates
+            to boolean False, no switching function will be used. Illegal values
+            (e.g., less than 0) will raise a ValueError
```
peastman: This description doesn't say what `switchDistance` *is*.  It makes it sound like just a boolean flag, where all that matters is the sign.

swails: Oh, I didn't know you wanted docstrings to be *descriptive*...

swails:
How's the new version?

You can also cancel the travis tests for 35adccb (it'll speed up the last commit testing by probably 30-40 minutes or so).

The last commit before the docstring changes passed Travis (failed appveyor), so the tests should definitely pass.

peastman:
Looks good.

swails:
The same multi-site ion test failed in the 2 builds that failed in Travis.  This looks unrelated to the changes I made (particularly since it worked before I just changed the docstring).

Is this a test that can fail randomly?  Can we merge this anyway and deal with potential issues in the multisite ion test in a different PR?

peastman:
I don't know of any reason for that to fail, but I agree it doesn't seem to have anything to do with this PR.  I'll look into it.

PR #1279, opened 2015-12-01
Support nucleic acids in AMOEBA 2009

peastman:
For most purposes, the AMOEBA 2009 force field is deprecated and people should use AMOEBA 2013 instead.  But there's one exception: AMOEBA 2013 doesn't yet include parameters for nucleic acids, so for that you still need to use 2009.  But our version didn't actually include the parameters them.  This adds that, along with the needed changes to the conversion script.

This has been tested somewhat, but could definitely use more testing.  On one particular system, I've verified that it produces forces and energy identical to Tinker.


swails:
Thanks @peastman!

PR #1285, opened 2015-12-08
PDBFile - change behavior of connectivity to metals to that of PDBxFile

rafwiewiora:
Here's the thinking:

1. I scanned the pdbx files of the PDB for all `metalc` bonds. 

2. By looking for bonds with atom.name == residue.name I identified large majority of the metals involved. 

3. Looking at the remaining bonds and checking what atom names started with, removing digits etc. a few further metals were identified. 

4. Wanting to change to coding conditions by element rather than names, I've scanned the whole PDB with both PDBFile and PDBxFile looking for atoms for which topology was returning element as None. Nicely and as you would expect, the only residues for which that happened were `{'ASX', 'GLX', 'UNK', 'UNL', 'UNX'}`. There will be no problems with missing elements. 

5. For final confirmation I had not missed any elements I looked at all elements found in `PDBxFile` topologies for files containing `metalc` sections - all metals were found ok. 

6. For all metal elements identified above, I checked over the PDB whether `PDBxFile` was returning any bonds (i.e. whether those elements were present in `covale` or `modres` sections). That did happen for some of the elements, and looking at the residues and atoms those bonds were to: there wasn't any cases where a rule could be developed for deciding whether the bond was `metalc` or `covale` / `modres` - i.e. same bond pairs are present in both categories. Not clear if this is PDB's mistakes or chemical knowledge was used manually, maybe both. To decide what to do with these I compared the total numbers of bonds these metals had in `metalc` vs `covale` / `modres`:

    metalc:

   ``` 
    {'AG': 52,
     'AU': 147,
     'BE': 123,
     'CR': 27,
     'EU': 212,
     'FE': 43179,
     'GA': 12,
     'HG': 4147,
     'LU': 44,
     'NA': 62969,
     'OS': 7,
     'PR': 89,
     'RE': 14,
     'RU': 75,
     'TB': 55}
   ```

    covale / modres:

   ```
    {'AG': 31,
     'AU': 38,
     'BE': 87,
     'CR': 28,
     'EU': 61,
     'FE': 5,
     'GA': 22,
     'HG': 2,
     'LU': 4,
     'NA': 1,
     'OS': 6,
     'PR': 69,
     'RE': 21,
     'RU': 91,
     'TB': 26}
   ```

    I decided - for Fe, Na, Hg - numbers clearly suggest the covale / modres entries are mistakes in the     pdbx files - keep in the list, for the others: numbers too similar, without having more understanding of these ions / models for them, better to keep the current PDBFile behavior. So:
    ```
['Ag', 'Au', 'Be', 'Cr', 'Eu', 'Ga', 'Lu', 'Os', 'Pr', 'Re', 'Ru', 'Tb']
    ```

    were removed from `metalElements`. 

7. The last issue is deciding what to do with bonds within small-molecule ligands / cofactors that contain any of these metal elements. I was expecting here that `covale` would contain bonds within ligands, but this doesn't seem to be the case. For example, for HEM, the most common metal containing cofactor, PDBxFile scanning over the whole PDB had only 2 files with topologies containing any bonds within HEM, while there are 3756 structures containing HEM. By looking at a few pdbx files I didn't not see any where this connectivity would be written in any other section, though it seems like [chem_comp_bond](http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v40.dic/Categories/chem_comp_bond.html)  is meant for this ? Also, I could not find any justification in the PDBx materials for omitting connectivity in small molecules, but this appears to be the what they've done. 

    I run the updated `PDBFile` as in the PR over the whole database to see what kind of ligands had CONECT bonds to these metal elements. The three most common are: HEM (heme) (27% bonds), SF4 (iron - sulfur cluster) (17%), CLA (chlorophyll A) (14%), all others below 5%.  While one could argue the metal - nitrogen bonds in HEM and CLA should not be added (though I wonder whether covalent bonds are the only way of forcing the unusual coordinations there), covalent bonds are probably a good idea for iron-sulfur clusters. 

    As there is big ambiguity here, I think a sane decision is to retain current behavior of `PDBFile` - i.e. adding all CONECT bonds within small molecule ligands / cofactors, and only change the behavior for inter-residue ligation. Hence code still adds the bonds if `atomByNumber[i].residue == atomByNumber[j].residue`. 

I retained the current behavior for atoms where element is None (but as I've shown this only happens within the database for truly unknown atoms). `PDBFile` will add such bonds, the `metalElements` check will only happen if both atoms in the bond have `element != None` to avoid 'Nonetype has no attribute symbol' exceptions. There is a very small amount of cases of `metalc` metal - unknown bonds I saw, but I think caring about not adding those is not worth the extra lines of code, as those topologies are broken anyway. 

Was planning to test this over the PDB by comparing numbers of bonds between `PDBFile` and `PDBxFile`, but the missing connectivity in small molecules situation in PDBx makes that nonsense.

### DIFF
--- a/wrappers/python/simtk/openmm/app/pdbfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbfile.py
@@ -71,7 +71,7 @@ class PDBFile(object):
         """
         
         metalElements = ['Al','As','Ba','Ca','Cd','Ce','Co','Cs','Cu','Dy','Fe','Gd','Hg','Ho','In','Ir','K','Li','Mg',
-        'Mn','Mo','Na','Ni','Pb','Pd','Pt','Rb','Rh','Ru','Sm','Sr','Te','Tl','V','W','Yb','Zn']
+        'Mn','Mo','Na','Ni','Pb','Pd','Pt','Rb','Rh','Sm','Sr','Te','Tl','V','W','Yb','Zn']
         
         top = Topology()
         ## The Topology read from the PDB file
--- a/wrappers/python/simtk/openmm/app/pdbfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbfile.py
@@ -73,6 +73,10 @@ class PDBFile(object):
         metalElements = ['Al','As','Ba','Ca','Cd','Ce','Co','Cs','Cu','Dy','Fe','Gd','Hg','Ho','In','Ir','K','Li','Mg',
         'Mn','Mo','Na','Ni','Pb','Pd','Pt','Rb','Rh','Sm','Sr','Te','Tl','V','W','Yb','Zn']
         
+        standardResidues = ['ALA', 'ASN', 'CYS', 'GLU', 'HIS', 'LEU', 'MET', 'PRO', 'THR', 'TYR',
+                            'ARG', 'ASP', 'GLN', 'GLY', 'ILE', 'LYS', 'PHE', 'SER', 'TRP', 'VAL',
+                            'A', 'G', 'C', 'U', 'I', 'DA', 'DG', 'DC', 'DT', 'DI', 'HOH']
+
         top = Topology()
         ## The Topology read from the PDB file
         self.topology = top
@@ -155,7 +159,7 @@ class PDBFile(object):
         self.topology.createDisulfideBonds(self.positions)
         self._numpyPositions = None
 
-        # Add bonds based on CONECT records. Bonds between metals of element specified in metalElements and standard residues are not added. 
+        # Add bonds based on CONECT records. Bonds between metals of elements specified in metalElements and residues in standardResidues are not added.
 
         connectBonds = []
         for connect in pdb.models[-1].connects:
@@ -165,9 +169,9 @@ class PDBFile(object):
                     if atomByNumber[i].element is not None and atomByNumber[j].element is not None:
                         if atomByNumber[i].element.symbol not in metalElements and atomByNumber[j].element.symbol not in metalElements:
                             connectBonds.append((atomByNumber[i], atomByNumber[j])) 
-                        elif atomByNumber[i].element.symbol in metalElements and atomByNumber[j].residue.name not in top._standardBonds:
+                        elif atomByNumber[i].element.symbol in metalElements and atomByNumber[j].residue.name not in standardResidues:
                             connectBonds.append((atomByNumber[i], atomByNumber[j])) 
-                        elif atomByNumber[j].element.symbol in metalElements and atomByNumber[i].residue.name not in top._standardBonds:
+                        elif atomByNumber[j].element.symbol in metalElements and atomByNumber[i].residue.name not in standardResidues:
                             connectBonds.append((atomByNumber[i], atomByNumber[j]))     
                     else:
                         connectBonds.append((atomByNumber[i], atomByNumber[j]))         
--- a/wrappers/python/simtk/openmm/app/pdbfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbfile.py
@@ -162,7 +162,7 @@ class PDBFile(object):
             i = connect[0]
             for j in connect[1:]:
                 if i in atomByNumber and j in atomByNumber:
-                    if atomByNumber[i].element != None and atomByNumber[j].element != None:
+                    if atomByNumber[i].element is not None and atomByNumber[j].element is not None:
                         if atomByNumber[i].element.symbol in metalElements or atomByNumber[j].element.symbol in metalElements:
                             if atomByNumber[i].residue == atomByNumber[j].residue:
                                 connectBonds.append((atomByNumber[i], atomByNumber[j])) 
--- a/wrappers/python/simtk/openmm/app/pdbfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbfile.py
@@ -59,7 +59,7 @@ class PDBFile(object):
     _residueNameReplacements = {}
     _atomNameReplacements = {}
 
-    def __init__(self, file):
+    def __init__(self, file, removeBondsToMetals=False):
         """Load a PDB file.
 
         The atom positions and Topology can be retrieved by calling getPositions() and getTopology().
@@ -68,6 +68,9 @@ class PDBFile(object):
         ----------
         file : string
             the name of the file to load
+        removeBondsToMetals : bool
+            if True, those inter-residue CONECT bonds that indicate coordination to metals 
+            (cf. metalc connectivity in mmCIF/PDBx) are not added to the topology
         """
         
         metalElements = ['Al','As','Ba','Ca','Cd','Ce','Co','Cs','Cu','Dy','Fe','Gd','Hg','Ho','In','Ir','K','Li','Mg',
@@ -161,7 +164,9 @@ class PDBFile(object):
         for connect in pdb.models[-1].connects:
             i = connect[0]
             for j in connect[1:]:
-                if i in atomByNumber and j in atomByNumber:
+                if i in atomByNumber and j in atomByNumber and not removeBondsToMetals:
+                    connectBonds.append((atomByNumber[i], atomByNumber[j]))  
+                elif i in atomByNumber and j in atomByNumber and removeBondsToMetals:    
                     if atomByNumber[i].element is not None and atomByNumber[j].element is not None:
                         if atomByNumber[i].element.symbol in metalElements or atomByNumber[j].element.symbol in metalElements:
                             if atomByNumber[i].residue == atomByNumber[j].residue:
--- a/wrappers/python/simtk/openmm/app/pdbfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbfile.py
@@ -59,7 +59,7 @@ class PDBFile(object):
     _residueNameReplacements = {}
     _atomNameReplacements = {}
 
-    def __init__(self, file, removeBondsToMetals=False):
+    def __init__(self, file):
         """Load a PDB file.
 
         The atom positions and Topology can be retrieved by calling getPositions() and getTopology().
@@ -68,9 +68,6 @@ class PDBFile(object):
         ----------
         file : string
             the name of the file to load
-        removeBondsToMetals : bool
-            if True, those inter-residue CONECT bonds that indicate coordination to metals 
-            (cf. metalc connectivity in mmCIF/PDBx) are not added to the topology
         """
         
         metalElements = ['Al','As','Ba','Ca','Cd','Ce','Co','Cs','Cu','Dy','Fe','Gd','Hg','Ho','In','Ir','K','Li','Mg',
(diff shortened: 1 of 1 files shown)


peastman:
I'm worried this logic may be a little too permissive.  I know there are people using OpenMM to simulate a lot of things other than proteins, including condensed matter systems.  Maybe this logic works for the files in RCSB, but it could easily do the wrong thing with files that contain different sorts of molecules.

You said there were only a few protein residues that ever have metalc bonds?  If so, let's add a check and only omit bonds when they connect to one of those residues.  That way, if someone loads a file containing a completely different sort of system (in fact, anything that isn't a protein), there's no risk of this breaking it.

```Python
if atomByNumber[i].element != None and atomByNumber[j].element != None:
```

The better way to check for `None` is with `is not None`.  Using != can do unexpected things depending on how that operator is defined for the object on the left.

rafwiewiora:
I see, seems tricky though - there's quite a few protein residues that are in metalc (any residue can really coordinate with the backbone oxygens and nitrogens) - so for that we could just do a check for any protein residues. But there is a big variety of small molecule ligands and ions too, with no good way to check against.  

I'm not sure how to deal with this then. Can we just check that the topology contains 'a protein'? How about an argument to turn this on (`removeMetalBonds`, False default)?

rafwiewiora:
> The better way to check for None is with is not None. Using != can do unexpected things depending on how that operator is defined for the object on the left.

Good to know! Changing now.

jchodera:
> I'm worried this logic may be a little too permissive. I know there are people using OpenMM to simulate a lot of things other than proteins, including condensed matter systems. Maybe this logic works for the files in RCSB, but it could easily do the wrong thing with files that contain different sorts of molecules.

What about just having an optional flag `removeBondsToMetals=True`, and if users find that this messes up their attempts to read in crazy off-label uses of PDB file formats to represent metal-organic frameworks or whatnot, they can just set the flag to `False`?

rafwiewiora:
Yeah, I think that's a good solution - not sure whether True or False should be default - intuitively my default expectation would be that all CONECT bonds are added, so maybe False.

jchodera:
Sounds reasonable.  Setting it to `False` by default also preserves existing behavior.

rafwiewiora:
Exactly. Ok, updated with `removeBondsToMetals`.

peastman:
> Can we just check that the topology contains 'a protein'?

The question is not whether the topology contains a protein.  The question is whether the particular bond connects to a protein.

> so for that we could just do a check for any protein residues.

That sounds reasonable to me.

> What about just having an optional flag removeBondsToMetals=True

That's expecting users to know and specify something that, in practice, they usually don't know about and aren't qualified to set.  Nearly everyone will ignore that flag, not understand what it does, and leave it as the default value.  At which point it might as well not exist.

rafwiewiora:
> That sounds reasonable to me.

But that does not do anything about bonds to ions / small molecules I mentioned. 

> That's expecting users to know and specify something that, in practice, they usually don't know about and aren't qualified to set. Nearly everyone will ignore that flag, not understand what it does, and leave it as the default value. At which point it might as well not exist.

I don't see how anyone would attempt simulating a metal containing system without knowing what they're doing and expecting anything useful. Fine, we won't add the bonds to metals and then what? Are they going to know which metal ffxml to pass to `Forcefield` even though they didn't even stop to think what's happening to the connectivity data in their PDB file? 

Also, CONECT is the only mechanism by which to add non-standard bonds to topology via `PDBFile`. I expect that any user would have a strong intuitive expectation that whatever they write in CONECT will be added as bonds to the topology. What about covalent metal models? Those have been used and people might want to use them in OpenMM - set `removeBondsToMetals` to False and you can easily load in such systems, don't have the flag and all these rules about whether it's a protein or not - no way to load those in. While, there will be a way in `PDBxFile` - just add them to `covale`. So, while the point here is to make `PDBFile` more similar to `PDBxFile`, in fact we would limit the PDB format further. 

I think an important thing here is accepting the limitations of the PDB format. There are no subsections of CONECT as in PDBx and a sane expectation from any software would be to do exactly what's in the file. `PDBFile` just has to be worse than `PDBxFile`, because they're tied to their respective file formats. Now, we can add further functionality like removing these metal bonds, but that should be directed at people who know what they're doing and while accepting the limitations of a PDB file and still deciding to describe their system in that format, they are given these extra tools to get exactly what they want in the topology. 

Aside from all these thoughts - the `removeBondstoMetals` is the only way of dealing with people loading in condensed matter systems etc. that has been proposed here. As I said - we can check for standard protein residues, but I can think of no way of checking for other ions and small molecules in the coordination sphere. This seems like a very hacky territory to be fair...

peastman:
You're completely right that PDB is a limited format, and we have to work within the limitations of that.  So we need to consider potential use cases, then determine what is the best thing to do in each case.  Most importantly, we need to decide what the *default* behavior should be in each case.  We can always add options to override that.  And users always have the option of adding or removing bonds themselves.  But as much as possible, the default behavior should do the right thing.  That's one of the core design principles of OpenMM: if a users does the obvious thing, that should usually be the right thing.

One common use case is loading a file from RCSB.  What's the best thing to do in that situation?  There are a few options.  One is to simply add a covalent bond for every CONECT record.  But we know that will frequently produce incorrect results.  Not so good.  Can we do better?

Well, we can try to identify bonds that shouldn't be included.  We need to be careful here though.  We don't want to risk omitting "real" bonds.  If we have a high level of confidence that a CONECT record does *not* represent a covalent bond, we should omit it.  That makes the default behavior more correct, and increases the fraction of the time that "doing the obvious" does the right thing.  But if we can't be confident about it, then removing the bond is dangerous.  We could potentially make the default behavior less correct, not more.

In my opinion, the initial version of this PR went too far.  It was too aggressive about removing bonds, even in situations where we can't be confident it's the right thing to do.  But that doesn't mean we need to throw out the whole concept and have the default behavior revert back to not removing any at all!  If you see a CONECT record between a HIS sidechain and a monatomic Zn ion, it's a pretty solid bet that it represents metal coordination, not a covalent bond.  As long as we can be confident that removing a bond is the right thing to do, it should be the default behavior.

rafwiewiora:
Ok, I can see now where the philosophy behind this is coming from!

So let's talk more about these systems that you are worried this could break them, do you have any examples handy? I'd like to understand in what cases and why the CONECT records that this commit would have not added as bonds, should be added as bonds there. Maybe we can go from there with changing the logic here.

jchodera:
Any chance we could arrange a Skype call to quickly wrap up all these loose threads?

rafwiewiora:
Yes please!

peastman:
I could do a Skype call tomorrow.  I'm not in the office right now.

I don't have any example files.  This is just based on talking to users at workshops and the like.  But if we only remove bonds between an amino acid and a monatomic ion, I think that should be pretty safe?

jchodera:
Great!  How about 1.00P EST / 10.00A PST tomorrow?

rafwiewiora:
> But if we only remove bonds between an amino acid and a monatomic ion

Ok, let's do that. It will still leave ions / waters / small molecules bonded, but I guess it will at least be fully useful for metal coordinations where all ligands are amino acids, like the zinc cases we work on.

rafwiewiora:
Commit updated - bonds are now not added only between metal elements in metalElements and residues in top._standardBonds. This good?

peastman:
The content of `_standardBonds` isn't fixed.  Anyone can add entries to it by calling `loadBondDefinitions()`, so we can't rely on that containing only amino acids.  Even with just the default definitions, it contains nucleotides as well.  So if we only want standard amino acids, we need to hardcode the list of standard amino acids.

rafwiewiora:
Ah I see. Ok, so I have all the residue names from 'Protein' section of pdbNames.xml:
```
standardProteinResidues = ['GLY','ALA','VAL','LEU','ILE','SER','THR','CYS','PRO','PHE','TYR','TRP','HIS','ASP','ASN','GLU','GLN','MET','LYS','ARG','ORN','AIB','PCA','FOR','ACE','NME','NH2']
```

You want to limit that any further i.e. to only the very standard proteinogenic set, that'd be everything up to ARG, or is this good?

peastman:
@jchodera what do you think?  Can pretty much any amino acid coordinate ions (even things like ACE)?  Or should we restrict it to just a subset of them?

rafwiewiora:
Total of metalc bonds in PDBx files: 719687

Bonds which any of the residues listed above participates in (no correction for alternative names though): 393573 (55%)

Counts for each residue:
```
{'ACE': 3,
 'AIB': 0,
 'ALA': 4646,
 'ARG': 3121,
 'ASN': 16261,
 'ASP': 96420,
 'CYS': 59200,
 'FOR': 3,
 'GLN': 6559,
 'GLU': 64046,
 'GLY': 9851,
 'HIS': 74667,
 'ILE': 4354,
 'LEU': 5006,
 'LYS': 3761,
 'MET': 3506,
 'NH2': 0,
 'NME': 8,
 'ORN': 29,
 'PCA': 0,
 'PHE': 2988,
 'PRO': 1731,
 'SER': 10959,
 'THR': 13079,
 'TRP': 851,
 'TYR': 6349,
 'VAL': 6175}
```

So from those 'extra' ones, ORN highest at 29. If they would cause any trouble we should remove them, otherwise it's probably inconsequential either way.

rafwiewiora:
Btw, this is interesting data for our physics vs forcefield discussion this morning. Any residue can technically be a ligand via backbone oxygens and nitrogens, a tougher question to answer is are any such backbone coordinations structurally important or is that just a random ion clinging around in the crystal. Something I can definitely look into in the future.

rafwiewiora:
Ok, so here's the counts of all coordinating residues in the metalc records of PDBx, for atom names== residue names == element.upper(), using the list of metal elements as in this commit. 

https://gist.github.com/rafwiewiora/fee2f1fb94b5c6606981

rafwiewiora:
So our options here for the residues to which we won't add CONECT:

- the standard, proteinogenic amino acids

- all in the Protein section of pdbNames.xml, as listed above

- the already hard-coded standardResidues here: https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/pdbfile.py#L367

The last option includes nucleotides and water - from the gist above I'd argue: water should have bonds removed too, as the most common coordinating residue, and I don't see why not remove bonds to nucleotides too as those shouldn't be covalent either. Unless the addition of water and nucleotides would introduce any problems for those non-protein systems we've mentioned?

rafwiewiora:
Just a reminder for @jchodera to weight in on this, @peastman mentioned in the chat on Monday he was waiting for your thoughts

jchodera:
If we are agreed that the bond-removal process should happen in `PDBFile`, then removing bonds to the `standardResidues` sounds like the most sensible thing to do, since this will certainly cause problems downstream if we don't do this.

We do have to be cognizant that some residue definitions for cofactors like heme groups [do include bonds to `Fe`](http://personalpages.manchester.ac.uk/staff/Richard.Bryce/amber/cof/heme_all.in), so we shouldn't remove *all* bonds (as @peastman suggests).

rafwiewiora:
> We do have to be cognizant that some residue definitions for cofactors like heme groups do include bonds to Fe, so we shouldn't remove all bonds (as @peastman suggests).

Exactly, this is why in the first version of the PR I had a check for whether a bond was in the same residue and wasn't removing those. This won't be necessary anymore though, if we only remove bonds between metals and `standardResidues`. 

So are we settled on the `standardResidues` as hard-coded already?

peastman:
`standardResidues` would be reasonable.

rafwiewiora:
Updated. Looks ok now?

peastman:
Looks good!

PR #1286, opened 2015-12-10
WIP: Debug OpenCL on Travis

rmcgibbo:
It looks like the test failure in https://github.com/pandegroup/openmm/pull/1153 with OpenCL is caused by the migration of Travis-CI's services to a different cloud, reported https://github.com/travis-ci/travis-ci/issues/5221#issuecomment-162824870

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ before_install:
     fi
   - if [[ "$OPENCL" == "true" ]]; then
       sudo apt-get -yq update > /dev/null 2>&1 ;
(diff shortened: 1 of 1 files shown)


rmcgibbo:
```
Package fglrx is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
E: Package 'fglrx' has no installation candidate
```

I don't think there's anything we can do to fix this ourselves, other than just waiting for them to fix it.

PR #1288, opened 2015-12-11
Appveyor OpenCL

rmcgibbo:


rmcgibbo:
Surprisingly, this seems to work. The OpenCL tests can't run without a device (it looks like one is just hanging), but it does compile.

PR #1296, opened 2015-12-14
Bug fixes

peastman:
This fixes two errors.  One was a mistake in #1290.  The other was a memory leak.


jchodera:
Does this also affect the openmm-core bugfix?

peastman:
No.  The error was only in CUDA, not OpenCL.

peastman:
I did, however, just find a different error in that patch.  It will fail if you run it in mixed precision mode.  Does FAH ever do that, or does it always run in single precision?

jchodera:
As of 0.0.14, fah can run in mixed precision mode, so we will need to fix
this.

PR #1297, opened 2015-12-15
Backported fix for error in energy computation to 6.3

peastman:


jchodera:
Thanks!  Will this be 6.3.1?

peastman:
We still need to update the version number.  I think that's the only remaining change?

jchodera:
Do we have a version-controlled `CHANGELOG.md` or release notes?

peastman:
The release notes for each version are on simtk.org.

PR #1300, opened 2015-12-15
[CI] Use ccache

rmcgibbo:


peastman:
What is this for?

rmcgibbo:
You've never used ccache?

swails:
No... I'm also curious :)

rmcgibbo:
CCache is a compiler wrapper that caches each object file based on a hash of the contents of the .c/.cpp files and the command line flags. So if you compile the same file twice, it can just pull it from the cache without recompiling. Travis-CI also has a caching system where you can persist one or more directories between builds.

I think the net result should be that, if for example in each PR only 1% of the .cpp files are apt to be modified, that 99% of the compilation work done on travis can be skipped. So it should speed up the builds a lot. But I'm not sure totally. I think this could also work on Appveyor too. https://github.com/inorton/cclash

rmcgibbo:
In 2e01f6a on travis, everything hit the cache and the compilation took only like 1 minute :)

rmcgibbo:
Meh, overall on travis it doesn't seem to make too much of a difference. Maybe it'll be more on Appveyor.

rmcgibbo:
@peastman can you try enabling the "Rolling build" [feature](http://www.appveyor.com/docs/build-configuration#rolling-builds) in Appveyor? Apparently it needs to be done through the web UI at https://ci.appveyor.com/project/peastman/openmm/settings

jchodera:
NICE

peastman:
Ok, enabled.

rmcgibbo:
For some reason the exeutables produced when using ccache with clang are like ~5x slower than without  ccache, but this doesn't happen with gcc. It's like maybe ccache, when using clang, is losing the optimization flags. With gcc, the speed is unchanged, as expected.

rmcgibbo:
Oh nevermind, I see what I did wrong that caused the clang problem.

rmcgibbo:
Okay, ready to merge.

Issue #1193, opened 2015-10-20
Dummy atoms in OpenMM


leelasd:
Is it possible to have the option of element==None for dummy atoms when loading PDB files ?

jchodera:
We realized today at the OpenMM Jülich workshop that we don't seem to have a codepath in the `PDBFile` loader that would allow one to specify some atoms should match dummy atoms.

peastman:
The question is, how would you specify it?  Ideally we'd just say that if the "element" field of the ATOM record is empty, that's a dummy atom.  Except a lot of PDB files found in the wild leave that field empty, so instead it tries to guess the element based on the atom name.

I suppose we could make up some magic code that you put in that field to indicate a dummy atom.  Do you know how other programs handle this?

swails:
At least in Amber (and probably other programs) the dummy atom is part of the residue template. 

Of course all that differentiates a dummy atom from a regular one is a type that has no charge or LJ terms.

peastman:
But what does it expect to be in the PDB file?  How does it recognize which atoms in the PDB are dummy atoms?

leelasd:
@swails  Dummy atoms can also be a point charge/LJ sites that people are using for simulating halogen bonds, lone pairs of nitrogen and sulphur. It would be great if OpenMM can have a **X** atom type that can have any mass, charge or LJ values.

jchodera:
Here's another idea: In PDB files that specify the element column for *any* atom record, assume that records missing the element are dummy atoms.

swails:
> Here's another idea: In PDB files that specify the element column for any atom record, assume that records missing the element are dummy atoms.

Are you sure that all tools that add atoms (like modeling in missing loops, hydrogens, etc.) will add the element column for you?  Rosetta?  reduce?

swails:
> @swails Dummy atoms can also be a point charge/LJ sites that people are using for simulating halogen bonds, lone pairs of nitrogen and sulphur. It would be great if OpenMM can have a X atom type that can have any mass, charge or LJ values.

I usually see this done with virtual sites, which are different (and far superior for this type of thing).  OpenMM provides very flexible support for virtual sites, and I can't really see much of an argument for using a dummy atom to model lone pairs and off-center charges.

jchodera:
> Are you sure that all tools that add atoms (like modeling in missing loops, hydrogens, etc.) will add the element column for you? Rosetta? reduce?

[...]

PR #1302, opened 2015-12-16
Allow specification of extra particles in PDB files via element field and extraParticlesIdentifier

rafwiewiora:
Fixes https://github.com/pandegroup/openmm/issues/1193

Please check if the code looks ok and if I didn't break anything accidentally. 

Will update docstrings when code approved.


peastman:
I don't think we want to modify how we write PDB files, only how we read them.  This is basically a nonstandard extension to the format that we're defining, but all files we write should conform to the standard.  If a particle doesn't have an element, we should just leave the element column blank (as we do now).

I would also change these lines:

```Python
elif not self.element_symbol and extraParticleIdentifier is None:
    self.element = 'EP'
```

If `extraParticleIdentifier` is `None`, I would expect that to mean that nothing should be marked as an extra particle.  If you want an empty element column to be interpreted as an extra particle, you can already say `extraParticleIdentifier=''`.

Two things are still missing: documentation and testing.  Add the following to the docstring for the `PDBFile` constructor:

```
extraParticleIdentifier : string='EP'
    if this value appears in the element column for an ATOM record, the Atom's element will be set to None to mark it as an extra particle
```

Then you should add a new test case to TestPdbFile.py to verify this feature works correctly.

rafwiewiora:
> I don't think we want to modify how we write PDB files, only how we read them. This is basically a nonstandard extension to the format that we're defining, but all files we write should conform to the standard. If a particle doesn't have an element, we should just leave the element column blank (as we do now).

Good point, I totally missed conforming to the standard. But is it still useful to have `extraParticleIdentifier=''` for writing? If I've loaded in a file with EP for extra particle elements I probably care about those staying that way when I write it out, so should have an option to do that? 

> If extraParticleIdentifier is None, I would expect that to mean that nothing should be marked as an extra particle. If you want an empty element column to be interpreted as an extra particle, you can already say extraParticleIdentifier=''.

Alright. But keep the None case in the meaning you described or get rid of it? Will anyone have a file with 'EP' in elements and want to load that in without that meaning an extra particle? Maybe..

Working on the documentation and testing soon.

peastman:
> But keep the None case in the meaning you described or get rid of it?

That's what it will already do.  If you specify `extraParticleIdentifier=None` then nothing will match it, so PDBFile will try to select an element for anything it possibly can (as it does now).

rafwiewiora:
Ah that's true, alright!

rafwiewiora:
> Good point, I totally missed conforming to the standard. But is it still useful to have extraParticleIdentifier='' for writing? If I've loaded in a file with EP for extra particle elements I probably care about those staying that way when I write it out, so should have an option to do that?

What about this though?

peastman:
Sure, that seems reasonable.

rafwiewiora:
All fixed / added here. 

For testing I've added tip5p.pdb, but changed the names of the virtual sites from M1/2 to D1/2 (as in Dummy), because M would result in None element even if this EP code wasn't working ok, while D gives deuterium. 

Looks good?

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
```diff
@@ -123,7 +123,7 @@ class PdbStructure(object):
     """
 
 
-    def __init__(self, input_stream, load_all_models=False):
+    def __init__(self, input_stream, load_all_models=False, extraParticleIdentifier='EP'):
```
jchodera: Maybe `extraParticleElementName`, to make it clearer that it will appear in the `element` field of the ATOM record?

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
```diff
@@ -795,11 +798,14 @@ def __init__(self, pdb_line, pdbstructure=None):
         try: self.formal_charge = int(pdb_line[78:80])
         except ValueError: self.formal_charge = None
         # figure out atom element
-        try:
-            # Try to find a sensible element symbol from columns 76-77
-            self.element = element.get_by_symbol(self.element_symbol)
-        except KeyError:
-            self.element = None
+        if self.element_symbol == extraParticleIdentifier:
+            self.element = 'EP'
```
jchodera: I thought we wanted to set `element = None` in this case, since `atom.element` is either an `Element` object or `None` (indicating a dummy particle).  Allowing it to be a string too seems weird.  Or am I misunderstanding how downstream processing works?

wrappers/python/simtk/openmm/app/pdbfile.py
```diff
@@ -237,7 +241,7 @@ def _parseResidueAtoms(residue, map):
                 map[atom.attrib[id]] = name
 
     @staticmethod
-    def writeFile(topology, positions, file=sys.stdout, keepIds=False):
+    def writeFile(topology, positions, file=sys.stdout, keepIds=False, extraParticleIdentifier=' '):
```
jchodera: Shouldn't the defaults for reading/writing be the same?

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
rafwiewiora: Hmm, both sound ok to me, will leave to you and @peastman to decide.

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
rafwiewiora: The way I did it allows for `PDBStructure` to separately return 'extra particle' and 'atom for which element could not be determined by `self.element = element.get_by_symbol(self.element_symbol)`, and these are passed respectively as `element = 'EP'` and `element = None`. 

`PDBFile` then picks this up and applies some extra rules to try to work out elements: https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/pdbfile.py#L119

For doing what you're proposing, you'd need to change the meaning of `None` at `PDBFile` from "unknown element - do some more working out" to "extra particle", so we can't apply these extra rules in `PDBFile` and we're getting element `None` returned in topology by `PDBFile` where elements used to be worked out by this code. 

On a practical note for what this EP code is meant to add: think e.g. loading a PDB file that has both 'EP' elements and some blank / all other blank element fields - user marks EPs but hopes the other atom element's will be determined by code. 

We can add `EP' as an `Element` object if not comfortable with this being a string in `PDBStructure`?

wrappers/python/simtk/openmm/app/pdbfile.py
rafwiewiora: Check out the discussion above where @peastman points out we need to write out PDB files according to the PDB specification - i.e. `None` element needs to be written out as blank. If you want to write your own-style PDB file that does not conform to specification, you'll have to set this to 'EP' yourself.

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
peastman: PdbStructure is an internal class, so it isn't a huge problem if its API is slightly clunky.  Though possibly we should just move the element guessing code into PdbStructure?

rafwiewiora: Yeah, I like the moving everything to PdbStructure idea. Are we agreed on that? I can fix later today.

peastman: It's ok with me.

rafwiewiora: Ok great.

rafwiewiora: I'm not sure this is worthwhile anymore - because all the name replacement code (https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/pdbfile.py#L97 and so on) will also need to move to PDBStructure. Worth doing this and taking the risk of breaking something or should we just leave it as is?

peastman: If you think it's safer to leave things as they are, that's ok with me.  Like I said, it's ok for an internal API to be slightly clunky.

rafwiewiora: Yeah, I think it's good as it is. So if everything looks good you're good to merge.

rafwiewiora:
Just a reminder for @jchodera and @swails to maybe have a final look through this before merging, as you were the authors of the idea in #1193, some discussion to have a look at in the line notes.

jchodera:
Just go for it.  We'll fix it if it doesn't work, but we can't really test it thoroughly until it's incorporated.

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
peastman: The docstring on PdbStructure still claims it will set the element to None rather than 'EP'.

rafwiewiora:
Alright!

wrappers/python/simtk/openmm/app/internal/pdbstructure.py
rafwiewiora: Fixed!

PR #1310, opened 2015-12-23
Change functions in `customgbforces.py` to subclasses

swails:
This is part of making these classes a little more consistent and isolating the support functionality and parameter definition specific to these models inside these models themselves.  Right now, there is GB-model-specific code scattered in several places that can make it difficult to maintain.

The main thing that needs to happen now is for the radius assignment (that is currently in `charmmpsffile.py`) to be moved to `customgbforces.py`, and possibly correcting the handling in `amber_file_parser.py` to use the "correct" radii rather than the radii in the prmtop file.  That last one requires some discussion, probably.

That can be done either in another PR or added to this one.

### DIFF
--- a/wrappers/python/simtk/openmm/app/charmmpsffile.py
+++ b/wrappers/python/simtk/openmm/app/charmmpsffile.py
@@ -47,7 +47,7 @@ import simtk.unit as u
 from simtk.openmm.app import (forcefield as ff, Topology, element)
 from simtk.openmm.app.amberprmtopfile import HCT, OBC1, OBC2, GBn, GBn2
 from simtk.openmm.app.internal.customgbforces import (GBSAHCTForce,
-                GBSAOBC1Force, GBSAOBC2Force, GBSAGBnForce, GBSAGBn2Force, convertParameters)
+                GBSAOBC1Force, GBSAOBC2Force, GBSAGBnForce, GBSAGBn2Force)
 from simtk.openmm.app.internal.unitcell import computePeriodicBoxVectors
 # CHARMM imports
 from simtk.openmm.app.internal.charmm.topologyobjects import (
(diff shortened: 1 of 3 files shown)


jchodera:
I think creating subclasses is great, but I am still very uncomfortable with either (1) having the parameters defined inside the classes, and (2) overriding radii in the `prmtop` file.  Maybe these could just be optional  behaviors?

Separately, I'm working on trying to use the `<script>` functionality to assign GB parameters.  This could of course be adapted to use your classes.

swails:
These changes are not targeted to the ffxml files, they are targeted to making these models available to CharmmPsfFile, AmberPrmtopFile, and GromacsTopologyFile if desired.  Currently, the parameter assignment is implemented in CharmmPsfFile (and would need to either be imported from there or implemented in GromacsTopologyFile to support the GBSA parameters through those files).

Right now [the parameters ARE defined and hard-coded in the source code](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py#L669-L749).  That code would be *much* better off being directly in the subclasses.  It would be much easier to plug future parsers/system creators into the GB models in the future (GromacsTopologyFile, Desmond files, etc.) if all of the functionality was encapsulated where the model itself is defined and would eliminate the need to duplicate code.

> Separately, I'm working on trying to use the <script> functionality to assign GB parameters. This could of course be adapted to use your classes.

Sure, but that's for ffxml, which is irrelevant to the other parsers.  The upside is that you'd get to test this system out for the ffxml route and see if you like it.  If not, implement something better.  It's still what I think we should do for the other parsers.

jchodera:
> Right now the parameters ARE defined and hard-coded in the source code. That code would be much better off being directly in the subclasses.

Whoa.  You are totally, totally right sir!  Please carry on!  :)

PR #1311, opened 2015-12-31
Move all of the logic for GB parameter assignment into the new subclasses

swails:
Previously, parameters were defined in multiple places -- amber_file_parser.py and charmmpsffile.py.

### DIFF
--- a/wrappers/python/simtk/openmm/app/internal/amber_file_parser.py
+++ b/wrappers/python/simtk/openmm/app/internal/amber_file_parser.py
@@ -34,8 +34,7 @@ DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
 OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
-from __future__ import absolute_import
-from __future__ import print_function
+from __future__ import absolute_import, print_function
 
 #=============================================================================================
 # GLOBAL IMPORTS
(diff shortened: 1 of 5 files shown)


jchodera:
Just curious what the status of this is, @swails.

swails:
The tests failed after the initial revision so I tabled it.  I've been meaning to come back to it, but other priorities keep popping up.

swails:
OK, I think I fixed the outstanding issues here.  When Travis finishes, this should be good to merge.

wrappers/python/simtk/openmm/app/internal/amber_file_parser.py
```diff
@@ -1070,7 +987,29 @@ def readAmberSystem(topology, prmtop_filename=None, prmtop_loader=None, shake=No
...
+        else:
+            screen = [gb_parm[1] for gb_parm in gb_parms]
+        radii = [float(r)/10 for r in prmtop._raw_data['RADII']]
+        warned = False
+        for i, (r, s) in enumerate(zip(radii, screen)):
+            if abs(r - gb_parms[i][0]) > 1e-4 or abs(s - gb_parms[i][1]) > 1e-4:
+                if not warned:
+                    warnings.warn('Non-optimal GB parameters detected for GB '
+                                  'model %s' % gbmodel)
+                    warned = True
```
peastman: What is this warning for?  In what situations would it be issued?  What does "non-optimal" mean?

swails: Each GB model has associated with it a "suggested" set of intrinsic radii for which the parameters were derived.  Each of the `CustomAmberGBForce` objects know what these parameters are.  However, the Amber prmtop file stores these parameters (which are assigned by LEaP).

If these disagree, that means that the user did not specify the *correct* radii that were used in the parametrization inside tleap.

I forget which issue this was discussed, but @jchodera raised concerns about always using the *correct* GB radii for Amber topologies, since that's a marked change in behavior of what used to be done.  I would personally rather always use the suggested parameters, but in the interest of maintaining backwards compatibility and not having results change from version to version, I kept the "old" behavior and was verbose with a warning if the parameters differed from "ideal" in the prmtop for the requested GB model.

swails:
AFAICT, the failures here have nothing to do with the changes (OpenCL failures on Travis and who knows what appveyor is complaining about now...)

wrappers/python/simtk/openmm/app/internal/amber_file_parser.py
peastman: So a user would get this message if, when they created their prmtop file, they somehow specified incorrect radii for the GB model they're using?

swails: Correct.

peastman: Ok.  I just wanted to make sure I understood what was going on.

Issue #488, opened 2014-06-20
Extending ForceField with hooks for parameterizing residues without associated parameters


jchodera:
It would be useful to extend the `ForceField` class in the `app` layer to allow it to try to call out to registered "hooks" when it encounters residues that do not appear in any loaded forcefield.

For example, suppose a ligand is bound to a protein in a PDB file, or cofactors are present, and the user loads in this via `PDBFile` and passes the topology to `ForceField`.  It would be very useful if we could, for example, register `gaff2xml` via these hooks and have `ForceField` call out to `gaff2xml` to generate an ffxml file for any residues that it encounters without existing parameters.  This would initially be used only for small molecules, but could eventually be used to assign parameters to polymeric residues without parameters as well.  Future versions could use [ParamChem](https://www.paramchem.org/), [ATB](http://compbio.biosci.uq.edu.au/atb/), or other parameterization schemes to come.

We would need to standardize an interface for this.  Suggestions?

jchodera:
Thanks to these additions, it should be pretty straightforward to add functionality to register some handlers to generate templates for missing small molecules or polymeric residues.

In re-reading this thread, I think I now disagree that `createSystem` should not be able to add the additional residue templates to the `ForceField` object so that they are available the next time `createSystem` is called.  The reasons are multifold:
* Residue parameterization is an expensive task, and incurring the cost of parameterizing the *same* residue when calling `createSystem()` multiple times is not efficient.  Adding the residue definition to `ForceField` after handlers have explicitly been registered can be considered a form of memoization---a common form of optimization by caching the results of computations. 
* Some handlers may end up producing stochastic differences in parameters produced for a given residue due to the use of numerical optimizers or stochastic conformation generation schemes (e.g. with charges). I'm presuming users don't want to call `createSystem()` multiple times on the same `ForceField` object and get different parameters each time even if the underlying scheme is not fully deterministic.
* We may want to somehow serialize `ForceField` (presumably by writing out the resulting `ffxml` file) after generating lots of parameters to save time next time (e.g. if we have parameterized a small molecule library).   

At first, I thought perhaps adding an *option* as to whether calling `createSystem()` should modify `ForceField` by adding the residue definitions, but I think that it's safe to say that this behavior is the most desirable if any residue parameterization handlers have been added.

rmcgibbo:
Are we making a square-peg-round-whole mistake in trying to force all of this small molecule parameterization stuff into the existing OpenMM `ForceField` and `createSystem` functions?

[...]

PR #1312, opened 2016-01-01
Add hooks for parameterizing unmatched residues in ForceField

jchodera:
This is the beginning of an implementation of pandegroup/openmm#488 (also related to pandegroup/openmm#1298).

I added a `registerTemplateGenerator` method to allow external residue parameterization engines (here called "template generators") to be registered. These generators are called when a residue is found in the topology object passed to `ForceField.createSystem()` that has no parameters in the current forcefield.

Here's the docstring:
```
Register a residue template generator that can be used to parameterize residues that do not match existing forcefield templates.
This functionality can be used to add handlers to parameterize small molecules or unnatural/modified residues.
Parameters
----------
template_generator : function
    A function that will be called when a residue is encountered that does not match an existing forcefield template.

When a residue without a template is encountered, the `template_generator` function is called with:

  success = template_generator(forcefield, residue)

where `residue` is a simtk.openmm.app.topology.Residue object.

`template_generator` should either register a residue template directly with `forcefield.registerResidueTemplate(template)`
or it should call `forcefield.loadFile(file)` to load residue definitions from an ffxml file.
Residue template generators are called in the order in which they were added.
The first residue template generator that can provide a template for an unmatched residue has this residue template added to the forcefield.
The residue template generator must return `True` if it could successfully generate a template, and `False` if it could not.
```
Implementing this as a PR within OpenMM because it is easier to test this way.


jchodera:
I'm presuming travis errors like [this one](https://travis-ci.org/pandegroup/openmm/jobs/99733057#L139) are not actual errors.
```
E: Version '2:8.960-0ubuntu1' for 'fglrx' was not found
```

jchodera:
The travis error seems to be due to the fact that there is [no `fglrx` package found](https://travis-ci.org/pandegroup/openmm/jobs/99746962#L138-L139):
```
E: Package 'fglrx' has no installation candidate
```
I wonder if `fglrx` was removed from the [list of whitelisted packages](https://github.com/travis-ci/apt-package-whitelist).

jchodera:
Some observations so far:
* The programmatic API for adding residue templates is very nice! 
* If we preload `gaff.xml` atom type and parameter definitions, I think this can work for parameterizing molecules with GAFF with very little modification.
* There is a programmatic API for adding atom types, but we don't have a simple way of querying whether atom types already exist or if we have already defined some parameters.  For example, we may want to load/define GAFF atom types and force parameters from within the residue template generator, or we may need to generate new force parameters for bonds/angles/torsions connecting a newly-parameterized residue to other residues with existing parameters.  In both of these cases, we may need to know if we have already defined a parameter.  It's possible we could use `try..except` for this, but we would want to make sure that we throw a special kind of exception that says the parameter has already been defined so we don't catch exceptions too broadly.

jchodera:
The `ForceField` and `Topology` additions pass tests now.

There are new, unrelated errors with [some OpenCL tests](https://travis-ci.org/pandegroup/openmm/jobs/99882075#L1067-L1072) that have appeared because of the change in travis-ci infrastructure:
```
        Start  76: TestOpenCLCustomManyParticleForceSingle
 76/134 Test  #76: TestOpenCLCustomManyParticleForceSingle .........***Failed   20.54 sec
        Start  77: TestOpenCLCustomManyParticleForceMixed
 77/134 Test  #77: TestOpenCLCustomManyParticleForceMixed ..........***Exception: SegFault 21.00 sec
        Start  78: TestOpenCLCustomManyParticleForceDouble
 78/134 Test  #78: TestOpenCLCustomManyParticleForceDouble .........***Failed   21.17 sec
```
There may actually be something to these errors.  We'll see what the test reruns say.

jchodera:
There is one test condition which is timing out (and experiencing those unrelated OpenCL failures).

I should note that I did some very minor refactoring of some parts of the `ForceField` internal API here, and added some methods to `Topology.Residue` to retrieve bonds associated with the residue.

See #488 for extended discussion of the main feature.

jchodera:
I've also added some documentation to the User Guide a CAUTION block that the feature is experimental and the API is subject to change.

jchodera:
Just updated `.travis.yml` with the version that @rmcgibbo fixed in #1317

jchodera:
OK, that one failure is the unrelated segfault from #1319 
```
Rerunning failing tests...
1/1 Test #77: TestOpenCLCustomManyParticleForceMixed ...***Exception: SegFault 24.77 sec
```

jchodera:
Any thoughts/feedback on this?

peastman:
> we don't have a simple way of querying whether atom types already exist

How about `'XYZ' in forcefield._atomTypes`?

> or if we have already defined some parameters.

That's a more complicated question, since parameters are defined separately by each force generator, or possibly by a script.  But is that really an issue?  If the types already exist, it seems reasonable to expect that your TemplateGenerator isn't responsible for defining parameters for those types.

This brings up a conceptual question though.  Is this really just a "template generator", that is, an object whose sole job is to create templates?  Or is it a much more generic "force field generator", something that can (and is expected to) add arbitrary features to the force field (atom types, force parameters, possibly even new forces)?

jchodera:
> How about `'XYZ' in forcefield._atomTypes`?

Great!  But that's a private interface.  Can we make an experimental draft public interface for this?  Or make `atomTypes` public?  The template generators are supposed to be accessible through---and use---the public API (though it can still be experimental at this stage).

> That's a more complicated question, since parameters are defined separately by each force generator, or possibly by a script. But is that really an issue? If the types already exist, it seems reasonable to expect that your TemplateGenerator isn't responsible for defining parameters for those types.

That's not quite true, unfortunately.  The parameterization with GAFF/AM1-BCC by `antechamber` is a two step process:
* `antechamber` is run to assign GAFF atom types and charges
* `parmchk` or `parmchk2` is run to see if there are parameters that would be required but are missing from `gaff.dat`; any missing parameters that would be needed are generated by analogy

One could argue that a more expressive type matching system *should* have been used to eliminate the second step, but to support GAFF/antechamber as it exists now, the residue template generator may also need to add some parameters if they do not exist.

> This brings up a conceptual question though. Is this really just a "template generator", that is, an object whose sole job is to create templates? Or is it a much more generic "force field generator", something that can (and is expected to) add arbitrary features to the force field (atom types, force parameters, possibly even new forces)?

I think we should think of it as a thing that may need to generate additional parameters (or even atom types) in order to parameterize a new residue to create a template.  Its job is to create a template, but in doing its job, it may need to add additional parameters or types.

peastman:
> Great! But that's a private interface. Can we make an experimental draft public interface for this? Or make atomTypes public?

"Private" and "public" aren't clearly defined here, because we actually have three levels of code.  There's the ForceField class itself.  There's extensions that plug into it (force generators, and now template generators as well).  And there's fully external code that just creates a ForceField and uses it to build Systems.  And `_atomTypes` is a field that external code shouldn't access, but it's entirely reasonable for extension code to use it.  If you look through the existing force generators, you'll find lots of cases of them accessing fields and classes whose names begin with `_`.  They basically behave as internal code, not external code.

This is very similar to how plugins work.  We have lots of "internal" APIs that ordinary client code shouldn't access, but plugins can.  Also note that we explicitly do not guarantee backward compatibility for plugins.  They may need updating to work with new releases.  For example, this PR changes what type of values are stored in `_atomTypes`.  If we made it public, we couldn't do that!  But as long as it's an "internal" API, we reserve the right to change it.  Plugins can still access it, but with the knowledge that it might change in the future.

jchodera:
It seems to me there is some value in trying to define a "semi-public" API that could at least make sure these user-defined generators would still work even if there is internal refactoring of how internal data is stored.  Otherwise, we are tied down more than we should be in terms of avoiding even simple refactoring because we are worried about breaking fragile user code.

You've already done nearly all the work here, with making `registerAtomType()` public as part of exposing a programmatic type/parameter definition API.  Adding a method to query types is a simple extension of this, and maybe adding public methods for defining parameters isn't conceptually beyond this.

We certainly don't have to do that now, but maybe something to thing about in a separate PR?

jchodera:
Anything else to be done here?  Or can we merge this and continue the discussion of further refinements to support additional parameter definitions?

peastman:
I'm ok with moving ahead on it.

jchodera:
Thanks!

Issue #1313, opened 2016-01-02
Travis tests failing to due inability to install fglrx package


jchodera:
Noticed this in #1312:
```
E: Version '2:8.960-0ubuntu1' for 'fglrx' was not found
```
Just noted a warning link at the top of the travis-ci pages about implementing a new "legacy `precise`" infrastructure:
https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future

PR #1314, opened 2016-01-02
Fix travis following migration to legacy precise infrastructure.

jchodera:
Addresses #1313.

A change in the way travis-ci handled the Legacy Precise infrastructure meant that restricted ubuntu packages (such as `fglrx`) were not available unless the `restricted` repository was explicitly added.


rmcgibbo:
Great!

Issue #1315, opened 2016-01-02
Travis build timing out


jchodera:
Now, after the fix from #1314, one of the travis builds is timing out due to no output received:
https://travis-ci.org/pandegroup/openmm/builds/99777387

Specifically, [`TestOpenCLAndersenThermostatSingle`](https://travis-ci.org/pandegroup/openmm/jobs/99777388#L1088-L1094) seems to be timing out:
```
 93/134 Test  #93: TestOpenCLCustomIntegratorDouble ................   Passed  141.44 sec
        Start  94: TestOpenCLAndersenThermostatSingle
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

jchodera:
From reading [this blog post](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future) in more detail, the best option might be to have @peastman (or whoever's account is running Travis) email travis-ci to opt in to the older Blue Box system.

rmcgibbo:
I think you can tell ctest about the timeout so it'll skip the test if it goes for too long, no?

jchodera:
Good point. Maybe this flag?
>* --timeout <seconds>*: Set a global timeout on all tests.
> This option will set a global timeout on all tests that do not already
have a timeout set on them.

https://cmake.org/cmake/help/v2.8.8/ctest.html

jchodera:
There seems to be an issue with [the whole set of tests timing out](https://travis-ci.org/pandegroup/openmm/jobs/99882075#L1118) for OpenCL on Linux:
```
The job exceeded the maxmimum time limit for jobs, and has been terminated.
```
I thought the tests were supposed to be limited to 25 min of wall clock time, so I'm a bit confused as to how the whole job is timing out.

rmcgibbo:
I wouldn't worry about it too much. Might need to play around with some of the settings like the timeout, or the number of tests to run in parallel (`-j`), etc.

jchodera:
Taking a quick stab at it in #1316, but may need your help, @rmcgibbo.

I do wonder if [those OpenCL failures](https://github.com/pandegroup/openmm/pull/1312#issuecomment-168438851) could possibly be real.

jchodera:
Seems like we're good again.

PR #1316, opened 2016-01-03
[WIP] Attempt to fix travis timeouts

jchodera:
Fiddling with a fix for #1315.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -146,8 +146,10 @@ script:
 
   - # run all of the tests, making sure failures at this stage don't cause travis failures
   - # set a timeout of $CTEST_TIMEOUT seconds for each individual test
+  - echo `date`
+  - echo $CTEST_STOP_TIME
   - CTEST_TIMEOUT=180
-  - python devtools/run-ctest.py -j2 --schedule-random --stop-time $CTEST_STOP_TIME --timeout $CTEST_TIMEOUT --output-on-failure
+  - python devtools/run-ctest.py -j1 --schedule-random --stop-time $CTEST_STOP_TIME --timeout $CTEST_TIMEOUT --output-on-failure
   - if [[ ! -z "${DOCS_DEPLOY}" && "${DOCS_DEPLOY}" = "true" ]]; then
       pip install sphinx sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen;
       make C++ApiDocs PythonApiDocs;
--- a/.travis.yml
+++ b/.travis.yml
@@ -146,6 +146,7 @@ script:
 
   - # run all of the tests, making sure failures at this stage don't cause travis failures
   - # set a timeout of $CTEST_TIMEOUT seconds for each individual test
+  - CTEST_TIMEOUT=180
   - python devtools/run-ctest.py -j2 --schedule-random --stop-time $CTEST_STOP_TIME --timeout $CTEST_TIMEOUT --output-on-failure
   - if [[ ! -z "${DOCS_DEPLOY}" && "${DOCS_DEPLOY}" = "true" ]]; then
       pip install sphinx sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen;
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,7 +130,7 @@ before_install:
     fi
 
 script:
-  - CTEST_STOP_TIME=$(python -c "from datetime import datetime, timedelta; import sys; sys.stdout.write((datetime.now() + timedelta(minutes=25)).strftime('%H:%M:%S'))")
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Might also help to add `--output-on-failure` (https://cmake.org/cmake/help/v2.8.8/ctest.html#opt:--output-on-failure)

.travis.yml
```diff
@@ -145,7 +145,8 @@ script:
     fi
 
   - # run all of the tests, making sure failures at this stage don't cause travis failures
-  - python devtools/run-ctest.py -j2 --schedule-random --stop-time $CTEST_STOP_TIME
+  - # set a timeout of $CTEST_TIMEOUT seconds for each individual test
+  - python devtools/run-ctest.py -j2 --schedule-random --stop-time $CTEST_STOP_TIME --timeout $CTEST_TIMEOUT
   - if [[ ! -z "${DOCS_DEPLOY}" && "${DOCS_DEPLOY}" = "true" ]]; then
```
rmcgibbo: It doesn't look like like the environment variable CTEST_TIMEOUT was ever set though?

jchodera: Whoops!

jchodera:
Either `ctest` is ignoring the command-line arguments, or it isn't actually being passed them.

rmcgibbo:
There could be something wrong with devtools/run-ctest.py, but I looked and didn't see anything.

jchodera:
Well, I can only conclude that `ctest` is completely broken, since it doesn't seem to be honoring any of its command-line arguments.

rmcgibbo:
Or `devtools/run-ctest.py`?

jchodera:
The argument array is definitely being built. I have it being output to the terminal.

jchodera:
Oh.  I think `subprocess.call` should *either* be getting an argument list *or* be getting a single command along with the `shell=True` argument, but not both.

See http://stackoverflow.com/questions/16645761/subprocess-call-doesnt-work-as-expected

jchodera:
That was it.  Hopefully can finish this up soon.

rmcgibbo:
I think the second invocation of ctest from the python script is supposed to not receive the `--stop-time` argument.

jchodera:
Well, what's our goal here?  Do we want to finish whatever tests we can by the time travis allots us runs out, or do we want to run into the travis limit by letting those test reruns run without limits?

rmcgibbo:
We'd like to finish before the job times out. If you pass the same 25 minute timeout to both invocations, then the 1st invocation will get killed after running for 25 minutes and the second invocation will die immediately, because 25 minutes will already have elapsed.

rmcgibbo:
Let me just rewrite it.

rmcgibbo:
Is `-j4` really the best here?

jchodera:
> We'd like to finish before the job times out. If you pass the same 25 minute timeout to both invocations, then the 1st invocation will get killed after running for 25 minutes and the second invocation will die immediately, because 25 minutes will already have elapsed.

Yes, but how do we apportion the available runtime?  Should we timeout individual tests more quickly?

Really, we should probably set the max test run time at the very beginning when travis starts and make sure travis won't time out on us.  Then we could just complete as many tests as we have time for in that time, keeping the per-test timeout to be pretty short.

jchodera:
> Let me just rewrite it.

Go for it!

> Is -j4 really the best here?

I was just trying other things.  It says there is ~2 cores available for this class of VM.

Compilation only takes ~3 min, though.

jchodera:
More importantly, we only want to rerun tests that did not time out.  Is there an easy way to do that?

rmcgibbo:
Oh, good idea. Not sure. I think ctest can generate xUnit output, which might contain it...

jchodera:
It definitely prints this information to the terminal:
```
The following tests FAILED:
	 71 - TestOpenCLEwaldMixed (Timeout)
	 72 - TestOpenCLEwaldDouble (Timeout)
	 77 - TestOpenCLCustomManyParticleForceMixed (SEGFAULT)
	 78 - TestOpenCLCustomManyParticleForceDouble (SEGFAULT)
	100 - TestOpenCLMonteCarloBarostatSingle (Timeout)
	101 - TestOpenCLMonteCarloBarostatMixed (Timeout)
	117 - TestOpenCLRpmdDouble (Timeout)
	118 - TestOpenCLDrudeSCFIntegratorSingle (Timeout)
	119 - TestOpenCLDrudeSCFIntegratorMixed (Timeout)
	124 - TestOpenCLDrudeLangevinIntegratorSingle (Timeout)
	125 - TestOpenCLDrudeLangevinIntegratorMixed (Timeout)
	126 - TestOpenCLDrudeLangevinIntegratorDouble (Timeout)
```

rmcgibbo:
http://stackoverflow.com/a/21688776/1079728 says that you can get xml output.

jchodera:
Closing this in favor of #1317

PR #1317, opened 2016-01-04
Rewrite run-ctest.py

rmcgibbo:
Fixes #1316

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,6 +93,7 @@ matrix:
            CMAKE_FLAGS=""
 
 before_install:
+  - START_TIME=$(date +%s)
   - wget https://anaconda.org/omnia/ccache/3.2.4/download/${TRAVIS_OS_NAME}-64/ccache-3.2.4-0.tar.bz2
   - mkdir -p $HOME/ccache && tar xf ccache-3.2.4-0.tar.bz2 -C $HOME/ccache
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
@@ -130,7 +131,6 @@ before_install:
     fi
 
 script:
-  - CTEST_STOP_TIME=$(python -c "from datetime import datetime, timedelta; import sys; sys.stdout.write((datetime.now() + timedelta(minutes=25)).strftime('%H:%M:%S'))")
   - cmake . $CMAKE_FLAGS -DCMAKE_INSTALL_PREFIX=$HOME/OpenMM
   - make -j2 install
   - if [[ "$OPENCL" == "true" ]]; then ./TestOpenCLDeviceQuery; fi
@@ -144,8 +144,9 @@ script:
       (cd python/tests && py.test -v);
     fi
 
-  - # run all of the tests, making sure failures at this stage don't cause travis failures
-  - python devtools/run-ctest.py -j2 --schedule-random --stop-time $CTEST_STOP_TIME
+  # Run the tests, and rerun any failing tests.
+  - python devtools/run-ctest.py --start-time $START_TIME
+
   - if [[ ! -z "${DOCS_DEPLOY}" && "${DOCS_DEPLOY}" = "true" ]]; then
       pip install sphinx sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen;
       make C++ApiDocs PythonApiDocs;
--- a/devtools/run-ctest.py
+++ b/devtools/run-ctest.py
@@ -1,29 +1,105 @@
+"""
+Run test suite through CTest, with some options set for the CI environment.
+
+- Runs with a per-test and overall timeout which can be governed by the
+  avialable time on the CI system.
+- Reruns tests which fail (does not rerun tests which merely timeout).
+
+"""
 from __future__ import print_function
 import sys
+import os.path
+import shutil
+import time
+from glob import glob
 from os.path import join, exists
 from subprocess import call
+from argparse import ArgumentParser
+from datetime import datetime, timedelta
+from xml.etree import ElementTree
+
 
 def main():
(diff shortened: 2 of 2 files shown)


rmcgibbo:
lol python2 is terrible. you can't multiply a timedelta and a float.

jchodera:
Relevant: https://www.destroyallsoftware.com/talks/wat

rmcgibbo:
@jchodera one of the errors is an OpenCL segfault, so it's probably something. I thought we'd previously ironed out all of the OpenCL CPU platform segfaults, but maybe not.

```
    Start 76: TestOpenCLCustomManyParticleForceSingle
1/1 Test #76: TestOpenCLCustomManyParticleForceSingle ...***Exception: SegFault 22.26 sec
```

jchodera:
The the segfaults *are* real!

jchodera:
How did you make the segfaults go away?  Did you just limit the overall test runtime to be very short, and then retrigger travis a few times?  

Maybe we should let the tests run for a bit longer?  We're only using 23 / 50 travis minutes right now, I think...

jchodera:
Or maybe we should split up the OpenCL tests into multiple parallel travis instances?
https://docs.travis-ci.com/user/speeding-up-the-build/#Parallelizing-your-builds-across-virtual-machines

rmcgibbo:
> How did you make the segfaults go away?

Just cause it's stochastic, and I rebased and forced push to clean up the git history before merging.

Obviously there's a tension between having the tests be exhaustive and complete quickly. I think we get pretty good coverage with the current settings, in that we're aware of the failures pretty quickly.

rmcgibbo:
The most useful thing to actually fix or investigate the segfault would be to get a stacktrace.

jchodera:
> Just cause it's stochastic, and I rebased and forced push to clean up the git history before merging.

Actually, it doesn't look like the failure is stochastic, but the order in which tests are run is stochastic due to the `--schedule-random`.  The test that previously segfaulted (`TestOpenCLCustomManyParticleForceSingle`) wasn't even run this time.

rmcgibbo:
Yes, that's what I meant by "it".

jchodera:
> Obviously there's a tension between having the tests be exhaustive and complete quickly. I think we get pretty good coverage with the current settings, in that we're aware of the failures pretty quickly.

Could we use the `-I start,end,stride` argument and a couple of extra build matrix entries to split up the tests into, say, 4 groups that run in parallel?
https://cmake.org/cmake/help/v2.8.8/ctest.html#opt:-IStartEndStridetesttestTestfile--tests-information

rmcgibbo:
We hashed out the pros and cons of testing a subset random order in #1212, and I discussed it with peter before that was merged. I'd prefer not to beat a dead horse. Splitting up the OpenCL with multiple jobs is also an option, but it's more complex, and isn't worth it, IMO.

jchodera:
> The most useful thing to actually fix or investigate the segfault would be to get a stacktrace.

Now that there is a [more unified test infrastructure](https://github.com/pandegroup/openmm/tree/master/tests), we could have the tests [install a `SIGSEGV` handler](http://stackoverflow.com/questions/77005/how-to-generate-a-stacktrace-when-my-gcc-c-app-crashes) that will print a useful stacktrace if they segfault.

Right now, the tests all have their own `main()`, but we could presumably add a single header file that, for `unix/osx`, defines this handler and a corresponding `#define` that could be included at the beginning of each `main()`:
```C++
#include <stdio.h>
#include <execinfo.h>
#include <signal.h>
#include <stdlib.h>
#include <unistd.h>


void handler(int sig) {
  void *array[10];
  size_t size;

  // get void*'s for all entries on the stack
  size = backtrace(array, 10);

  // print out all the frames to stderr
  fprintf(stderr, "Error: signal %d:\n", sig);
  backtrace_symbols_fd(array, size, STDERR_FILENO);
  exit(1);
}
```
and does nothing for `win`.

rmcgibbo:
I feel like this is something ctest should be able to do for you, but maybe not.

jchodera:
> We hashed out the pros and cons of testing a subset random order in #1212, and I discussed it with peter before that was merged. I'd prefer not to beat a dead horse.

Oh, I have no problem with the random order testing if we're only testing everything in a single travis instance.

> Splitting up the OpenCL with multiple jobs is also an option, but it's more complex, and isn't worth it, IMO.

That's what I was suggesting, if there was a simple way to do it with the `-I` argument to CTest.  We can definitely explore that in a separate PR, though.

rmcgibbo:
I thought it might help to get the exact same OpenCL CPU driver version, but I can't reproduce the test failure :(

```
root@33af72743cee:/openmm/build# ./TestOpenCLDeviceQuery
OpenCL devices:

OpenCLPlatformIndex 0, OpenCLDeviceIndex 0: "Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz"
    CL_DEVICE_VENDOR                 = GenuineIntel
    CL_DEVICE_MAX_COMPUTE_UNITS      = 8
    CL_DEVICE_GLOBAL_MEM_SIZE        = 12558561280
    CL_DEVICE_MAX_MEM_ALLOC_SIZE     = 3139640320
    CL_DEVICE_MAX_CLOCK_FREQUENCY    = 2451
    estimatedSpeed                   = 19608

root@33af72743cee:/openmm/build# ./TestOpenCLCustomManyParticleForce single
Done
```

jchodera:
This should be ready to merge, yes?

jchodera:
Also might be good to open a separate issue thread about the OpenCL segfaults.

rmcgibbo:
Sure, yeah. This is ready to merge.

Issue #1292, opened 2015-12-12
Compiling 7.0 using CUDA 7.5 


msultan:
I have been trying to get the master branch to compile with CUDA 7.5 but I keep running into the following issue with the wrappers. 

```python
 python -m simtk.testInstallation
Failed to import OpenMM packages; OpenMM will not work.
Make sure OpenMM is installed and the library path is set correctly.

Error message: /home/msultan/software/miniconda/lib/python2.7/site-packages/simtk/openmm/_openmm.so: undefined symbol: _ZTVN6OpenMM23CustomCentroidBondForceE

```
The tests are being passed so I think it compiled alright. 
```bash
make test
Running tests...
Test project /home/msultan/software/openmm/build_openmm
        Start   1: TestReferenceRandom
  1/269 Test   #1: TestReferenceRandom ............................   Passed    2.84 sec
        Start   2: TestReferenceBrownianIntegrator
  2/269 Test   #2: TestReferenceBrownianIntegrator ................   Passed    0.22 sec
        Start   3: TestReferenceCompoundIntegrator
  3/269 Test   #3: TestReferenceCompoundIntegrator ................   Passed    0.05 sec
        Start   4: TestReferenceVariableVerletIntegrator
  4/269 Test   #4: TestReferenceVariableVerletIntegrator ..........   Passed    5.13 sec
        Start   5: TestReferenceCustomHbondForce
  5/269 Test   #5: TestReferenceCustomHbondForce ..................   Passed    0.04 sec

```

I have already tried 
1). deleting the ```wrappers``` folder in ```openmm_build``` and running ```make PythonInstall`` a second time around. 
2). Manually removing the simtk folder from the site-packages and remaking the python wrappers 

Doing the same procedure with 6.3 release using CUDA 7.5 doesn't produce this issue.  Any one have any recommendations on what to do ?

peastman:
I would guess that Python is linking to the OpenMM 6.3 library instead of the master branch version.  Do you in fact have two different copies of OpenMM on your computer?  Is one of them in your library path?

msultan:
Yea, deleting the entire build folder and the python site-packages simtk folder seemed to have fixed it. Thanks @peastman!

PR #1320, opened 2016-01-05
Reorganize linking / install paths

rmcgibbo:
This changes around `setup.py` to copy all of the `libOpenMM.{so,dylib,dll}` libraries into the Python build directory, such that any bdist produced by `setup.py` like bdist_wheel or bdist_winist always contains both all of the python code and all of the compiled C++ code, and there's a fixed relative path between the two.
~~Still WIP.~~


wrappers/python/setup.py
```diff
@@ -5,63 +5,86 @@
...
+                 'OpenMMRPMD',
+                 'OpenMMDrude']
+
+    if 'OPENMM_USE_DEBUG_LIBS' in os.environ:
+        if platform.system() == "Windows":
+            raise Exception("use of OpenMM debug libs not supported on Win OS")
         else:
-            os.remove(path)
-    os.rmdir(dir)
+            sys.stdout.write("WARNING: using debug libs:\n")
```
mpharrigan: why not `print`?

wrappers/python/setup.py
```diff
@@ -5,63 +5,86 @@
...
+            for ii in range(len(libraries)):
```
mpharrigan: why not `for lib in libraries`?

wrappers/python/setup.py
```diff
@@ -5,63 +5,86 @@
...
+    elif platform.system() == 'Darwin':
+        return 'lib%s.dylib' % libName
+    elif platform.system() == 'Linux':
+        return 'lib%s.so' % libName
+    else:
+        raise ValueError('Unknown platform: "%s"' % platform.system())
+
+
+def getLibraries():
+    libraries = ['OpenMM',
```
mpharrigan: won't copying fail if these optional libraries are turned off in cmake?

rmcgibbo: no, cmake won't let you build the python wrappers if those libraries are turned off.

peastman:
Could you describe in more detail what this does and how it would work?  Would this mean users would end up with two different copies of the OpenMM libraries on their computers in two different places?  If so, that could cause a lot of problems.

rmcgibbo:
1. It changes the Python library to be fully self-contained, in the sense that all of the necessary code is bundled together. For instance, on OS X, it creates a directory structure like this. (Note that `libOpenMM.dylib` and all of the plugins are included.) So `_openmm.so` doesn't have any linker dependencies on libraries installed somewhere else.

```
simtk/__init__.py
simtk/testInstallation.py
simtk/openmm/__init__.py
simtk/openmm/_openmm.cpython-35m-darwin.so
simtk/openmm/amd.py
simtk/openmm/libOpenMM.dylib
simtk/openmm/libOpenMMAmoeba.dylib
simtk/openmm/libOpenMMDrude.dylib
simtk/openmm/libOpenMMRPMD.dylib
simtk/openmm/mtsintegrator.py
simtk/openmm/openmm.py
simtk/openmm/vec3.py
simtk/openmm/version.py
simtk/openmm/app/__init__.py
simtk/openmm/app/amberinpcrdfile.py
simtk/openmm/app/amberprmtopfile.py
simtk/openmm/app/charmmcrdfiles.py
simtk/openmm/app/charmmparameterset.py
simtk/openmm/app/charmmpsffile.py
simtk/openmm/app/checkpointreporter.py
simtk/openmm/app/dcdfile.py
simtk/openmm/app/dcdreporter.py
simtk/openmm/app/desmonddmsfile.py
simtk/openmm/app/element.py
simtk/openmm/app/forcefield.py
simtk/openmm/app/gromacsgrofile.py
simtk/openmm/app/gromacstopfile.py
simtk/openmm/app/modeller.py
simtk/openmm/app/pdbfile.py
simtk/openmm/app/pdbreporter.py
simtk/openmm/app/pdbxfile.py
simtk/openmm/app/simulation.py
simtk/openmm/app/statedatareporter.py
simtk/openmm/app/topology.py
simtk/openmm/app/data/absinth.xml
simtk/openmm/app/data/amber03.xml
simtk/openmm/app/data/amber03_obc.xml
simtk/openmm/app/data/amber10.xml
simtk/openmm/app/data/amber10_obc.xml
simtk/openmm/app/data/amber96.xml
simtk/openmm/app/data/amber96_obc.xml
simtk/openmm/app/data/amber99_obc.xml
simtk/openmm/app/data/amber99sb.xml
simtk/openmm/app/data/amber99sbildn.xml
simtk/openmm/app/data/amber99sbnmr.xml
simtk/openmm/app/data/amber99Test.xml
simtk/openmm/app/data/amoeba2009.xml
simtk/openmm/app/data/amoeba2009_gk.xml
simtk/openmm/app/data/amoeba2013.xml
simtk/openmm/app/data/amoeba2013_gk.xml
simtk/openmm/app/data/charmm_polar_2013.xml
simtk/openmm/app/data/hydrogens.xml
simtk/openmm/app/data/iamoeba.xml
simtk/openmm/app/data/pdbNames.xml
simtk/openmm/app/data/residues.xml
simtk/openmm/app/data/spce.pdb
simtk/openmm/app/data/spce.xml
simtk/openmm/app/data/swm4ndp.xml
simtk/openmm/app/data/test.pdb
simtk/openmm/app/data/tip3p.pdb
simtk/openmm/app/data/tip3p.xml
simtk/openmm/app/data/tip3pfb.xml
simtk/openmm/app/data/tip4pew.pdb
simtk/openmm/app/data/tip4pew.xml
simtk/openmm/app/data/tip4pfb.xml
simtk/openmm/app/data/tip5p.pdb
simtk/openmm/app/data/tip5p.xml
simtk/openmm/app/internal/__init__.py
simtk/openmm/app/internal/amber_file_parser.py
simtk/openmm/app/internal/customgbforces.py
simtk/openmm/app/internal/pdbstructure.py
simtk/openmm/app/internal/unitcell.py
simtk/openmm/app/internal/charmm/__init__.py
simtk/openmm/app/internal/charmm/_charmmfile.py
simtk/openmm/app/internal/charmm/exceptions.py
simtk/openmm/app/internal/charmm/topologyobjects.py
simtk/openmm/app/internal/pdbx/__init__.py
simtk/openmm/app/internal/pdbx/reader/__init__.py
simtk/openmm/app/internal/pdbx/reader/PdbxContainers.py
simtk/openmm/app/internal/pdbx/reader/PdbxParser.py
simtk/openmm/app/internal/pdbx/reader/PdbxReader.py
simtk/openmm/app/internal/pdbx/reader/PdbxReaderTests.py
simtk/openmm/app/internal/pdbx/reader/PdbxReadWriteTests.py
simtk/openmm/app/internal/pdbx/writer/__init__.py
simtk/openmm/app/internal/pdbx/writer/PdbxWriter.py
simtk/openmm/app/internal/pdbx/writer/PdbxWriterTests.py
simtk/openmm/plugins/libOpenMMAmoebaReference.dylib
simtk/openmm/plugins/libOpenMMCPU.dylib
simtk/openmm/plugins/libOpenMMDrudeOpenCL.dylib
simtk/openmm/plugins/libOpenMMDrudeReference.dylib
simtk/openmm/plugins/libOpenMMOpenCL.dylib
simtk/openmm/plugins/libOpenMMPME.dylib
simtk/openmm/plugins/libOpenMMRPMDOpenCL.dylib
simtk/openmm/plugins/libOpenMMRPMDReference.dylib
simtk/unit/__init__.py
simtk/unit/basedimension.py
simtk/unit/baseunit.py
simtk/unit/constants.py
simtk/unit/doctests.py
simtk/unit/mymatrix.py
simtk/unit/prefix.py
simtk/unit/quantity.py
simtk/unit/standard_dimensions.py
simtk/unit/unit.py
simtk/unit/unit_definitions.py
simtk/unit/unit_math.py
simtk/unit/unit_operators.py
```

2. If you have two different [conda environments](http://conda.pydata.org/docs/using/envs.html) or [virtualenvs](http://docs.python-guide.org/en/latest/dev/virtualenvs/), you could have different versions of OpenMM installed and they wont conflict -- they won't be competing over who's libraries are supposed to be in `/usr/local/openmm`.

3. It also makes it possible to improve the way we distribute the pre-compiled Python packages, using [wheels](http://pythonwheels.com/). This will solve the Linux C++11ABI issue https://github.com/pandegroup/openmm/issues/1307#issuecomment-165925356, and simplify the process, because no second compilation step is required to install the wheel. It just needs to be unpacked (the wheel is a zip archive that contains all the Python code, all of the libraries, and data files), so `pip install openmm` just unpacks it and copies the tree into your Python `site-packages/` folder.

peastman:
But are we still installing libraries in /usr/local/openmm?  OpenMM isn't just for Python!  And what about documentation, header files, etc.?  Where do they get installed?

swails:
> But are we still installing libraries in /usr/local/openmm? OpenMM isn't just for Python! And what about documentation, header files, etc.? Where do they get installed?

It copies them, so they're still installed in the CMake prefix (look for `shutil.copytree` in the setup.py diffs).  I don't think this is necessarily a bad thing, though, since the libs in the python site-packages/simtk directory are almost certainly not going to be touched by any loader when building against the C++/C/Fortran APIs.

mpharrigan:
I'm +1 on these changes. I know just a couple weeks ago @msultan spent a long time trying to figure out what libraries were being loaded (#1292) that would be solved if everything was in python site_packages

I'm +1 having the libraries installed in two places: the install dir for people who use it as a library and as a python package for people who use it as a python package

rmcgibbo:
> But are we still installing libraries in /usr/local/openmm? OpenMM isn't just for Python! And what about documentation, header files, etc.? Where do they get installed?

This doesn't change anything about the current cmake behavior on `make install`. When you run `make install`, everything gets copied to `CMAKE_INSTALL_PREFIX` as it does now.

If you're compiling from source, the difference is that when your run `make PythonInstall`, the resulting Python extension installed in your site-packages folder won't have a RPATH entry pointing to CMAKE_INSTALL_PREFIX, it'll just have the libraries it needs copied into site-packages. No headers though.

rmcgibbo:
IMO it would be ideal for `make PythonInstall` to not depend on having executed `make install` and be able to copy the libs directly out of the build directory, but I couldn't figure out how to do that technically.

peastman:
I can guarantee you this will lead to problems.  I couldn't count how many user issues I've dealt with that ultimately turned out to be caused by having multiple copies of the libraries in various places on their computers.  There will be people who added /usr/local/openmm/lib to their LD_LIBRARY_PATH, and then Python picked up the wrong one.  And people who "make install" and then get really confused about why Python scripts aren't seeing any of their changes.  There will be people writing C/C++/Fortran programs who also bring up a Python interpreter for simple testing and debugging, not realizing it's using a different version of the libraries.  And people who set `OPENMM_PLUGIN_DIR`, not realizing there are now two different plugin directories in two different places.  And so on.

rmcgibbo:
None of your examples make any sense to me.

>  There will be people who added /usr/local/openmm/lib to their LD_LIBRARY_PATH, and then Python picked up the wrong one.

1) RPATH overrides LD_LIBRARY_PATH, so this can't happen.

> And people who "make install" and then get really confused about why Python scripts aren't seeing any of their changes.

2) `make install` never installed the python package. If you want to install your modified copy of the librar, you'lve always run `make PythonInstall`.
 
> There will be people writing C/C++/Fortran programs who also bring up a Python interpreter for simple testing and debugging, not realizing it's using a different version of the libraries.

3) But to "bring up the python interpreter", they have always run `make PythonInstall`...

In my world, the problem that comes up virtually _every single week_ for me is that multiple versions of OpenMM conflict with one another with errors like #1292 because different versions of OpenMM are fighting over who controls /usr/local/openmm (or similar).

Having a Python package that's "split" across multiple parts of the filesystem is a very poor design. It impossible to manage multiple different versions of OpenMM, because they have these RPATH links scattered across the filesystem. It makes `pip uninstall openmm` never work because the Python package manager only knows about half of the package. It requires crazy fragile Windows hacks like https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/__init__.py#L14-L17.

rmcgibbo:
Also, [our story](http://docs.openmm.org/6.3.0/userguide/application.html#installing-on-windows) about how to install binary distributions of openmm is really *exceptionally* poor compared to every other Python package I've used. Especially on windows. This is my main motivating factor. It requires manually copying files around, manually adding stuff to the PATH (this is very hard and unfamiliar Windows). The UX here is very poor.

The root cause, of course, is that the package is split over multiple places on the filesystem. This is so fragile.

peastman:
> 2) make install never installed the python package. If you want to install your modified copy of the librar, you've always run make PythonInstall.

That's a non-sequitor.  Of course they've installed the Python packages.  And then at some later point they make changes to OpenMM and type "make install".  And then they run a script, and it doesn't see any of their changes.

> 3) But to "bring up the python interpreter", they have always run make PythonInstall...

Yes, see above.

> Also, our story about how to install binary distributions of openmm is really exceptionally poor compared to every other Python package I've used.

OpenMM *is not just a Python package*.  I realize that's the only language you work with it from, but there are a lot of people using the C++, C, and Fortran interfaces.  We need to take into account how this will affect them.

Besides, what you just said simply isn't true.  If you're only interested in Python, our story about how to install it is trivial: "conda install openmm" and you're done.  It doesn't get much simpler than that.

rmcgibbo:
> We need to take into account how this will affect them.

It has no effect.

My basic contention is that: *If you're only a Python programmer, you should be able to interact with OpenMM as if it's just a Python package*. It shouldn't be some special flower just because has other interfaces.

> Besides, what you just said simply isn't true. If you're only interested in Python, our story about how to install it is trivial: "conda install openmm" and you're done. It doesn't get much simpler than that.

What's not true? I was reading from the user manual (where you will not find `conda` mentioned) under the heading "Installing on Windows".

mpharrigan:
I don't think it's too much to ask developers who are tinkering with openmm to run `make PythonInstall` to see their changes show up from python

peastman:
> If you're only a Python programmer, you should be able to interact with OpenMM as if it's just a Python package.

That's a reasonable goal, but that's a very big "if" it starts with.  Especially the "only".  There's big difference between a person who uses Python, and a person who *only* uses Python.  For example, when I speak at workshops, I always advise people that even if they're calling OpenMM from another language, they still use Python as a tool for development and debugging.  A large fraction of people who call OpenMM from other languages *also* use the Python interface.

So let's clarify exactly what our goals are, then see if there's a way to achieve them without creating new problems for significant classes of users.

mpharrigan:
The classes of users are

1. c++ (et. al) users using openmm as a library in their application. These users probably aren't changing and recompiling openmm but rather using a released version. They will be totally unaffected [link against libs in make install, drop into python for development and debugging]

2. python users using openmm as an md application. They will benefit from these changes because they don't have to be concerned about where the openmm c++ libraries are. They can have multiple python environments whose openmm libraries don't clash

3. Developers of openmm. This class of user would be slightly disadvantaged by having to run `make PythonInstall` instead of just `make install` to see their changes from python

Do you think this is a reasonable classification @peastman ?

rmcgibbo:
My personal goal is essentially: *OpenMM-as-a-python-package* should behave like any other Python package. That means, it should follow same workflow, as much as possible, as every other Python package using the standard package-management tools provided by the language.
   * `pip install openmm` should *just work*.
   * `pip uninstall openmm` should *just work*.
   * Having multiple different versions of OpenMM installed on your machine should be not cause any problems. You should be able to have different interpreters or virtualenvs that have different OpenMM's installed without link time errors.
   * Our instructions for "here's how to install *OpenMM-as-a-python-package*" for new users should look immediately familiar to any user who has previously installed other Python packages. It should not require modifying the `PATH`, setting any environment variables, running a script called `install.sh`, or other non-standard things like that.
   * Errors like "Make sure OpenMM is installed and the library path is set correctly. Error message: DLL load failed: The specified module could not be found." https://github.com/pandegroup/openmm/issues/1307#issuecomment-168843431 should be impossible.

Also, we could add `make PythonDevelop` as an alternative to `make PythonInstall` for developers, as an analogue to https://pythonhosted.org/setuptools/setuptools.html#development-mode.

jchodera:
Maybe the simple solution is to eliminate Python support from the binary installers?  Force users to compile/install from source or go the conda route?

mpharrigan:
These changes would help (unbiased verb: impact) installing from source with python bindings as well.

peastman:
> c++ (et. al) users using openmm as a library in their application. These users probably aren't changing and recompiling openmm but rather using a released version.

No, that's not correct.  People who use it as a library very frequently start modifying the source code.  Then I have to convince them they should really extract their changes and turn them into a plugin!  But if they're just making changes for their own use, they usually just hack it into the existing code, since that's the path of least resistance.

>  pip install openmm should just work.

How important is it to support both pip *and* conda?

> Having multiple different versions of OpenMM installed on your machine should be not cause any problems.

Keep in mind that you aren't typical of most users.  You *want* to have multiple versions of OpenMM on your computer at once, and you're looking for a way to make them easier to manage.  The majority of users don't want or need multiple versions at once, and for them, having multiple versions at once is just asking for trouble.

> Our instructions for "here's how to install OpenMM-as-a-python-package" for new users should look immediately familiar to any user who has previously installed other Python packages.

We need to add a discussion of conda to the manual.  If we do that, is this point sufficiently well addressed?  And conda already installs its own copy of the libraries in its own internal directory.

> Maybe the simple solution is to eliminate Python support from the binary installers? Force users to compile/install from source or go the conda route?

That just creates exactly the same problems I pointed out before.  There's a significant class of users who access OpenMM from multiple languages, and want all those languages to be using the same version.  Right now, those users are well served.  Let's see if we can find a way to make things easier for the Python-only users *without* making things harder for any other classes of users.

mpharrigan:
> No, that's not correct. People who use it as a library very frequently start modifying the source code. Then I have to convince them they should really extract their changes and turn them into a plugin! But if they're just making changes for their own use, they usually just hack it into the existing code, since that's the path of least resistance.

Ok, so there's no class of users who use vanilla openmm

1. python users using openmm as an md application. They will benefit from these changes because they don't have to be concerned about where the openmm c++ libraries are. They can have multiple python environments whose openmm libraries don't clash

2. Developers of openmm or openmm hacks. This class of user would be slightly disadvantaged by having to run make PythonInstall instead of just make install to see their changes from python

rmcgibbo:
I'm honestly baffled that anyone thinks having the package split up in separate places on the filesystem by default is a good idea.

> Developers of openmm or openmm hacks. This class of user would be slightly disadvantaged by having to run make PythonInstall instead of just make install to see their changes from python

If you want to modify any of the OpenMM python code, and see your changes reflected in the interpreter, you already currently have to run `make PythonInstall`, or similar. So I don't even really see this as a change in behavior.

mpharrigan:
> If you want to modify any of the OpenMM python code, and see your changes reflected in the interpreter, you already currently have to run make PythonInstall, or similar. So I don't even really see this as a change in behavior.

I agree, but to see changes in the c++ behavior, you could have run `make install` in the past. Although maybe *that* only worked if you were sure not to change any of the function signatures

rmcgibbo:
If the inconvenience that this PR creates for developers is that `make install` no longer makes changes to the C++ code available from Python, we could fix that pretty easily.

We could actually just remove the need for a second command to install the Python wrappers and have `make install` copy both the C++ libraries to CMAKE_INSTALL_PREFIX and the install Python libraries (if `OPENMM_BUILD_PYTHON_WRAPPERS` is set). by having `install` run the PythonInstall target under the hood -- I think it would require a little bit of tinkering in cmake but not much).

mpharrigan:
> We could actually just remove the need for a second command to install the Python wrappers and have make install copy both the headers & libs to CMAKE_INSTALL_PREFIX and the install Python libraries (if OPENMM_BUILD_PYTHON_WRAPPERS is set), by having install run the PythonInstall target under the hood -- I think it would require a little bit of tinkering in cmake but not much).

I like this idea in general

peastman:
> I'm honestly baffled that anyone thinks having the package split up in separate places on the filesystem by default is a good idea.

That's because you think of OpenMM exclusively as a Python package, whereas I think of it as a library that has multiple interfaces and is designed to be called from multiple languages.

The real problem is that Python wants to be a complete self contained world, all on its own.  The files needed to call OpenMM from C, C++, and Fortran are all found in a single directory.  Ideally I'd like to have the Python interface in exactly the same place.  But Python doesn't play well with others.  It wants you to stick everything in locations it provides within its Python-specific directory structure.  That sucks, but it's what we have to deal with.  So we somehow need to compromise between the "Python world" and the "everything else world"

rmcgibbo:
> That's because you think of OpenMM exclusively as a Python package, whereas I think of it as a library that has multiple interfaces and is designed to be called from multiple languages.

I don't think that's a fair characterization.

My point is that I want each language-specific interface to behave as a first-class citizen in that languages ecosystem, and obey that languages conventions.

peastman:
Agreed.  But the conventions of Python directly conflict with the conventions of most other languages.  C, C++, and Fortran don't provide their own directory structure that all libraries are supposed to go in.  Neither, for that matter, do Java, C#, Objective C, Swift, or most other languages.  Python is the outlier here.

Consider how most other libraries deal with this: FFTW, OpenGL, OpenCL, etc.  All of these are native libraries that can be called from multiple languages.  And for all of them, the normal way to install them is to put a single copy in a central location (On Linux, typically /usr/lib if it's installed with the standard package manager, or /usr/local if it isn't.  On OS X, typically /System/Library/Frameworks if it's been packaged up as an OS X framework, or /usr/local if it hasn't).  All of those libraries also have Python bindings.  Those bindings do not bundle a new copy of the library.  Instead, they link to the existing copy in the central location.

rmcgibbo:
If OpenMM were split into two separate packages called e.g. `libopenmm`, `python-openmm` and  -- one of which was a C++ library and the other which provided Python bindings, and the two different packages had different release cycles, and you could mix and match versions, etc, then I might agree with you. That would make it more similar to FFTW or OpenCL or something.

But that's not the case. The "libopenmm" API changes all the time, and as such the language-specific bindings like "python-openmm" need to be extremely careful about linking to the proper version. The headaches like #1292 greatly increase the value of having the Python library be self-contained.

And It's not like Python libraries that wrap native libraries always reach across the filesystem to find them -- e.g. https://pypi.python.org/pypi/pyFFTW embeds the fftw DLLs. R packages are also a good example. For example if you download and unpack any of the fftw packages from CRAN, you'll see that they contain all of fftw https://cran.r-project.org/web/packages/fftw/index.html -- they don't expect to find fftw in /usr/lib or /usr/local, or depend on the users being savy enough to set their LD_LIBRARY_PATH/DYLD_LIBRARY_PATH/PATH. In fact, I suspect this convention is standard for all dynamic languages that include a first-class package manager (npm, rubygems, pip, julia, luarocks, etc).

rmcgibbo:
Anyways, by integrating `PythonInstall` into `install`, I think I can get all of the benefits of this without the drawback that `make install` won't make changes available to the Python interpreter.

rmcgibbo:
Okay, e3d5c89 9057aff make some significant changes to this PR.
 - `make install` now installs both the C/C++/Fortran libraries to CMAKE_INSTALL_PREFIX and the Python libraries to their native locations.
 - You can now run `make PythonInstall` without having previously run `make install`, if you prefer that.

So if you have `OPENMM_BUILD_PYTHON_WRAPPERS` set on and make a change to the code, `make install` will install your changes for whatever language you're using.

rmcgibbo:
Also fixes https://github.com/omnia-md/conda-recipes/issues/313, which was a bug here.

mpharrigan:
neat! +1

jchodera:
Awesome!

peastman:
Closing abandoned PRs.

PR #1322, opened 2016-01-05
Python 3 does not have dict.itervalues

swails:
Really we should move to using six and importing the iterator versions of dict
methods and range/zip from there, but for now just use "values" since using a
raw list doesn't seem like it will cause memory or performance problems here.

This fixes an `AttributeError` I got working with Python 3.


rmcgibbo:
If I'd be useful, we can just copy six.py over to internal/ or something.

swails:
> If I'd be useful, we can just copy six.py over to internal/ or something.

My vote is to wait on this until someone can sit down and do it "right".  I would've done that if I'd had time to go through and replace the instances of `range`, `zip`, etc. in the Python layer :).

But :+1: on this plan in the long run

swails:
I'll point out that this plan is what both mdtraj *and* parmed do, and I've been pretty happy with how that has worked out in both cases so far.

PR #1327, opened 2016-01-07
AmoebaTorsionTorsionForce avoids singularity in acos()

### LINKED ISSUE #1323: Amoeba simulations crashing on GTX 980 Ti in single precision
[I've put the files you can download on dropbox](https://www.dropbox.com/s/qwu3y79awhy8efb/4yula_amoeba.tgz?dl=0)

You need ParmEd installed, and you can run the script via the command `./run.sh`.  What I've observed so far:

- On a GTX 980 Ti in the default (single) precision, within 0-3k steps, the simulation flounces off to NaNdyland and never comes back. Before crashing, simulation looks "normal"
- Using mixed precision fixes the problem.  I have not checked double yet (running now).
- Problem occurs on every GTX 980 Ti we've tried (we have a box with 8 of them)
- Problem does not occur on the GTX 980, GTX 780, and GTX 680 cards we have
- Problem does not occur if we use the exact same coordinates/PDB, but use a system.xml file built with amber99sbildn instead of amoeba13.

peastman:
Fixes #1323.

### DIFF
--- a/plugins/amoeba/platforms/cuda/src/kernels/amoebaTorsionTorsionForce.cu
+++ b/plugins/amoeba/platforms/cuda/src/kernels/amoebaTorsionTorsionForce.cu
@@ -40,8 +40,18 @@ real rcb = SQRT(xcb*xcb + ycb*ycb + zcb*zcb);
 real cosine1 = (rtru != 0 ? (xt*xu+yt*yu+zt*zu)/rtru : (real) 0);
 cosine1 = (cosine1 > 1 ? (real) 1 : cosine1);
 cosine1 = (cosine1 < -1 ? (real) -1 : cosine1);
+real angle1;
+if (cosine1 > 0.99f || cosine1 < -0.99f) {
+    // We're close to the singularity in acos(), so take the cross product and use asin() instead.
 
-real angle1 = RAD_TO_DEG * ACOS(cosine1);
+    real3 cross_prod = cross(make_real3(xt, yt, zt), make_real3(xu, yu, zu));
+    angle1 = ASIN(SQRT(dot(cross_prod, cross_prod)/(rt2*ru2)));
+    if (cosine1 < 0.0f)
+        angle1 = M_PI-angle1;
+}
+else
+   angle1 = ACOS(cosine1);
+angle1 = RAD_TO_DEG*angle1;
 real sign = xba*xu + yba*yu + zba*zu;
 angle1 = (sign < 0 ? -angle1 : angle1);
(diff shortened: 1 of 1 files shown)


swails:
This isn't working for me.  When I try running, I get:

```
  File "./runmd.py", line 87, in <module>
    mm.Platform.getPlatformByName('CUDA'))
  File "/usr/local/miniconda/envs/py27/lib/python2.7/site-packages/simtk/openmm/app/simulation.py", line 86, in __init__
    self.context = mm.Context(system, integrator, platform)
  File "/usr/local/miniconda/envs/py27/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 2861, in __init__
    this = _openmm.new_Context(*args)
Exception: Error compiling program: (1455): error: identifier "M_PI" is undefined

(1475): error: identifier "M_PI" is undefined

2 errors detected in this compilation.
```

peastman:
Try now.

swails:
It got past that compilation step.  It's running now, I'll let you know if I hit a blowup.

As an aside, how would you have included `cmath` in that file if you didn't just switch to using `RAD_TO_DEG`?  I couldn't for the life of me see how `M_PI` would be available in some of the other kernels, but not the torsion-torsion one...

swails:
It's already hit past 2000 steps, so I suspect this has fixed the issues.

peastman:
Some compilers define M_PI automatically, others don't.  I generally just define it myself if I need it in a kernel.

swails:
what if you need something from a header file?... or that just doesn't happen?

swails:
Anyway, this seems to have fixed it!

peastman:
CUDA automatically provides most of the math functions normally found in cmath.  They're just part of the language.  What other things are you thinking of that you would need from a header file?

swails:
well, M_PI, for instance :).  I was just curious as to how the namespace inside the CUDA kernels was populated (since they don't appear to be standalone-compilable files).

peastman:
It depends on the file.  What we ultimately send to the compiler has to be a standalone file.  In the case of bonded forces (like AmoebaTorsionTorsionForce), for example, CudaBondedUtilities constructs a kernel, inserting into the middle of it the code for any bonded forces that have been registered with it.  It also prepends the contents of vectorOps.cu to get all the vector operations that CUDA doesn't include.  Then it sends the kernel to CudaContext::createModule(), which adds a bunch of typedefs and defines before finally sending it to the compiler.

swails:
thanks for the explanation

PR #1328, opened 2016-01-08
[WIP] Attempt at parallel conda openmm-dev build/push

jchodera:
This attempts to bring the `openmm-dev` conda package build back into the `openmm` repo.

Currently, only the `linux` package is built in a separate `travis` branch.  The package is always built in PRs to make sure things work, but the package is only pushed in commits to the actual repo.


.travis.yml
```diff
@@ -78,6 +84,7 @@ matrix:
       env: ==PYTNON_2==
```
swails: PYTHON? (I know this is a pre-existing typo)

mpharrigan:
why is sudo required?

still -1 on having this in the openmm repo (or at all, see removal of msmbuilder-dev and mdtraj-dev packages)

jchodera:
> why is sudo required?

It isn't.  I'm just tweaking things bit by bit.

> still -1 on having this in the openmm repo (or at all, see removal of msmbuilder-dev and mdtraj-dev packages)

Can you elaborate (or provide links)?

The current situation (having the recipe [here](https://github.com/omnia-md/openmm-dev), which has become hopelessly out of sync with the actual [openmm recipe](https://github.com/omnia-md/conda-recipes/tree/master/openmm).  The build is currently done for linux only by Jenkins [here](https://jenkins.choderalab.org/job/conda-openmm-dev-linux-vagrant-2/)) is very, very, very unreliable.   See how many times it fails to upload a package?  It often ends up deleting the existing one, which makes it impossible to automatically test things with it.

In short, *anything* is better than the current situation.

jchodera:
Oh---`sudo` is required for installing CUDA and OpenCL, it seems.

mpharrigan:
Here's the discussion https://github.com/msmbuilder/msmbuilder/pull/683

swails:
I still use the `openmm-dev` package frequently, and lament how unreliable it is.  Development on several fronts is ongoing and fairly rapid, and there are quite a few places for me where installing `openmm-dev` is far more convenient (or even the only way!) of getting the latest snapshot of master.

msm-builder may have a stable API, but OpenMM is challenging enough to build, and conda easy enough to deploy, that keeping this package is a big win for me, personally.

jchodera:
I'm 100% with @swails on this one.

jchodera:
Also, manually building `msmbuilder` and building `openmm` are freakishly, horrifically different from each other.

mpharrigan:
ok, I see the benefit

swails:
@mpharrigan -- I had already typed this out and was about to post it before your response.  I'll still post it, though, since it's a concrete example of where the API/code is rapidly evolving (to my needs in particular).

For instance, I'm using Amoeba a lot, and there are numerous little bugs and fixes and stuff getting added (see #1327, the extrapolated Amoeba stuff, etc.) where the release schedule is far too slow to keep up with the code I need to be using.

And building OpenMM on one of my supercomputers is proving to be very annoying (see #1329), and for some reason on my Mac Mini, seemingly impossible (not sure why, but it doesn't matter because I do my development on a different Mac).

The other front where development is occurring rapidly is related to FFXML development that @jchodera's lab is (at least largely) spearheading (I'm too lazy to link the appropriate issues and PRs here).  Again -- deploying for testing in this case is infinitely easier if we have a `openmm-dev` package.  And a reliable one is essential for CI testing through Travis.

jchodera:
@rmcgibbo: Can you help me figure out what to do with the new OpenMM API docs?  The travis branch that compiles `openmm-dev` is [failing with](https://travis-ci.org/pandegroup/openmm/jobs/101154918#L3543-L3551):
```
+ cd /home/travis/miniconda/envs/_build/docs
+ mkdir openmm
+ mv '*.html' 'api-*' openmm/
mv: cannot stat `*.html': No such file or directory
mv: cannot stat `api-*': No such file or directory
Command failed: /bin/bash -x -e /home/travis/build/pandegroup/openmm/devtools/conda-recipe/build.sh
```

rmcgibbo:
Uggh. I'm still -1 on this stuff being in this repo.

But it looks like you're using `make DoxygenApiDocs`, which isn't a make target anymore. I documented the make targets [here](https://github.com/pandegroup/openmm/blob/master/docs-source/usersguide/library.rst#building-the-documentation-optional).

jchodera:
> Uggh. I'm still -1 on this stuff being in this repo.

It's either this or just blow away the `openmm-dev` conda package entirely, I think.  Whatever is broken, we're not able to get the continuum folks to fix it anytime soon.

> But it looks like you're using make DoxygenApiDocs, which isn't a make target anymore. I documented the make targets here.

jchodera:
Silly question: Shouldn't we be compiling the HTML and/or PDF docs in travis to test that we haven't broken the docs?

rmcgibbo:
> It's either this or just blow away the openmm-dev conda package entirely, I think. Whatever is broken, we're not able to get the continuum folks to fix it anytime soon.

I guess that there's something persistently or intermittently with your `openmm-dev` package, related to how uploading it to anaconda cloud works, or something.

What leads you to believe that this problem is going to be fixed by moving this code into `github.com/pandegroup/openmm`?

jchodera:
> I guess that there's something persistently or intermittently with your openmm-dev package, related to how uploading it to anaconda cloud works, or something.

Something funny between the package, our Jenkins server, MSK's firewall, and the anaconda cloud servers.

> What leads you to believe that this problem is going to be fixed by moving this code into github.com/pandegroup/openmm?

Because we have never had any of these problems with building packages via travis.

jchodera:
I kind of need a functional `openmm-dev` package to test the GAFF residue template generator written in another package (`openmoltools`).

rmcgibbo:
> Because we have never had any of these problems with building packages via travis.

Not trying to be too much of a nagging nancy, but presumably you could switch the CI system for the `github.com/omnia-md/openmm-dev` repo from jenkins to travis without any changes to this repo.

jchodera:
> Not trying to be too much of a nagging nancy, but presumably you could switch the CI system for the github.com/omnia-md/openmm-dev repo from jenkins to travis without any changes to this repo.

And then manually trigger the travis build each time `pandegroup/openmm` is updated?  No thanks.

Unless there's a way to automate the triggering of builds in that repo everything `pandegroup/openmm` is updated?

jchodera:
I'm certainly open to other ideas!  But they would have to actually work.

jchodera:
I guess we could develop some way to use the travis JSON web trigger API to have `pandegroup/openmm` tell `omnia-md/openmm-dev` to run?
https://docs.travis-ci.com/user/triggering-builds

rmcgibbo:
My 2 cents:

1. The current `pandegroup/openmm` testing infrastructure is already both quite complex and quite slow. I don't want to make it more complex and slower.
2. Packaging and distribution of binaries is hard. I think the evidence for this claim is plain in all the problems you've been having, all of the complexity of `omnia-md/conda-recipes`, and all of the number of bugs I've filed against `conda`. I think `conda` is pretty good, but I definitely wouldn't call it rock-solid stable.
3. I would like to keep the travis-ci build status indicator on `pandegroup/openmm` working (that is, properly signalling green when the tests pass). Increasing the complexity and number of tasks that this testing infrastructure is supposed to accomplish (i.e. packaging) will negatively impact the reliability of this.
4. I'm not convincing that all of the problems you've been having getting your jenkins configuration for openmm-dev to work come from Jenkins or MSKCC. I suspect some of them come from `conda` or anaconda.org. The implication of this is that when this PR gets merged, it's not going to fix any of those bugs, it's just going to "import" them into this repo.
5. I have been the primary one maintaining the `pandegroup/openmm` travis-ci infrastructure, and that is one reason I would prefer its scope not be expanded. Inevitably when it breaks, I'm probably going get pinged to fix it.
7. Finally, I don't really think automated packaging and distribution of each development snapshot is a very good idea in the first place, as I explained [here](https://github.com/msmbuilder/msmbuilder/pull/683#issuecomment-155172922). I know that not everyone agrees with that though.

jchodera:
I do agree with many of your points (though notably not number 6---the `openmm-dev` packages are a huge timesaver for us because of the slow release cycle and huge hurdles to inclusion), but the current problem is that nobody has proposed a better solution.  

Including this in `pandegroup/openmm` is currently the best option among the current set of even worse options.

rmcgibbo:
> but the current problem is that nobody has proposed a better solution.

This doesn't quite seem fair to me. I mean, you created this `openmm-dev` infrastructure primarily to serve your needs. That's fine. And then it was unreliable and broken, and you had trouble fixing it. And then you propose, from my perspective (point 5), something that feels like "Why don't we just move somewhere so that Robert will fix it!".

jchodera:
> This doesn't quite seem fair to me. I mean, you created this openmm-dev infrastructure primarily to serve your needs. That's fine. And then it was unreliable and broken, and you had trouble fixing it. And then you propose, from my perspective (point 5), something that feels like "Why don't we just move somewhere so that Robert will fix it!".

That *would* be totally unfair, but we don't expect you to do that!  I did ask for your help here to figure out how to fix the parts of the [`openmm-dev`](https://github.com/omnia-md/openmm-dev/tree/master/openmm-dev) recipe that broke when you overhauled the API docs (which is a huge win!)---thanks!  But the new `.travis.yml` structure is actually really nice and modular, so I should be able to maintain these components.  We should also be able to easily turn this off and have the impact go to zero if we find it is disruptive.  

Given the issues we had been having, I *think* this will solve the primary problems we have with some unusual travis-MSK-anaconda interactions that do not seem to appear anywhere else in our automated package push infrastructure.  (Believe me, I've tried to track down the networking and anaconda/binstar issues and have got nowhere.)

Lots of our other projects push dev packages via travis, and that seems to be working OK!

I can give Jenkins one more try once I fix the docs building in the [`openmm-dev`](https://github.com/omnia-md/openmm-dev/tree/master/openmm-dev) recipe, but that infrastructure has become an enormous pain by comparison.

jchodera:
For example, our Jenkins build box became nonresponse over break and nobody was around to reboot it.  Maybe we need [this](http://www.gizmag.com/microbot-push-robotic-finger-home-automation/39097/).

rmcgibbo:
Maybe you could put a little CURL to the travis-ci API inside this build https://docs.travis-ci.com/user/triggering-builds that would trigger a buid on the master branch of `omnia-md/openmm-dev`?

swails:
@rmcgibbo makes some convincing points.  I think it's worth holding off on this until we find something that doesn't have such marked drawbacks on the OpenMM Travis-CI system.

> Finally, I don't really think automated packaging and distribution of each development snapshot is a very good idea in the first place, as I explained here. I know that not everyone agrees with that though.

IMO, OpenMM warrants an exception here, but this has been discussed before.

>> And then you propose, from my perspective (point 5), something that feels like "Why don't we just move somewhere so that Robert will fix it!".

> That would be totally unfair, but we don't expect you to do that!

No, but it's a possible (bordering on probable) result of the work here.  I've experienced it with Amber, myself.  Someone wants to do something, dumps a whole bunch of half-finished or broken stuff into the tree, and I end up having to fix it.  Not because that's what they intended, but because they were too slow fixing it or didn't know how, and I needed it fixed for my own work.

The same thing could easily happen here, and I think it's worth avoiding.

jchodera:
> Maybe you could put a little CURL to the travis-ci API inside this build https://docs.travis-ci.com/user/triggering-builds that would trigger a buid on the master branch of omnia-md/openmm-dev?

I've been looking for examples of that.  I think this might be difficult for me to debug, since I don't have direct commit access and it might require a few direct test commits (as well as some new encrypted environment variables?) to get working.

rmcgibbo:
Someone has written a script for this I think: http://stackoverflow.com/a/34273424/1079728, https://github.com/mernst/plume-lib/blob/master/bin/trigger-travis.sh

rmcgibbo:
You could also just set a cron job on your workstation (or on something like https://www.setcronjob.com/) to make empty commits / pushes to `omnia-md/openmm-dev` every night, or every hour.

mpharrigan:
> I've been looking for examples of that. I think this might be difficult for me to debug, since I don't have direct commit access and it might require a few direct test commits (as well as some new encrypted environment variables?) to get working.

On this note, a benefit of keeping a separate repo is you (and others interested in the dev package) can commit to your hearts' content without polluting the openmm tree

mpharrigan:
As an alternative to the bash script curling a post request to travis, you can set up a webhook on the openmm repository to trigger a travis build of the openmm-dev repository via github's "webhooks" extensions

https://developer.github.com/webhooks/

jchodera:
A cron job on a physical workstation doesn't get away from the fragility issues associated with having these services dependent on physical hardware here at MSKCC, but [this `traviscron` web service](http://traviscron.pythonanywhere.com) might be a viable alternative.

> As an alternative to the bash script curling a post request to travis, you can set up a webhook on the openmm repository to trigger a travis build of the openmm-dev repository via github's "webhooks" extensions

That would be awesome!  Would this be easy for one of the devs to set up?

rmcgibbo:
The traviscron service seems perfect.

jchodera:
> On this note, a benefit of keeping a separate repo is you (and others interested in the dev package) can commit to your hearts' content without polluting the openmm tree

Definitely a benefit, but with the associated drawback of allowing changes to be made to the main `openmm` repo without anyone noticing that this breaks the downstream `openmm-dev` recipe, like what happened with the API docs changes.

rmcgibbo:
Or I found https://nightli.es/

rmcgibbo:
> ... without anyone noticing that this breaks the downstream

Well, Travis-CI can send an email/slack/irc/etc to tell you when it fails.

jchodera:
> The traviscron service seems perfect.

Definitely not perfect.  Just not as bad as Jenkins, but potentially not as good as having the `openmm-dev` package build/push in the `pandegroup/openmm` travis, depending on how you weight things like "is it OK for someone to break the `openmm-dev` recipe and have nobody notice for weeks?".

rmcgibbo:
I don't see why nobody would notice.

jchodera:
> Well, Travis-CI can send an email/slack/irc/etc to tell you when it fails.

OK, will give this option a try once I finish debugging the `openmm-dev` recipe for the docs build changes.

Since we'll be triggering travis twice daily, I wonder if it makes sense to set up an `omnia-md/conda-dev-recipes` repo with `openmm-dev` as one of potentially several packages we want `osx` and `linux` dev packages regularly built for.

rmcgibbo:
I just activated https://nightli.es/, and it was one click to turn it on for mdtraj/mdtraj. I guess I have to wait for the night to see if it works though.

jchodera:
> I don't see why nobody would notice.

Nobody did notice when it was on jenkins.  The utility of travis is generally that PRs aren't merged until all test failures are cleared.  Without the tests reporting back into the PR, this can easily be missed.

jchodera:
This is odd: The makefile target `C++ApiDocs` doesn't appear on `osx`:
https://travis-ci.org/pandegroup/openmm/jobs/101180671#L3728-L3729

I confirm that I don't see that target when I build on my `osx` machine.  Am I missing something needed beyond the [instructions here](https://github.com/pandegroup/openmm/blob/master/docs-source/usersguide/library.rst#building-the-documentation-optional), @rmcgibbo?

jchodera:
Ah, sorry, I had missed the CMake option:
```
OPENMM_GENERATE_API_DOCS=ON
```
I'll add that to the docs.

rmcgibbo:
> Nobody did notice when it was on jenkins. The utility of travis is generally that PRs aren't merged until all test failures are cleared. Without the tests reporting back into the PR, this can easily be missed.

I see what you mean, but it I don't think this is a blocker. The number of PRs that require corresponding changes to your build recipe is pretty small. It's essentially just `make install`, `make PythonInstall`, and whatever the make targets for the docs are. Since 6.3, I think that's only #1238. And if you have a badge or email notification that pings someone who has an interest in these packages being updated, it seems like they could make the changes required to keep it current without much trouble.

rmcgibbo:
`OPENMM_GENERATE_API_DOCS=ON`, yeah. Sorry, I didn't realize that was missing from the docs. You probably also need to install the mentioned sphinx extensions.

rmcgibbo:
(You also don't really need to have the docs in the conda package, but whatever. I'm not sure if anyone would think to look in inside there, instead of just looking on the web.)

jchodera:
> (You also don't really need to have the docs in the conda package, but whatever. I'm not sure if anyone would think to look in inside there, instead of just looking on the web.)

Eh.  We've been distributing the docs with other release packages, so why not?

Where *are* the dev docs available on the web, by the way?

And can we do something about the fact that the top google hits for "openmm docs" (at least for me) are still [this less-than-useful SimTK page](Simtk.org: OpenMM: Documents) and [the 6.1.0 release docs](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=2&ved=0ahUKEwjai924wpvKAhVB2hQKHbBbBGoQFggjMAE&url=http%3A%2F%2Fdocs.openmm.org%2F6.1.0%2Fuserguide%2F&usg=AFQjCNG5py3yb_paSNTbnQq6RXRP7Lo27w&sig2=LsFa0VGustkxkIRnpoX4Yw)?

jchodera:
> I don't think this is going to work. Stuff that you install with pip into the base conda env generally isn't available in the conda-build build environment.

Oof.  And there's no way to specify pip-installed packages as a `build:` prerequisite, huh.

rmcgibbo:
> Eh. We've been distributing the docs with other release packages, so why not?

Yeah, it obviously doesn't hurt, but I just don't think it really helps either. A user that runs `conda install openmm` might not really know to go hunting inside ~/home/anaconda to find the OpenMM docs. Or maybe they would -- I'm not sure. google is usually the first place I look.

> And can we do something about the fact that the top google hits for "openmm docs" (at least for me) are still this less-than-useful SimTK page and the 6.1.0 release docs?

This is a good point. I'm not sure how to do this, but maybe someone does. There's probably something with the Google Webmaster tools that could be done to let the googlebot know more about `openmm.org` and `docs.openmm.org`, or maybe we could add a sitemap or something (https://support.google.com/sites/answer/100283?hl=en).

rmcgibbo:
But if the googlebot is crawling this page, the dev docs are at

- http://docs.openmm.org/development/api-python/
- http://docs.openmm.org/development/api-c++/

jchodera:
Maybe we could pull @mpiercy in to help with the google site optimization of http://docs.openmm.org?

jchodera:
Closing this in favor of handling the build separately through https://github.com/omnia-md/conda-dev-recipes

Thanks for the suggestions and help, everyone!

PR #1332, opened 2016-01-09
Also build the HTML user guide in travis

jchodera:
I just found I missed that I had added some syntax errors to the user guide because none of the travis branches are building/testing the guide.

This PR also builds the user guide in the same branch that the API docs are built, but the user guide currently isn't deployed anywhere.  Do we also want to push that to the S3 docs?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,7 +148,8 @@ script:
   - python devtools/run-ctest.py --start-time $START_TIME
 
   - if [[ ! -z "${DOCS_DEPLOY}" && "${DOCS_DEPLOY}" = "true" ]]; then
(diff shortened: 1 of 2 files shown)


rmcgibbo:
Yes, that would be good. Currently, the deploy section at the bottom of .travis.yml copies the local directory called `api-docs` (it doesn't need to be called this, that's just where the stuff to be pushed is [moved in](https://github.com/pandegroup/openmm/blob/master/.travis.yml#L153-L155)) to the directory in S3 called `development` on `http://docs.openmm.org`, which leads to 
- http://docs.openmm.org/development/api-python/, and
- http://docs.openmm.org/development/api-c++/

being available.

jchodera:
OK, I *think* this is working now.

PR #1334, opened 2016-01-09
Fix push of dev user and developer guide HTMLs to S3

jchodera:
I think the HTML user and dev guides were not being correctly moved into the directory uploaded to S3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -152,8 +152,8 @@ script:
       make sphinxhtml;
       make C++ApiDocs PythonApiDocs;
       mkdir -p api-docs;
-      mv sphinx-docs/userguide api-docs;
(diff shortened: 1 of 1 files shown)


rmcgibbo:
No, this isn't right.

rmcgibbo:
They got pushed successfully from the last master branch to http://docs.openmm.org/development/userguide/html/ and
http://docs.openmm.org/development/developerguide/html/

the extra "html" in the path is a little annoying perhaps.

jchodera:
Oh!  Sorry---just fixed this to remove the `html/` since the URL isn't correct from http://openmm.or

jchodera:
Thanks!  All the links seem to work now!

Issue #1319, opened 2016-01-04
Possible CPU OpenCL errors on Travis-CI


rmcgibbo:
There may be a couple CPU OpenCL errors on Travis. They may have arisen with their [move to new infrastructure](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future), or not.

- https://travis-ci.org/pandegroup/openmm/jobs/100031954#L979-L985
- https://travis-ci.org/pandegroup/openmm/jobs/99979017#L1078
- https://travis-ci.org/pandegroup/openmm/jobs/99990558#L1079

The apparent errors are segfaults in `TestOpenCLCustomManyParticleForce`.

I previously tried, but was unable to reproduce the bug on my machine, either in Ubuntu 15.04 or in a 12.04 container, using the CPU OpenCL platform (AMD drivers). So, ¯\_(ツ)_/¯.

cc: https://github.com/pandegroup/openmm/pull/1317#issuecomment-168567674, https://github.com/pandegroup/openmm/issues/1315#issuecomment-168448636, https://github.com/pandegroup/openmm/pull/1312#issuecomment-168438851

rmcgibbo:
It doesn't always segfault. [Here](https://travis-ci.org/pandegroup/openmm/jobs/100437216), it segfaulted once and then failed an assertion in `TestCustomManyParticleForce:testCentralParticleModeLargeSystem` the second time.

```
16/134 Test  #77: TestOpenCLCustomManyParticleForceMixed ..........***Exception: SegFault 22.71 sec
[...]

******************************
Rerunning failing tests...
******************************
Test project /home/travis/build/pandegroup/openmm
    Start 77: TestOpenCLCustomManyParticleForceMixed
1/1 Test #77: TestOpenCLCustomManyParticleForceMixed ...***Failed   24.50 sec
exception: Assertion failure at TestCustomManyParticleForce.h:705.  Expected 107521, found 2.8418e+06
```

jchodera:
I'm still seeing these failures (e.g. [here](https://travis-ci.org/pandegroup/openmm/jobs/107378907#L1015-L1021)).  

Before we release 7.0, I think it is essential that we debug this (if it is a bug) or disable that OpenCL kernel on OpenCL implementations we think are buggy.

rmcgibbo:
[...]

PR #1289, opened 2015-12-11
Add OpenCL to Appveyor and fix nosetests

rmcgibbo:
Incorporates #1287, #1288.

PR #1287, opened 2015-12-10
Change Python test runner on appveyor

rmcgibbo:

PR #1336, opened 2016-01-11
More acos protections in Amoeba

swails:
We still had some Amoeba simulations go off to NaNdyland, although it happened ~once per day rather than every couple minutes since #1327 was merged.  This is hoping that the in-plane angle force (or, less likely, the stretch-bend force) occasionally hits this singularity.

I'll test this out and see if it (hopefully) eliminates our Amoeba blowups.  If it does, I'll remove the [WIP] label.

### DIFF
--- a/plugins/amoeba/platforms/cuda/src/kernels/amoebaInPlaneForce.cu
+++ b/plugins/amoeba/platforms/cuda/src/kernels/amoebaInPlaneForce.cu
@@ -43,11 +43,19 @@ real dot = xap*xcp + yap*ycp + zap*zcp;
 real product = SQRT(rap2*rcp2);
 real cosine = (product > 0 ? (dot/product) : 0);
 cosine = max(min(cosine, (real) 1), (real) -1);
-real angle = ACOS(cosine);
+real angle;
+if (cosine > 0.99f || cosine < -0.99f) {
+    real3 cross_prod = cross(make_real3(xap, yap, zap), make_real3(xcp, ycp, zcp));
+    angle = ASIN(SQRT(dot(cross_prod, cross_prod)/(rap2*rcp2)))*RAD_TO_DEG;
+    if (cosine < 0.0f)
+        angle = 180-angle;
+}
+else
+    real angle = ACOS(cosine)*RAD_TO_DEG;
 
 // if product == 0, set force/energy to 0
 
-real deltaIdeal = (product > 0 ? (angle*RAD_TO_DEG - angleParams.x) : 0);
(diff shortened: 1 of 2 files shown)


peastman:
Any idea why AppVeyor failed?  I don't see any error messages, it just stopped part way through the Python tests.  Maybe it took too long?

rmcgibbo:
I don't think it's a timeout. If you hover your mouse over the linenumbers it shows the time at which that line was emmitted -- the last line is about 24 minutes in. Remember that a couple months ago Appveyor (before #1287 / #1289) would do a similar thing and just stop without any message when it got to running the Python tests. That was "fixed" by replacing `nosetests` with `py.test` as the test runner, but I never had _any idea why_ that worked. I just tried it randomly and it seemed to.

peastman:
Ok.  Anyway, it's pretty clearly unrelated to this PR, since AppVeyor doesn't even build the CUDA platform.

swails:
Same with the Travis failure (which is OCL-related).  Amoeba doesn't have an OCL implementation, so this PR is clearly unrelated as well :).  I think this is something @rmcgibbo has mentioned before (and has an open issue regarding a potential CPU-OCL segfault IIRC).

So far one of the simulations that used to NaN hasn't yet, but it hasn't run long enough for me to tell whether that's because it hasn't happened yet or if this PR really fixed it.

rmcgibbo:
> Same with the Travis failure (which is OCL-related).

Yes, there's #1319 for the Travis CPU OpenCL issue. I don't think we've made any progress on that bug. I can't reproduce it on my machine (even with the same OpenCL driver). It could be an OpenMM bug. It could realistically be a bug in the OpenCL driver problem. ¯\_(ツ)_/¯

swails:
> ¯\\\_(ツ)_/¯

I need to save that for later use...

rmcgibbo:
http://www.theatlantic.com/technology/archive/2014/05/the-best-way-to-type-__/371351/

> The Best Way to Type ¯\_(ツ)_/¯
> Every time you shrug, you don’t need to Google, then copy, then paste.

But that is _exactly_ I do every time.

peastman:
But fear not!  There's an official "shrug" emoji coming in Unicode 9.0!

http://emojipedia.org/shrug

swails:
OK -- the simulation has been running for ~20 hours now and still no crash, which is well past where it died last time I ran it (before this PR).  This seems to have fixed at least some of the 980 Ti-related Amoeba instabilities in our simulations.

Should be good to go on my end.

PR #1337, opened 2016-01-14
AMOEBA extrapolated polarization

peastman:
Finally done!  Many thanks @andysim both for the reference implementation and the test cases.  I doubt I'd ever have managed to get this debugged without them!  Here are the current speeds in ns/day for the DHFR benchmark on a Titan X:

mutual: 7.0
extrapolated: 14.5
direct: 32.5


rmcgibbo:
Nice work, Peter. This is very impressive.

andysim:
I concur; this is excellent stuff, @peastman .  I found a small out-of-bounds bug the other day that I'll commit a patch for when I get back to a clean branch.  It's a very trivial fix.  Also, I'll send you the latest and greatest coefficients for mixing the various orders of dipoles.  ExPT is quite good, but it turns out you can get an order of magnitude improvement with some fitting, as we've discussed by email.  The timings will be exactly the same as the benchmarks that you've run, but the method will give much better agreement with the fully iterative method.  Just out of interest, how tightly were the dipoles converged in your benchmarks?

peastman:
Thanks!  That will be great.  The mutual benchmark used 1e-5 as the tolerance.

PR #1338, opened 2016-01-14
Fixed a bug in serializing CustomCentroidBondForce

peastman:


jchodera:
Do we need more test coverage to test serialization/deserialization?

peastman:
I updated the test as well.  The coverage is generally pretty good, but this was a particular situation that missed getting tested.

PR #1339, opened 2016-01-14
Improvements to addHydrogens()

### LINKED ISSUE #1301: addHydrogens - action when specified variant can't be prepared
Follow up from our chat with @peastman and @jchodera yesterday. 

Currently: https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L796

addHydrogens only does a simple check of existing vs expected number of hydrogen atoms bonded to a parent atom. There's no warning or exception if, due to the residue already being in a more 'hydrogen-rich' variant, the variant specified by user cannot be applied. 

HIS is the most worrying case in current behavior, where by running addHydrogens with variant HIE and then running again with variant HID, you will add both HD1 and HE2 hydrogens and achieve a positively charged HIS (HIP). (A more realistic case would be running with None for that residue, then remembering you wanted a particular variant and running with variant 'opposite' to the one automatically chosen in first call). 

I think throwing an exception here would be a good idea. 

- The easiest way seems to:

   Add an else to this if:
 https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L713

 that will do checks for each possible residue with variants analogical to this check for HIS:
https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L729-L736

   and throw an exception for cases where achieving this variant won't be possible. 

- Alternatively, to avoid hard-coding checks for all residues with possible variants we could, in addition to `actualVariants`, have another list `userVariants` and add an `elif len(existing) > len(expected):` to this if: https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L796
 that will throw an exception if this residue's variant is not None in `userVariants`. We also need to repeat the check for HIS though as an else to the `if variant is None:` to avoid the HIS-catastrophy as I described above. 

On a separate note, is there a need for `deleteHydrogens` function? I imagine the exception would direct the user to run deleteHydrogens, followed by repeating addHydrogens?

peastman:
Fixes #1301.  Also adds a HIN variant.

### DIFF
--- a/wrappers/python/simtk/openmm/app/data/hydrogens.xml
+++ b/wrappers/python/simtk/openmm/app/data/hydrogens.xml
@@ -220,6 +220,7 @@
   <Variant name="HID"/>
   <Variant name="HIE"/>
   <Variant name="HIP"/>
+  <Variant name="HIN"/>
   <H name="H" parent="N"/>
   <H name="H2" parent="N" terminal="N"/>
   <H name="H3" parent="N" terminal="N"/>
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -6,7 +6,7 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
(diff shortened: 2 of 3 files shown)


rafwiewiora:
Awesome, thanks!

PR #1348, opened 2016-01-18
Enable wildcard use for Atom Types

rafwiewiora:
A test would be good too - let me complete that.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -320,11 +320,15 @@ class ForceField(object):
                     types.append(None) # Unknown atom class
                 else:
                     types.append(self._atomClasses[attrib[classAttrib]])
-            else:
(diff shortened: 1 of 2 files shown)


rafwiewiora:
Reference https://github.com/pandegroup/openmm/issues/1347

peastman:
Looks good.  A test will also help though.

jchodera:
@rafwiewiora: Just wanted to make sure that you saw the request to add tests.

rafwiewiora:
Yep, thanks @jchodera!

rafwiewiora:
Tests added!

peastman:
Looks like there's an error in your test case:

```
E     File "/home/travis/build/pandegroup/openmm/python/tests/TestForceField.py", line 453
E       ff = ForceField(StringIO(xml))
E                                    ^
E   IndentationError: unindent does not match any outer indentation level
```

rafwiewiora:
Whoops, yeah just noticed the fail, fixing!

rafwiewiora:
Should be good now.

PR #1349, opened 2016-01-19
Fix error in ionicStrength definition in Modeller.addSolvent()

jchodera:
The `ionicStrength` argument was incorrectly being interpreted in determining how many ion pairs to add to the system [here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L522-L529).

Ionic strength is defined as
image
(via [wikipedia](https://en.wikipedia.org/wiki/Ionic_strength#Quantifying_ionic_strength)), but the factor of 1/2 had been omitted from the calculation.

For monovalent ions, the concentration of monovalent salt pairs should equal the ionic strength.

Later, we can generalize this to correctly compute the appropriate number of pairs for more complex salts as well, but we would need to extend the API or find some way to determine the charge on each ionic species.


jchodera:
Whoops---I had also forgotten to change the tests.

jchodera:
Fixed!

PR #1350, opened 2016-01-19
Add methods to retrieve residues without matching forcefield residue templates

jchodera:
I've added two methods to `ForceField` that will make it much easier to both diagnose issues with unmatched residue templates and generate missing residue templates:
* `ForceField.getUnmatchedResidues`: Get a list of `Residue`s in the `topology` that do not have matching residue templates in the forcefield. This list would be useful in debugging issues with missing or incorrect residue templates.
* `ForceField.getUniqueUnmatchedResidues`: Same as above, but returns a unique set of residues. This set would be useful in generating new residue templates.


jchodera:
Also, my apologies---the Atom editor seems to have removed a number of trailing whitespace characters in `TestForcefield.py`, causing it to look like there are more changes than there actually are.

peastman:
Looks like your test case is failing.

jchodera:
> Looks like your test case is failing.

Yep, still a WIP!  Almost got it.

jchodera:
Actually, since we're building `_TemplateData` objects anyway for returning a unique list of missing residues to build templates from, we might as well return these too.

jchodera:
Looks like this is working now.

jchodera:
OK, this works now.  The only failures are unrelated:
```
The following tests FAILED:
	 76 - TestOpenCLCustomManyParticleForceSingle (SEGFAULT)
	 78 - TestOpenCLCustomManyParticleForceDouble (Timeout)
	120 - TestOpenCLDrudeSCFIntegratorDouble (Timeout)
```
I added another test where we use the ability to get blank templates for residues for which templates are missing in order to parameterize new residues before calling `createSystem()`, rather than through the `residueTemplateGenerator()` method.  So now it should be very easy for us to either pre-generate templates (either registering them through the programmatic API or writing out an `ffxml` file) or generate templates on the fly.

All the documentation for these features has warning boxes stating that the API is experimental and subject to change.

jchodera:
After this is merged, I can update our former `gaff2xml` scripts to use this to generate templates to go with the ParmEd-generated `gaff.xml` and we will have small molecule support!

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -569,6 +569,98 @@ def _getResidueTemplateMatches(self, res, bondedToAtom):
...
+
+        Parameters
+        ----------
+        topology : Topology
+            The Topology whose residues are to be checked against the forcefield residue templates.
+
+        Returns
+        -------
+        unmatched_residues : list of Residue
+            List of residue templates from `topology` for which no forcefield residue templates are available.
```
peastman: That should be, "List of Residue objects", not residue templates.

peastman:
I don't understand exactly what `getUniqueUnmatchedResidues()` is supposed to do.  The docstring says, "Returns a unique list of Residue objects".  But since every `Residue` object is unique (by definition), what does that mean?  The real purpose of this method seems to be to generate new templates?  So maybe it would make sense to call it `generateTemplatesForUnmatchedResidues()` or something like that, and revise the docstring to be clearer?

jchodera:
Excellent points.  `generateTemplatesForUnmatchedResidues()` is a much better name for its current use.  I'll fix the method name and docstrings.  Thanks!

wrappers/python/simtk/openmm/app/forcefield.py
jchodera: Thanks!  FIxed.

jchodera:
The failures are more unrelated segfaults.  
```
1/1 Test #77: TestOpenCLCustomManyParticleForceMixed ...***Exception: SegFault 20.09 sec
```
Should be good to merge once the other tests finish.

peastman:
> All the documentation for these features has warning boxes stating that the API is experimental and subject to change.

Are you sure?  I see that in the manual, but not in any of the API docs.

jchodera:
Whoops!  Added.

jchodera:
This should be good to go.  Apologies again for Atom removing trailing whitespaces---it looks like the changes are more extensive than they are.

PR #1353, opened 2016-01-20
Update README.md badges and links

jchodera:
* Update OpenMM URL to http://openmm.org
* Update docs URL to http://docs.openmm.org (previously they pointed to 6.0 docs)
* Update badges

Has anyone else noticed that the anaconda badge points to 6.2.dev0?  I fear the anaconda precedence thing might be accidentally thinking that 6.2.dev0 is the latest version---maybe we should delete just that version from anaconda cloud (but keep other actual releases)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,13 +3,13 @@
 Introduction
 ------------
 
-[OpenMM](https://simtk.org/home/openmm) is a toolkit for molecular simulation. It can be used either as a stand-alone application for running simulations, or as a library you call from your own code. It
+[OpenMM](http://openmm.org) is a toolkit for molecular simulation. It can be used either as a stand-alone application for running simulations, or as a library you call from your own code. It
 provides a combination of extreme flexibility (through custom forces and integrators), openness, and high performance (especially on recent GPUs) that make it truly unique among simulation codes.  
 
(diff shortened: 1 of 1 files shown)


jchodera:
That OpenCL test is still segfaulting:
```
1/2 Test #77: TestOpenCLCustomManyParticleForceMixed ....***Exception: SegFault 23.68 sec
    Start 78: TestOpenCLCustomManyParticleForceDouble
2/2 Test #78: TestOpenCLCustomManyParticleForceDouble ...***Exception: SegFault 23.91 sec
0% tests passed, 2 tests failed out of 2
Total Test time (real) =  47.60 sec
The following tests FAILED:
	 77 - TestOpenCLCustomManyParticleForceMixed (SEGFAULT)
	 78 - TestOpenCLCustomManyParticleForceDouble (SEGFAULT)
```
Everything else passes.

peastman:
Yeah, I've been trying to figure that out.  I haven't been able to reproduce it on any computer I've tried, which makes it difficult.  I'm wondering if there's something strange about the virtual environment travis uses.  Or maybe it's something specific to the type of CPUs they use?  Do we know what model they are?

PR #1355, opened 2016-01-21
Created property for disabling the PME stream

peastman:
If you set this property to "true", that guarantees the separate stream will not be used.  If you set it to "false" (the default), that does *not* guarantee that it *will* be used.  It still will decide for itself based on the GPU.  That's why I called the property "disable" instead of "use".


jchodera:
Awesome, thanks!

PR #1356, opened 2016-01-21
Updated documentation on installing to discuss conda

peastman:


docs-source/usersguide/application.rst
```diff
@@ -38,18 +38,61 @@ Follow these instructions to install OpenMM.  There also is an online
 troubleshooting guide that describes common problems and how to fix them
 (http://wiki.simtk.org/openmm/FAQApp).
 
+
+Installing with Conda
+*********************
+
+There are two ways to install OpenMM: using the Conda package manager, or with
```
jchodera: Perhaps link to http://conda.pydata.org/docs/ when you mention 'Conda'?

jchodera: Should the paragraph "There are two ways to install..." come *before* the "Installing with Conda" subsection?

docs-source/usersguide/application.rst
```diff
@@ -38,20 +38,64 @@ Follow these instructions to install OpenMM.  There also is an online
...
 **********************
 
-OpenMM works on Mac OS X 10.7 or later.
-
-.. note::
-   The OpenCL implementations on all recent versions of Mac OS X contain serious
-   bugs that make them unsuitable for use with OpenMM.  GPU acceleration is
-   therefore only supported with the CUDA platform.  This limits it to only Nvidia
-   GPUs, not AMD or Intel GPUs.
+OpenMM works on Mac OS X 10.7 or later.  OpenCL is supported on OS X 10.10.5 or
```
jchodera: Does the OpenCL platform check the OS version and not initialize itself if earlier versions are used?

If not, we may want to add a bigger `.. warning::` around the "OpenCL is supported on OS X 10.10.5".

peastman: Yes, the OpenCL platform won't register itself if the OS version is too low.

Issue #1357, opened 2016-01-21
Update .travis.yml and openmm-dev builds to use CUDA 7.5


jchodera:
Buried in #1352 is the suggestion that OpenMM 7.0 will use CUDA 7.5.

We should first update `.travis.yml` and the `openmm-dev` build environments in https://github.com/omnia-md/conda-dev-recipes to use CUDA 7.5.

jchodera:
Working on `.travis.yml` in #1358.

Updating `conda-dev-recipes` and our `conda-recipes` build systems to use CUDA 7.5 seems like it will be much harder:
* NVIDIA has changed where they hide the `osx` network install files
* We will need to update the `Dockerfile` build scripts for `rmcgibbo/centos5-build-box` to use CUDA 7.5.  Given how hard it was to get 7.0 on there, it may take some tinkering.

jchodera:
OK, `.travis.yml` has been upgraded in #1358.

What about updating which version of the AMD APP SDK we use?  We're currently on 2.9.1 for building conda packages---should that be updated to 3.0?

peastman:
Sure, that would be reasonable.

jchodera:
I think I've done this now.  conda packages rebuilding.

Issue #1282, opened 2015-12-03
Final features for 7.0


peastman:
I think we're getting close to the point where we declare 7.0 feature complete and put out a beta.  Are there any remaining features we want to wait for?  I know of a few things that are in progress that we should consider.

1. Extrapolated polarization for AMOEBA (#1153).
2. Convert API docs to Sphinx (#1238).
3. Changes to how we protonate residues that coordinate ions (#1176).
4. Changes to support constant pH (#1278).

Are there any others we should consider?

Do people have estimates of how long each of the above will take to complete?  For extrapolated polarization, I'm still waiting for @andysim to complete the reference implementation.  Once that's done, it should take me very little time to port it to CUDA.  How about the others?

andysim:
The polarization code that I pushed is correct, but I'm working right now to make it a little more efficient.  I know the reference implementation is not supposed to be fast, but I want to provide a sensible algorithm to base the CUDA version on.  I expect to have it done this week.

peastman:
Great, thanks!  Also, having test cases for it will be essential.  That's even more important than efficiency.

andysim:
I couldn't agree more.  I know how to add to the OpenMM unit tests, but may need some help hooking into the finite difference test machinery.

peastman:
Do you mean for checking consistency between forces and energies?  Here's a good example of a test case that does that:

https://github.com/pandegroup/openmm/blob/master/tests/TestCustomManyParticleForce.h#L118-L138

andysim:
Oh, that's easy.  I found [this page](https://github.com/pandegroup/openmm/issues/575) when I was searching for OpenMM's validation mechanisms, so I thought I'd have to do something to hook into it.

jchodera:
My wishlist is:

### Pretty close
* Multisite ions for Amber (#1032) - @rafwiewiora is working on this with @peastman
* AmberTools15 forcefields (#503) - @rafwiewiora is working on this with @swails
* CHARMM 36 forcefields (#685) - @chayast is working on this with @swails

### Less close
* support for small molecules via `ForceField` and `ffxml` route, which needs better mix-and-match support (especially for GB) (#1168)
* constant-pH simulation support (#172; #494; #1278) - @bas-rustenburg is working on this, but it will take a great deal of work

### No idea
* Support for external hooks for `ForceField` to call out to parameterize small molecules (#488); ideally, we would include support for `antechamber` in `ambermini`
* Deterministic force and energy computation (necessary for #916)
[...]

PR #1358, opened 2016-01-21
Update travis to use CUDA 7.5

jchodera:
cc: #1282 #1357 

This requires updating from the travis `precise` (ubuntu 12.04) build environment to `trusty` (ubuntu 14.04) since the CUDA 7.5 distribution is not available via the same mechanism for 12.04:
https://docs.travis-ci.com/user/trusty-ci-environment/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,7 +125,7 @@ before_install:
 
   - if [[ "$CUDA" == "true" ]]; then
       wget "http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/cuda-repo-ubuntu1404_${CUDA_VERSION}_amd64.deb";
-      sudo dpkg -i cuda-repo-ubuntu1204_${CUDA_VERSION}_amd64.deb;
+      sudo dpkg -i cuda-repo-ubuntu1404_${CUDA_VERSION}_amd64.deb;
       sudo apt-get update -qq;
       export CUDA_APT=${CUDA_VERSION%-*};
       export CUDA_APT=${CUDA_APT/./-};
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,10 +37,11 @@ matrix:
       addons: {apt: {packages: []}}
 
     - sudo: required
+      dist: trusty
       env: ==CUDA_COMPILE==
            CUDA=true
            OPENCL=false
-           CUDA_VERSION="7.0-28"
+           CUDA_VERSION="7.5-18"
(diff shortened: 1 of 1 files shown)


jchodera:
Wow, I think that actually worked.

peastman:
I don't think it's actually compiling any CUDA code.  I see that there's a build called "CUDA_COMPILE" and that it specifies "CUDA=true".  But looking through the build log, I don't see any of the CUDA platform getting built.

swails:
haha.  Straight out of a movie :)

jchodera:
Good point.  Added some more CUDA paths, and am printing out `CMakeCache.txt` to see if we can debug what is going on.

jchodera:
The resulting `CMakeCache.txt` has
```
//Build OpenMMCuda library for Nvidia GPUs
OPENMM_BUILD_CUDA_LIB:BOOL=OFF
```
Do I need to explicitly switch that on?

peastman:
If the FindCuda script is able to find where CUDA is installed, that should be enabled by default.  You could enable it manually (along with all the similar ones like OPENMM_BUILD_AMOEBA_CUDA_LIB), but you'll also need to set various CMake variables to tell it where the CUDA libraries are found.  A better approach is probably to try to figure out why CUDA isn't getting found in the first place.

jchodera:
Here's what we are defining:
```
+             -DOPENCL_LIBRARY=/usr/local/cuda-7.5/lib64/libOpenCL.so
 +             -DCUDA_CUDART_LIBRARY=/usr/local/cuda-7.5/lib64/libcudart.so
 +             -DCUDA_NVCC_EXECUTABLE=/usr/local/cuda-7.5/bin/nvcc
 +             -DCUDA_SDK_ROOT_DIR=/usr/local/cuda-7.5/
 +             -DCUDA_TOOLKIT_INCLUDE=/usr/local/cuda-7.5/include
 +             -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-7.5/"
```
> A better approach is probably to try to figure out why CUDA isn't getting found in the first place.

Any ideas on how to do this?

jchodera:
Found the problem.  Removed the debug stuff.  Should be good to merge after the latest round of tests finishes.

jchodera:
OK, the only failure is the same segfault that has been bugging us:
```
1/1 Test #77: TestOpenCLCustomManyParticleForceMixed ...***Exception: SegFault 19.60 sec
```
Things seem to compile with CUDA now.

PR #1359, opened 2016-01-21
Add a method for retrieving list of matching templates to ForceField

jchodera:
This adds one more experimental convenience function `getMatchingTemplates(topology)` to `ForceField`.

This method returns the list of templates that match the corresponding residues in `topology`.

This is valuable in debugging which templates are matching which residues, as well as testing that changes to `Topology` were implemented correctly for transforming a residue into another recognized residue.

Example:
```python
    pdb = PDBFile('input.pdb')
    forcefield = ForceField('amber99sb.xml', 'tip3p.xml')
    templates = forcefield.getMatchingTemplates(topology)
    for (residue, template) in zip(pdb.topology.residues(), templates):
        print "Residue %d %s matched template %s" % (residue.id, residue.name, template.name)
```

@juliebehr: Can you check if this API would work for you too?

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -1976,7 +1976,7 @@ residue template defined in the :class:`ForceField`.
     forcefield = ForceField('amber99sb.xml', 'tip3p.xml')
     unmatched_residues = forcefield.getUnmatchedResidues(topology)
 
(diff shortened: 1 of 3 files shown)


juliebehr:
+1, this should be all I need!

PR #1360, opened 2016-01-22
Debugging OpenCL errors on travis

peastman:
This isn't meant to be merged.  I'm just using this PR as a way to run debugging code on travis to try to figure out what's causing the errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
 matrix:
   include:
     - sudo: required
+      dist: trusty
       env: ==CPU_OPENCL==
            OPENCL=true
            CUDA=false
@@ -33,7 +34,9 @@ matrix:
            -DOPENMM_BUILD_AMOEBA_PLUGIN=OFF
            -DOPENMM_BUILD_PYTHON_WRAPPERS=OFF
            -DOPENMM_BUILD_C_AND_FORTRAN_WRAPPERS=OFF
-           -DOPENMM_BUILD_EXAMPLES=OFF"
+           -DOPENMM_BUILD_EXAMPLES=OFF
+           -DOPENCL_INCLUDE_DIR=$HOME/AMDAPPSDK/include
+           -DOPENCL_LIBRARY=$HOME/AMDAPPSDK/lib/x86_64/libOpenCL.so"
       addons: {apt: {packages: []}}
 
     - sudo: required
@@ -107,9 +110,16 @@ before_install:
       sudo easy_install pytest;
     fi
   - if [[ "$OPENCL" == "true" ]]; then
-      sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu $(lsb_release -sc) main universe restricted multiverse";
-      sudo apt-get -yq update > /dev/null 2>&1 ;
-      sudo apt-get install -qq fglrx=2:8.960-0ubuntu1 opencl-headers;
+      wget https://jenkins.choderalab.org/userContent/AMD-APP-SDKInstaller-v3.0.130.135-GA-linux64.tar.bz2;
+      tar -xjf AMD-APP-SDK*.tar.bz2;
+      AMDAPPSDK=${HOME}/AMDAPPSDK;
+      export OPENCL_VENDOR_PATH=${AMDAPPSDK}/etc/OpenCL/vendors;
+      mkdir -p ${OPENCL_VENDOR_PATH};
+      sh AMD-APP-SDK*.sh --tar -xf -C ${AMDAPPSDK};
+      echo libamdocl64.so > ${OPENCL_VENDOR_PATH}/amdocl64.icd;
+      export LD_LIBRARY_PATH=${AMDAPPSDK}/lib/x86_64:${LD_LIBRARY_PATH};
+      chmod +x ${AMDAPPSDK}/bin/x86_64/clinfo;
+      ${AMDAPPSDK}/bin/x86_64/clinfo;
     fi
   # Install swig for Python wrappers. However, testing CUDA and OpenCL, we
   # skip the Python wrapper for speed. We're not using anaconda python,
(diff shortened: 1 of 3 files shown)


swails:
This would be pretty awesome to merge though :)

peastman:
Well, I found one problem and now it isn't segfaulting anymore.  Unfortunately it's still giving the wrong answer...

peastman:
This really looks to me like a bug in the OpenCL implementation.  See https://github.com/peastman/openmm/blob/debugcl/platforms/opencl/src/kernels/customManyParticle.cl.  Line 270 never gets executed for two atoms (464 and 465).  You can see that from the log at https://travis-ci.org/pandegroup/openmm/jobs/104710149.  There's no output for those two atoms from the printf at line 269.  On the other hand, the printf at the start of the loop (line 185) does produce output for them.  There's also a printf in the middle of the loop (line 259) that only gets executed for atom 464, and it produces the correct output.

I've been trying to figure out what version of AMD's OpenCL we're using.  Here's the command we use to install it:

```
sudo apt-get install -qq fglrx=2:8.960-0ubuntu1 opencl-headers
```

Is 8.960 the version number?  If so, that's a very old version.  Perhaps we need to upgrade?

peastman:
Ok, what did I do wrong?  Here's the code I added:

```
  - if [[ "$OPENCL" == "true" ]]; then
      wget https://jenkins.choderalab.org/userContent/AMD-APP-SDKInstaller-v3.0.130.135-GA-linux64.tar.bz2;
      tar -xjf AMD-APP-SDK*.tar.bz2;
      AMDAPPSDK=${HOME}/AMDAPPSDK;
      export OPENCL_VENDOR_PATH=${AMDAPPSDK}/etc/OpenCL/vendors;
      mkdir -p ${OPENCL_VENDOR_PATH};
      sh AMD-APP-SDK*.sh --tar -xf -C ${AMDAPPSDK};
      echo libamdocl64.so > ${OPENCL_VENDOR_PATH}/amdocl64.icd;
      export LD_LIBRARY_PATH=${AMDAPPSDK}/lib/x86_64:${LD_LIBRARY_PATH};
      chmod +x ${AMDAPPSDK}/bin/x86_64/clinfo;
      ${AMDAPPSDK}/bin/x86_64/clinfo;
    fi
```

And here's the log: https://api.travis-ci.org/jobs/108140361/log.txt?deansi=true.  You can see where it executed the first line of that block and (successfully, as far as I can tell) downloaded the file.  But there's no indication it executed any of the rest of that block.

swails:
> But there's no indication it executed any of the rest of that block.

What do you mean?  Isn't the output of clinfo clearly there?  What were you expecting to see that you didn't?

jchodera:
Travis expects a single line of `bash` shell script per YAML line, and does weird things to process multi-line things, which makes `if` statements crazy hard to debug:
https://github.com/travis-ci/travis-ci/issues/497

jchodera:
I'd also add the `--quiet` line to `wget` to suppress the `####...` output.

jchodera:
Whoa, it looks like it actually printed out the output from `${AMDAPPSDK}/bin/x86_64/clinfo` in that multi-line `if` block, though!

peastman:
Well, I just copied most of that block from the repository you referenced.  I was expecting the installer to produce some output, but maybe one of those command line flags told it not to.  I now think the problem is that CMake isn't locating it, so I'm working on that now.

swails:
You can add the -v flag to the tar command to get a verbose dump of the tarball contents.  That should at least produce some output you can check against.  But it's not obvious to me that the commands between the wget and the clinfo actually produce any output...

peastman:
Ok!  I have it successfully installing the newer OpenCL.  And... the test still fails.

Sigh.

jchodera:
What if I have the platform tests packaged (or at least that test) packaged into the  openmm-dev conda package. That would make it trivial to try on any platform we can install conda.

jchodera:
Still, progress!

Let's find another machine to reproduce the bug on...

swails:
I was able to repro on a Vagrant machine I built:

```
vagrant@precise64:~/openmm$ ./TestOpenCLCustomManyParticleForce mixed 
Segmentation fault
vagrant@precise64:~/openmm$ ./TestOpenCLCustomManyParticleForce single
Segmentation fault
```

I'll check out this branch to test it out.

swails:
Still fails on this branch -- log file here: https://www.dropbox.com/s/x3g06h3auuaptdf/multiparticle_force_single.log?dl=0

Anything you want me to do or try?  Has anyone been able to repro this before?

swails:
If it'll help, I can give @peastman access to this machine until it goes down on Friday.

rmcgibbo:
What OpenCL driver are you using?

swails:
```
vagrant@precise64:~/openmm$ ./TestOpenCLDeviceQuery 
OpenCL devices:

OpenCLPlatformIndex 0, OpenCLDeviceIndex 0: "AMD FX(tm)-6100 Six-Core Processor             "
    CL_DEVICE_VENDOR                 = AuthenticAMD
    CL_DEVICE_MAX_COMPUTE_UNITS      = 2
    CL_DEVICE_GLOBAL_MEM_SIZE        = 4147060736
    CL_DEVICE_MAX_CLOCK_FREQUENCY    = 3311
    estimatedSpeed                   = 6622
```

Does this have the information you're looking for?  The output looks quite similar to what Travis prints.  I'm running the full test suite now, but the first couple tests have already passed.  So I suspect it's a reliable repro VM for this issue.

peastman:
I finally found a way to reproduce this locally!  It seems to be related to the number of threads.  If I tell it to pretend my CPU only has one core, then I can reproduce it.  And after a lot of experimenting, I found that if I add a barrier at a particular place, the error goes away.  *I have no idea why that should make a difference*.  I'm still inclined to think it's a compiler bug, and this is just working around it.  But for now, I think that's good enough.  So I'm going to clean up the debugging code, rebase to get a cleaner version history, and then merge it.

swails:
How much does the barrier slow stuff down?

peastman:
Completely negligible.  It's at the end of an outermost loop.

swails:
:). Puzzling. 

Awesome work!

swails:
OK!  This is done!  The failure is the same error I was seeing on my Mac.  So I think this is ready to merge, unless you want me to try my hand at fixing the errors.

swails:
Sorry, last post was meant for a different repo... :-[

PR #1368, opened 2016-01-27
Compare 1-4 scaling factors by looking at size of difference

### LINKED ISSUE #1367: ForceField requires coulomb14scale and lj14scale in <NonbondedForce> match *exactly*
Cross-posted this to https://github.com/ParmEd/ParmEd/issues/575

`simtk.openmm.app.ForceField` seems to require that `coulomb14scale` and `lj14scale` match *exactly* between `ffxml` files.

Trying to combine the old `amber99sb.xml` with a newly-converted `gaff.xml` results in:
```
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/choderalab/openmoltools/openmoltools.jchodera/openmoltools/tests/test_forcefield_generators.py", line 208, in test_gaffResidueTemplateGenerator
    forcefield = ForceField('amber99sb.xml', 'tip3p.xml', 'gaff.xml')
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/simtk/openmm/app/forcefield.py", line 124, in __init__
    self.loadFile(file)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/simtk/openmm/app/forcefield.py", line 199, in loadFile
    parsers[child.tag](child, self)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/simtk/openmm/app/forcefield.py", line 1608, in parseElement
    raise ValueError('Found multiple NonbondedForce tags with different 1-4 scales')
ValueError: Found multiple NonbondedForce tags with different 1-4 scales
```
This is because the `amber99sb.xml` file contains
```XML
 <NonbondedForce coulomb14scale="0.833333" lj14scale="0.5">
```
while the newly-converted `gaff.xml` contains:
```
 <NonbondedForce coulomb14scale="0.833333333333" lj14scale="0.5">
```
If we convert all of the files with the *same* version of the code, then we should be fine, but we may want to either (1) modify the behavior of `ForceField` to have a tolerance for this check, or (2) truncate writing of these to some fixed precision to not cause problems down the road.

swails:
Fixes #1367 

This commit also changes a bunch of old-style Python classes to new-style (Py3 only has the latter).

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -345,7 +345,7 @@ class ForceField(object):
             index += 1
         return torsion
 
-    class _SystemData:
+    class _SystemData(object):
         """Inner class used to encapsulate data about the system being created."""
         def __init__(self):
             self.atomType = {}
@@ -371,7 +371,7 @@ class ForceField(object):
                 self.constraints[key] = distance
                 system.addConstraint(atom1, atom2, distance)
 
-    class _TemplateData:
+    class _TemplateData(object):
         """Inner class used to encapsulate data about a residue template definition."""
         def __init__(self, name):
             self.name = name
@@ -413,7 +413,7 @@ class ForceField(object):
             atom = self.getAtomIndexByName(atom_name)
             self.addExternalBond(atom)
 
-    class _TemplateAtomData:
+    class _TemplateAtomData(object):
         """Inner class used to encapsulate data about an atom in a residue template definition."""
         def __init__(self, name, type, element, parameters={}):
             self.name = name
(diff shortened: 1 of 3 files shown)


peastman:
I take it you're still making changes to this?  Let me know when it's ready.

swails:
There aren't any more changes I want to make, but I don't know why I'm getting the failures I'm getting.  [Here's one](https://travis-ci.org/pandegroup/openmm/jobs/105235602#L4810) associated with multi-site ion treatment in Modeller?  Is that stochastic and can occasionally fail?  How is that in any way related to my changes?

That failed twice, on different versions of Python, but passed in other tests... I can't make sense out of that.  The third failure was the standard many-particle OpenCL issue that's known on Travis.

peastman:
I don't see any way your changes could have caused that failure.  But it isn't something I've seen failing recently, so to have it fail twice is very suspicious.  I'm going to try rerunning all the tests and see what happens.

peastman:
This time everything passed, except for the standard OpenCL error.  Since I'm pretty sure the other errors were unrelated to this, I'll go ahead and merge it.

PR #1370, opened 2016-01-27
Create alternative constructor for Simulation from XML files

swails:
This replaces #872.

The actual changes are actually pretty small.  `Simulation` now accepts `None` for its topology argument, and falls back to `system.usesPeriodicBoundaryConditions` to determine the value of `enforceBox` in `Context.getState` calls.  It also accepts either a `System` or `str` input for its `system` and `Integrator` or `str` for its `integrator` arguments.  A string is interpreted as a XML file name in both cases.

The only reason I bothered with an alternate constructor is because `topology` is a *required* positional argument to the `Simulation` initializer, and the `fromXmlFiles` constructor is a clearer API for instantiating a `Simulation` from XML files (even if you can do the exact same thing with the standard constructor).

The `State` XML file is optional.

What do people think?

### DIFF
--- a/wrappers/python/simtk/openmm/app/simulation.py
+++ b/wrappers/python/simtk/openmm/app/simulation.py
@@ -74,9 +74,12 @@ class Simulation(object):
         platformProperties : map=None
             If not None, a set of platform-specific properties to pass to the
             Context's constructor
-        state : XML file name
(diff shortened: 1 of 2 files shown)


swails:
Failure here is still the OpenCL failure.  Should be good to go.

PR #1372, opened 2016-01-28
Marked usesPeriodicBoundaryConditions() as const

peastman:


swails:
Good call.  My bad...

PR #1373, opened 2016-01-28
Residue template overloading / User can specify template if multiple match / Refactor XML loading / No multiple gens

rafwiewiora:


jchodera:
This definitely seems like an improvement!  But deferred processing of the ffxml files might be even better, since we can still load additional files afterwards by calling `ForceField.load_file()`.

rafwiewiora:
What do you mean by deferred processing?

jchodera:
"Deferred processing" would mean caching the XML files and then processing them (as you do here) all at once by section just before we need to use them to create a system.  Only when `createSystem()` is called would we actually parse the XML files.

This has some advantages, including satisfying @peastman's desire that the files could really be loaded in any order.  On the other hand, it has disadvantages, such as making it much harder to track down errors---for example, a syntax error in the XML file would no longer occur during `load_file()` but instead during `createSystem()`.

rafwiewiora:
Ah I see!

peastman:
I think we first need to figure out the question of how/if to handle overriding.  This change is compatible with some of the approaches that have been suggested, but not with others.

rafwiewiora:
Sure thing, let's leave it hanging for now!

rafwiewiora:
@peastman this is done and ready for you to review. 

We have:
* collect all xml trees from all files passed in a list, then load in by section: Atom Type - Residues - Forces - Scripts from each file 
* protected generators against creating copies for the same type of force

The big diff in TestForceField.py is because I moved two tests that were recently added to `AmoebaTestForceField` class mistakenly, to the `TestForceField` class.

peastman:
This looks good.  Since it's a 7.1 feature, though, I'm going to wait to merge this until after 7.0 is released.

rafwiewiora:
ok!

rafwiewiora:
@peastman I have not committed:
* overloading
* user-specifies which template to use via `residueTemplates` dictionary arg to `createSystem`

here. https://github.com/rafwiewiora/openmm/commit/557faa7d6c08aa39051985306a33fb60342610c9 for just that diff. 

Can you review please? I will add tests when we're okayed on the final shape. 

I called it overloading, seemed to fit best from a choice of overloading, overwriting and overriding. 
The variable carried in the template objects is `overloadLevel` and is specified in the XML in the residue tag like such:
```
<Residue name='Ag' overload='1'/>
```

default `overloadLevel` is 0 and does not require specifying in XML. 

Residues to 'participate' in the overloading must have the same name. The overloading is done during XML loading, rather than on `createSystem` - if residue of same name is found only the one with higher `overloadLevel` is kept. If they have the same level - exception. 

On `createSystem()` if >1 templates match a residue - exception. UNLESS user has specified which one to use via the `residueTemplates` argument. `residueTemplates = {residue_object : template_name}`. Match of the user-specified template is of course still checked - if doesn't match - exception. 

Also tagging @jchodera - can you have a look whether this is good for what you envisaged?

rafwiewiora:
Fixed a thing and overwrote commit - new one is this https://github.com/rafwiewiora/openmm/commit/c0e1335302bf89bd37fc947585bfd845d7474a3c. Please change to this one if opened the previous one before seeing this comment.

peastman:
That looks really good.  The one thing that's missing, of course, is a test case.

rafwiewiora:
> That looks really good. The one thing that's missing, of course, is a test case.

Glad you like it! :) Definitely, writing it now.

rafwiewiora:
Tests added.

wrappers/python/tests/TestForceField.py
```diff
@@ -448,6 +448,179 @@ def test_getMatchingTemplates(self):
...
+ </NonbondedForce>
+</ForceField>"""
+
+        pdb_string = "ATOM      1 FE    FE A   1      20.956  27.448 -29.067  1.00  0.00          Fe"
+        ff = ForceField(StringIO(ffxml))
+        pdb = PDBFile(StringIO(pdb_string))
+
+        self.assertRaises(Exception, lambda: ff.createSystem(pdb.topology))
+        ff.createSystem(pdb.topology, residueTemplates={list(pdb.topology.residues())[0] : 'FE2'})
+        ff.createSystem(pdb.topology, residueTemplates={list(pdb.topology.residues())[0] : 'FE'})
```
peastman: This doesn't verify that it used the correct template, just that it didn't throw an exception.  How about checking the charge on the atom to make sure it's correct?

jchodera:
This is good stuff, but I think we still haven't fixed the primary problem that processing *must* be deferred until right before `createSystem` is called if we want behavior to be independent of the order in which files our read.  The residue definition overloading behavior checks to see which definition has higher priority, but this check happens when `loadFile` is called.

Suppose we did this:
```
forcefield = ForceField(priority1file)
forcefield.loadFile(priority2file)
forcefield.loadFile(priority3file)
```
or instead did
```
forcefield = ForceField(priority1file)
forcefield.loadFile(priority3file)
forcefield.loadFile(priority2file)
```
we would get different behavior, wouldn't we?

wrappers/python/tests/TestForceField.py
rafwiewiora: Sure, good point!

rafwiewiora:
@jchodera you will get the same behavior. Note that:
a) every single residue template event calls `ForceField.registerResidueTemplate` (https://github.com/rafwiewiora/openmm/blob/forcefield_xml/wrappers/python/simtk/openmm/app/forcefield.py#L204)

b) `registerResidueTemplate` on such every call 'does the overloading' (https://github.com/rafwiewiora/openmm/blob/forcefield_xml/wrappers/python/simtk/openmm/app/forcefield.py#L252).

peastman:
It shouldn't matter.  At the end of those calls, the ForceField contains only one template with that name, and it's the one with the highest priority.  That's true with either order.

jchodera:
Ah, OK---I had read too quickly. Thanks!

rafwiewiora:
Also note that with this PR order of files in the `ForceField` call no longer matters as far as atom types having had to be loaded before loading in residue templates that use them etc. 

The only thing is that your `.loadFile` calls can't 'look into the future' (of course). But you can pass a tuple of multiple files to `.loadFile` if you have types and templates spread between two for example.

I think the actual point where this pre-processing vs `createSystem` processing will start to matter will be when I get to the type/class refactoring and we'll have to decide how belonging to each group depends on the loading history. But let's leave that for now. Refactoring the whole thing to have `createSystem`-time processing looks like a lot of work anyway, and I value having the final data available from the `ForceField` object before creating system too. (plus with the createSystem-time processing that call would have to modify the `ForceField` object itself I imagine - and that doesn't sound so pretty to me).

rafwiewiora:
Charge checking in test added.

rafwiewiora:
Travis pass, AppVeyor - no idea.

jchodera:
@peastman: Rather than hold up development by keeping multiple PRs on hold for weeks while we release an OpenMM 7.0 RC and final release, can we create a new `openmm71` branch and merge these PRs into there? That will allow us to create a conda dev package with forcefield modifications we can test and continue to develop with.

rafwiewiora:
+1

peastman:
I think we're just about ready to build the release candidate.  Once we do that, let's make a branch that any further 7.0 changes will go into, and move the main branch to 7.1.

jchodera:
> I think we're just about ready to build the release candidate. Once we do that, let's make a branch that any further 7.0 changes will go into, and move the main branch to 7.1.

Thanks!

jchodera:
The RC is cut. Can we create this branch and merge the PR now?

jchodera:
@peastman: Let's get this merged now that the RC is out.

PR #1376, opened 2016-02-02
Hopefully improve the docstring so the generated docs will improve

swails:
See http://docs.openmm.org/development/api-python/generated/simtk.openmm.mtsintegrator.MTSIntegrator.html#
for the rendered results that are less than ideal.

### DIFF
--- a/wrappers/python/simtk/openmm/mtsintegrator.py
+++ b/wrappers/python/simtk/openmm/mtsintegrator.py
@@ -47,9 +47,9 @@ class MTSIntegrator(CustomIntegrator):
     of the force group and the frequency (as a fraction of the outermost time step) at
     which to evaluate it.  For example:
 
-    <pre>
+    ```
     integrator = MTSIntegrator(4*femtoseconds, [(0,1), (1,2), (2,8)])
-    </pre>
+    ```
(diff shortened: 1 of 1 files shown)


peastman:
Does Sphinx recognize \`\`\` syntax?  For what it's worth, I think the "recommended" way of marking code examples is just to indent them four spaces.

rmcgibbo:
I think it will also come out correctly if you put it inside `<tt><pre>`, instead of just `<pre>`

rmcgibbo:
But yeah, @peastman is right that "```" is Markdown syntax, not reStructuredText.

swails:
Yea, I was hoping they'd be enough alike... :).  I'll switch to indentation.

Also, remember this isn't going through the doxygen translator, I don't think (since it's all living in pure Python).

swails:
How does that look?

rmcgibbo:
Right, but it does get piped through this code: https://github.com/pandegroup/openmm/blob/master/docs-source/api-python/process-docstring.py

rmcgibbo:
untitled

peastman:
Ok, that look good.

PR #1378, opened 2016-02-03
Update OpenCL in Travis CI

rmcgibbo:
Perhaps this will fix the Travis CPU OpenCL bug. Or perhaps not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,7 +110,7 @@ before_install:
   - if [[ "$OPENCL" == "true" ]]; then
       sudo add-apt-repository "deb http://archive.ubuntu.com/ubuntu $(lsb_release -sc) main universe restricted multiverse";
       sudo apt-get -yq update > /dev/null 2>&1 ;
-      sudo apt-get install -qq 2:15.200-0ubuntu0.5 opencl-headers;
+      sudo apt-get install -qq fglrx=2:15.200-0ubuntu0.5 opencl-headers;
     fi
   # Install swig for Python wrappers. However, testing CUDA and OpenCL, we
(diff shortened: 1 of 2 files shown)


peastman:
I did also find one genuine error in the code.  This is the fix for it:

https://github.com/peastman/openmm/commit/2e53f824f87b7eb77eb371e6e6e36110ab64bf25

However, even after fixing it, the test still failed.  Try adding that fix to this PR.  Perhaps the combination of it plus the newer OpenCL will work.

rmcgibbo:
Okay, I cherry-picked that in. From looking at the travis log, I suspect they don't have so many trusty VMs, because both the branches of the matrix that are supposed to run on trusty VMs are just waiting to start.

rmcgibbo:
(https://www.traviscistatus.com/incidents/lpgbdfcmjhpk)

peastman:
```
E: Unable to locate package 2
E: Couldn't find any package by regex '2:15.200-0ubuntu0.5'
```

rmcgibbo:
oh, I probably shouldn't have removed the string `fglrx`. :-)

rmcgibbo:
Okay, this is obviously still broken but in a different way, and I don't know how to fix it. Someone else can give it a shot if they want.

peastman:
I'll mess around a bit and see what I can figure out.

jchodera:
https://jenkins.choderalab.org/userContent/AMD-APP-SDKInstaller-v3.0.130.135-GA-linux64.tar.bz2

peastman:
Thanks!  That should hopefully make it easier.

peastman:
Closing, since #1360 did this.

PR #1381, opened 2016-02-04
Parallelized the code for downloading positions

peastman:
Querying the atom positions involves a nontrivial amount of processing on the CPU.  This parallelizes that.  In most cases the effect is negligible, but in cases where you need to query the positions at every time step, it does make a difference.  (This came up in the context of the PLUMED plugin.)


jchodera:
Cool!

PR #1385, opened 2016-02-05
bugfix on the ansiotropic polarizability for Drude force field

jing-huang:
The anisotropic polarizability in Drude FF is defined using four particles, first anisotropic term aniso12 using particle 1 and 2, and second term aniso34 using particle 3 and 4. There is a bug in the cuda and opencl platforms that the aniso34 are defined wrongly using the direction of particle 2 and particle 3. This bug escapes our previous tests using water and protein systems because of symmetry in the anisotropy definition in Drude protein force field, but surfaces when we compared energies and forces of nucleic acid systems with Drude FF. 

The reference platform is correct.


peastman:
Thanks!

PR #1384, opened 2016-02-04
Fixed error in computing global variables

peastman:
Fixes #1380.  This error was introduced post-6.3, so it isn't in any released version.

PR #1386, opened 2016-02-05
Fixed another error in CustomIntegrator

peastman:
This is closely related to #1384.  It caused the wrong value to be used for "dt" in ComputeGlobal steps.


jchodera:
I've been wondering if there is a more automated way to thoroughly test `CustomIntegrator`.  I know you're adding tests as you uncover issues, but is there potentially a way to combinatorially generate tests?  And would it be useful to do so?

peastman:
What did you have in mind?  There's a lot of step types, and a lot of variables (and types of variables) that can be used in each step, and bugs can arise from subtle interactions between successive steps.  So it's a huge search space.  Can you think of a systematic way to produce useful tests?

jchodera:
I was wondering if a combinatorial approach of having a set of operations on a common set of variables and attempting to validate all permutations might be useful.  But it may not be worth the effort.

PR #1389, opened 2016-02-06
Prevent Travis from reporting a failed build if the OpenCL platform tests fail

swails:
At the moment, it fails every time and we always ignore it. Once we stop
*expecting* it to fail, we should remove this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
 
 matrix:
   allow_failures:
-      - env: ==CPU_OPENCL==
+      - env: OPENCL=true
   include:
     - sudo: required
       env: ==CPU_OPENCL==
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ matrix:
            -DOPENMM_BUILD_C_AND_FORTRAN_WRAPPERS=OFF
            -DOPENMM_BUILD_EXAMPLES=OFF"
       addons: {apt: {packages: []}}
+      allow_failure: true
 
     - sudo: required
       dist: trusty
@@ -97,10 +98,6 @@ matrix:
            CC=$CCACHE/gcc
            CXX=$CCACHE/g++
            CMAKE_FLAGS=""
-
-matrix:
-  allow_failures:
-      - env: OPENCL=true
 
 before_install:
   - START_TIME=$(date +%s)
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,6 +97,8 @@ matrix:
            CC=$CCACHE/gcc
            CXX=$CCACHE/g++
            CMAKE_FLAGS=""
+  allowfailures:
+    - env: ==CPU_OPENCL==
 
 before_install:
   - START_TIME=$(date +%s)
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ matrix:
            -DOPENMM_BUILD_C_AND_FORTRAN_WRAPPERS=OFF
            -DOPENMM_BUILD_EXAMPLES=OFF"
       addons: {apt: {packages: []}}
-      allow_failure: true
+      allow_failures: on
 
     - sudo: required
       dist: trusty
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ env:
 
 
 matrix:
+  allowfailures:
+      - env: OPENCL=true
   include:
     - sudo: required
       env: ==CPU_OPENCL==
@@ -97,10 +99,6 @@ matrix:
            CC=$CCACHE/gcc
            CXX=$CCACHE/g++
            CMAKE_FLAGS=""
-
-matrix:
-  allowfailures:
-    - env: OPENCL=true
 
 before_install:
   - START_TIME=$(date +%s)
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ env:
 
 
 matrix:
-  allowfailures:
-      - env: OPENCL=true
+  allow_failures:
+      - env: ==CPU_OPENCL==
   include:
     - sudo: required
       env: ==CPU_OPENCL==
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,6 +98,7 @@ matrix:
            CXX=$CCACHE/g++
            CMAKE_FLAGS=""
 
+matrix:
   allow_failures:
       - env: OPENCL=true
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ env:
 
 
 matrix:
-  allow_failures:
-      - env: OPENCL=true
   include:
     - sudo: required
       env: ==CPU_OPENCL==
@@ -99,6 +97,9 @@ matrix:
            CC=$CCACHE/gcc
            CXX=$CCACHE/g++
            CMAKE_FLAGS=""
+
+  allow_failures:
+      - env: OPENCL=true
 
 before_install:
   - START_TIME=$(date +%s)
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,8 +97,10 @@ matrix:
            CC=$CCACHE/gcc
            CXX=$CCACHE/g++
            CMAKE_FLAGS=""
+
+matrix:
   allowfailures:
-    - env: ==CPU_OPENCL==
+    - env: OPENCL=true
 
 before_install:
   - START_TIME=$(date +%s)


jchodera:
No no no no no!  This is the opposite of what we want!

We need to either disable the many body force kernel on buggy OpenCL
implementations (if we think the opencl implementation is the problem) or
fix the bug.

swails:
How can you say this is the *opposite* of what we want when it's what we're actually doing by hand?

I actively ignore this build because it fails every. time.  And it is irrelevant for almost every PR that has been submitted to OMM since that element started failing except for the ones @peastman sent explicitly to try and fix it.  I understand why you may not like it, but it's a slight variant on the status quo (and one that I find less irritating, personally).

But I can't get it to work, anyway, so I'll just close it.

jchodera:
I 100% agree with you that we need to do something about this ASAP, but we need to either (1) decide for certain the AMD CPU OpenCL implementation is buggy and exclude the `CustomManyBodyForce` kernels from being available on it, or (2) actually work on fixing the bug.  Up to now, I think @peastman was thinking "maybe it's a bug, but maybe not", but given that we are about to go for a OpenMM 7.0 release candidate, we need to make that decision immediately.

cc https://github.com/pandegroup/openmm/issues/1319#issuecomment-180770745

peastman:
My best guess is, "The test works fine on AMD CPU OpenCL, except on the truly ancient version that's being used on travis."  We haven't been able to reproduce the error anywhere else, including on computers using AMD CPU OpenCL.  @rmcgibbo spent a while trying to get it to use a more recent version, but got stuck.

jchodera:
Can we exclude the kernel from loading on AMD CPU OpenCL versions we think are old enough to have this bug?

jchodera:
Presumably you can just use [`clGetPlatformInfo`](https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/clGetPlatformInfo.html) to query the platform name and version?

peastman:
If we can figure out which versions that is.

jchodera:
Well, what's the earlier version you are certain things work correctly for?  Why don't we use that?  That way, there is zero probability we will compute things incorrectly.

rmcgibbo:
> Well, what's the earlier version you are certain things work correctly for?

What earlier version? I don't recall one.

FWIW, the output of `TestOpenCLDeviceInfo` is
```
OpenCLPlatformIndex 0, OpenCLDeviceIndex 0: "Intel(R) Xeon(R) CPU @ 2.30GHz"
    CL_DEVICE_VENDOR                 = GenuineIntel
    CL_DEVICE_MAX_COMPUTE_UNITS      = 2
    CL_DEVICE_GLOBAL_MEM_SIZE        = 7844462592
    CL_DEVICE_MAX_CLOCK_FREQUENCY    = 2299
    estimatedSpeed                   = 4598
```

jchodera:
Sorry---*earliest*.

rmcgibbo:
¯\_(ツ)_/¯

rmcgibbo:
If you want to try to get other versions installed on Travis and test them, that would be great. I didn't get very far in https://github.com/pandegroup/openmm/pull/1378, where I tried to upgrade to Ubuntu Trusty and use a later `fglrx` from `apt-get`, but was getting errors like

```
      Start   8: TestOpenCLFFTMixed
 5/18 Test   #8: TestOpenCLFFTMixed ...........................***Failed    1.03 sec
Setting of real/effective user Id to 0/0 failed
modprobe: ERROR: ../libkmod/libkmod-module.c:809 kmod_module_insert_module() could not find module by name='fglrx'
modprobe: ERROR: could not insert 'fglrx': Function not implemented
Error! Fail to load fglrx kernel module! Maybe you can switch to root user to load kernel module directly
Setting of real/effective user Id to 0/0 failed
modprobe: ERROR: ../libkmod/libkmod-module.c:809 kmod_module_insert_module() could not find module by name='fglrx'
modprobe: ERROR: could not insert 'fglrx': Function not implemented
Error! Fail to load fglrx kernel module! Maybe you can switch to root user to load kernel module directly
exception: Assertion failure at TestOpenCLFFT.cpp:94.  Expected 0.740309, found 15.0063
```

rmcgibbo:
Obviously the best solution would be to install a newer OpenCL on Travis. Second best would be to perhaps just skip the problematic test cases but leave the other ones running. But I think I agree with @swails that turning on `allow_failure` is better than nothing.

jchodera:
I'm not sure I understand why the consensus seems to be that we should upgrade OpenCL on travid *and* allow this kernel to run on broken implementations of OpenCL.  Yes, I agree that upgrading OpenCL on travis would be great, but shouldn't we also not be distributing code that doesn't work correctly on certain OpenCL implementations if we can help it?

rmcgibbo:
Yeah, I would be good with that, if someone made such a check. I just don't know how to implement it.

peastman:
Yes, if we can figure out what version this was fixed in, then we can put a check for that.  We just need to figure out what version that was.  Assuming the issue actually *is* with the OpenCL version, which is currently just a guess.  All we actually *know* is that this has only ever been observed on travis, nowhere else, and travis is using a really ancient version.  It's likely those are connected, but that's still just a guess.

Nonetheless, I'm not worrying too much about this issue for a few reasons:

1. CustomManyParticleForce is a rarely used feature.

2.  Running the OpenCL platform on a CPU is a rare thing to do.  If you're using a CPU, you're much more likely to use the CPU platform.

3. When this error happens, it's completely obvious that it's happened.  You either get a segfault, or you get forces that are just wildly wrong.  So it isn't the sort of bug that's going to subtly corrupt people's simulations.

swails:
It would be nice if this test would stop failing, or stop reporting that the build was broken.  Either through `allow_failure` or something else (disable this specific test in Travis?).  Particularly since the effort involved with figuring out what is wrong here in what way appears to be immense compared to the payoff.

The benefit of allow_failure is that you can always go look into the build logs if you're trying to target the problem later, and it doesn't complicate the Travis script by bending over backwards to avoid a specific test.

The benefit of just disabling the failing test is that it still reports on other OCL failures (if they're real), but makes the Travis setup possibly more complicated.

jchodera:
> Yes, if we can figure out what version this was fixed in, then we can put a check for that. We just need to figure out what version that was. 

We don't have to figure out what version this was fixed in.  We just have to exclude version numbers less than the *earliest known working version*.  If AMD APP SDK 2.9 is the earliest version we can demonstrate working OpenCL function for this kernel, let's just exclude that kernel from running on earlier AMD APP SDK CPU OpenCL implementations.  It's really simple: This way, we will never have someone get the wrong result.  We may exclude more people than we had anticipated, but if we ever find an earlier working version, we can update the exclusion rule.

> Nonetheless, I'm not worrying too much about this issue for a few reasons:
> CustomManyParticleForce is a rarely used feature.
> Running the OpenCL platform on a CPU is a rare thing to do. If you're using a CPU, you're much more likely to use the CPU platform.
> When this error happens, it's completely obvious that it's happened. You either get a segfault, or you get forces that are just wildly wrong. So it isn't the sort of bug that's going to subtly corrupt people's simulations.

It sounds like there should be very little reason *not* to exclude AMD CPU OpenCL earlier than APP SDK 2.9 for this particular kernel, then.

peastman:
The OpenCL runtime comes with the driver, not the APP SDK.  And like I said, we don't even know for sure this error is related to the driver version.  The only way to be absolutely certain is to move travis to a newer version and see the error go away.

jchodera:
Could [something like this](https://github.com/JuliaGPU/OpenCL.jl/blob/master/.travis.yml#L18-L27) work to get a more recent AMD OpenCL runtime working on travis?

rmcgibbo:
Good find! Only one way to find out, but that looks very promising.

jchodera:
Does that mean we can convince you to give it a try, @rmcgibbo?  ;)

rmcgibbo:
Probably not.

peastman:
I followed the link to the repository that script originally came from: https://github.com/gregvw/amd_sdk.  It says in their readme:

> This script no longer works as AMD has changed their website to add another page where you must accept terms and conditions to download.

rmcgibbo:
Could probably manually download the file and then upload it to a hidden
URL on openmm.org

On Tue, Feb 9, 2016 at 12:58 PM, peastman <notifications@github.com> wrote:

> I followed the link to the repository that script originally came from:
> https://github.com/gregvw/amd_sdk. It says in their readme:
>
> This script no longer works as AMD has changed their website to add
> another page where you must accept terms and conditions to download.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pandegroup/openmm/pull/1389#issuecomment-182064975>.
>



-- 
-Robert

jchodera:
I have these files hosted on Jenkins. Just a moment.

jchodera:
https://jenkins.choderalab.org/userContent/AMD-APP-SDKInstaller-v3.0.130.135-GA-linux64.tar.bz2

PR #1399, opened 2016-02-19
NBFix generator

ChayaSt:
I modeled the NBFix generator on ParmEd's implementation here:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2479-L2587

It creates a type lookup table with  and then uses it to create the A and B coefficient arrays for the `customNonbondedForce`. If the pair of atom types are in NBFix types, it uses the nbfix rmin and emin for the A and B coefficient array. Then it wipes out the L-J parameters in the standard `NonbondedForce`

The NBFix types are written out in the ffxml:
```
<NBFixForce>
 <NBFix type1="CLA" type2="SOD" emin="0.350933" rmin="0.3731"/>
</NBFixForce>
```

@swails, I will open a PR in ParmEd once this PR is merged.


swails:
Do you have a test case for this that you can add?

Does it work?

Well done!

ChayaSt:
> Do you have a test case for this that you can add?
> 
> Does it work?

yes, it works. 
There is a test case here:
https://github.com/ChayaSt/openmm/blob/nbfix/wrappers/python/tests/TestForceField.py#L539-L596

peastman:
I'm a little uncomfortable with the way this entangles itself with the NonbondedGenerator.  It assumes you already have a `<NonbondedForce>` tag defining most of your LJ settings.  It copies most of its information from that generator, adds in a little information of its own, and then modifies the NonbondedGenerator to remove all the parameters it's copied out.  Problems with that approach include:

- It won't always work, since there's no guarantee about which generator will be processed first.
- It's very brittle, since changes to the NonbondedGenerator class could easily break it.
- `createSystem()` is supposed to be a "const" function: calling it shouldn't alter the ForceField it's called on.
- It's conceptually confusing, since the tags defining different forces are supposed to be self contained and generally independent.

So I'd suggest changing this to be fully self contained.  Include all parameters needed for this force inside its own tag, not relying on getting them from a different force.  In any force field that includes this force, and that also includes a NonbondedForce, the NonbondedForce would of course specify `epsilon="0"` for all particles.  But that should be shown explicitly in the XML file, not patched up behind the scenes.

Also, why did you specify the LJ parameters as emin and rmin, rather than sigma and epsilon like NonbondedForce?

jchodera:
> So I'd suggest changing this to be fully self contained. Include all parameters needed for this force inside its own tag, not relying on getting them from a different force. 

That simply isn't practical.  You would need to specify *every* combination of *every* possible set of atom types or classes.  It breaks the ability to split anything across multiple files.

Is there a different approach that isn't completely unworkable?

I could imagine the simplest way to check for the failure cases you are worried about is to simply throw an `Exception` if the nonbonded forces cannot be found for the other atom types in the system when the force is being created inside the `NBFixGenerator`.

peastman:
>  You would need to specify every combination of every possible set of atom types or classes. It breaks the ability to split anything across multiple files.

What are you talking about?  It has nothing to do with that.  I'm just saying that the definition of this force should be in its own tag (or set of tags split across multiple files).

jchodera:
Here's an example of the information in the `NBFixForce` XML tag:
```XML
<NBFixForce>
 <NBFix type1="CLA" type2="SOD" emin="0.350933" rmin="0.3731"/>
</NBFixForce>
```
What this means is that *only* between `CLA` and `SOD` types do we use this modified LJ term.  For all other atom type interactions, however, we are forced to use the original LJ terms.  Where do we get them from? 

It sounds like you are suggesting we define *all* LJ type terms *again* in `<NBFixForce>` since they will be needed for computing the interactions of `CLA` and `SOD` with any other atom types.  Is that correct?  If so, this requires duplicating a huge amount of information between `<NonbondedForce>` and `<NBFixForce>`, and does not present a workable way to fuse parameter information between multiple `<NonbondedForce>` tags split across different files.

Perhaps we're misunderstanding what you are recommending?  Do you want to pull `NBFix` functionality *within* `NonbondedForceGenerator` so that this isn't split up into two places that must necessarily share the same information?

ChayaSt:
> Do you want to pull NBFix functionality within NonbondedForceGenerator so that this isn't split up into two places that must necessarily share the same information?

:+1: . This is probably makes the most sense conceptually since the NBFix is a fix to the NonbondedForce.

> Also, why did you specify the LJ parameters as emin and rmin, rather than sigma and epsilon like NonbondedForce?

This is the way the NBFIX is defined in the Charmm. Emin is equal to epsilon so I can change that, but rmin is sigma * 2^(1/6). I can convert it to sigma when writing it out to the xml file and I guess convert it back to rmin for the a and b coefficients?

jchodera:
> This is the way the NBFIX is defined in the Charmm. Emin is equal to epsilon so I can change that, but rmin is sigma * 2^(1/6). I can convert it to sigma when writing it out to the xml file and I guess convert it back to rmin for the a and b coefficients?

It would actually be convenient to have `<NonbondedForce>` and friends accept all popular forms (`sigma`, `epsilon`, `emin`, `rmin`, `A`, `B`, etc.).

peastman:
> It sounds like you are suggesting we define all LJ type terms again in `<NBFixForce>` since they will be needed for computing the interactions of CLA and SOD with any other atom types. Is that correct?

I'm suggesting that information *only* be defined in the NBFixForce section (but let's come up with a better name for it than that—we can discuss that later).  In the `<NonbondedForce>` tag, it will simply specify `epsilon="0"` for every atom type, indicating that the NonbondedForce is not to calculate any LJ forces.

jchodera:
I still do not understand how you propose this to work.

@ChayaSt, can you make a complete example of charmm36.xml available for
@peastman in its current form?

@peastman: Can you modify this file to your proposed spec? Making your
proposal concrete will facilitate our understanding.

swails:
> I still do not understand how you propose this to work.

Why not?  It seems pretty simple to me.  (Possible name `LennardJonesMatrixForce`).  Right now, @ChayaSt stores the pairs and pair-specific `epsilon` and `sigma` for the modified pairs.  @peastman is simply suggesting that in addition to that, we *also* store all of the sigmas and epsilons for each atom type in the generator as well instead of grabbing the parameters from `NonbondedForce` (I think).  That approach makes good sense to me.

swails:
Note that here, @ChayaSt has already done the hard work!  She actually gets to simplify her code a little bit :).  Still good work, and it's only a slight modification for improved consistency and scalability.

ChayaSt:
> Possible name LennardJonesMatrixForce

:+1: 

> we also store all of the sigmas and epsilons for each atom type in the generator as well instead of grabbing the parameters from NonbondedForce (I think)

Just to make sure I understand what you are suggesting. We don't list all the L-J parameters again under the ```<NBFix>``` tag, but have the generator also parse the Nonbonded parameters and store it? But then how will we then turn off the nonbondedForce without having the NBFix generator modifying it?

swails:
> Just to make sure I understand what you are suggesting. We don't list all the L-J parameters again under the <NBFix> tag, but have the generator also parse the Nonbonded parameters and store it? But then how will we then turn off the nonbondedForce without having the NBFix generator modifying it?

No.  In the XML file, `NonbondedForce` will simply have 0 for all atom type epsilons and 1 (or something non-zero) for all atom type sigmas so that the L-J contribution for `NonbondedForce` that is generated by `NonbondedForceGenerator` is 0.  Instead, put those parameters inside the `LennardJonesMatrixForce` so that the generator can create the appropriate `CustomNonbondedForce` without ever having to look through another `Force`.

ChayaSt:
> No. In the XML file, NonbondedForce will simply have 0 for all atom type epsilons and 1 (or something non-zero) for all atom type sigmas so that the L-J contribution for NonbondedForce that is generated by NonbondedForceGenerator is 0. Instead, put those parameters inside the LennardJonesMatrixForce so that the generator can create the appropriate CustomNonbondedForce without ever having to look through another Force.

Ah, I get it now. So we don't change the xml file but have LennardJonesMatrixForce read all nonbonded parameters while Nonbonded sets all epsilons to 0 and something else for sigma. But that is still conceptually confusing because LennardJonesMatrixForce will be using parameters not explicitly defined in its own tag. If we do change this in the xml file (set all epsilon to 0 and have all nonbonded parameters defined in LennardJonesMatrixForce), we will run into problems when we have some parts of the forcefield defined in another file that doesn't have NBFIX. Unless we always write it out the same way regardless if the input parameter files have NBFIX or not.

swails:
> So we don't change the xml file

no -- we *do* change the XML file.  We put the sigma and epsilon for each atom type into the `LennardJonesMatrix` force as well as a set of modified pair-specific terms.  In the `NonbondedForce` we simply zero-out the L-J parameters.  That maintains complete parameter encapsulation.

ChayaSt:
> no -- we do change the XML file. We put the sigma and epsilon for each atom type into the LennardJonesMatrix force as well as a set of modified pair-specific terms. In the NonbondedForce we simply zero-out the L-J parameters. That maintains complete parameter encapsulation.

So then this behavior should be the same for writing out the ```NonbondedForce``` parameters regardless if the input files have NBFIX or not. And the L-J force will always be implemented with ```LennardJonesMatrix```. 
Did I get it right this time?

swails:
For the CHARMM FF, yes.  If there are any NBFIXes declared in the force field XML file, then a `CustomNonbondedForce` will always be used to compute L-J interactions.

peastman:
What I have in mind is something like this:

```XML
<LennardJonesForce lj14scale="0.5">
  <Atom type="0" sigma="0.324999852378" epsilon="0.71128"/>
  <Atom type="1" sigma="0.106907846177" epsilon="0.0656888"/>
  ...
  <AtomTypePair type="0" type="1" sigma="0.21" epsilon="0.132"/>
  ... 
</LennardJonesForce>
```

Note that I've called it "LennardJonesForce", because that's all it really is: a different implementation of LJ.  And it defines parameters for atoms in the usual way.  Mostly it looks just like a `<NonbondedForce>` tag without the Coulomb specific parts.  The only difference is that it also lets you override the combining rules for specific pairs of types or classes with an `<AtomTypePair>` tag.  (I'm not too attached to that name.  We can probably come up with something better.)

ChayaSt:
I'm ready to get back to this now. 

@swails, I can add a ```_write_omm_LennardJonesForce``` function that will only be called for the CharMM forcefield or for any  force field file that has NBFIX. Any pointers on that? 

Once this is done I'll fix up the generator so that it's self contained.

ChayaSt:
I cleaned up the generator a bit. However, I am not sure how to deal with the ```lj14scale``` and ```bondCutoff```. Previously the exclusions were transferred from the exceptions of the ```NonbondedForce```. Here I add the exclusions from bonds and hard coded in the ```bondCutoff```. Alternatively we can add that as an attribute to the ```<LennardJonesForce>``` tag. But how do I add the exception for the 1-4 L-J scaling?

swails:
CustomBondForce for the 1-4 scaled L-J interactions.

ChayaSt:
@swails, thanks for your input.
@peastman, do you want to check if you're happy with the generator now? I cleaned it up as you suggested.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,186 @@ def postprocessSystem(self, sys, data, args):
...
+            # Multiple <LennardJonesForce> tags were found, probably in different files
+            generator = existing[0]
+
+        for LJ in element.findall('Atom'):
+            generator.registerLennardJones(LJ.attrib)
+        for Nbfix in element.findall('NBFixPair'):
+            generator.registerNBFIX(Nbfix.attrib)
+
+    def createForce(self, sys, data, nonbondedMethod, nonbondedCutoff, args):
+        """ Ewald and PME will be interpreted as CutoffPeriodic for the CustomNonbondedForce """
```
peastman: I don't think you really meant that to be the docstring for the method did you?

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,186 @@ def postprocessSystem(self, sys, data, args):
...
+                atype2 = data.atomType[data.atoms[j]]
+                if lj_indx_list[j] > 0: continue
+                if atype2 == atype:
+                    lj_indx_list[j] = num_lj_types
+                elif not (atype in self.types1) or (atype in self.types2):
+                    # only non NBFIX types can be compressed
+                    values = self.lj_types.paramsForType[atype2]
+                    ljtype2 = (values['sigma']* 2**(1.0/6) / 2, abs(values['epsilon']))
+                    if ljtype == ljtype2:
+                        lj_indx_list[j] = num_lj_types
```
peastman: You're using an O(n^2) loop, which could be slow.  You can convert this to an O(n) loop by using a dict to map parameters to type indices.  Something like this:

```Python
if values in typeMap:
    lj_indx_list[i] = typeMap[values]
else:
    typeMap[values] = num_lj_types
    lj_indx_list[i] = num_lj_types
    num_lj_types += 1
```

The code is also a lot simpler that way.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,186 @@ def postprocessSystem(self, sys, data, args):
...
+        # Now everything is assigned. Create the A- and B-coefficient arrays
+        acoef = [0 for i in range(num_lj_types*num_lj_types)]
+        bcoef = acoef[:]
+        for m in range(num_lj_types):
+            for n in range(num_lj_types):
+                namem = lj_type_list[m]
+                namen = lj_type_list[n]
+                if range(len(self.types1)):
```
peastman: I gather that range() evaluates to false if its argument is 0, true otherwise?  That's... perhaps not the clearest way of writing this. :)  How about `if len(self.types1) > 0`?

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,186 @@ def postprocessSystem(self, sys, data, args):
...
+                    for k in range(len(self.types1)):
+                        types1 = self.types1[k]
+                        types2 = self.types2[k]
+                        if (namem == types1 and namen == types2) or (namem == types2 and namen == types1):
```
peastman: This can also be implemented more efficiently and simply with a lookup in a dict rather than a loop over types.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,186 @@ def postprocessSystem(self, sys, data, args):
...
+        for i in lj_indx_list:
+            self.force.addParticle((i-1,))
+        self.force.setUseLongRangeCorrection(True)
+        self.force.setCutoffDistance(nonbondedCutoff)
+
+        # Create customBondForce for the 1-4 scaled L-J interactions
+        lj14ScaleForce = mm.CustomBondForce('(a/r^6)^2-(b/r^6); r6=r2*r2*r2; r2=r^2;'
+                                            'a=acoef;'
+                                            'b=bcoef;')
+        lj14ScaleForce.addGlobalParameter('lj14scale', self.lj14scale)
```
peastman: I don't think this is ever used.

wrappers/python/simtk/openmm/app/forcefield.py
ChayaSt: This was supposed to be a comment. Will change that.

wrappers/python/simtk/openmm/app/forcefield.py
ChayaSt: Thanks! I'll change it.

wrappers/python/simtk/openmm/app/forcefield.py
ChayaSt: This is used for the 1-4 scaled L-J.
https://github.com/ChayaSt/openmm/blob/charmm36/devtools/forcefields/charmm/tests/test_ffconversion.py

Here's the output of that script:
```
pamed energies: [('HarmonicBondForce', 8.26018942073294), ('HarmonicAngleForce', 0.2714968886612339), ('PeriodicTorsionForce', 1.0799992422864986), ('HarmonicBondForce', 0.007913540341948437), ('NonbondedForce', -50.08503469213911), ('CustomNonbondedForce', -0.09783352740061724), ('CMMotionRemover', 0.0)]
omm eneriges [('HarmonicBondForce', 8.268102961074886), ('HarmonicAngleForce', 0.271496888687008), ('PeriodicTorsionForce', 1.0799992422864986), ('CustomTorsionForce', 0.0), ('NonbondedForce', -50.07130056020566), ('CustomNonbondedForce', -0.09783352740099871), ('CustomBondForce', -0.013734131933562961), ('CMMotionRemover', 0.0)]
``` 
If you look at the ```energy_decomposition_system```, you can see that the ```CustomBondForce``` computes the 1-4 scaled LJ. In ParmEd, ```NonbondedForce``` calculates that with the exceptions. But since we are not using ```NonbondedForce``` here, I needed to add the ```CustomBondForce```.

peastman: I was referring to the call to `addGlobalParameter()`.  You create a parameter called `lj14scale`, but that parameter doesn't appear anywhere in the force's energy expression.

ChayaSt: Ah, yes. You're right.

ChayaSt:
@peastman, did you have a look at the changes?

Also, I am not sure what the travis failure is about. It is apparently not finding the a file that is in the PR. But this nosetest passes locally.

peastman:
Sorry for the delay.  More comments coming!

> It is apparently not finding the a file that is in the PR.

Are you sure you remembered to add it?  I don't see toppar_water_ions.str there.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+        self.lj14scale = lj14scale
+        self.lj_types = ForceField._AtomTypeParameters(forcefield, 'LennardJonesForce', 'Atom', ('sigma', 'epsilon'))
+
+    def registerNBFIX(self, parameters):
+        types = self.ff._findAtomTypes(parameters, 2)
+        if None not in types:
+            type1 = types[0][0]
+            type2 = types[1][0]
+            emin = _convertParameterToNumber(parameters['emin'])
+            rmin = _convertParameterToNumber(parameters['rmin'])
```
peastman: You switched it to use `sigma` and `epsilon` for regular parameters, but it's still using `emin` and `rmin` here.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+        # We need a CustomNonbondedForce to implement the NBFIX functionality.
+        # First derive the lookup tables
+        lj_indx_list = [0 for atom in data.atoms]
+        num_lj_types= 0
+        lj_type_list = []
+        type_map = {}
+        for i, atom in enumerate(data.atoms):
+            atype = data.atomType[atom]
+            values = self.lj_types.paramsForType[atype]
+            if lj_indx_list[i]: continue # ljtype has been assigned an index
```
peastman: You initially filled `lj_indx_llist` with 0, and then here you rely on the fact that 0 evaluates to false.  But 0 is also a valid index value!  It would be better to fill it with `None`, then check for `if lj_indx_list[i] is not None`.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+            if values in type_map.values() and not atype in [type for key in self.nbfix_types.keys() for type in key]:
```
peastman: `type_map` should map from values to indices, not the other way around.  Checking to see if something is in a map's list of values is an O(n) operation!  And then what's going on with the second clause?  I can't even parse what that's supposed to be doing.  And while checking membership in a `dict` *should* be an O(1) operation, you've managed to do it an an O(n) way here.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+                        rij6 = rij**6
+                        wdij = math.sqrt(type_map[m+1]['epsilon'] * type_map[n+1]['epsilon'])
+                        acoef[m+num_lj_types*n] = math.sqrt(wdij) * rij6
+                        bcoef[m+num_lj_types*n] = 2 * wdij * rij6
+                else:
+                    rij = (type_map[m+1]['sigma'] + type_map[n+1]['sigma']) * 2**(1.0/6) / 2
+                    rij6 = rij**6
+                    wdij = math.sqrt(type_map[m+1]['epsilon'] * type_map[n+1]['epsilon'])
+                    acoef[m+num_lj_types*n] = math.sqrt(wdij) * rij6
+                    bcoef[m+num_lj_types*n] = 2 * wdij * rij6
```
peastman: This second `else` block isn't needed.  It's just unnecessary code duplication.  You can get rid of the check for `if len(self.nbfix_types) > 0`.  If the length is 0, then of course the check for `if pair in self.nbfix_types` will fail, and so it will be handled correctly.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+        self.force = mm.CustomNonbondedForce('(a/r6)^2-b/r6; r6=r2*r2*r2; r2=r^2; '
```
peastman: It's better to write `a/r^12-b/r^6`.  OpenMM is pretty smart about figuring out the most efficient way to compute integers powers, and writing them out the way you have will just interfere with its optimizations.  Also, you shouldn't be squaring `a`.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+        # Add the particles
+        for i in lj_indx_list:
+            self.force.addParticle((i-1,))
+        self.force.setUseLongRangeCorrection(True)
+        self.force.setCutoffDistance(nonbondedCutoff)
+
+        # Create customBondForce for the 1-4 scaled L-J interactions
+        lj14ScaleForce = mm.CustomBondForce('(a/r^6)^2-(b/r^6); r6=r2*r2*r2; r2=r^2;'
+                                            'a=acoef;'
+                                            'b=bcoef;')
```
peastman: Same as above.  Also, why not just call the parameters `a` and `b` directly, rather than copying them to extra intermediate variables?

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1674,6 +1674,171 @@ def postprocessSystem(self, sys, data, args):
...
+        lj14ScaleForce.addPerBondParameter('acoef')
+        lj14ScaleForce.addPerBondParameter('bcoef')
+        lj14Pairs = []
+
+        # Add bonds for 1-4 scaled L-J from dihedrals
```
peastman: This won't work correctly.  See the code below in `postprocessSystem()`.  There are various cases like virtual sites and Drude particles that require special handling when identifying 1-4 pairs.

wrappers/python/simtk/openmm/app/forcefield.py
ChayaSt: For the second clause, I can add a ```set()``` of nbfix atom types to ```LennardJonesForce``` so then checking membership will be fast. Something like this:

```
def __init__(self, forcefield):
    ...
    self.nbfix_set = set()
```
in ```registerNBFIX```:

 ```
registerNBFIX(self, parameters):
    ....
    self.nbfix_set.update([type1, type2])
```
Then the second clause become:
```atype not in self.nbfix_set:```

If I do ```type_map``` the other way around, mapping values to indices, How will I populate ```acoef``` and ```bcoef```? Right now I loop through the indices and get their values. I can use an ```OrderedDict``` but that will be slow.

Comment (peastman) [inline wrappers/python/simtk/openmm/app/forcefield.py]: It doesn't have to be that complicated.  All you need is a fast way to identify whether a type has any nbfix parameters, since if it does, you need to treat it as unique.  So create a set of all types involved in nbfix terms:

```Python
nbfix_type_set = set().union(*self.nbfix_types)
```

You can build that set at the start of `createForce()`, just before entering the loop.  Then within the loop, you can just write `type in nbfix_type_set` to quickly determine whether a type has any nbfix parameters.

I don't understand what your concern is about building the coefficient arrays.  I see lots of straightforward ways to do that, none of which requires an OrderedDict.  You could iterate over the entries in `type_map`.  Or you could build an inverse map.  What's the compilcation?

wrappers/python/simtk/openmm/app/forcefield.py
jchodera: This scheme was probably selected in ParmEd (since it is identical to what ParmEd does) so that the tabulated values have the same order of magnitude. Otherwise, numerical error might kill us.

swails: You give me too much credit :).  This scheme is used in ParmEd because #628 (particularly [this comment](https://github.com/pandegroup/openmm/pull/628#issuecomment-56537401)) led me to believe that ^3 and ^2 specifically were optimized (since there's a `square` and `cube` function in Lepton not documented in the API docs, but apparently nothing higher).

So I thought that keeping things as ^2 was the "right" thing to do.  Also, in ParmEd, the `a` coefficient is sqrt-ed before added to the table.  Not sure what's done here...  If that's the wrong thing to do, I should adjust ParmEd lest it serve as a misleading example (as it seems it has here).

peastman: If you're interested in how custom forces compute integer powers, take a look at http://www.sciencedirect.com/science/article/pii/B9780123859631000290.

swails: Can you email me a PDF?  I *would* like to read it, but don't have academic access to journals anymore :(.

ChayaSt:
@peastman, I addressed most of your comments besides the one on handling the 1-4 interactions. Do you want to take over now?

jchodera:
@ChayaSt: Maybe you can give a detailed summary of what has been done and what remains to be done?

ChayaSt:
@jchodera, here's a summary:

The ```LennardJonesGenerator``` was introduced to handle NBFIX. 

https://github.com/ParmEd/ParmEd/pull/633 adds a ```<LennardJonesForce>``` tag which looks like the ```<NonbondedForce>``` tag without the Coulomb force and with an extra tag ```<NBFixPair>```. This allows overriding the combining rule.

In the ```LennardJonesGenerator```, a ```CutomNonbondedForce``` is used to implement NBFIX override with the ```Dicrete2DFunction```. First a lookup table is derived where every Lennard Jones type is assigned an index without compressing the NBFix types. Then it creates an A and B coefficient array with the combination rule if the Lennard Jones types are not both NBFix types, otherwise the NBFix parameters are used. 

In ```postprocessSystem```, ```createExclusionFromBonds``` is used to create the exceptions for 1-2 and 1-3 and 1-4 interactions. However, the 1-4 scaling interactions have not been implemented. Initially I used a CustomBondForce for that, but as @peastman and testing pointed out, this didn't work.

jchodera:
Thanks!

@peastman: Let us know if you need any more info from @ChayaSt in order for
you to take this over.

peastman:
I can't take over the OpenMM side of it.  Someone else will need to handle the ParmEd side.

jchodera:
> I can't take over the OpenMM side of it.

Can't?

ChayaSt:
> Someone else will need to handle the ParmEd side.

I can do that.

peastman:
Sorry, that should have been "can". :)

PR #1405, opened 2016-02-29
Fix 3 member rings

pgrinaway:
This adds just a bit of code to ensure that the set of atoms being added to a proper torsion does not include a duplicate of one of the angle atoms. Let me know if anything seems off or should be fixed.

Thanks!


pgrinaway:
Also, addresses #1405

jchodera:
To be clear: This fixes a bug in which molecules containing three-membered rings (which occur in many drugs!) fail.

peastman:
Thanks!

PR #1407, opened 2016-03-04
Updated default coefficients for extrapolated polarization

peastman:
These values were provided by @andysim.  They produce more accurate results on a range of different system types.  There's no reference for them yet, since they're still working on the paper.  Once it's published, we should add a reference to it.


jchodera:
It would be great to also provide a rationale for their selection.

peastman:
They're writing up a paper on it.

PR #1416, opened 2016-03-11
Can build manuals on Windows

peastman:
For this to work, you need the following packages installed on your computer and added to the PATH:

https://sourceforge.net/projects/gnuwin32/files/make/3.81
http://miktex.org


jchodera:
> For this to work, you need the following packages installed on your computer and added to the PATH:

Can you document this in a `README.md` somewhere?

PR #1420, opened 2016-03-14
C and Fortran wrappers support groups argument to getState()

peastman:


jchodera:
What about adding tests?

peastman:
We don't have any tests at all for the C and Fortran wrappers (except verifying that they compile).  This came up recently in a different thread.  It would be good to add some, but that will take a bunch of new infrastructure.

PR #1421, opened 2016-03-14
Debugging errors on AppVeyor

peastman:


peastman:
@rmcgibbo The errors seem to be caused by your changes in #1289.  I don't understand how it's supposed to work.  I don't see it installing any kind of OpenCL SDK anywhere.  It just downloads the icd, which doesn't include any headers.

peastman:
The fix is in #1518.

Issue #1418, opened 2016-03-13
Document domain for `phase` in PeriodicTorsionForce


jchodera:
The domain (is it `[-pi,+pi)`?) for `phase` in `PeriodicTorsionForce` is not specified in the docstring.
```
     * @param phase        the phase offset of the torsion, measured in radians
```
Lines that need to be changed:
* https://github.com/pandegroup/openmm/blob/master/openmmapi/include/openmm/PeriodicTorsionForce.h#L70
* https://github.com/pandegroup/openmm/blob/master/openmmapi/include/openmm/PeriodicTorsionForce.h#L84
* https://github.com/pandegroup/openmm/blob/master/openmmapi/include/openmm/PeriodicTorsionForce.h#L97
*

jchodera:
Addressed by #1422.

PR #1422, opened 2016-03-15
Add explicit statement of domain of torsion phase to docs

jchodera:
Addresses #1418.

### DIFF
--- a/openmmapi/include/openmm/PeriodicTorsionForce.h
+++ b/openmmapi/include/openmm/PeriodicTorsionForce.h
@@ -67,7 +67,7 @@ public:
      * @param particle3    the index of the third particle forming the torsion
      * @param particle4    the index of the fourth particle forming the torsion
      * @param periodicity  the periodicity of the torsion
-     * @param phase        the phase offset of the torsion, measured in radians
+     * @param phase        the phase offset of the torsion, measured in radians, on domain [-pi,pi)
      * @param k            the force constant for the torsion
      * @return the index of the torsion that was added
      */
@@ -81,7 +81,7 @@ public:
      * @param[out] particle3    the index of the third particle forming the torsion
      * @param[out] particle4    the index of the fourth particle forming the torsion
      * @param[out] periodicity  the periodicity of the torsion
-     * @param[out] phase        the phase offset of the torsion, measured in radians
+     * @param[out] phase        the phase offset of the torsion, measured in radians, on domain [-pi,pi)
      * @param[out] k            the force constant for the torsion
      */
     void getTorsionParameters(int index, int& particle1, int& particle2, int& particle3, int& particle4, int& periodicity, double& phase, double& k) const;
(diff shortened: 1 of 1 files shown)


jchodera:
@peastman : Is this actually necessary?  I am having trouble figuring out what domain the torsion phase is defined on, or if it doesn't matter since periodic functions that are insensitive to the deviation are guaranteed to be used on all platforms.

It looks like [`Reference` uses periodic functions](https://github.com/pandegroup/openmm/blob/master/platforms/reference/src/SimTKReference/ReferenceProperDihedralBond.cpp#L135-L138) so that the domain doesn't matter, but is this guaranteed to be true of all platforms?

peastman:
It's unnecessary, because PeriodicTorsionForce is, well, periodic. :)  On the other hand, it is relevant for CustomTorsionForce, since that can implement non-periodic functions.  Its range is documented in the manual: http://docs.openmm.org/6.3.0/userguide/theory.html#customtorsionforce

jchodera:
Great. I will change this PR to note this in the headers of these files.

jchodera:
I've fixed the headers.

@peastman: The manual says the domain is `[-pi,+pi]`.  Shouldn't this be `[-pi,+pi)`?

peastman:
No, that's intentional.  In practice the difference is usually irrelevant, so I didn't want to set a requirement on which one platforms had to do.

jchodera:
> No, that's intentional. In practice the difference is usually irrelevant, so I didn't want to set a requirement on which one platforms had to do.

So, which domain should users specify torsions on?  If the code could fail to work properly on some platforms if the user specifies `+pi` or `-pi` while the API allows this, then that's a bug, isn't it?

peastman:
If you've written a force expression where this actually matters, you've almost certainly done something wrong!  Remember that regardless of how this is implemented, you can get angles that are arbitrarily close to both -pi and pi.  So you need to be handling those in any case.  In the very rare event that the angle is precisely +/- pi, you should be prepared for it to be reported as either one.

If your energy function is periodic, then -pi and pi are the same thing.  If it's not periodic, then you have a discontinuity at that point and you can't assume what the value will at that one exact point where it jumps discontinuously.

swails:
This PR is now 7+ years old.  Can we close it or merge it?

jchodera:
Since `CustomTorsionForce` does not throw an exception if you've written a non-periodic force, it's necessary to document the domain.

swails:
> Since `CustomTorsionForce` does not throw an exception if you've written a non-periodic force, it's necessary to document the domain.

Then let's update this PR or open a new one?

Leaving stale PRs around makes it harder for me to clean up Jenkins (aside from the aesthetics of having a PR open so long).

jchodera:
> Then let's update this PR or open a new one?

Update how? To my knowledge, it's still relevant, correct, and does not conflict.

> Leaving stale PRs around makes it harder for me to clean up Jenkins (aside from the aesthetics of having a PR open so long).

My understanding is that it's @peastman's call to merge or close PRs.

peastman:
Thanks for bumping this!  It just needs to have `[-pi,+pi)` changed to `[-pi,+pi]` as discussed above.  Then it can be merged.

PR #1423, opened 2016-03-17
Add residue name and template name matching

kpolok:
In original version, when a template for a residue is searched, first
match for the elements-bonds graph is selected. This is problematic,
when one has ffxml with separate residues for different isotopologues,
where the elements are the same, but atom masses differ. If a file with
different isotopologues is read in, all of them will be assigned the
same template, which is an undesired behavior.
This commit adds residue name matching. If the name matching fails, the
template matching is not influenced.

	modified:   wrappers/python/simtk/openmm/app/forcefield.py

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -371,7 +371,10 @@ class ForceField(object):
                 matches = None
                 signature = _createResidueSignature([atom.element for atom in res.atoms()])
                 if signature in self._templateSignatures:
-                    for t in self._templateSignatures[signature]:
+		    templates = [t for t in self._templateSignatures[signature] if t.name == res.name]
+		    if templates == []:
+		        templates = self._templateSignatures[signature]
+                    for t in templates:
                         matches = _matchResidue(res, t, bondedToAtom)
                         if matches is not None:
                             template = t


peastman:
We've actually implemented (but not yet merged) code that makes lots of changes to how template matching is done, including adding an argument to `createSystem()` that lets you specify what template to use in cases like this.  See #1373.  This is part of a large collection of interconnected force field improvements that will be coming in 7.1.

peastman:
I'm closing this, since it's addressed by #1373.

jchodera:
Thanks for the contribution, @kpolok!  Can you elaborate on the cases you wanted to model where the old scheme would break down?  Was this for isotopically labeled residues?  We've been to find other cases where we may run into this behavior.

kpolok:
[CCl4i_8_init.pdb.txt](https://github.com/pandegroup/openmm/files/210167/CCl4i_8_init.pdb.txt)
[SEB.xml.txt](https://github.com/pandegroup/openmm/files/210172/SEB.xml.txt)

I have attached a PDB file, that I use to start the simulation and a force field file. The simulation involves CCl4 molecules with all possible combinations of Cl35 and Cl37 atoms. The templates for different CCl4 isotopologues are considered identical by OpenMM 6.3, the residue name in PDB is not considered by the software either. This is what I have addressed in my patch.

peastman:
When you call `createSystem()`, there is now a `residueTemplates` argument you can specify to tell it which template to use for each one.

PR #1432, opened 2016-03-24
Require SWIG 3.0.5 or later

### LINKED ISSUE #1431: CustomNonbondedForce with InteractionGroup
```python
from simtk import openmm
force = openmm.CustomNonbondedForce('r')
force.addInteractionGroup([0,1,2], [4,5,6])
```

  File "t.py", line 3, in <module>
    force.addInteractionGroup([0,1,2], [4,5,6])
  File "/net/isjoung/.local/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 10208, in addInteractionGroup
    args[0] = list(args[0])
TypeError: 'tuple' object does not support item assignment

Is this a bug or is my syntax wrong?

peastman:
Fixes #1431.


rmcgibbo:
We may want to remove the code path for the previous swig versions from the
script that builds rhe swig input file

swails:
I was just about to suggest that.  The SWIG bug that made all Python wrappers use `compactdefaultargs` forced @rmcgibbo to add some really ugly code IIRC, and getting rid of it should help things moving forward.  (Not to put down your work, Robert -- quite the contrary, it was a great way to migrate as the new SWIG matured.)

rmcgibbo:
Most of the complex code (e.g. the swig code that uses the Python C api to strip the units) was dealing with the new way of doing things. The amount of code related to the old way is actually much smaller in comparison. But it would streamline things to remove it.

swails:
Yea, doing the equivalent of `stripUnits` in C was a lot clunkier since you had to get Python functions/objects in C and call them, and you needed typemaps for all the possible kinds of arguments you could have.  But on top of that, you also had to handle the case where all the arguments were condensed to a single tuple and pull those out.

It'll just be cleaner and easier to maintain if that stuff was gone.

peastman:
I'd prefer to wait until after 7.0 to remove that code.  No point in risking breaking something when we're just about to release.  I'll open an issue on it so we don't forget.

swails:
:+1:

PR #1446, opened 2016-04-02
Support negative residue numbers in PSF files.

swails:
See https://simtk.org/forums/viewtopic.php?f=161&t=6512&p=16820#p16820 for more
information

Is it too late to get this into 7.0?  Sorry, I haven't really been following the timeline as closely as I normally do...


peastman:
Unless someone feels strongly about it, I don't think it's worth building a new release candidate just for this.  If we end up having to do a new one for another reason, we can include this.  Otherwise, I think it can wait for 7.1.

swails:
Fine with me.  An easy workaround is to use ParmEd's parsers instead, which are API-compatible with OpenMM's.

PR #1447, opened 2016-04-04
Removed code for old swig versions

### LINKED ISSUE #1433: Remove code for old SWIG versions
Now that we require SWIG 3.0.5 or later, we should remove the code that was for earlier versions.  See #1432.

peastman:
Fixes #1433.


peastman:
It looks like on OS X only, SWIG is still generating old style arguments:

```Python
def addForce(self, *args):
```

I can't tell why.  If it were somehow installing too old a version of SWIG, then CMake ought to report an error, but it doesn't.

peastman:
Anyone have an idea what could be causing this?

rmcgibbo:
I see "Swig version must be 3.0.5 or greater! (You have 3.0.2)" in the build log.

mpharrigan:
You need 
```
message(FATAL_ERROR "swig blah blah")
```

https://github.com/pandegroup/openmm/pull/1432/files#diff-177c12b0bb0f80226b9c5a4768013024R113

mpharrigan:
(and also to install a newer swig on mac travis)

peastman:
You're right - it printed that message, but then still went ahead and built anyway.  We ought to make that an error, not just a message.

It's being installed with `conda install swig`, so it should get the latest version.  And at https://anaconda.org/omnia/swig/files it lists the version as being 3.0.7.  So where is it getting 3.0.2 from?

rmcgibbo:
https://github.com/pandegroup/openmm/blob/master/.travis.yml#L66-L74?

peastman:
Wow, that file has some really strange code in it that I didn't know was there.  I removed that line.  Let's see if it now finds the version installed with conda.

peastman:
No, that still doesn't work.

In `.travis.yml` I see this line:

    brew install doxygen swig fftw;

And in `prepare.sh` I see this line:

    conda install --yes --quiet swig fftw3f pip

This raises a few different questions:

1. Why are we installing two different versions each of SWIG and FFTW?  (I assume that wasn't intentional...)
2. Why is the work of installing dependencies split between two different files?

peastman:
I tried removing the extra installs of SWIG and FFTW.  Now it just doesn't find SWIG at all, despite the fact that it's still *supposed* to be getting installed with conda.  I really don't understand what the intended logic behind these scripts is supposed to be.

rmcgibbo:
I don't think `prepare.sh` is used inside travis.yml -- I think that's something else that isn't used in the testing. Inside travis.yml, miniconda/conda isn't used in any way. Swig is always installed manually, either through `brew` on OS X, or by manually `wget`[https://github.com/peastman/openmm/blob/swig/.travis.yml#L128-L132]ing a tarball and unpacking it.

peastman:
Ok, that would explain why I was confused.  So I guess the next question is: why is brew installing an old version of SWIG?  I've never used it, but as far as I can tell from their website, it ought to be installing 3.0.8.  That's the version they define at https://github.com/Homebrew/homebrew-core/blob/master/Formula/swig.rb.

Issue #909, opened 2015-05-11
Standardize approach for handling platform precision


jchodera:
This might be an OpenMM 7 feature due to its potential to break backward compatibility (slightly) or introduce API changes, but it would be great to standardize the way that `Platform` objects are queried for which precision models they support and the way that the user selects among available precision models.

Currently, we have the following logic:
- For all platforms, you can query `supportsDoublePrecision()` to see if it supports double precision, but there is no corresponding `supportsSinglePrecision()`.  
- The `Reference` platform only supports double precision, so you need to call `platform.getName()` to check the name to see that it does not support single precision.
- The `CPU` platform only supports single precision
- Selecting among precision models for CUDA and OpenCL is different, an requires you check the platform name to see whether you should set the platform variable `CudaPrecision` or `OpenCLPrecision`
- Querying which precision model is currently selected is similarly bewildering
- Some platforms support `mixed` precision, but there is no way to query for that

Perhaps there is a very simple way to both query and select among precision models?  This doesn't necessarily _have_ to break backwards compatibility, but it may be conceptually more straightforward to do so.  For example, a `platform.getAvailablePrecisionModels()` and `platform.setDefaultPrecisionModel()` may be sufficient.

peastman:
I've been meaning to deprecate the supportsDoublePrecision() method.  It was created back when there was a clear distinction: Reference did everything in double precision, OpenCL and CUDA did everything in single precision.  But today that isn't true.  It doesn't really capture how Platforms work today.

Also, there are three distinct things you might want to know:

1. What precision modes is a Platform capable of?

2. What precision modes is a particular device capable of?

3. What precision mode is a particular Context using?

The definitions of various modes aren't at all standardized.  For example, the CPU platform uses a mixed precision mode, not single.  But it isn't the same as the mixed precision mode used by CUDA and OpenCL.  It uses single precision only for the nonbonded forces, while computing bonded forces in double.  In contrast, CUDA and OpenCL compute all forces in single precision.  The distinction isn't usually important: bonded forces usually contribute less error than nonbonded ones, so if you're using single for nonbonded, there's not much benefit to using double for bonded.  But they also have little effect on performance, so I haven't bothered to created optimized versions of them for the CPU platform.  But even so, "mixed precision" isn't uniquely defined.  And it's entirely possible its behavior could change in the future, for example if I create an optimized CPU implementation of bonded forces, or find a new way to significantly improve accuracy with little performance cost.

jchodera:
[...]

PR #1448, opened 2016-04-04
Added property to request deterministic forces

### LINKED ISSUE #1283: Implement deterministic energy/force computation
I just realized there wasn't an open thread with this enhancement request.

peastman:
Fixes #1283.


jchodera:
Why call it `CudaDeterministicForces`? Won't it raise an exception if the Platform doesn't support it?

Do we want other platforms to also support this parameter (even though all are deterministic) so that code for selecting deterministic forces is simplified and doesn't have to include special cases for CUDA?

peastman:
Currently, *all* platform specific properties begin with the name of the platform.  It would be worth discussing alternate ways of specifying them.  But if we do that, it should be done in a coherent way, not as a one-off for this particular feature.

jchodera:
> Currently, all platform specific properties begin with the name of the platform. It would be worth discussing alternate ways of specifying them. But if we do that, it should be done in a coherent way, not as a one-off for this particular feature.

OK.  CCing #909 as well.  I'll create a new issue to propose harmonizing platform properties.

PR #1453, opened 2016-04-07
Support patches in force fields

peastman:
This is fully functional but could use more testing.  It includes the simplification I described at https://github.com/pandegroup/openmm/issues/1335#issuecomment-206613163: no residue can be affected by more than one multi-residue patch.

I also fixed a bug in the code for overriding templates.  And I renamed the attribute for doing that from "overload" to "override", since that's a more accurate description of what it does.

I haven't written documentation for this yet.


peastman:
Documentation added.

ChayaSt:
Should I push the test systems to @peastman's patches branch? 

I have the following systems
Phosphorylated SER, THR, TYR and CY5 dye bound to a CYS.

peastman:
Just post them somewhere I can download them.  I'll create tests based on them and determine whether any code changes are needed to support them.

ChayaSt:
Here are the pdb files and str files with the patches. 
https://github.com/ChayaSt/Charmm36_FF/tree/master/test_systems

 All have `NTER` and `CTER` patch.

peastman:
In the README it says, "PS1 was used in sep.pdb."  But there is no PS1 patch.  Did you mean SP1?

All of these are single residue patches.  Do you have any examples that involve multi-residue patches?

ChayaSt:
> In the README it says, "PS1 was used in sep.pdb." But there is no PS1 patch. Did you mean SP1?

Yes, It's the SP1 patch. Sorry for that typo.

I added a [multi-residue patch](https://github.com/ChayaSt/Charmm36_FF/blob/master/test_systems/Cy5r_nter.pdb). Both a CY5R and NTER was applied to CYS 7.

peastman:
That's two single-residue patches being applied to the same residue.  A multi-residue patch means a single patch that affects multiple residues.  Those are the case that's hard to deal with.

ChayaSt:
Ok. I'll add a patch for a cyclic protein which links the nter and cter so
it's applied to 2 residues.

peastman:
Thanks!

ChayaSt:
I ended up adding a [GFP](https://github.com/ChayaSt/Charmm36_FF/blob/master/test_systems/top_all36_prot.rtf#L1776-L1798) patch which cyclizes Ser-Tyr-Gly to form gfp fluorophore. [Here's](https://github.com/ChayaSt/Charmm36_FF/blob/master/test_systems/gfp.pdb) the pdb. The patch applies to three residues.

peastman:
All of these examples should work with the design I created.  They either involve only single-residue patches, or else only one multi-residue patch.  None of them requires overlapping multi-residue patches.  So assuming these really are representative of the range of use cases, we should be in good shape.

cing:
I have a patch that joins the N-terminus of two peptides (dioxolane-linked gramicidin). Any reason why this wouldn't work?

peastman:
That should be fine.

peastman:
I'm going to go ahead and merge this, since it sounds like we're in good shape supporting all the required use cases.  Once the CHARMM force fields are converted we can check that more thoroughly.

PR #1455, opened 2016-04-10
Update README.md

tcyrus:
Change Badge from PNG to SVG


peastman:
What's the reason for this?

tcyrus:
@peastman, SVG's are vectors, so they scale when zooming in the browser or viewed on a retina screen

peastman:
Ok.  SVG used to be less well supported than PNG in web browsers, but I assume all modern browsers will have no trouble with it.

PR #1461, opened 2016-04-13
Allow periodic boundary conditions for bonded forces

### LINKED ISSUE #1411: Feature Request: Covalent bonds through periodic boundary conditions
Hi there,

I'm running a kind of calculation where a molecule is covalently bonded to itself through a periodic box (think of an infinitely long carbon nanotube or something like that).  I don't think OpenMM currently supports covalent bonds going through the periodic boundary, because when I run the simulation the system folds in on itself (i.e. atoms A and B on opposite sides of the periodic box, which are bonded to each other, both move toward the center of the box).

I think this situation could be fixed if we applied the minimum image convention before computing the displacement vectors that go into our bonded energy terms, and from the user's perspective it should be a simple on/off switch.  Is this a suggestion that makes sense?  How hard would it be to implement?

Thanks,

- Lee-Ping

peastman:
Fixes #1411.  For @leeping, who I hope is very appreciative, because *wow* this was a tedious feature to implement! :)

### DIFF
--- a/openmmapi/include/openmm/CMAPTorsionForce.h
+++ b/openmmapi/include/openmm/CMAPTorsionForce.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2010-2015 Stanford University and the Authors.      *
+ * Portions copyright (c) 2010-2016 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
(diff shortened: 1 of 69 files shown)


jchodera:
Wow, that's a lot of files changed.

leeping:
Wow, thank you so much.  That was indeed a lot of changes.  I will start using the feature right away.  :)

leeping:
Hi Peter,

We took a look at the periodic bonded forces over the past couple of days, and it seems this is still missing from AMOEBA.  Would it be easy to add this for the AMOEBA bonded forces too?  

Thanks a lot,

- Lee-Ping

peastman:
grumble.  sigh.

:p

peastman:
To make sure I don't miss anything, you want the same feature added to the following classes, right?

AmoebaAngleForce
AmoebaBondForce
AmoebaInPlaneAngleForce
AmoebaOutOfPlaneBendForce
AmoebaPiTorsionForce
AmoebaStretchBendForce
AmoebaTorsionTorsionForce

leeping:
I think AmoebaUreyBradleyForce is another one.  I have a phone call in a minute, but I will look through the list carefully as soon as I can.  :p

leeping:
Actually, the UreyBradleyForce is just a HarmonicBondForce, so I think that may not be needed.

PR #1462, opened 2016-04-13
Optimized energy minimization when adding hydrogens

### LINKED ISSUE #1456: Allow minimization tolerance used in Modeller.addHydrogens to be specified
We're having an issue where adding hydrogens to a protein on the `CPU` platform takes so long (>10 min) that it times out on travis.

It would be useful if adding hydrogens did not take this long, or at least there was some way to specify a more realistic tolerance to the minimizer used in this step.

peastman:
Fixes #1456.


jchodera:
Thanks!

PR #1464, opened 2016-04-15
Serialization works for periodic boundary conditions on bonded forces

peastman:
I forgot to do this part when implementing #1461.


swails:
Does the usesPeriodicBoundaryConditions get updated? 

I haven't had a chance to look at the code yet, but it's an easy thing to forget!

peastman:
Yes.

Issue #1440, opened 2016-03-26
Modifying simulation temperature, pressure, and parameters in an existing Context


jchodera:
I'm trying to make sure I have the correct scheme for modifying temperature, pressure, and global parameters within an existing simulation `Context` and `Integrator`, and would appreciate suggestions/advice.  There are a number of complications due to interface nonuniformity, and I'm not even sure if there's really a way to do this correctly at all.

When we want to modify the **temperature**, we have to ensure that we update
* OpenMM integrators, which unfortunately do not all have a `setTemperature()` API call; 
* If we are using a `CustomIntegrator`s, which would have to define/expose the temperature through a global parameter, but I don't think these global parameters can actually be updated via a call to `Context.setParameter()` since this only affects `Force` parameters (?)
* `MonteCarloBarostat`, which has a `setTemperature()` method, but I am unsure of whether this updates the temperature used within the `Context`; unfortunately, the API is also different from setting the pressure, which is done via a global context parameter whose name is retrieved via `MonteCarloBarostat.Pressure()`
* if any `Force` objects used a `temperature` global parameter (such as a temperature-dependent GB model), we'd have to update the `temperature` global context parameter via `Context.setParameter()`

When we want to modify the **pressure**, we would update
* `MonteCarloBarostat`, which exposes the pressure via a global context parameter whose name is retrieved via `MonteCarloBarostat.Pressure()`

When we want to update simulation context parameters that affect the energy computation, this is actually very straightforward: We can call `Context.setParameter().

Here's my best attempt at putting all of this together into code:
```
        if integrator is not None:
            if hasattr(integrator, 'setTemperature'):
                integrator.setTemperature(self.temperature)
            if hasattr(integrator, 'getNumGlobalVariables'):
                integrator_parameters = [ integrator.getGlobalVariableName(index) for index in range(integrator.getNumGlobalVariables()) ]
                if 'temperature' in integrator_parameters:
                    integrator.setGlobalVariableByName('temperature', self.temperature.value_in_unit_system(unit.md_unit_system))
                if 'kT' in integrator_parameters:
                    integrator.setGlobalVariableByName('kT', self.kT.value_in_unit_system(unit.md_unit_system))

        # Set pressure
        if self.pressure:
            forces = { self.system.getForce(index).__class__.__name__ : self.system.getForce(index) for index in range(self.system.getNumForces()) }
            barostat = forces['MonteCarloBarostat']
[...]

PR #1468, opened 2016-04-21
Changed mechanism for setting barostat temperature

peastman:
This addresses at least part of #1440.  The barostats now have a "default temperature" attribute instead of "temperature", and they define a context parameter for changing the temperature in the middle of a simulation.

Note that this breaks backward compatibility!  Hopefully it will affect very few people, but it's still important to be aware of.

### DIFF
--- a/openmmapi/include/openmm/MonteCarloAnisotropicBarostat.h
+++ b/openmmapi/include/openmm/MonteCarloAnisotropicBarostat.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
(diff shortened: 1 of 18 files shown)


jchodera:
Yay!  But does this still run the risk of having someone easily do the wrong thing by changing the integrator temperature (via `integrator.setTemperature`) but neglecting to change the barostat temperature (via context variable `temperature`), or vice-versa?  What if there was a way to link the two, or have the barostat get the integrator's current temperature?

peastman:
It doesn't do anything about that.  This only addresses one part of #1440.  We still have a variety of ways that temperature can be defined: by `setTemperature()` on a LangevinIntegrator, by a context parameter defined by AndersenThermostat, by a user-defined global parameter in a CustomIntegrator.

jchodera:
How about we brainstorm ways to make it easier for users to "do the right
thing"?

peastman:
Go for it!

PR #1469, opened 2016-04-22
Unified names of platform properties

### LINKED ISSUE #1450: Harmonize platform properties
We might consider harmonizing how we handle platform properties.

Currently, each property has its platform name as a prefix, which requires that any use of properties have a lot of platform-specific code. Not all capabilities are available on all platforms, and using properties that are not available will result in an exception, but exception handling can not be effectively used because of the inconsistencies in property names.

What about
* retain old property names for compatibility but mark them as "deprecated"---they might be removed in future
* introduce new property names which are coherent among platforms, and a `setProperty` flag that allows us to ignore exceptions if the property is not available
* introduce an API call that allows us to query if the platform has this property/capability

* CC #909: Standardize approach for handling precision

peastman:
Fixes #1450.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -746,7 +746,7 @@ double precision:
 ::
 
     platform = Platform.getPlatformByName('CUDA')
-    properties = {'CudaDeviceIndex': '0,1', 'CudaPrecision': 'double'}
+    properties = {'DeviceIndex': '0,1', 'Precision': 'double'}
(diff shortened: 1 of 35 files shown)


peastman:
The test failure is a segfault in an OpenCL test.  It doesn't seem to be connected to this PR.  I verified that I can reproduce it on the main branch.  But it only happens sometimes, not always, and only when using the CPU device.  So I'll merge this and then see if I can debug the error.

Issue #1237, opened 2015-11-03
GPU LBFGS Minimizer


rmcgibbo:
[ This is kind of a wishlist item, and it might not be worth the development effort. ]

It would be cool to have a Platform-specific LFBGS minimizer that can run on the GPU. I looked around and found https://github.com/jwetzl/CudaLBFGS, so I think all the external machinery is there, at least for CUDA. OpenCL might be another story, since IIRC they use cuBLAS.

The trickier part is that I'm not sure how you'd implement it other than as an `Integrator`. And since a context can have only one integrator, using a minimizer that's implemented as an integrator is a little bit less convenient than using `LocalEnergyMinimizer` if you want to do everything in one script, since you'd need to trash and recreate your context after you finish minimizing.

Also, perhaps this is a separate discussion, but I wonder if it's possible to handle constraints more efficiently during minimization by, instead of adding harmonic restraints, using the constraint solver to project out the component of the gradient along the constraints before returning the gradient to the solver.

rmcgibbo:
> ... using the constraint solver to project out the component of the gradient along the constraints

FWIW, I believe this is what's done in gromacs.

peastman:
If we wanted each platform to provide its own implementation, we could simply change how LocalEnergyMinimizer is implemented.  That isn't a problem.  But of course, that's only useful if we actually create separate implementations.  Also, despite the *many* times this question has come up, I don't think anyone has ever actually profiled the current implementation to determine how efficient or inefficient it is.  That seems like the first step.

Constrained optimization is a *hard* problem.  There are different algorithms designed for doing that.  For example, Simbody includes IPOpt.  They tend to be slower and less robust than unconstrained optimizers.

rmcgibbo:
I guess I don't quite see how `LocalEnergyMinimizer` can into the context and add a new kernel, but I trust you :)

I've profiled it. I implemented Nesterov as a CustomIntegrator. On CUDA, each iteration was quite a bit faster. The overall time to solution was about the same, because it got less improvement per iteration but each iteration was faster.

rmcgibbo:
This is what I did: https://github.com/rmcgibbo/openmm-cookbook/blob/master/02-nesterov-minimization.ipynb

rmcgibbo:
[...]

PR #1470, opened 2016-04-22
Improved minimization with strict constraint tolerance

peastman:
Instead of making the harmonic restraints enforce the constraints to the full tolerance, it only requires them to produce roughly correct lengths (1e-4 tolerance).  It then does a projection to remove any remaining error.

This has two benefits.  First, it means that minimization is faster when using a very strict tolerance.  Second, if the tolerance is stricter than can actually be achieved (less than about 1e-6 in single precision), this prevents the minimization from failing.  See #1237.


jchodera:
Awesome!  Thanks!

PR #1471, opened 2016-04-25
Optimized CPU implementation of CustomIntegrator

peastman:
I switched it over to using CompiledExpression instead of ExpressionProgram, which makes it a lot faster.


jchodera:
Woohoo!

PR #1479, opened 2016-05-05
Changing box vectors does not change positions

### LINKED ISSUE #1478: Prmtop simulation fails with CUDA, works ok with CPU
Hi guys,

I am trying to extend a NAMD 25ns simulation I did months ago in my desktop computer (i7-4790, 16GB RAM, Nvidia 970). I am using OpenMM 7.0-rc1 (conda build), CUDA 7.5 as provided by an Arch-Linux installation with Nvidia drivers 364.19. I am also using parmed to load the binary coordinates and velocities.

However, the simulation always fails with ValueErrors ("Energy is NaN") with CUDA. After a long debugging session, I found out that it works OK if I just change the platform from CUDA to CPU.

For example, with CUDA, the simulation fails after 50 fs if I impose no constraints nor rigid water. However, it fails immediately with rigid water and HBonds constraints. This is not the case with the CPU implementation.

During those 50 fs, the CUDA implementation reports absurd temperatures and energies, that continue rising and rising, until they hit the NaN threshold. CPU, however, lists correct temperatures (~300K).

I've attached a Zip containing the input files and an IPython notebook with the minimal code. 

Am I doing something wrong here? Thanks in advance!

[AmberRestartNaNEnergies.zip](https://github.com/pandegroup/openmm/files/247315/AmberRestartNaNEnergies.zip)

peastman:
This fixes #1478.  If positions have internally been wrapped to the periodic box boundaries, it unwraps them before changing the box dimensions.  This ensures the absolute positions do not change.

### DIFF
--- a/olla/include/openmm/kernels.h
+++ b/olla/include/openmm/kernels.h
@@ -187,7 +187,7 @@ public:
      * @param b      the vector defining the second edge of the periodic box
      * @param c      the vector defining the third edge of the periodic box
      */
-    virtual void setPeriodicBoxVectors(ContextImpl& context, const Vec3& a, const Vec3& b, const Vec3& c) const = 0;
+    virtual void setPeriodicBoxVectors(ContextImpl& context, const Vec3& a, const Vec3& b, const Vec3& c) = 0;
     /**
      * Create a checkpoint recording the current state of the Context.
      * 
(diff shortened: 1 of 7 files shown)


mpharrigan:
Were the potential numerical issues deemed not a factor here?

peastman:
I just couldn't come up with anything better.  At least it's a big improvement over the old behavior.

swails:
You could always do that imaging in double precision on the host of precision is a concern.

PR #1484, opened 2016-05-11
Fix error check for GBneck parameters

swails:
This doesn't lead to incorrect results, but it does erroneously flag some radii
as being illegal when they aren't (making it impossible to run some suggested
systems).

The parameters need to be compared to the table limits *before* subtracting off
the offset, whereas it's currently being done after.

Reported by Alberto Perez in a private email to me.

### DIFF
--- a/wrappers/python/simtk/openmm/app/internal/customgbforces.py
+++ b/wrappers/python/simtk/openmm/app/internal/customgbforces.py
@@ -335,7 +335,7 @@ def _mbondi3_radii(topology):
 
 def _createEnergyTerms(force, solventDielectric, soluteDielectric, SA, cutoff, kappa, offset):
     # Add the energy terms to the CustomGBForce.  These are identical for all the GB models.
-    
+
     params = "; solventDielectric=%.16g; soluteDielectric=%.16g; kappa=%.16g; offset=%.16g" % (solventDielectric, soluteDielectric, kappa, offset)
     if cutoff is not None:
         params += "; cutoff=%.16g" % cutoff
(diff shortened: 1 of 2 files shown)


jlmaccal:
Great! I've been meaning to look into this.

F-AH:
hello, I got the same error and tried to use your edited code .. but I am still getting the same error again. 
"ValueError: Radii must be between 1 and 2 Angstroms for neck lookup"

what I do is, I replace the customgbforces.py file with this edited one 
am I doing something wrong ??

F-AH:
hello, I got the same error and tried to use your edited code .. but I am still getting the same error again.
"ValueError: Radii must be between 1 and 2 Angstroms for neck lookup"

what I do is, I replace the customgbforces.py file with this edited one
am I doing something wrong ??

swails:
This PR was merged over 4 years ago, so this bug should be fixed in any version you're using.

If you're experiencing this issue, it's probably because your radii really *are* outside the legal range of the lookup table.

F-AH:
> This PR was merged over 4 years ago, so this bug should be fixed in any version you're using.
> 
> If you're experiencing this issue, it's probably because your radii really _are_ outside the legal range of the lookup table.

thanks a lot for your reply, I am new to openMM and cant figure out what could I do to solve this issue. I need to optimize the protein structure using implicit solvation model. but my protein has about 4 NAG molecules so that I used antechamber in ambertools to build the prmtop file and to use it for openMM. am I missing something ??

swails:
Use a different gb model. The radii on some of your atoms are incompatible with the GB model you are trying to use.

PR #1500, opened 2016-05-26
Platform::getOpenMMVersion() includes patch version

peastman:


mpharrigan:
Why not always include it? It would simplify parsing for consumers of the string

jchodera:
+1

peastman:
Because I like '7.0' better than '7.0.1'? :)

peastman:
Sorry, that should have been '7.0.0'.

PR #1504, opened 2016-05-28
[WIP] Precompute GBNeck terms

jlmaccal:
@swails Can you take a look at this incomplete PR?

The speed has improved again, but I'm not getting results that match the previous version. I suspect the problem is in `customgbforces.py` in the `_createUniqueTable` function. I'm guessing I screwed up computing the interpolated values, but I can't seem to spot the mistake.

### DIFF
--- a/wrappers/python/simtk/openmm/app/internal/customgbforces.py
+++ b/wrappers/python/simtk/openmm/app/internal/customgbforces.py
@@ -8,7 +8,7 @@ Medical Research, grant U54 GM072970. See https://simtk.org.
 
 Portions copyright (c) 2012-2015 University of Virginia and the Authors.
 Authors: Christoph Klein, Michael R. Shirts
-Contributors: Jason M. Swails, Peter Eastman
+Contributors: Jason M. Swails, Peter Eastman, Justin L. MacCallum
 
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the "Software"),
@@ -29,19 +29,26 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 
+
 from __future__ import division, absolute_import
 
 from collections import defaultdict
 import copy
+import numpy as np
+from scipy import interpolate
 from simtk.openmm.app import element as E
-from simtk.openmm import CustomGBForce, Continuous2DFunction
+from simtk.openmm import CustomGBForce, Discrete2DFunction
 import simtk.unit as u
 
+
 def strip_unit(value, unit):
+    """Strip off any units and return value in unit"""
     if not u.is_quantity(value):
         return value
     return value.value_in_unit(unit)
 
+
+# GBn and GBn2 lookup tables
 d0=[2.26685,2.32548,2.38397,2.44235,2.50057,2.55867,2.61663,2.67444,
     2.73212,2.78965,2.84705,2.9043,2.96141,3.0184,3.07524,3.13196,
     3.18854,3.24498,3.30132,3.35752,3.4136,
@@ -105,8 +112,6 @@ d0=[2.26685,2.32548,2.38397,2.44235,2.50057,2.55867,2.61663,2.67444,
     3.22855,3.28307,3.33751,3.39188,3.4462,3.50046,3.55466,3.6088,
     3.6629,3.71694,3.77095,3.82489,3.8788,3.93265,3.98646,4.04022,
     4.09395,4.14762,4.20126,4.25485,4.3084]
-
-
 
 m0=[0.0381511,0.0338587,0.0301776,0.027003,0.0242506,0.0218529,
     0.0197547,0.0179109,0.0162844,0.0148442,0.0135647,0.0124243,
@@ -192,10 +197,10 @@ m0=[0.0381511,0.0338587,0.0301776,0.027003,0.0242506,0.0218529,
     0.0332033,0.030322,0.0277596,0.0254732,0.0234266,0.0215892,
     0.0199351,0.018442,0.0170909,0.0158654,0.0147514,0.0137365,
     0.0128101,0.0119627,0.0111863]
-# Rescale to nm
-for i in range (len(d0)):
-    d0[i]=d0[i]/10
-    m0[i]=m0[i]*10
+# Rescale to nanometers
+d0 = [d / 10. for d in d0]
+m0 = [m * 10. for m in m0]
+
 
 def _get_bonded_atom_list(topology):
     """ Returns a list of atoms bonded to each other atom in a dict """
@@ -204,6 +209,7 @@ def _get_bonded_atom_list(topology):
         bondeds[a1].append(a2)
         bondeds[a2].append(a1)
     return bondeds
+
 
 def _is_carboxylateO(atom, all_bonds):
     if atom is not E.oxygen: return False
@@ -223,6 +229,7 @@ def _is_carboxylateO(atom, all_bonds):
         return False
     # If we got here, must be a carboxylate
     return True
+
 
 def _bondi_radii(topology):
     """ Sets the bondi radii """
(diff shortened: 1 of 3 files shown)


wrappers/python/simtk/openmm/app/internal/customgbforces.py
```diff
@@ -192,10 +196,10 @@ def strip_unit(value, unit):
     0.0332033,0.030322,0.0277596,0.0254732,0.0234266,0.0215892,
     0.0199351,0.018442,0.0170909,0.0158654,0.0147514,0.0137365,
     0.0128101,0.0119627,0.0111863]
-# Rescale to nm
-for i in range (len(d0)):
-    d0[i]=d0[i]/10
-    m0[i]=m0[i]*10
+# Rescale to nanometers
+d0 = [d / 10. for d in d0]
+m0 = [m / 10. for m in m0]
```
swails: This should be *10, yes? At least that's what the last version was.

jlmaccal:
@swails Thanks, I missed that.

The remaining issue is to decide on linear or cubic interpolation. I would suggest linear, as this matches Amber. But this would lead to a small inconsistency with previous versions of OpenMM. 

@peastman @jchodera Any thoughts?

For reference, here are the maximum and mean of the norm of the difference of the forces on each atom before and after this commit.

Model | Interpolation | Max (kJ/mol/nm) | Mean (kJ/mol/nm)
---------|----------|------------------------|------------------------
GBn | cubic | 0.002 | 1e-4
GBn | linear | 0.14 | 0.008
GBn2 | cubic | 0.002 | 9e-5
GBn2 | linear | 0.12 | 0.006

swails:
Definitely do what Amber does if it's easy. That's the reference implementation. 

The switch to the continuous function was a "break" from previous behavior, but more correct.

jchodera:
I still think an option to select between "reproduce amber" and "better
match the spirit of the original work" would be useful here if these are
different things.

jlmaccal:
> 
> I still think an option to select between "reproduce amber" and "better
> match the spirit of the original work" would be useful here if these are
> different things.

That should be easy. I’ll add an option for cubic or linear, with a default of linear to match Amber.

jchodera:
> That should be easy. I’ll add an option for cubic or linear, with a default of linear to match Amber.

That sounds awesome.

swails:
The difference between the two is fairly small, and would almost certainly be negligible if the parametrization scheme was applied to each one. 

But I have a problem characterizing the cubic spline approximation as "more correct" when IMO it is actually the opposite. The only reason I went with that was because it was better than no interpolation by my tests. But linear interpolation is better than both. 

I'm personally -1 on allowing a constructor parameter turn this into a subtly different GB model when all work and parameters are tuned to linear interpolation. And since other implementations use linear interpolation, parameters developed with that scheme would be q lot more easily transferable.

jlmaccal:
> 
> I'm personally -1 on allowing a constructor parameter turn this into a subtly different GB model when all work and parameters are tuned to linear interpolation. And since other implementations use linear interpolation, parameters developed with that scheme would be q lot more easily transferable. 
> 
> 

I guess the argument in favor is for cases where one is trying to reproduce or reanalyze earlier OpenMM runs. The constructor in question is in a class that’s clearly marked as being internal. The parameter won’t be accessible through the standard `openmm.app` interface, so I don’t see a downside. In 99.9% of cases, people will use the default linear option.

swails:
Oh yeah.  Good point.  I withdraw my objection then.  Thanks!

jlmaccal:
If someone wants to give this a quick once-over, this should be ready to merge. I also updated / added some docstrings.

wrappers/python/simtk/openmm/app/internal/customgbforces.py
```diff
@@ -377,36 +393,59 @@ def _createEnergyTerms(force, solventDielectric, soluteDielectric, SA, cutoff, k
...
     return _SCREEN_PARAMETERS[None]
 
+
 class CustomAmberGBForce(CustomGBForce):
+    """Base class for all of the Amber custom GB forces.
+
+    Cannot be instantiated.
+
+    """
+    __metaclass__ = abc.ABCMeta
```
swails: You can't use metaclasses in OpenMM without adding the `six` module.  Python 3 adds metaclasses via a `metaclass=` argument to the class declaration.  Without running this code through `2to3`, Python 3 will choke on this.

If you have `six`, you can do `@add_metaclass(abc.ABCMeta)` in the class decorator.  But that's why I didn't consider doing anything like this before.  Is this level of complexity really necessary?  Or can we get rid of the abstract base class stuff and go back to how it was?

swails: Well, Python 3 won't *choke* on it, but it won't actually apply the metaclass.  This is what I get for Python 3:

```python
Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:24:55) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import abc
>>> class MyClass(object):
...     __metaclass__ = abc.ABCMeta
... 
>>> type(MyClass)
<class 'type'>
```

And for Python 2 where this syntax is used for metaclasses:

```python
Python 2.7.11 (default, May  3 2016, 22:20:22) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import abc
>>> class MyClass(object):
...     __metaclass__ = abc.ABCMeta
... 
>>> type(MyClass)
<class 'abc.ABCMeta'>
```

jlmaccal:
I thought we ran things through 2to3, which does change this? I didn't use six because I don't see it used anywhere else.

jlmaccal:
FWIW, I switched to using abc because:
1. It's the idiomatic way to do this
2. It prevents you from instantiating an abstract class
3. It allows you to flag methods as abstract
4. However, these methods can be called by subclasses, like I did with `__init__`.

If it's not being handled correctly by 2to3, then by all means I can take it out.

jlmaccal:
Not sure why I thought 2to3 was being used...

It's probably just easier to remove this than to add a dependency on six.

jlmaccal:
OK, I removed the ABC stuff since we're not using 2to3 and it's not worth adding a dependency on `six` just for that. I also renamed to `CustomAmberGBForceBase` to better reflect that it is a base class.

peastman:
We used to run the code through 2to3, but #1083 removed that.

Is there really a reason to have the option for linear or cubic interpolation?  If the model was parametrized with linear interpolation, and all other codes use linear, and all public APIs give you that, then what's the value in having an option at all?  "Reproducing results from earlier versions" really doesn't apply here.  If you created a system in an earlier version and serialized it to XML, you'll still get the same result.  If you recreate your system using the public API, you'll get different results.  Adding an unused option to an internal API has no effect on that one way or another.  If you really need to recreate your system and get the same result you got before, then "reinstall OpenMM 7.0" is as reasonable an answer as "rewrite your script to use an internal API."

jlmaccal:
OK, I removed the option. It's only linear, which matches Amber.

wrappers/python/simtk/openmm/app/internal/customgbforces.py
```diff
@@ -29,19 +29,26 @@
 USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 
+
 from __future__ import division, absolute_import
 
 from collections import defaultdict
 import copy
+import numpy as np
+from scipy import interpolate
```
peastman: OpenMM doesn't require numpy or scipy, so you can't assume they'll be available.  That's why the test cases are failing.

peastman:
I suggest we get this cleaned up and merged.  Between this change and disabling the runtime compiler, I now get roughly the same performance as in earlier releases, so I think we've found all the problems.

peastman:
@jlmaccal can we get this finished up?  I sent you a PR a while ago to fix the build errors.

jlmaccal:
@peastman 

Sorry about that, I was having github email end up in spam...

I merged your PR. I'm assuming you tested the radii that are computed. If so, then this should be good to merge.

peastman:
Thanks!  I'm just waiting for the tests to run now.  Assuming they pass, I'll go ahead and merge it.

PR #1505, opened 2016-05-29
[WIP] Tests to detect small molecule parameterization failures

jchodera:
This test ensures that a protonation state of imatinib can be correctly parameterized.

On my machine, this test should fail.


jchodera:
Whoops, not sure how those torsion docs changes made it in here.

PR #1512, opened 2016-06-08
Fix periodic box vector writing in the DCD reporter

### LINKED ISSUE #1510: DCDReporter with a truncated octahedron box
DCDReporter writes trajectories in an orthogonal box even if the box is truncated octahedron. I used AMBER topology file (prmtop) with a truncated octahedron box and saved trajectories using `DCDReporter`. cpptraj program in AmberTools also gives a warning in the process of the tracjectory.
```
Warning: Trajectory box type is 'Orthogonal' but topology box type is 'Trunc. Oct.'.
```
I checked visually using VMD and the boundary of the water box looks orthogonal.

swails:
Fixes #1510


peastman:
Thanks!

PR #1518, opened 2016-06-20
Fix OpenCL on AppVeyor

peastman:


peastman:
I've fixed the OpenCL problem, but appveyor has somehow gotten *really* slow.  It runs for an hour, then times out having gotten less than half way through compiling.  It used to be a lot faster than that.  Any idea what's going on?

peastman:
I'm going to merge this, since it at least fixes the OpenCL problem.  But we still have more issues to deal with before AppVeyor will be functional.

jchodera:
I wonder if using a build cache would help?
https://www.appveyor.com/docs/build-cache

peastman:
We already are.

jchodera:
We might ask here for help/ideas:
http://help.appveyor.com/discussions/problems

Others seem to have good feedback:
http://help.appveyor.com/discussions/problems/2548-slow-build-tends-fail-because-of-test-timeout

PR #1519, opened 2016-06-21
Fixed an exception in addSolvent()

peastman:
If you gave it an empty Topology and asked it to build a water box with a fixed number of waters, it threw an exception.  This fixes it.


jchodera:
Thanks!

PR #1526, opened 2016-06-29
Added check for illegal particle index

### LINKED ISSUE #1525: Segfault in NonbondedForce::createExceptionsFromBonds()
This construct is illegal according to the API, but should throw an exception instead of a segfault:
```python
>>> from simtk import openmm
>>> force = openmm.NonbondedForce()
>>> force.createExceptionsFromBonds([(0, 1)], 1.0, 1.0)
Segmentation fault
```

peastman:
Fixes #1525.

### DIFF
--- a/openmmapi/src/CustomNonbondedForce.cpp
+++ b/openmmapi/src/CustomNonbondedForce.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2008-2014 Stanford University and the Authors.      *
+ * Portions copyright (c) 2008-2016 Stanford University and the Authors.      *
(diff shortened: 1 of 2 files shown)


frank-stonybrook:
A OpenMM simulation restart error leads me here. I am curious what's the physical meaning of an illegal particle index error.

peastman:
It means you passed a particle index to a method, but that index was either negative or greater than the number of particles.

frank-stonybrook:
Thanks! What could be the reason for an incorrect particle index? The exact simulation has been restarted two times before without any problem.

peastman:
This is a user error.  It means somewhere you called a method with an incorrect particle index.  From the exception you should be able to see exactly where you called it and figure out how the arguments were selected.

frank-stonybrook:
Thanks Peter! Seems that it is related to the mpi4py module on the Frontera. I was able to restart the same simulation using our in-house cluster. Their mpi library drives us crazy.

PR #1533, opened 2016-06-30
Made AMOEBA API more consistent

### LINKED ISSUE #435: AmoebaVdwForce uses getCutoff()/setCutoff() while NonbondedForce uses getCutoffDistance()/setCutoffDistance()
It would be great to unify/harmonize the API here at some point, even if this just means extending the Amoeba*Force API with functions that match the standard OpenMM NonbondedForce conventions.

peastman:
Fixes #435, among other things.  I replaced several methods to make the AMOEBA API more consistent with the rest of OpenMM.  The old methods are still there but marked as deprecated.


jchodera:
Woohoo!

PR #1534, opened 2016-06-30
Added functions to get per atom lab frame permanent dipoles and per atom total dipole moments for AMOEBA

saurabhbelsare:


plugins/amoeba/openmmapi/include/openmm/AmoebaMultipoleForce.h
```diff
@@ -326,6 +332,14 @@ class OPENMM_EXPORT_AMOEBA AmoebaMultipoleForce : public Force {
     void getInducedDipoles(Context& context, std::vector<Vec3>& dipoles);
 
     /**
+     * Get the total dipole moments of all particles.
+     *
+     * @param context         the Context for which to get the induced dipoles
+     * @param[out] dipoles    the induced dipole moment of particle i is stored into the i'th element
```
peastman: The above two lines should say "total" instead of "induced".  To make the function of this method completely clear, it might also be good to change the summary to, "Get the total (fixed plus induced) dipole moments of all particles."

plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp
```diff
@@ -1791,6 +1812,48 @@ void CudaCalcAmoebaMultipoleForceKernel::getInducedDipoles(ContextImpl& context,
...
+    int numParticles = cu.getNumAtoms();
+    dipoles.resize(numParticles);
+    const vector<int>& order = cu.getAtomIndex();
+    if (cu.getUseDoublePrecision()) {
+        vector<double4> posqVec;
+        vector<double> labDipoleVec;
+        vector<double> inducedDipoleVec;
+        vector<double> totalDipoleVecX(numParticles);
+        vector<double> totalDipoleVecY(numParticles);
+        vector<double> totalDipoleVecZ(numParticles);
```
peastman: I don't think you actually need the totalDipoleVec arrays, since you just set one element, use it immediately, and then never look at it again.  Those could just be local variables inside the loop.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.h
```diff
@@ -608,6 +677,10 @@ class AmoebaReferenceMultipoleForce {
             RealOpenMM thole;
             RealOpenMM dampingFactor;
             RealOpenMM polarity;
+
+            std::vector<RealOpenMM> xDipoleTraj;
+            std::vector<RealOpenMM> yDipoleTraj;
+            std::vector<RealOpenMM> zDipoleTraj;
```
peastman: What are these for?  I don't think they get used anywhere.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.h
```diff
@@ -650,6 +723,7 @@ class AmoebaReferenceMultipoleForce {
     std::vector<RealVec> _fixedMultipoleField;
     std::vector<RealVec> _fixedMultipoleFieldPolar;
     std::vector<RealVec> _inducedDipole;
+    std::vector<RealVec> _labFramePermanentDipole;
```
peastman: This never gets used.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.h
```diff
@@ -974,6 +1048,13 @@ class AmoebaReferenceMultipoleForce {
      */
     virtual void calculateInducedDipoles(const std::vector<MultipoleParticleData>& particleData);
 
+    /*
+     * Calculate permanent dipoles rotated in the lab frame.
+     * 
+     * @param particleData      vector of particle positions and parameters (charge, labFrame dipoles, quadrupoles, ...)
+     */
+ //   virtual void calculateLabFramePermanentDipoles(const std::vector<MultipoleParticleData>& particleData);
+
```
peastman: This can be deleted.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.h
```diff
@@ -1222,6 +1303,13 @@ class AmoebaReferenceGeneralizedKirkwoodMultipoleForce : public AmoebaReferenceM
      */
     void calculateInducedDipoles(const std::vector<MultipoleParticleData>& particleData);
 
+    /*
+     * Calculate fixed dipoles rotated in the lab frame.
+     * 
+     * @param particleData      vector of particle positions and parameters (charge, labFrame dipoles, quadrupoles, ...)
+     */
+ //   void calculateLabFramePermanentDipoles(const std::vector<MultipoleParticleData>& particleData);
+
```
peastman: This can be deleted.

peastman:
This looks great!  I made some comments on the code, but they're all really minor things.  You're missing one essential element though: a test case.  We need to verify that the new methods are working correctly.

Also, travis and appveyor are failing to build because you have conflicts with the main branch.  If you pull the latest changes and then resolve the conflicts, it will be able to build.

saurabhbelsare:
Hi Dr. Eastman,

I'll fix these things, and resubmit a pull request. I tested the output
manually, comparing it to a TINKER output I had ready, but didn't make a
test case because I wasn't sure how to go about it. I'll pull the changes
on travis and appveyor (I'm not sure what they are, I've never used them)
and submit another pull request?

Thank you.

Saurabh

peastman:
Travis and appveyor are what we use for continuous integration.  (See below where it says, "All checks have failed.")  You don't need to do anything with them: they'll run automatically once the conflicts are resolved.  Just do a "git pull" into your repository.  And you don't need to do a new PR.  Notice at the top where it says, "saurabhbelsare wants to merge 38 commits into pandegroup:master from saurabhbelsare:master"?  That means that any commits you add to saurabhbelsare:master will automatically be added to this one.

You can add your tests to TestCudaAmoebaMultipoleForce.cpp and TestReferenceAmoebaMultipoleForce.cpp.  Just create a simple system where you know what the correct answers are (for example by having done the same calculation in Tinker) and verify the the new methods match to a reasonable level of precision.

saurabhbelsare:
So I'll do a git pull, commit the changes that are needed, including the
test I need to add, and the push to my remote. That should do it right?

Thanks.

Saurabh

peastman:
Correct.

peastman:
Now you've completely deleted build.sh.  It's essential for packaging.  We need it back, in its original form.

saurabhbelsare:
I've added the original build.sh from the original github repo now.

​Thanks.

Saurabh

peastman:
Now you've replaced the content of build.sh with an html file?

saurabhbelsare:
Sorry I had downloaded the original build.sh incorrectly. It downloaded as
an html file instead of the actual script. I think I've fixed that now.
Does it look ok?

Thank you.

Saurabh

peastman:
Almost, but github is still showing that the file mode has change (from 100755 to 100644).  That file is a script, so it needs to be marked as executable.  You can fix that by typing

```
chmod +x build.sh
```

Then verify that `ls -l` lists the flags for it as `-rwxr-xr-x`.

saurabhbelsare:
I've marked it as an executable, and pushed it.


Thanks.

Saurabh

plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp
```diff
@@ -2011,6 +2032,48 @@ void CudaCalcAmoebaMultipoleForceKernel::getInducedDipoles(ContextImpl& context,
...
+        double totalDipoleVecX;
+        double totalDipoleVecY;
+        double totalDipoleVecZ;
+        inducedDipole->download(inducedDipoleVec);
+        labFrameDipoles->download(labDipoleVec);
+        cu.getPosq().download(posqVec);
+        for (int i = 0; i < numParticles; i++) {
+            totalDipoleVecX = posqVec[i].x * posqVec[i].w + labDipoleVec[3*i] + inducedDipoleVec[3*i];
+            totalDipoleVecY = posqVec[i].y * posqVec[i].w + labDipoleVec[3*i+1] + inducedDipoleVec[3*i+1];
+            totalDipoleVecZ = posqVec[i].z * posqVec[i].w + labDipoleVec[3*i+2] + inducedDipoleVec[3*i+2];
```
peastman: Is that really what you want to be computing?  You include the lab frame fixed dipole moment, and the induced dipole moment.  But then you also add x\*q, which is the dipole moment of a point charge relative to the origin.  That isn't what I would have expected it to return.  I would expect `getTotalDipoles()` to simply return the sum of the values from `getInducedDipoles()` and `getLabFramePermanentDipoles()`.  That is, I would expect the dipole moment of each particle relative to its own position, not relative to the origin.

saurabhbelsare: ​The reason I included the x*q term, is that when I write out the per atom
dipole moments for each atom, and then sum over all the atoms, I should be
able to reproduce the total dipole moment of the molecule. Ideally I
understand, that should be (x-xcm)​, but for a net neutral molecule, which
I was working with, the xcm contribution was zero. I could change this
function to just have the lab frame permanent dipole and induced dipole,
and add the monopole contribution manually in my python script on top, if
you feel that that would be a better way to do it.

Thanks.

Saurabh

peastman: I think that would be more intuitive behavior.  For example, for a non-polarizable point charge I'd expect it to return 0 for the dipole moment.

saurabhbelsare: I think that makes more intuitive sense. I was only thinking from the point
of view of the system I was dealing with. I'll remove the monopole
contribution and push the modified code then.

Thanks.

Saurabh

peastman: Thanks!  Remember to update the test cases as well.

saurabhbelsare: Hi Dr. Eastman,

I've modified the definitions of the functions in
plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp
and plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp
and changed the test values in
plugins/amoeba/platforms/cuda/tests/TestCudaAmoebaMultipoleForce.cpp
and plugins/amoeba/platforms/reference/tests/TestReferenceAmoebaMultipoleForce.cpp

Thanks.

Saurabh

peastman: Thanks!  Let's wait for the tests to run.  If everything passes, I think it's ready to merge.

peastman:
I think we're ready to merge.  Thanks!

saurabhbelsare:
Hi Dr. Eastman,

Thanks a lot for walking me through this. Uptill now, I had only worked
with my own git repos, so submitting a pull request to a public repo was an
entirely new process for me.

If I need to access the functions now, should I compile OpenMM from source
on the machines I'm using? Or would they be available in the conda release?

​Thanks a lot!

Saurabh

peastman:
If you need to use them before OpenMM 7.1 is released, you should build from source.

saurabhbelsare:
Is there some estimated date when this would be happening? If it going to
be in some time, I'll compile from source then.

Thanks.

Saurabh

peastman:
It'll probably be a few months.  We're still waiting on a few features to be implemented.  Then we'll put out a beta, and the release will probably follow about a month after that.

saurabhbelsare:
Thanks a lot! I'll compile from source for the moment, and then move to the
conda version when the new version is released.

Saurabh

PR #1542, opened 2016-07-18
[Incomplete] Charmm36 conversion scripts 

ChayaSt:
@peastman, [here](https://github.com/ChayaSt/openmm/tree/charmm36/devtools/forcefields/charmm) are the scripts for converting the charmm36 forcefield. 

 The `yaml` file specifies which files to convert and the references for the different forcefields and `convert_charmm.py` is where the conversion happens. There are some simple test systems in `tests/` but since this is incomplete, the testing is far from comprehensive. 

Writing out the `<patch>` tag has not yet been implemented in ParmEd, so this tag doesn't appear in the `charmm36.xml` file.


jchodera:
@peastman: I think @swails and I can help with the ParmEd changes required to write `<patch>` fields, but we'll need your help to organize the conversion scripts in the openmm repo.

The original idea was for these conversion driver scripts---which drive the production of the CHARMM `ffxml` files from a YAML file and the original CHARMM forcefield files---to live inside somewhere like `openmm/devtools/forcefield-conversion/` and have a travis branch that tests them to make sure (1) they don't break, and (2) the resulting files are usable.

peastman:
Sounds like a good plan.

jchodera:
We'll migrate this to https://github.com/choderalab/openmm-forcefields

ChayaSt:
I migrated the files here to https://github.com/choderalab/openmm-forcefields

I will continue the discussion on what still needs to get done over there so we can close this PR.

Issue #1451, opened 2016-04-05
[FR] Allow Lepton's derivative capability to extend to arbitrary variables


Lnaden:
Lepton has the ability to compute dU/dr for force evaluation. This feature request is to allow users to take derivative to arbitrary variables (or a specified set) in the energy expression to compute gradients in non-Cartesian variables such as global parameters.

jchodera:
We would definitely appreciate the ability to retrieve derivatives of the energy with respect to context parameters, and potentially even per-particle parameters.

This would be helpful for numerous applications, including
* lambda dynamics
* forcefield parameterization
* orthogonal-space random walk (OSRW)-like methods

peastman:
Could you give some examples of how this might work?  What would you do with the derivative?  Would this be a feature of a custom force?  Or a custom integrator?  Maybe give some code showing what you imagine the API might look like.

Lnaden:
I was thinking it would be declared in the custom force as explicit global variables or per particle variables, then accessed through the context and custom integrator, something like:

```python
#Declarations
expression = "myGlobalVar^3 + myPerPart;"
expression += "myPerPart = myPerPart1 * myPerPart2"
cnbf = mm.CustomNonbondedForce(expression)
cnbf.addGlobalParameter('myGlobalVar', 1)
cnbf.addPerParticleParameter('myPerPart')
#Specify which variables to compute derivative of and how to name/invoke
cnbf.addGlobalDerivative('myGlobalVar', 'dMGV') #Name of variable, name of variable to invoke in custom integrator
cnbf.addPerParticleDerivative('myPerPart', 'dMPP')

...
#Invoke its derivative in integrator by variable specified at force declaration
customIntegrator.addComputeGlobal('myGlobalVar', 'myGlobalVar + dt*dMGV')
customIntegrator.addComputePerDof('v', 'v + dt*f*dMPP/m') #per-particle treated like mass in this 

...
#Return the additional derivatives with the getForces
>>> state = context.getState(getForces=True)
>>> state.getForces()
# x, y, z, dMPP
[( 0, 0, 0, 0), ...]
>>> state.getGlobalForces() #returns dMGV
```

Not every Force will have the variable, so they would just return 0

Not sure how this would work cross force groups (e.g. `invoke energy5 + dMGV5` if that variable appears in multiple Forces + Force Groups)

Just some ideas though, @jchodera anything you think I missed in scope/need?

jchodera:
[...]

PR #1547, opened 2016-07-29
Energy derivatives with respect to global parameters

peastman:
This implements part of #1451.  You can compute derivatives of the potential energy with respect to global parameters, and either query them directly or use them in a CustomIntegrator.  Currently the following forces support it: CustomBondForce, CustomAngleForce, CustomTorsionForce, CustomNonbondedForce, CustomGBForce, CustomCompoundBondForce, CustomCentroidBondForce.

### DIFF
--- a/docs-source/usersguide/theory.rst
+++ b/docs-source/usersguide/theory.rst
@@ -1086,6 +1086,24 @@ is exactly equivalent to
 The definition of an intermediate value may itself involve other intermediate
 values.  All uses of a value must appear *before* that value’s definition.
 
+Parameter Derivatives
+*********************
+
+Many custom forces have the ability to compute derivatives of the potential energy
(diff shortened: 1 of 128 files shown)


Lnaden:
This is AWESOME! Thank you @peastman!

jchodera:
Awesome! Looks great!

jchodera:
In case anyone is looking for an easy way to try this out, note that `openmm-dev` builds are no longer called `openmm-dev`, but instead a development version of the forthcoming `openmm 7.1.0` is being pushed only to the `dev` label of the `openmm` channel. This allows you to use
```
conda install -c https://conda.anaconda.org/omnia/label/dev openmm
```
to get the latest version, and have it properly *uninstall* the previous 7.0.1 version.

PR #1548, opened 2016-08-05
Bug fixes to reading Gromacs files

peastman:


swails:
Is this why @leeping wasn't getting as good agreement b/w OMM and GROMACS with the some of the test systems?  (The systems set up by Amber were truncated octahedron if I recall correctly...)

peastman:
It's why he was using the ParmEd Gromacs reading classes instead of the OpenMM ones.  Previously it threw an exception trying to read the top file.

swails:
Ah, got it.

Issue #933, opened 2015-05-26
Custom Ewald force


peastman:
I've been looking into the possibility of creating a PME implementation for custom forces, since several people have asked about that.  It looks to me like that would be challenging at best, but perhaps one of you knows something about it I don't.

From what I've read, it looks like PME has only been derived for one very limited class of potentials: basically a_1*a_2/r^b, where a is a per-particle parameter and b>=1.  And even for that class, it hasn't been derived in a way that would be easy for a single program to support arbitrary values of b.  There isn't a simple formula you plug b into to get the equations.  It's more that there's a procedure to follow, involving a lot of algebra, for any value of b.  Furthermore, if b<=3, the sum is only conditionally convergent, so you have to use tricks to eliminate the infinity, and you can get different results (that have different physical interpretations) depending on how you apply them.

Can anyone provide more insight on this?

jchodera:
Thanks for looking into this, @peastman!

This new paper is relevant in justifying the utility of supporting PME for Lennard-Jones:
http://pubs.acs.org/doi/abs/10.1021/acs.jctc.5b00190

This uses the gromacs implementation, described in:
http://dx.doi.org/10.1021/ct400140n

which in turn references Appendix A of the Essmann et al paper (with Tom Darden):
http://dx.doi.org/10.1063/1.470117

If the straightforward generalization to other forms of a_1*a_2/r^b is of concern, I've found it much easier to start with the formulation in [this paper](http://dx.doi.org/10.1063/1.1324708).  Following [Smith](http://www.ccp5.ac.uk/infoweb/knowledge_center/smith.pdf)---which can also be extended to multipoles---everything can be computed using a family of functions B_n(r) are defined (Eq. 2.8) that have a simple recurrence relation where B_{n+1} = - (1/r) (d/dr B_n(r)).  

Here's my thought: If you let the user define an interaction kernel f(r), where B(r) = erf(alpha*r)*f(r), the recurrence series could presumably be computed with Lepton.

Not every choice of kernel f(r) would be convergent, but with some care from the user in this choice, it would probably be still of tremendous value.

peastman:
Thanks.  I've read the Gromacs paper, and of course the Essmann one.  I'll take a look at the Toukmaji paper and see if their derivation would be easier to work with.

jchodera:
I'll send you some notes that collect some of the relevant ideas.

peastman:
[...]

PR #1556, opened 2016-08-11
Dispersion PME

andysim:
Description
=========

This partly addresses the issues raised in #933, but is hardwired for the Lennard-Jones potential (I'll address the arbitrary power version in a plugin, later).  The method uses Lorentz–Berthelot combination rules in real space, but the geometric mean in reciprocal space, as developed in [this paper](http://dx.doi.org/10.1021/acs.jctc.5b00726).

Status
=====

This is far from ready, but I wanted to submit a PR early, to get feedback from anybody that would prefer a different approach to implementing this.

- [ ] Ready to go!

TODO
=====

- [x] Reference implementation, with exclusions.
- [x] Test against large cutoffs.
- [ ] Test case for forces.
- [ ] Properly handle special 1-4 interactions.
- [ ] Convert the CPU platform code, which is just ripped of from NonbondedForce right now.

### DIFF
--- a/olla/include/openmm/kernels.h
+++ b/olla/include/openmm/kernels.h
@@ -49,6 +49,7 @@
 #include "openmm/CustomTorsionForce.h"
 #include "openmm/GayBerneForce.h"
 #include "openmm/GBSAOBCForce.h"
+#include "openmm/DPMENonbondedForce.h"
 #include "openmm/HarmonicAngleForce.h"
 #include "openmm/HarmonicBondForce.h"
 #include "openmm/KernelImpl.h"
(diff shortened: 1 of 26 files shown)


jchodera:
Cool!

Is it possible to make the real-space combining rules selectable, in case we wanted to use geometric combining rules throughout?

andysim:
Sure, that shouldn't be a problem.  This version was, ahem, "inspired" by the NonbondedForce code (you can see the cut 'n' paste action in the current CPU version), which only handles LB rules.  The PME method becomes exact in the case of geometric rules, so it'd be quite a nice addition now you mention it.

peastman:
That's awesome!

I don't think we need a whole force class for this.  I'd make it just an option in NonbondedForce.  Add a method `setUsePMEForLennardJones()`.  If the nonbonded method is PME and that option is enabled (it would default to false), it uses PME for LJ as well as Coulomb.  That way, switching it on and off is very simple.  Having it as a different force class would make it a lot more difficult: you would need to build your system in a different way, which would require having separate force fields with and without LJ PME.

In the latest versions of Gromacs they handle the combining rules a bit differently.  They compute the reciprocal space part using geometric combining rules, as before.  Then in direct space they subtract off the part that came from reciprocal space, then add on the full interaction with the correct combining rules.  The result is that you get the correct combining rules with no approximation all the way out to the cutoff, and you only have the geometric combining rules beyond the cutoff.  See equations 4.186-4.188 at http://manual.gromacs.org/documentation/2016/manual-2016.pdf.

andysim:
@peastman , that Gromacs approach is what I've implemented; it's a very neat trick.  I didn't want to interfere with the NonbondedForce code, by adding too much logic, which is why I forked it originally.  Having seen how little needs to be done, I agree that having a single force class is better.  I'll migrate the new routines back over, and remove the new classes, once I shore up the tests.

jchodera:
Having this in NonbondedForce would indeed give the most utility for the
least amount of new userspace code.

@peastman: What would you think about allowing selection of the combining
rule in NonbondedForce---at least allowing the geometric sigma combining
rule as an option?

peastman:
Here's an even better option for the API: just add `DispersionPME` or `LJPME` or something like that as another value for nonbonded method.

> What would you think about allowing selection of the combining rule in NonbondedForce

I don't see any problem with that.  Is there a lot of demand for it?

jchodera:
>> What would you think about allowing selection of the combining rule in
NonbondedForce
>
> I don't see any problem with that. Is there a lot of demand for it?

* Dispersion PME is exact for this rule
* The OPLS family of forcefields use it
* Active forcefield parameterization efforts are exploring its use
* It will be easier to add more combining rule options later

andysim:
Thanks for the feedback @jchodera and @peastman.  I think it's cleanest if I start again from master, and then port my changes into NonbondedForce.  I will leave this PR open, but will be working on a different branch temporarily.

andysim:
Ok, I'm making good progress on the switch, but could use a little help with a design decision.  Before, I could only provided DPME tests for the platforms that implemented it.  Because all platforms implement NonbondedForce, this is no longer possible.  There are two different approaches I can take:-

* Implement the extra platform-specific tests in the .cpp files for the platforms that implement the new method.

* Have a method that the user (and tests) can query, to find out if the platform implements the method, skipping tests for those that don't.  This approach would see the new tests live in the header file.

I can see merits for both, but it seems like the first approach is encouraged, given the way the tests are set up.  Any thoughts?  Eventually this will all be moot when all platforms provide the new code (which shouldn't be _that_ much work), but I don't want to break things in the mean time.

peastman:
Let's just assume that all platforms will implement it.  That means some tests will fail while it's still being implemented, but that's exactly what they should do.  We won't merge it into the master branch until the feature is fully implemented, which is to say, until it works on all platforms.

By the way, I'm happy to help with this if you don't want to do all the platforms.  Reference is the most critical one.  As long as you can get that version fully working, I can then port it to any of the other platforms that you don't want to do.

andysim:
Thanks, @peastman.  I'll stick the tests in the header, and they should help with porting to other platforms.  If you could help with CUDA, that'd be great.   I have access to a CUDA machine, but need SSH, which is not convenient for train/plane travel.  I may also need help with OpenCL, but I'm not sure about that yet; I have an OpenCL-compatible graphics card, but I don't think it's a high enough spec for OpenMM.  The last time I checked, I think a number of OpenCL tests failed.

peastman:
That's fine.  For what it's worth, the CUDA and OpenCL platforms are very similar to each other.  Once one of them is implemented, porting it to the other is largely a matter of find-and-replace.

PR #1557, opened 2016-08-12
Dispersion PME

andysim:
# Description

This partly addresses the issues raised in #933, but is hardwired for the Lennard-Jones potential (I'll address the arbitrary power version in a plugin, later).  The method uses Lorentz–Berthelot combination rules in real space, but the geometric mean in reciprocal space, as developed in [this paper](http://dx.doi.org/10.1021/acs.jctc.5b00726).
# Status

This is far from ready, but I wanted to submit a PR early, to get feedback from anybody that would prefer a different approach to implementing this.
- [x] Ready to go!
# TODO
- [x] Reference implementation, with exclusions.
- [x] Test against large cutoffs.
- [x] Test case for forces.
- [x] Properly handle special 1-4 interactions, and add tests for these.
- [x] CUDA implementation
- [x] OpenCL implementation
- [x] Make sure CUDA plays nicely with cpupme, and all CUDA FFT pathways
- [x] Convert the CPU platform code, which is just ripped of from NonbondedForce right now.


andysim:
Apparently GitHub didn't like me replacing the old dpme branch, so I had to launch a new PR.  I followed your advice, and have added some preliminary tests, along with a working reference implementation in NonbondedForce.  I would like to run some Gromacs comparisons, but have never run GMX before so, if you already have some inputs available, I'd appreciate any help you can offer.  If not, I'll figure it out; their documentation is very good.

I still need to make sure that scaled/replaced 1-4 interactions get implemented properly, but I think the reference code is good to go otherwise.  I will look into CPU kernels over the next few days; feel free to hack around with CUDA and/or OpenCL at your leisure.  If you don't have time to do those ports, I'll get to them one fine day.

openmmapi/include/openmm/NonbondedForce.h
```diff
@@ -208,6 +213,16 @@ class OPENMM_EXPORT NonbondedForce : public Force {
...
     /**
+     * Get the parameters to use for dispersion term in LJ-PME calculations.  If alpha is 0 (the default),
+     * these parameters are ignored and instead their values are chosen based on the Ewald error tolerance.
+     *
+     * @param[out] dalpha   the separation parameter
+     * @param[out] dnx      the number of dispersion grid points along the X axis
+     * @param[out] dny      the number of dispersion grid points along the Y axis
+     * @param[out] dnz      the number of dispersion grid points along the Z axis
+     */
+    void getLJPMEParameters(double& dalpha, int& dnx, int& dny, int& dnz) const;
```
peastman: Can we omit the d in front of the argument names?  It's fine to use that to distinguish the internal variables, but no reason to expose it in the public API.

andysim: Sure, that's fine with me.  My intention was to make it absolutely clear that this only sets the dispersion PME parameters, but it is probably cleaner without the prefixes.  I'll remove them during my next set of commits.

andysim:
Ok, I figured out how to get GMX going, and have updated the tests to directly compare to GMX (input files are in comments in the tests).  The code I had written was almost correct, but just needed a tiny C12 shift term to match GMX.  I still need to work on the 1-4 code to make sure they're handled properly, and will add tests for those terms.  The CPU platform is next on my list after that.

andysim:
I'm hitting a strange issue that's most likely user error.  If I break in [here](https://github.com/pandegroup/openmm/blob/master/platforms/cpu/src/CpuNonbondedForceVec4.cpp#L154) and print the contents of `blockAtom` and `atom`, running [this](https://github.com/andysim/openmm/blob/398bcc8ccea17bc734b94cc4b05477169cb7d9c5/tests/TestNonbondedForce.h#L648-L702) test, I should see 9 pairs present, but only see 8.  I thought the 9th pair (0-5) that is missing would be included as a block padded with zeros.  Is there some special function to handle the incomplete blocks?

peastman:
I don't understand what you mean.  Your test has two water molecules, so six atoms.  Normally you would excluded interactions between atoms in the same molecule, giving 9 interactions.  But you don't, so every atom interacts with every other atom, giving a total of 15 interactions.

calculateBlockIxnImpl() should get called twice, once for a block of four atoms and once for the remaining two atoms.  `neighbors` will list all the atoms that anything in the block interacts with.  That would be all atoms, except that some are excluded due to symmetry.

andysim:
Peter, I am using exclusions; they're set [here](https://github.com/andysim/openmm/blob/398bcc8ccea17bc734b94cc4b05477169cb7d9c5/tests/TestNonbondedForce.h#L667).  If I add printing, I only see one block of interactions, which are {1,2,3,4}.  The triplets of numbers below are the atom number in the blocks, the neighbors to those atoms, and a junk number, respectively:-
```
In CpuNonbondedForce::threadComputeDirect, neighborList->getNumBlocks() = 1
In CpuNonbondedForceVec4::calculateBlockEwaldIxnImpl, BlockIndex = 0
Neighbors.size() 6
3  2  0.000351303
4  2  7.72058e-07
2  2  0
1  2  0
3  0  0.43928
4  0  0.00083444
2  0  0
1  0  0
3  1  0.000235163
4  1  5.56518e-07
2  1  0
1  1  0
3  5  0
4  5  0
2  5  5.56519e-07
1  5  3.3085e-07

EXCL 0  1
EXCL 0  2
EXCL 1  2
EXCL 3  4
EXCL 3  5
EXCL 4  5
```
The EXCL printout is just to confirm that the exclusion code is looping correctly, after the direct pairs have been processed.  The reference platform gives me all 9 pairs correctly, but I don't understand what I'm doing wrong to only generate a single block of interactions in the CPU code.  I tried increasing the cutoffs and dropping the number of cores to 1, but that didn't change anything.  I'm going to keep poking around but, if any ideas come to you, I'd appreciate any advice.  The number of neighbors is correctly identified as 6, but there is no block containing either 0 or 5, which is why that pair is absent.

peastman:
Are you certain calculateBlockIxnImpl() is only being called once?  It should get called twice (on two different threads), once for each block.

andysim:
That's why I went to a single core, but it didn't help.  This line of output `In CpuNonbondedForce::threadComputeDirect, neighborList->getNumBlocks() = 1` makes me think that only one will be called.  I tried setting the NonBondType to `PME` instead of `LJPME`, and the loops look the same.  I'm probably screwing something up in the setup, but I just can't see what that is right now.

andysim:
Mea culpa!  Sorry to waste your time.  I had hacked the code to use the Vec4 path for development, but looking at `CpuNeighborList::computeNeighborList`, the blocksize is 8 there.  I messed up hacking the vector length.

peastman:
That would do it. :)  If you want to force it to use vec4, the safest way is to modify isVec8Supported() in CpuNonbondedForceVec8.cpp.  If you make that return false, it will use vec4 everywhere.

andysim:
Sorry again for wasting your time earlier.  I *think* I have all the fast CPU stuff up and running for vector lengths 4 and 8, which prompts a couple of questions.  1) Are there any CI tests that explicitly check both SSE and AVX code paths, or should I be more paranoid?  2) I don't seem to be able to get the CPU platform to detect the `CalcPmeReciprocalForce` plugin, which I need to be able to begin work on that.  The plugin is being compiled and installed correctly, but setting `OPENMM_PLUGIN_DIR` to its home doesn't cause the check for optimizedPme to pass in `cpukernels.cpp`.  Any clues?

peastman:
I haven't found a good way of separately testing the two code paths, except by running the tests on two different computers, one that supports AVX and one that doesn't.  Definitely not ideal.

If you call `Platform::getPluginLoadFailures()`, that may provide information about why the PME plugin isn't getting loaded.  Do you have FFTW in your library path?

andysim:
Thanks for the suggestion.  The `Platform::getPluginLoadFailures()` list is empty, so now I'm really confused.  I hijacked `Platform::registerKernelFactory` and I don't see any of the plugins registering themselves.  I'll keep looking into this one.  In terms of AVX/SSE testing, it would be quite straightforward to add an override AVX flag to CMake, and use it in the checks in `isVec8Supported`.  The downside to that approach is yet another entry in the Travis build matrix, but at least OpenMM builds and tests quite quickly.

peastman:
Are you actually loading the plugins?

Normally, at the start of a program that uses OpenMM you call

```C++
Platform::loadPluginsFromDirectory(Platform::getDefaultPluginsDirectory());
```

to tell it to load plugins.  The Python module does that automatically, so you don't need to do it in Python programs.  And most of the test cases don't rely on dynamically loading plugins.  Instead, they directly link to the relevant libraries.  That way, we don't need to worry about installation directories and such when running test cases.

For example, take a look at TestCudaDrudeForce.cpp.  At the top it declares the external function

```C++
extern "C" void registerDrudeCudaKernelFactories();
```

And then the very first thing `main()` does is to call that function.  The implementation of that function is in CudaDrudeKernelFactory.cpp.  All it does is verify that the CUDA platform has been registered, then call `registerKernelFactories()`.  When the plugin is loaded in the normal way, that gets called automatically.  But for test cases, we provide a separate registration function.

andysim:
@peastman, thanks a lot for the detailed explanation and pointers.  I mistakenly thought that all plugins were loaded when running the tests.  My plan is to modify the CpuPME plugin to handle the LJPME terms, so that the optimized PME code path works correctly.  To test this, I think it makes most sense to add some tests into `TestCpuNonbondedForce.cpp`, along the lines of the Drude test that you pointed out.  Does that sound reasonable, or would you prefer a different strategy?

andysim:
Alternatively, I could add some LJPME tests into the plugin's test directory, in `TestCpuPme.cpp`; that would allow me to test the reciprocal space code independently of the NonbondedForce class, but would prevent issues with importing plugins, as you said.

peastman:
Testing it in TestCpuPme.cpp is probably the easiest way of doing it.

jchodera:
Just wondering how this was coming along---lots of us are still really excited about this functionality!

andysim:
Sorry, @jchodera , I completely forgot to respond to your message.  The hard part is pretty much done at this point; the CPU and reference platforms both pass the tests, with and without exclusions.  I haven't addressed the geometric mean combination rules feature that you requested yet, but that should be very straightforward.  My only concern at this point is figuring out how to add the special 1-4 interactions that CHARMM uses, and making sure they work correctly.  I also need to test the scaled 1-4 stuff.  I'll make sure I finish this up in the next couple of weeks, now that my todo list is starting to clear up.

jchodera:
I just wanted to resurrect this thread to see if there is anything we can do to move it along!

andysim:
Quick update:  The CUDA code is pretty much done now, sorry for the long time on the backburner.  I still need to hook it into CPUPME, which should be easy enough to do, and I need to make sure that it works correctly without streams and without cuBLAS.  If anybody is interested in porting that code over to OpenCL, I'm happy to let them do it.  If not, I'll try to get to it whenever I can.  I think the only major thing left is figuring out how exactly the 'special' 1-4 groups are handled in CHARMM FFs; the 1-4 interactions have special LJ parameters so I'm concerned this may trigger a different code path.  If anybody has any GMX files for a small molecule with the special 1-4s, they would be extremely helpful.  Also, any pointers about how OpenMM is run with the special 1-4 terms would be great.  These terms won't matter for LJPME, but I want to try my best to make sure this code is absolutely correct, and fully matches other codes.

peastman:
Thanks!  I'll take a look at it shortly.  1-4 interactions are handled as "exceptions".  They get excluded from the main nonbonded calculations, just like 1-2 and 1-3 interactions.  The only difference is that we then have a separate loop to add them back in with the modified parameters.  Here's the relevant code in different platforms:

Reference and CPU: https://github.com/pandegroup/openmm/blob/master/platforms/reference/src/SimTKReference/ReferenceLJCoulomb14.cpp

CUDA: https://github.com/pandegroup/openmm/blob/master/platforms/cuda/src/kernels/nonbondedExceptions.cu

andysim:
Thanks, @peastman, that's exactly what I need.  I figured it would be simple, but it's good to get help with making sure I cover all possible code paths.  I'll make sure I add a test that includes the special 1-4 case.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -1650,14 +1668,29 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
...
+            C6s = CudaArray::create<double>(cu, cu.getPaddedNumAtoms(), "C6s");
+        else
+            C6s = CudaArray::create<float>(cu, cu.getPaddedNumAtoms(), "C6s");
+    }
     CudaArray& posq = cu.getPosq();
     vector<double4> temp(posq.getSize());
     float4* posqf = (float4*) &temp[0];
     double4* posqd = (double4*) &temp[0];
+    // The C6 coefficients for LJPME could be computed from sigma and epsilon, but it
+    // seems like a good idea to cache them and avoid many recomputations.
```
peastman: On a GPU, memory access is much more expensive than computation.  If it takes less than a few hundred flops to compute something, that's probably faster than loading it from memory.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -1796,21 +1880,44 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
...
                 directPmeGrid = new CudaArray(cu, gridSizeX*gridSizeY*gridSizeZ, cu.getComputeCapability() >= 2.0 ? 2*elementSize : 2*sizeof(long long), "originalPmeGrid");
                 reciprocalPmeGrid = new CudaArray(cu, gridSizeX*gridSizeY*gridSizeZ, 2*elementSize, "reciprocalPmeGrid");
                 cu.addAutoclearBuffer(*directPmeGrid);
+                if(doLJPME){
+                    directDispersionPmeGrid = new CudaArray(cu, dispersionGridSizeX*dispersionGridSizeY*dispersionGridSizeZ,
+                                            cu.getComputeCapability() >= 2.0 ? 2*elementSize : 2*sizeof(long long), "originalDispersionPmeGrid");
+                    cu.addAutoclearBuffer(*directDispersionPmeGrid);
+                    reciprocalDispersionPmeGrid = new CudaArray(cu, dispersionGridSizeX*dispersionGridSizeY*dispersionGridSizeZ,
+                                                                2*elementSize, "reciprocalDispersionPmeGrid");
+                }
```
peastman: We can reuse the same arrays for Coulomb and LJ PME.  Just make sure to allocate them large enough for whichever one is bigger, then clear the direct grid before starting the second one.  Likewise for pmeAtomDispersionGridIndex: we can reuse the same array.  This will save a lot of GPU memory.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -2045,8 +2185,80 @@ double CudaCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeF
             cuEventRecord(pmeSyncEvent, pmeStream);
             cu.restoreDefaultStream();
         }
+
+        // As written, we check only the Electrostatic grid pointer to get here.  We could separate them out, but for
+        // now we assume that LJPME can only be used if electrostatic PME is also active.
+        if(doLJPME){
+            if (usePmeStream)
+                cu.setCurrentStream(dispersionPmeStream);
```
peastman: We should only set the stream once at the start of doing PME, and only reset it and call `cuEventRecord()` once at the end of doing PME (so also remove lines 2184-2187 above).  Both PME calculations get done on the same stream, and both need to complete before the main stream can finish the step.

andysim: Thanks for the feedback.  Yeah, I went back and forth about whether to make a new stream or not, and eventually decided that having the extra stream to overlap communication and computation would be worth trying.  I'll consolidate the memory and streams, as you suggested.

platforms/cuda/src/CudaKernels.cpp
andysim: Will do.

platforms/cuda/src/CudaKernels.cpp
andysim: This is mainly for the spreading / probing routines.  We'd still have to pass sigma and epsilon anyway, so I think it's worth having some kind of C6 cache.  I actually thought about going further, and making this a posq analog, so that there would be fewer cache misses in the direct space code, but went with just a plain C6 array as a first pass.

platforms/cuda/src/CudaKernels.cpp
peastman: Those routines are already totally dominated by memory access.  A few extra flops to compute the coefficient will have no effect.

platforms/cuda/src/CudaPlatform.cpp
```diff
@@ -247,6 +247,10 @@ CudaPlatform::PlatformData::PlatformData(ContextImpl* context, const System& sys
         CHECK_RESULT(cuDeviceGetName(name, 1000, contexts[i]->getDevice()), "Error querying device name");
         deviceName << name;
     }
+    size_t printfsize;
+    cuCtxGetLimit(&printfsize, CU_LIMIT_PRINTF_FIFO_SIZE);
+    cuCtxSetLimit(CU_LIMIT_PRINTF_FIFO_SIZE, 10*printfsize);
```
peastman: Is this debugging code that got left in?

andysim: It is, yes.  I thought I had rolled the file back before committing.  Sorry about that.

platforms/cuda/src/CudaKernels.cpp
andysim: Ok, so I completely agree about recomputing the C6 coefficients in the real space code.  For the reciprocal space, I have to pass data [here](https://github.com/pandegroup/openmm/pull/1557/files#diff-6647053ad41d2b5cc13dce3d49c64b1aR25) and [here](https://github.com/pandegroup/openmm/pull/1557/files#diff-6647053ad41d2b5cc13dce3d49c64b1aR244).  My first pass was to make a new C6 array, but now I think it'd be much better to group the C6 and xyz coords together in a posq-like array instead.  Passing in xyz, sigma and epsilon separately makes no sense to me.  Passing C6 as a distinct array is far less cache efficient than bundling with xyz, as is done for posq.  Agree?  Btw, I granted write permission to my repo in case you want to mess around with any of this stuff.  I'm happy to make the changes, but wanted to give you an easy route to tweaking things if you'd rather do that too.

peastman: Let me muck around with it for a while, and then I'll send a PR to your branch.  While I'm at it, I'd like to see if I can eliminate some of the duplicated code.  `pme.cu` and `ljpme.cu` are nearly identical, so I'd like to see if I can combine them with just a few ifdefs or something similar for the varying parts.

andysim: Sounds good.  I agree about the nasty code duplication, and would rather avoid it.  If we pack C6 into a posq-like array, the spreading and probing codes become identical, but for the grid size macros.

peastman: I can also clean up the other things I mentioned at the same time.

peastman:
I just sent a PR with a bunch of cleanup.  Next on my list is to figure out how to compute the PME parameters based on the requested accuracy.  The calculation in `NonbondedForceImpl::calcPMEParameters()` is an empirical formula based on running a whole lot of tests with different systems to see how the accuracy varied.  Dispersion PME will probably behave differently, so we need to find a new formula for that.

andysim:
Sorry I thought I had added a TODO in the code for the LJPME PME parameter setup, but obviously overlooked that.  Perhaps we could do a similar trick to [this code](https://github.com/gromacs/gromacs/blob/master/src/gromacs/math/calculate-ewald-splitting-coefficient.cpp)?  Thanks for cleaning up my messy code; my emphasis was correctness over efficiency for the first pass, and I should've better communicated the WIP nature of this PR.  Getting closer though...

peastman:
It's something like that, except it's calculating something different.  In Gromacs, the user specifies the maximum allowed error at the cutoff for a single direct space interaction.  In OpenMM they specify the total error in the force on each atom, after summing all interactions and including both direct and reciprocal space.  That's a much more useful quantity, but we don't have any rigorous theory for calculating it, so we have to rely on empirical formulas.

I've found one that seems to work pretty well, so I'll implement that.  It's possible I'll change it in the future though.  As you'd expect, the grid size can be a lot smaller for LJ than for Coulomb.  That likely means the cost is totally dominated by charge spreading and the FFTs are negligible.  If so, it might be faster to use 4th order splines instead, which would reduce that cost but require a bigger grid.  I won't know for sure until I profile it.

First, though, I have lots of other things to do, including adding serialization, supporting this in the Python layer classes, and writing documentation.  I'll send another PR with those things in a bit.

andysim:
Thank you very much for all your help with this. I'm tied up for the next couple of days, but will be on the case again this weekend. I'm getting a gromacs test case together that will confirm the correctness of the 1-4 interactions. I'm happy to help with any other aspects of the coding/docs. Just let me know. I agree that spreading/probing will be a bottleneck, so I think it's worth making a duplicate posq array with c6 in the fourth slot, for those routines. More global memory used, but very cache friendly.

peastman:
> I think it's worth making a duplicate posq array with c6 in the fourth slot, for those routines.

Actually that turned out to be completely irrelevant, because it happens in the outer loop.  It only needs to compute or look up c6 once per atom, and then spreads that charge over 125 grid points.  So compared to the work done per grid point, it's negligible.

peastman:
I've found something I think is a bug, but I want to check to make absolutely sure (especially because the test cases are still passing).  In the CUDA implementation, you only compute the spline moduli (`pmeBsplineModuliX`, `pmeBsplineModuliY`, and `pmeBsplineModuliZ`) once, then use them for both Coulomb and LJ.  But they depend on the grid size, which can be different for the two.  Should it instead compute two different sets of moduli, or am I misunderstanding it?

peastman:
I've been trying to write some test cases to verify everything is working correctly, but the results keep being different from what I expect.  The following test creates a box of randomly positioned particles (but making sure they aren't too close together to avoid large energies), then computes the energy with short (1 nm) and long (3 nm) cutoffs, as well as with LJPME.

```C++
void testConvergence() {
    // Create a cloud of random point charges.

    const int numParticles = 1000;
    const double boxWidth = 6.0;
    System system;
    system.setDefaultPeriodicBoxVectors(Vec3(boxWidth, 0, 0), Vec3(0, boxWidth, 0), Vec3(0, 0, boxWidth));
    NonbondedForce* force = new NonbondedForce();
    system.addForce(force);
    vector<Vec3> positions(numParticles);
    OpenMM_SFMT::SFMT sfmt;
    init_gen_rand(0, sfmt);

    for (int i = 0; i < numParticles; i++) {
        system.addParticle(1.0);
        force->addParticle(0.0, 0.1+0.3*genrand_real2(sfmt), genrand_real2(sfmt));
        while (true) {
            Vec3 pos = Vec3(boxWidth*genrand_real2(sfmt), boxWidth*genrand_real2(sfmt), boxWidth*genrand_real2(sfmt));
            double minDist = boxWidth;
            for (int j = 0; j < i; j++) {
                Vec3 delta = pos-positions[j];
                minDist = min(minDist, sqrt(delta.dot(delta)));
            }
            if (minDist > 0.15) {
                positions[i] = pos;
                break;
            }
        }
    }
    
    // Compute the energy with short and long cutoffs, and compare them to LJPME.
    // The long cutoff should match the LJPME result better than the short cutoff.
    
    force->setNonbondedMethod(NonbondedForce::LJPME);
    force->setCutoffDistance(1.0);
    force->setEwaldErrorTolerance(1e-5);
    VerletIntegrator integrator(0.01);
    Context context(system, integrator, platform);
    context.setPositions(positions);
    double pmeEnergy = context.getState(State::Energy).getPotentialEnergy();
    force->setNonbondedMethod(NonbondedForce::CutoffPeriodic);
    context.reinitialize();
    context.setPositions(positions);
    double shortEnergy = context.getState(State::Energy).getPotentialEnergy();
    force->setCutoffDistance(3.0);
    context.reinitialize();
    context.setPositions(positions);
    double longEnergy = context.getState(State::Energy).getPotentialEnergy();
    printf("%.10g %.10g %.10g\n", pmeEnergy, shortEnergy, longEnergy);
}
```

I'd expect the long cutoff to agree with LJPME much better than the short cutoff.  But that isn't what I see.  Here's the output with the reference platform:

```
59317.46161 59286.18411 59286.11766
```

The short and long cutoffs give very similar energies, but the LJPME energy is very different.  Any idea what's going on?

andysim:
Good point about the spreading/probing communication to work ratio.  Even with 4X4X4=64 grid points to cover per atom, there should be sufficient work.  You're right about the splines - I completely overlooked that fact.  I figured that fixing the spline order, allowing the grid to vary, seemed like a reasonable strategy so I also decided to recycle the splines for the CUDA code (the reference and CPU platforms look correct because the call `pme_init`).

When I coded this in CHARMM I found that a reasonable rule of thumb is that, for fixed spline order, you can use a grid about 2/3 as dense in each dimension.  For this reason, I should have added a test case that uses grid density for electrostatic and dispersion terms, which would've caught the spline error in the CUDA LJPME terms.

andysim:
Regarding the test, I am not sure what's going wrong there.  In [this](https://github.com/pandegroup/openmm/pull/1557/files#diff-609bb01b364651cce3c10ebf36c0e55bR1138) test, I use a 30A cutoff, which matches the LJPME result very well.  I've seen some tests with high density boxes where I needed really long (~90A) cutoffs to get very good agreement.  Your problem looks like more serious issue though, because the LJPME energy should always be lower than a cutoff result (C6 is attractive) and that's not the case here.  I wonder if I screwed up the cutoff code somehow?  I'll keep digging because this is a great test case and clearly shows a bug.

peastman:
Ok, I'll send a PR with the fix for that and various other changes shortly.

andysim:
@peastman, the test is failing because the LJPME alpha, and grid dimensions are all being returned as zero, so there's something wrong with the default values.  I'll have to dig through the logic for that, but setting the alpha and grid dimensions explicitly fixes the problem.  I also added PBC to the distance check:
```cpp
    vector<Vec3> boxVecs;
    boxVecs.push_back(Vec3(boxWidth, 0, 0));
    boxVecs.push_back(Vec3(0, boxWidth, 0));
    boxVecs.push_back(Vec3(0, 0, boxWidth));
    system.setDefaultPeriodicBoxVectors(boxVecs[0], boxVecs[1], boxVecs[2]);
```

then further down...

```cpp
            for (int j = 0; j < i; j++) {
                Vec3 delta = pos-positions[j];
                delta -= boxVecs[2]*floor(delta[2]/boxVecs[2][2]+0.5);
                delta -= boxVecs[1]*floor(delta[1]/boxVecs[1][1]+0.5);
                delta -= boxVecs[0]*floor(delta[0]/boxVecs[0][0]+0.5);
                minDist = min(minDist, sqrt(delta.dot(delta)));
            }
```

andysim:
A couple of followup comments.  First, the PBC modifications make no difference for this test, but I think they should be present.  Also, the reason that a zero alpha doesn't return the same result as the regular short cutoff LJ method is because of the [potential shift](https://github.com/pandegroup/openmm/pull/1557/files#diff-01baa9645a5e123943d658a9c80d8c3eR464) used to account for the different combination rules at the cutoff.  If geometric mean rules were used, a zero alpha would yield identical results to the short cutoff vanilla LJ method.

peastman:
If `alpha` is zero (the default), that means to calculate the parameters based on the requested error tolerance.  See `NonbondedForceImpl::calcPMEParameters()`.

andysim:
I agree, but if I add 
```cpp
    double alpha;
    int gx, gy, gz;
    force->getLJPMEParameters(alpha, gx, gy, gz);
    printf("%f %d %d %d\n", alpha, gx, gy, gz);
```
to the test, I see zeros being returned.

peastman:
As it should.  Instead call

```C++
force->getLJPMEParametersInContext(context, alpha, gx, gy, gz);
```

That will tell you the values actually being used in a particular context.

andysim:
Ah, my bad.  Calling that reveals that the grid has 12 points per edge.  Experimenting with different grid sizes reveals the following:
```
12 89089.85501 89054.13374 89054.12254
24 89044.80685 89054.13374 89054.12254
48 89055.66622 89054.13374 89054.12254
96 89055.74461 89054.13374 89054.12254
```
so I think the heuristic may be returning too coarse a grid.

jchodera:
> Part of the problem is that the long range dispersion correction is really ridiculously effective. I was finding almost no difference at all between a 0.5 nm cutoff and a 3.0 nm cutoff, which just didn't seem right. Then I realized that was because we were correcting for the difference.

Which system are you examining? For something like an kinase in water, the error between short and long cutoffs can be much larger. You can test out the [`SrcExplicit`](https://github.com/choderalab/openmmtools/blob/master/openmmtools/testsystems.py#L3485-L3516) example from `openmmtools.testsystems` if you want a convenient heterogeneous system.

peastman:
I was referring to the test case given above, so a random cloud of particles.  For a highly anisotropic system the correction would presumably work less well.

andysim:
I had forgotten about the LRC being on by default, and I had the same problem wrapping my head around the insensitivity w.r.t. cutoffs.  I think it makes more sense to target forces instead of energies, especially for LJPME, so your heuristic is probably good as-is.  The energy shift that is applied, to make sure the dimer potentials are continuous at the cutoff, do not affect the forces, so there's a bit of arbitrariness to the actual energies.  To that end, we may want to repeat your random particles test but with all sigma values set to 2, instead of a random number.  Then, because 2+2 = 2*2, the multiplicative and additive rules become equivalent and the LJPME result should *exactly* match the brute force result, even with very short cutoffs.

peastman:
Does anyone have a test case where we expect dispersion PME to actually make a difference?  I've been doing lots of testing, but can't find any situation where there's a measurable benefit from it.

As far as forces are concerned, it seems to be completely unnecessary.  Just using a simple cutoff, the forces are nearly identical with a 1.0 or 2.0 nm cutoff.  Any individual force beyond 1 nm is basically negligible.  And the sum of all forces beyond 1 nm is even more negligible, because they all tend to cancel each other out.

For energies there's more of a difference.  All the interactions beyond the cutoff add together with the same sign, so combined they make a measurable difference.  For isotropic systems, the long range correction does an almost perfect job of compensating for that, but presumably it would do less well in non-isotropic systems.

But as Andy pointed out, we don't care about constant shifts to the energy, only about how it varies.  Variations with particle positions will generally be negligible.  Even if your protein undergoes a conformational transition, it's still going to be a compact globular protein in solvent, or a protein embedded in a uniform membrane.  The sum of all those long range LJ energies isn't going to change significantly (which is just another way of saying it doesn't affect the forces).

In the Gromacs paper, they claimed it did make a difference for pressure regulation: it changed how the energy varied with box volume.  But my tests so far show the opposite.  I create a 6 nm cube containing 1000 random particles.  I evaluate the energy with cutoffs of 1 and 3 nm, and with LJPME.  Then I scale the box up to 6.1 nm, compute the energies again, and finally take the difference to see how the energies changed.  Here are the results using a fairly fine grid (36x36x36):

Box Size | 1 nm | 3 nm | LJPME
-------- | ---- | ---- | -----
6 nm | 59294.8373 | 59286.4556 | 59287.05576
6.1 nm | 48446.44577 | 48438.42659 | 48438.87017
change | 10848.39154 | 10848.029 | 10848.18559

The 3 nm cutoff and LJPME agree with each other almost perfectly in all cases.  The energy with a 1 nm cutoff is slightly different (though only slightly).  But it's mostly just an additive constant.  The change in energy when scaling the box is identical to the other cases.

Here's the same test, only using a coarser grid (24x24x24):

Box Size | 1 nm | 3 nm | LJPME
-------- | ---- | ---- | -----
6 nm | 59294.8373 | 59286.4555 | 59277.57722
6.1 nm | 48446.44576 | 48438.42705 | 48428.65362
change | 10848.39154 | 10848.02845 | 10848.9236

The LJPME energies now agree less well with the long cutoff.  But it's just shifted by a constant, independent of the box size.  The change in energy when scaling the box is unchanged.  As far as pressure regulation is concerned, there's no difference between a short cutoff, a long cutoff, PME with a coarse grid, or PME with a fine grid.

jchodera:
> Does anyone have a test case where we expect dispersion PME to actually make a difference? I've been doing lots of testing, but can't find any situation where there's a measurable benefit from it.

Here's a script to evaluate the errors w.r.t. a long cutoff for a relevant system: Src kinase in explicit solvent. Keep in mind this is for a single snapshot, but it shows how big the problem is:
```python
import numpy as np

# Create a test system                                                                                                                                                                                                                                                                                                      
testsystem = testsystems.SrcExplicit()
[system, positions] = [testsystem.system, testsystem.positions]

cutoffs_to_try = np.arange(2.0, 0.70, step=-0.05) * unit.nanometers

# Zero out charges                                                                                                                                                                                                                                                                                                          
    if force.__class__.__name__ == 'NonbondedCutoff':
        print('Zeroing charges...')
        for index in range(system.getNumParticles()):
            [charge, sigma, epsilon] = force.getParticleParameters(index)
            force.setParticleParameters(index, charge*0.0, sigma, epsilon)
        print('Disabling switch...')
        force.setUseSwitchingFunction(False)
        print('Turning on long-range dispersion correction...')
        force.setUseDispersionCorrection(True)

reference_potential = None
for cutoff in cutoffs_to_try:
    # Update cutoff                                                                                                                                                                                                                                                                                                         
        if hasattr(force, 'setCutoffDistance'):
            force.setCutoffDistance(cutoff)

    integrator = openmm.VerletIntegrator(1.0*unit.femtoseconds)
    potential = context.getState(getEnergy=True).getPotentialEnergy()

    if reference_potential is None:
        reference_potential = potential
    error = potential - reference_potential

    print('%8.3f nm : %16.8f kJ/mol | error: %16.8f kJ/mol' % (cutoff/unit.nanometers, potential/unit.kilojoules_per_mole, error/unit.kilojoules_per_mole))
```
Here's the output. Keep in mind that people generally run with a 0.9 nm cutoff:
```
mski1776:openmmtools choderaj$ python test_dispersion_correction.py
   2.000 nm : -907653.78159413 kJ/mol | error:       0.00000000 kJ/mol
   1.950 nm : -907670.46745610 kJ/mol | error:     -16.68586197 kJ/mol
   1.900 nm : -907651.22383811 kJ/mol | error:       2.55775602 kJ/mol
   1.850 nm : -907652.38034839 kJ/mol | error:       1.40124573 kJ/mol
   1.800 nm : -907666.92406484 kJ/mol | error:     -13.14247072 kJ/mol
   1.750 nm : -907662.05211800 kJ/mol | error:      -8.27052387 kJ/mol
   1.700 nm : -907658.03835591 kJ/mol | error:      -4.25676179 kJ/mol
   1.650 nm : -907652.33457311 kJ/mol | error:       1.44702102 kJ/mol
   1.600 nm : -907657.33348842 kJ/mol | error:      -3.55189430 kJ/mol
   1.550 nm : -907643.92169520 kJ/mol | error:       9.85989893 kJ/mol
   1.500 nm : -907657.07567634 kJ/mol | error:      -3.29408221 kJ/mol
   1.450 nm : -907632.17093115 kJ/mol | error:      21.61066297 kJ/mol
   1.400 nm : -907643.77166139 kJ/mol | error:      10.00993274 kJ/mol
   1.350 nm : -907649.92770502 kJ/mol | error:       3.85388911 kJ/mol
   1.300 nm : -907640.03433439 kJ/mol | error:      13.74725973 kJ/mol
   1.250 nm : -907631.86792548 kJ/mol | error:      21.91366864 kJ/mol
   1.200 nm : -907646.13474405 kJ/mol | error:       7.64685008 kJ/mol
   1.150 nm : -907620.16782411 kJ/mol | error:      33.61377002 kJ/mol
   1.100 nm : -907608.84353969 kJ/mol | error:      44.93805444 kJ/mol
   1.050 nm : -907616.72676218 kJ/mol | error:      37.05483195 kJ/mol
   1.000 nm : -907603.44579349 kJ/mol | error:      50.33580063 kJ/mol
   0.950 nm : -907537.47760337 kJ/mol | error:     116.30399076 kJ/mol
   0.900 nm : -907486.87144200 kJ/mol | error:     166.91015213 kJ/mol
   0.850 nm : -907513.56043700 kJ/mol | error:     140.22115712 kJ/mol
   0.800 nm : -907457.01125459 kJ/mol | error:     196.77033953 kJ/mol
   0.750 nm : -907299.94170507 kJ/mol | error:     353.83988906 kJ/mol
```

peastman:
> but it shows how big the problem is:

What is the problem it's supposed to be showing?  If all it's doing is adding a constant shift to the energy, that will have no effect either on dynamics or analysis.  It's only a problem if the shift varies with conformation.

jchodera:
> What is the problem it's supposed to be showing? If all it's doing is adding a constant shift to the energy, that will have no effect either on dynamics or analysis. It's only a problem if the shift varies with conformation.

True, but it does. This can be quantified by the variance of that error quantities. We found this to be large. Let me get statistics on that for you.

jchodera:
Here's a more complicated script (attached) that runs a simulation with 9A cutoff and quantifies the mean and stddev difference in the Lennard-Jones potential between 9A and a new cutoff (with dispersion correction on):
```
   2.000 nm : mean error:     -35.04119836 kT | error stddev:       8.19233053 kT
   1.950 nm : mean error:     -36.23484015 kT | error stddev:       7.94044903 kT
   1.900 nm : mean error:     -35.48341127 kT | error stddev:       8.03373766 kT
   1.850 nm : mean error:     -35.11914840 kT | error stddev:       8.51370531 kT
   1.800 nm : mean error:     -34.48976800 kT | error stddev:       7.74951817 kT
   1.750 nm : mean error:     -34.66615518 kT | error stddev:       7.72981236 kT
   1.700 nm : mean error:     -35.37102670 kT | error stddev:       8.04479569 kT
   1.650 nm : mean error:     -34.14559349 kT | error stddev:       7.74331924 kT
   1.600 nm : mean error:     -34.76932007 kT | error stddev:       7.08635773 kT
   1.550 nm : mean error:     -33.61615149 kT | error stddev:       7.15930411 kT
   1.500 nm : mean error:     -34.09999525 kT | error stddev:       6.74961246 kT
   1.450 nm : mean error:     -33.03803365 kT | error stddev:       6.72121262 kT
   1.400 nm : mean error:     -31.20494044 kT | error stddev:       6.48516368 kT
   1.350 nm : mean error:     -30.65402354 kT | error stddev:       6.76041545 kT
   1.300 nm : mean error:     -29.68634848 kT | error stddev:       5.62408763 kT
   1.250 nm : mean error:     -29.28121546 kT | error stddev:       5.93075867 kT
   1.200 nm : mean error:     -25.69135178 kT | error stddev:       5.01019289 kT
   1.150 nm : mean error:     -23.21106503 kT | error stddev:       5.51746380 kT
   1.100 nm : mean error:     -22.39513368 kT | error stddev:       5.45972817 kT
   1.050 nm : mean error:     -18.82758785 kT | error stddev:       4.04089714 kT
   1.000 nm : mean error:     -11.06097081 kT | error stddev:       3.54249853 kT
   0.950 nm : mean error:      -2.44814666 kT | error stddev:       4.18131598 kT
   0.900 nm : mean error:      -0.00209545 kT | error stddev:       0.01636701 kT
   0.850 nm : mean error:       6.96414078 kT | error stddev:       4.12495334 kT
   0.800 nm : mean error:      24.64786711 kT | error stddev:       5.12254322 kT
   0.750 nm : mean error:      43.26176613 kT | error stddev:       8.89713821 kT
```
The stddev between 0.9 nm (typical simulation cutoff) and 2 nm (where the isotropic dispersion correction is valid) is ~8 kT, meaning there is essentially no thermodynamic overlap between these two ensembles.

[test_langevin_splitting_integrator.py.zip](https://github.com/pandegroup/openmm/files/746454/test_langevin_splitting_integrator.py.zip)

peastman:
I think you attached the wrong file.

jchodera:
Whoops!  Here you go.
[test_dispersion_correction_variance.py.zip](https://github.com/pandegroup/openmm/files/746476/test_dispersion_correction_variance.py.zip)

peastman:
> meaning there is essentially no thermodynamic overlap between these two ensembles.

Why do you say that?  This is a big system (56k atoms, over 100k degrees of freedom).  So that standard deviation is less than 0.0001 kT/dof, which is tiny.

jchodera:
> Why do you say that? This is a big system (56k atoms, over 100k degrees of freedom). So that standard deviation is less than 0.0001 kT/dof, which is tiny.

If only thermodynamics respected relative energy differences instead of absolute ones.

How well does the dispersion PME handle something of this size? I believe we found that PME for electrostatics has very minimal stddev as a function of cutoff for this system.

Unfortunately, we don't have any membrane test systems easily accessible. Have you guys tried any of those?

peastman:
> If only thermodynamics respected relative energy differences instead of absolute ones.

Much of it does.  Any intensive quantity will only depend on relative differences.  Such as pressure, which is the main thing LJPME was proposed to address.

What is the physical quantity you want to calculate for which this is relevant?

jchodera:
> What is the physical quantity you want to calculate for which this is relevant?

Free energies of ligand binding using cutoffs large enough so that the computed free energy is insensitive to cutoff. These are log ratios of normalizing constants.

Free energies are extensive quantities.

jchodera:
Maybe you want to give a GPCR system in a membrane a try and see if it has an effect on the average box volume with a barostat? You can find some gromacs input files at the end of [this tutorial page](https://sites.google.com/site/anirbanzz/gpcr-gromacs-tutorial).

peastman:
Does LJPME produce answers that are more physically accurate (closer to experiment)?

To put it a different way: you care about the difference in free energy between the bound and unbound states.  If you compute the difference in energy (not free energy) for any one fixed conformation, that will scale with the size of the system, because every water molecule adds a tiny bit more error.  But the binding energy is determined mainly by local interactions, so if you hold the ligand and binding pocket fixed and average over all other degrees of freedom, the resulting free energy difference will probably *not* scale with the size of the system.

Do you actually get a different answer for the binding free energy with and without PME?

peastman:
Or even just with two different values for the cutoff, say 1 and 2 nm.  How much effect does it have on the final answer you're trying to compute?  (My guess is very little, but I'd have to see the data to know for sure.)

jchodera:
> Do you actually get a different answer for the binding free energy with and without PME?

Are you asking about PME for electrostatics or PME for dispersion?

> To put it a different way: you care about the difference in free energy between the bound and unbound states. If you compute the difference in energy (not free energy) for any one fixed conformation, that will scale with the size of the system, because every water molecule adds a tiny bit more error. But the binding energy is determined mainly by local interactions, so if you hold the ligand and binding pocket fixed and average over all other degrees of freedom, the resulting free energy difference will probably not scale with the size of the system.

Neglecting the dispersion correction entirely leads to horrifically large systematic errors in the free energy of binding that can be more than 10 kcal/mol. Using a dispersion correction with fixed 9 or 10A cutoff causes errors that are up to several kcal/mol. [[PDF]](http://www.choderalab.org/s/accurate-and-efficient-corrections-for-missing-dispersion-interactions-in-molecular-simulations.pdf).

We could run the simulation with a very large cutoff, but we're hoping to save time by not having to do that. Our simple trick (from that paper) of reweighting just the endpoints seems to break down for large systems like kinases.

peastman:
> Are you asking about PME for electrostatics or PME for dispersion?

Dispersion, of course.

peastman:
Note that there's an error in the test script.  The line

```Python
if force.__class__.__name__ == 'NonbondedCutoff':
```

should be

```Python
if force.__class__.__name__ == 'NonbondedForce':
```

That means the charges aren't getting zeroed.  Fixing it doesn't make a huge difference, but the deviations do get a little smaller:

```
   2.000 nm : mean error:     -38.20669066 kT | error stddev:       7.74905955 kT
   1.950 nm : mean error:     -38.10388001 kT | error stddev:       7.68378140 kT
   1.900 nm : mean error:     -37.97610240 kT | error stddev:       7.60562953 kT
   1.850 nm : mean error:     -37.83166809 kT | error stddev:       7.52182543 kT
   1.800 nm : mean error:     -37.65064430 kT | error stddev:       7.43318463 kT
   1.750 nm : mean error:     -37.43795621 kT | error stddev:       7.33129192 kT
   1.700 nm : mean error:     -37.18579189 kT | error stddev:       7.20859095 kT
   1.650 nm : mean error:     -36.87631394 kT | error stddev:       7.07528788 kT
   1.600 nm : mean error:     -36.49802234 kT | error stddev:       6.92123292 kT
   1.550 nm : mean error:     -36.03391315 kT | error stddev:       6.74678702 kT
   1.500 nm : mean error:     -35.48173727 kT | error stddev:       6.54765754 kT
   1.450 nm : mean error:     -34.79212800 kT | error stddev:       6.33300861 kT
   1.400 nm : mean error:     -33.95547456 kT | error stddev:       6.08175902 kT
   1.350 nm : mean error:     -32.97039438 kT | error stddev:       5.77984054 kT
   1.300 nm : mean error:     -31.76293778 kT | error stddev:       5.45425596 kT
   1.250 nm : mean error:     -30.33067156 kT | error stddev:       5.06348527 kT
   1.200 nm : mean error:     -28.65654094 kT | error stddev:       4.63142784 kT
   1.150 nm : mean error:     -26.41253353 kT | error stddev:       4.19105857 kT
   1.100 nm : mean error:     -23.51601542 kT | error stddev:       3.59210456 kT
   1.050 nm : mean error:     -19.86033885 kT | error stddev:       2.91940562 kT
   1.000 nm : mean error:     -15.01634984 kT | error stddev:       2.21180296 kT
   0.950 nm : mean error:      -8.47904685 kT | error stddev:       1.27554602 kT
   0.900 nm : mean error:       0.00000116 kT | error stddev:       0.00014247 kT
   0.850 nm : mean error:       9.05376193 kT | error stddev:       1.96239950 kT
   0.800 nm : mean error:      17.61077654 kT | error stddev:       3.94892119 kT
   0.750 nm : mean error:      30.66675850 kT | error stddev:       6.88258186 kT
```

I'm testing now to see what happens with LJPME.

peastman:
Using LJPME with a reasonably fine grid reduces the deviations further (but curiously makes the constant shift larger):

```
   2.000 nm : mean error:     -67.14549392 kT | error stddev:       2.03966550 kT
   1.900 nm : mean error:     -65.83468279 kT | error stddev:       1.97456446 kT
   1.800 nm : mean error:     -64.21186727 kT | error stddev:       1.91958214 kT
   1.700 nm : mean error:     -62.19010282 kT | error stddev:       1.83368451 kT
   1.600 nm : mean error:     -59.67444820 kT | error stddev:       1.72297364 kT
   1.500 nm : mean error:     -56.26485712 kT | error stddev:       1.63994134 kT
   1.400 nm : mean error:     -52.04550462 kT | error stddev:       1.49124203 kT
   1.300 nm : mean error:     -46.79519183 kT | error stddev:       1.35014624 kT
   1.200 nm : mean error:     -39.53406769 kT | error stddev:       1.12226305 kT
   1.100 nm : mean error:     -28.97901853 kT | error stddev:       1.17005596 kT
   1.000 nm : mean error:     -16.81528817 kT | error stddev:       1.70744947 kT
   0.900 nm : mean error:       0.00025120 kT | error stddev:       0.00121286 kT
   0.800 nm : mean error:      10.24223090 kT | error stddev:       3.36503273 kT
```

So I'm kind of puzzled about how I should interpret the error tolerance.  For Coulomb PME, we define it to be the average relative error in the forces.  But for LJPME, the errors in forces are tiny.  You can use a very coarse grid and still satisfy that requirement.  The errors in the energy are larger, but mostly it's just a constant shift independent of conformation.  Given a particular value for the error tolerance, what quantity should I try to keep below that tolerance?

peastman:
It looks like most of the error in the numbers above comes from the direct space part.  Further increasing the grid size doesn't help them.  But if I decrease the error tolerance to 1e-4 (which changes both the direct and reciprocal space parameters) they improve further:

```
   2.000 nm : mean error:     -25.24876865 kT | error stddev:       1.38985627 kT
   1.900 nm : mean error:     -24.77736378 kT | error stddev:       1.36242995 kT
   1.800 nm : mean error:     -24.15777702 kT | error stddev:       1.32727317 kT
   1.700 nm : mean error:     -23.39842569 kT | error stddev:       1.28597914 kT
   1.600 nm : mean error:     -22.43407407 kT | error stddev:       1.23350664 kT
   1.500 nm : mean error:     -21.18444081 kT | error stddev:       1.16004600 kT
   1.400 nm : mean error:     -19.52649094 kT | error stddev:       1.07326020 kT
   1.300 nm : mean error:     -17.41514203 kT | error stddev:       0.96261154 kT
   1.200 nm : mean error:     -14.54999730 kT | error stddev:       0.80400351 kT
   1.100 nm : mean error:     -10.77290755 kT | error stddev:       0.60319928 kT
   1.000 nm : mean error:      -5.80233580 kT | error stddev:       0.34381687 kT
   0.900 nm : mean error:       0.00000658 kT | error stddev:       0.00093045 kT
   0.800 nm : mean error:       3.45384163 kT | error stddev:       0.52473969 kT
```

jchodera:
> It looks like most of the error in the numbers above comes from the direct space part. Further increasing the grid size doesn't help them. But if I decrease the error tolerance to 1e-4 (which changes both the direct and reciprocal space parameters) they improve further:

This is fantastic!

peastman:
I just sent a PR revising the grid size calculation based on these tests.  @andysim what else still needs to be done?  There's the OpenCL platform of course, which I'm happy to do.  But I want to make sure everything on the CUDA platform is done before I start porting it.

peastman:
@andysim should I start on the OpenCL platform, or are there still other things that need to be done first?

andysim:
Sorry guys, my GH notifications got turned off somehow.  I *think* they're active again now.  The free energy discussion was really fascinating, and wasn't something I'd thought about as a test.  I'm surprised the forces were so insensitive to setup parameters, but surface tensions are another test that springs to mind (although it seems you already settled on a good solution).

There are a few things that need to be done for CUDA that I can think of:-

- Add hooks for CPUPME code (the LJPME itself is already in the CPUPME plugin).  This can't be tested directly from the test cases as far as I know.  This is an easy thing to do, but [it looks like I didn't finish it](https://github.com/andysim/openmm/blob/dpme/platforms/cuda/src/CudaKernels.cpp#L2322-L2334).
- Add a test for the new `CudaFFT3D` for LJPME, in case users don't have CuFFT available?
- Add the code to remove rec space contributions from 1-4 pairs, in the locations that you pointed out before.  This is a simple cut 'n' paste exercise, but I'm trying to get a decent test together.  My lack of familiarity with running Gromacs (and OpenMM, for that matter) has been a problem but I'll try to get one going today.

Apart from that, I think it's in great shape.  Thanks a lot for all your hard work to fix my janky first attempt, @peastman.

peastman:
Thanks!  I actually went ahead and did the OpenCL port.  I'll send a PR with that later today.

> Add hooks for CPUPME code (the LJPME itself is already in the CPUPME plugin).

I can do that.  I'll include it in the same PR with the OpenCL platform.

> Add a test for the new CudaFFT3D for LJPME, in case users don't have CuFFT available?

I already verified it works, but in practice it generally isn't used.  That code exists because CUDA 7.0 had a critical bug in the cufft library.  On more recent CUDA versions, we always use cufft.  But I'm keeping the code around, just in case we ever need it again.

> Add the code to remove rec space contributions from 1-4 pairs

Can you do that?  It sounds like you know exactly what needs to be done.

I think we're getting really close to having this ready to merge!

peastman:
Now that I think about it, I'm wondering whether we really want to use the CPU PME code here.  The point of that is to offload some work to the CPU, so it can be doing something useful at the same time as the GPU.  In practice, that's only useful on low end GPUs.  For high end ones, it's much faster to just let the GPU do everything.  So do we really want to offload twice as much work, which would make it even less likely to be useful?  Or would it be better to do only the Coulomb PME on the CPU, leaving the LJ PME on the GPU?

andysim:
I'll do the 1-4 stuff.  I have very little experience with benchmarking on different GPUs, so I will defer to you expertise to decide how the LJPME rec space should be handled.  I do agree with what you're saying though.  If it's a case of communicating the grid back to the CPU, evaluating, and communicating forces back to GPU that'd be particularly costly, compared to the cheap FFTs on a sleazy grid.  The cpupme code exists already for the CPU platform to use, but it's probably a good idea to skip it for CUDA.

peastman:
I think I'll do whatever looks easiest. :)  I'm probably worrying about an obscure case that will almost never come up.  The CUDA with CPU PME option isn't used very often.  And doing LJ PME on a low end GPU is also likely to be pretty rare.  So as long as it produces correct results, I won't worry about whether it's as fast as it could be.

andysim:
I agree with that.  More readable code that runs optimally on the best hardware available seems like a very good design goal.

peastman:
Ok, I took the incredibly lazy approach and just disabled the CPU PME option when using LJPME.  PR is submitted.

andysim:
Alrighty, I finally figured out how to get things set up in both GROMACS and OpenMM with replacement 1-4 interactions, and we get exact agreement between the two.  The good news is that no new code was needed.  Because we flag "special" 1-4 interactions as an exclusion, the reciprocal space term is already removed from them.  The replacement 1-4 LJ term is then added in and, because the 1-4 terms should always be within the cutoff, we don't need to worry about adding a potential shift term.  Easy :)  I added a test case, just to show that the code is correct w.r.t. GROMACS.  I noticed that GROMACS has some scale factors that can be universally applied to 1-4 interactions; does OpenMM have a similar feature?  If so I'll have to check that code path.  If scaled 1-4 are implemented by the same exclusion machinery as the special 1-4s then we're all good.  There were some merge conflicts with the master branch that I resolved, so the latest should be up to date now.

andysim:
I'm not sure what's going on with the one bad OpenCL test case.  If I run it on our cluster (in single precision), the offending force component is `[-0.00117642, 0.00118375, 0.000535846]`, which passes the test.

peastman:
> I noticed that GROMACS has some scale factors that can be universally applied to 1-4 interactions; does OpenMM have a similar feature?

The most common way of adding exceptions is by calling `createExceptionsFromBonds()`, which lets you specify scale factors.  But once the exceptions are created, it doesn't matter how you created them.  Everything gets handled the same way.

The failing test case may just be looking for unrealistic accuracy.  It's failing on atom 3, which has much smaller forces (by about three orders of magnitude) than any other atom.  If that's caused by a delicate cancellation of forces, it's going to be hard to reproduce that accurately in single precision mode.

peastman:
To fix the failure on appveyor, add this at the top of TestDispersionPME.h (just before the very first `#include`):

```C++
#ifdef _MSC_VER
    // Prevent Windows from defining macros that interfere with other code.
    #define NOMINMAX
#endif
```

andysim:
Ok, if all code paths are the same then I think the 1-4 code is all good to go.  What confused me about the OpenCL failure is that I can get it to pass in single precision just fine on a few of our systems.  Weird.  I guess I'll just dial back the precision required for the single precisions tests?  I had no clue on the appveyor problem, so thanks for that.  I'll make those switches now and push a patch.

peastman:
The OpenCL platform has a few different code paths depending on what sort of device it's running on: whether it's a CPU or GPU, and whether the device supports 64 bit atomic operations.  So accuracy can vary between devices even in the same precision mode.

Rather than turning down the accuracy, I wonder if it would be better to just skip that atom?  The problem is just that the force on that one is so much smaller than any other atom.

andysim:
Excellent idea - I'll do that

peastman:
Appveyor is still failing even with that change.  I think maybe you need to change `min` to `std::min`.

andysim:
I'll try that.  The timeout for mixed precision OpenCL is a little baffling.  This is what I get for a system running 2.3 GHz Xeon E5-2630.
```
tesla3:~/programming/openmm/build (dpme) > time ./TestOpenCLDispersionPME mixed
Done

real    0m30.927s
user    0m7.807s
sys     0m13.822s
```
Maybe we need to allow more than 100s for that test, to account for older hardware?

peastman:
You can ignore the timeout.  Travis doesn't give us enough time to get through all the OpenCL tests.  So we run them in a random order, let it get through as many as it can, and trust that each test gets run fairly often so problems will be caught quickly, even if not on every commit.

andysim:
But isn't the time limit 50 minutes?  It looks like it failed after just 27 minutes.

peastman:
The whole process gets 50 minutes, but that includes compiling, running python tests, etc..  We limit ctest to 30 minutes.

It looks like everything passing.  Just waiting for the OS X build, which can take a very long time to start.  Assuming that passes, is this all ready to merge?

andysim:
Yup, this one's finally good to go; the appveyor problem was just a missing `#include <algorithm>`.  Thanks a lot for all your help in getting this PR finalized, and sorry it took so long.

peastman:
Everything passed, which means this is now ready to merge!  Thanks so much for implementing this.  I think it will be a great feature!

jchodera:
Woohoo!

PR #1560, opened 2016-08-14
New code for impropers assignments

rafwiewiora:
@peastman @jchodera 

This is continuing from the discussion here: https://github.com/choderalab/openmm/issues/16

We decided:
- have default code for any FF, have special code for Amber FF - TODO: decide how to pass the `isAmber` info in the ffxml's
- for default code - if no wildcards or 1 wildcard - assigned in the order as loaded in ffxml, if any same types in the improper - sort by residue index and template atom index if same residue (in ascending order). 2 wildcards - additionally sort the first and second atom in the improper (i.e. the types that came through wildcards) on residue index and template atom index.
- for Amber code - here divide between no wildcard vs. has a wildcard - no wildcard same as above; has a wildcard - we sort on residue and template atom index if elements, rather than atom types, are the same; and then sort the first and second atoms of the improper on residue and template atom index no matter what their elements are. 

With this code, the Amber converted FFs have majority of improper energies at < 1e-5 relative error, with only 5 cases above 1e-5 but below the 'more tolerant' threshold of 1e-2. We also generate a list of all impropers that will be mismatched by the code (in the newer forcefields that's only one improper in RNA in my protein, DNA and RNA test systems; for all forcefields converted only 15 improper types were mismatched for protein, DNA and RNA over the whole conversion). 

TODO:
- decide how to include the `isAmber` (where to put the tag?) info in ffxml's


peastman:
Thanks!

> decide how to include the isAmber (where to put the tag?) info in ffxml's

How about a `useAmberOrdering` attribute on the `<PeriodicTorsionForce>` tag?

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1875,24 +1877,59 @@ def createForce(self, sys, data, nonbondedMethod, nonbondedCutoff, args):
...
+                            # the following is OpenMM default
+                            else:
+                                    if t2[0] == t4[0] and (r1 > r4 or (r1 == r4 and ta1 > ta4)):
+                                        (a1, a4) = (a4, a1)
+                                    if t3[0] == t4[0] and (r2 > r4 or (r2 == r4 and ta2 > ta4)):
+                                        (a2, a4) = (a4, a2)
+                                    if t2[0] == t3[0] and (r1 > r2 or (r1 == r2 and ta1 > ta2)):
+                                        (a1, a2) = (a2, a1)
+                                    if hasTwoWildcards and (r1 > r2 or (r1 == r2 and ta1 > ta2)):
+                                        (a1, a2) = (a2, a1)
```
peastman: Are you sure this is correct?  It's possible for both of the last two if's to be executed, in which case line 1931 will undo the swap that was done in line 1929.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -1875,24 +1877,59 @@ def createForce(self, sys, data, nonbondedMethod, nonbondedCutoff, args):
...
+                            e4 = data.atoms[a4].element
+                            # the following for AMBER only - TODO: decide how to pass this in ffxml's
+                            if isAmber:
+                                if not hasWildcard:
+                                    if t2[0] == t4[0] and (r1 > r4 or (r1 == r4 and ta1 > ta4)):
+                                        (a1, a4) = (a4, a1)
+                                    if t3[0] == t4[0] and (r2 > r4 or (r2 == r4 and ta2 > ta4)):
+                                        (a2, a4) = (a4, a2)
+                                    if t2[0] == t3[0] and (r1 > r2 or (r1 == r2 and ta1 > ta2)):
+                                        (a1, a2) = (a2, a1)
```
peastman: This code doesn't make sense to me.  Are some of the indices wrong?  For example, should line 1909 start out `if t1[0] == t4[0]`, not `if t2[0] == t4[0]`?

swails: This is looping over permutations, which suggests (to me, at least) that this logic is controlling improper assignment.  If that's the case, then `t2[0] == t4[0]` looks more correct to me (assuming that `t1` is the central "fixed" atom).

peastman: Yes, but then it checks `r1 > r4`.  The first comparison relates to atoms 2 and 4, but everything else in that line relates to atoms 1 and 4.

wrappers/python/simtk/openmm/app/forcefield.py
rafwiewiora: Yeah, these indices are confusing. Here's what's going on - we have two sets of indices; first comes from the order of atom types in the improper as found in the topology: `types1, types2, types3, types4`; second is the order which gets assigned as `match` - i.e. `types1` always goes as now 3rd atom, and the other 3 get assigned by all this logic. So at some point we need to do this 'translation' between the two orders:

lines 1891-3:
```python
a1 = torsion[t2[1]]
a2 = torsion[t3[1]]
a4 = torsion[t4[1]]
``` 

`a1`, `a2`, `a3` are then mixed around by the assignment logic, and `match` is defined:

line 1932
```python
match = (a1, a2, torsion[0], a4, tordef)
```

I could perhaps change the translation code to:
```python
(t1, t2) = (t2, t3)
a1 = torsion[t1[1]]
a2 = torsion[t2[1]]
a4 = torsion[t4[1]]
```

and change accordingly forward `t2` to `t1` and `t3` to `t2` to make this less confusing. What do you think?

wrappers/python/simtk/openmm/app/forcefield.py
rafwiewiora: Yeah, these indices are confusing. Here's what's going on - we have two sets of indices; first comes from the order of atom types in the improper as found in the topology: types1, types2, types3, types4; second is the order which gets assigned as match - i.e. types1 always goes as now 3rd atom, and the other 3 get assigned by all this logic. So at some point we need to do this 'translation' between the two orders:

lines 1891-3:

a1 = torsion[t2[1]]
a2 = torsion[t3[1]]
a4 = torsion[t4[1]]
a1, a2, a3 are then mixed around by the assignment logic, and match is defined:

line 1932

match = (a1, a2, torsion[0], a4, tordef)
I could perhaps change the translation code to:

(t1, t2) = (t2, t3)
a1 = torsion[t1[1]]
a2 = torsion[t2[1]]
a4 = torsion[t4[1]]
and change accordingly forward t2 to t1 and t3 to t2 to make this less confusing. What do you think?

wrappers/python/simtk/openmm/app/forcefield.py
rafwiewiora: Great point! Going to change the last `if` to `elif`.

wrappers/python/simtk/openmm/app/forcefield.py
rafwiewiora: So replying to your comments above, as @swails mentions `t1`, becomes the central, third in the new order atom, and @peastman : atom 2 from the 'old' order is atom 1 in the 'new' order initially (before all the switching).

rafwiewiora:
`useAmberOrdering` sounds good to me!

wrappers/python/simtk/openmm/app/forcefield.py
peastman: > I could perhaps change the translation code to:

Or alternatively

```Python
a2 = torsion[t2[1]]
a3 = torsion[t3[1]]
a4 = torsion[t4[1]]
```

The thing that makes it confusing is mixing two different numbering schemes.  Sticking to one or the other would be clearer.  It doesn't matter much which one, as long as it's consistent.  But if you write something like

```Python
(t1, t2) = (t2, t3)
```

that's clearly not being consistent.  It's repurposing the same set of variables that were initially named based on one numbering scheme, then reinterpreting the names as a different numbering scheme.

rafwiewiora: Ok, I like your suggestion, going with that!

rafwiewiora:
Ok made all those corrections we discussed in the line comments. And also noticed that the `r`'s and the `ta`'s had to be updated after the `a` swappings, so fixed that! Not sure why my energy tests were still ok with that mistake...

A few questions that came to mind for the `useAmberOrdering`:

- should we make it an attribute of `PeriodicTorsionGenerator` ? If so, what happens if you pass two  files to `ForceField` one with Amber ordering, one without? Should Amber ordering then be applied to all impropers? Or should we make it a property of every individual improper?

- Within the ffxml, you suggested having it as an attribute of the `PeriodicTorsionForce` tag, so should it be `<PeriodicTorsionForce useAmberOrdering=True/>` - does the `True` matter? I.e. do we also allow this to be `False` ? Maybe we should have an attribute that could be something else too: `useOrdering="amber"` ? Or maybe make it the first element, like for the `<UseAttributeFromResidue name="charge"/>` in `NonbondedForce`, i.e.:

```python
<PeriodicTorsionForce>
<UseAmberOrdering/>

**Comment by peastman** [comment]
> what happens if you pass two files to ForceField one with Amber ordering, one without?

Good question.  We definitely should *not* change the interpretation of one tag based on what's specified in a different one.  A simple option would be to throw an exception if you have two tags with inconsistent ordering, the way we do if you have two `<NonbondedForce>` tags with inconsistent scales.  But since the ordering is applied to every torsion independently, there's no real need for that restriction.  So I'd vote for making it a property of each torsion.

> Maybe we should have an attribute that could be something else too: useOrdering="amber" ?

I like that.  Or even more simply, just `ordering="amber"`.  The other allowed value would be `ordering="default"`, which is what it assumes if you omit the attribute.

**Comment by rafwiewiora** [comment]
Ok, got it, thanks!

**Comment by rafwiewiora** [comment]
@peastman added the ordering attribute code, have a look and see what you think - just need to add tests to this to be ready

**Comment by jchodera** [comment]
@rafwiewiora: Some test failures are due to travis cache weirdness, but you have some actual bugs in there. See [here](https://travis-ci.org/pandegroup/openmm/jobs/154893775#L6328):
```
___________________________ TestSimulation.testStep ____________________________
self = <TestSimulation.TestSimulation testMethod=testStep>
    def testStep(self):
        """Test the step() method."""
        pdb = PDBFile('systems/alanine-dipeptide-implicit.pdb')
>       ff = ForceField('amber99sb.xml', 'tip3p.xml')
TestSimulation.py:104: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/simtk/openmm/app/forcefield.py:128: in __init__
    self.loadFile(files)
/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/simtk/openmm/app/forcefield.py:289: in loadFile
    parsers[child.tag](child, self)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
element = <Element 'PeriodicTorsionForce' at 0x7fd2a6f4a9f8>
ff = <simtk.openmm.app.forcefield.ForceField object at 0x7fd2a71db240>
    @staticmethod
    def parseElement(element, ff):
        existing = [f for f in ff._forces if isinstance(f, PeriodicTorsionGenerator)]
        if len(existing) == 0:
            generator = PeriodicTorsionGenerator(ff)
            ff.registerGenerator(generator)
        else:
            generator = existing[0]
        for torsion in element.findall('Proper'):
            generator.registerProperTorsion(torsion.attrib)
        for torsion in element.findall('Improper'):
>           generator.registerImproperTorsion(torsion.attrib, element.attrib['ordering'])
E           KeyError: 'ordering'
/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/simtk/openmm/app/forcefield.py:1854: KeyError
```

**Comment by rafwiewiora** [comment]
Ah good call, thanks John!

**Comment by rafwiewiora** [comment]
Corrected, this looks to be passing. 

Still need to add new tests.

**Comment by rafwiewiora** [comment]
@peastman this is ready to go in. 

However, I have one concern about what the new default improper assignments code should be. We decided to replace the current code with a new default, which is the 'Amber' assignment code minus a few weird things. I liked that when I thought that we were going to also replace the ffxmls already in OpenMM with the newly converted versions. Now, that we are not doing that I think that we should keep the existing impropers assignment code. I have seen during testing of this that the results from the old code and the new default code on e.g. `amber99sbildn.xml` are not same (in terms of order of atom assignments in some of the impropers) - so essentially we'll have a slightly different force field once we've changed this code than we had before. That screams bad for reproducibility. I wouldn't want to make this change for my current work - can I compare results from the different simulations in good faith?

**Comment by swails** [comment]
> That screams bad for reproducibility. I wouldn't want to make this change for my current work - can I compare results from the different simulations in good faith?

It depends what you mean by "compare" here.  If the question is whether ensemble properties are comparable, I would be willing to vouch for the assertion that improper ordering in this case will exert negligible effect on the dynamics.

I'll give you an example of this kind of comparison.  If you use OpenMM to parametrize ff14SB and GROMACS to parametrize ff99SB-ILDN, and AMBER to parametrize ff99SB on the exact same system, rank-ordering the qualities based on agreement with experimental measurements is this kind of comparison.  I don't think there's any problem with doing this kind of comparison, and indeed this is done with regularity (for example, any time a GROMACS simulation claims to use an Amber force field!)

On the other hand, for doing careful validation of force field implementation (or conversion, as is the case with ParmEd), I'm of the opinion that the comparisons should be done a bit more strictly.  What I consider to be allowable differences are those arising from, for instance, default PME parameter assignments between different simulation engines, or very slightly different values for fundamental physical constants, or slightly different approximations for the long-range dispersion correction to name a few.  But certainly if you're converting topologies directly, improper ordering should most certainly be preserved (since it is easy to do).

**Comment by rafwiewiora** [comment]
Thanks for the examples! I didn't get though whether you're leaning then towards leaving the old code as default or introducing the new default? 

To avoid the problem we could add the `ordering='amber'` attribute to the Amber ffxmls that are already in OpenMM - this will also change their behavior, but at least it should make it better (should because I know it does with the ffxmls I converted, might for some reason behave a bit differently with the previously converted ffxmls, but probably not).

**Comment by peastman** [comment]
I'm not sure what you're talking about.  If you specify `order='amber'` you'll get the Amber ordering.  But if you specify `order='default'`, or just don't specify the order, shouldn't it behave the same way it did before?  I looked back through the discussion to try to figure out what you're referring to, but it wasn't clear to me.

**Comment by swails** [comment]
> I didn't get though whether you're leaning then towards leaving the old code as default or introducing the new default?

Leave it as-is (I presume you're talking about the existing Amber XML FFs).  The fact that it gives the "wrong" ordering (if we define tleap as "right" for the Amber force field) will result in a forcefield that is macroscopically correct (as in it gives correct thermodynamic predictions), but microscopically different (in that single-point energies and forces will differ *slightly*).

At this point, you have to ask yourself where do those microscopic differences matter the most.  If they don't matter when you calculate thermodynamics, then they'll only matter when you start comparing single-point energies.  In particular, it's reasonable to expect that different versions of OpenMM give the exact same behavior, so that you can do calculations that may actually be sensitive to these microscopic differences reliably across different versions (whatever those calculations may be).

**Comment by rafwiewiora** [comment]
Made the change as we discussed - the default is the same code we've had, new code added only for new Amber FFs, which will use the `ordering=amber` attribute on the impropers. 

Good to go in when tests complete.

**Comment by rafwiewiora** [comment]
@peastman this is good to merge

**Comment by peastman** [comment]
The code for ordering impropers appears in three classes: PeriodicTorsionForce, RBTorsionForce, and CustomTorsionForce.  You only changed it in PeriodicTorsionForce.  The other two need to be changed too.  Since it's a lot more code than it was before, though, let's extract it out into a function that can be called from all three.

Rather than embedding an entire PDB file in the Python code for the test case, you can put the file in wrappers/python/tests/systems.  That's what all the other tests do.

**Comment by rafwiewiora** [comment]
> The code for ordering impropers appears in three classes: PeriodicTorsionForce, RBTorsionForce, and CustomTorsionForce. You only changed it in PeriodicTorsionForce. The other two need to be changed too. Since it's a lot more code than it was before, though, let's extract it out into a function that can be called from all three

Why do the other two need the need code though? This is specifically a correction I've tested for PeriodicTorsionForce energies to match between `leap` and `openmm`. I don't know whether the other two need an `amber` option and whether it should be this code.

>Rather than embedding an entire PDB file in the Python code for the test case, you can put the file in wrappers/python/tests/systems. That's what all the other tests do.

Changed.

**Comment by peastman** [comment]
To give just one example, someone might want to reproduce an Amber force field using a CustomTorsionForce in place of the PeriodicTorsionForce, but multiplying it by a lambda parameter so they can turn it on and off (for lambda dynamics, TI, alchemical changes, etc.).

**Comment by rafwiewiora** [comment]
Ok!

**Comment by rafwiewiora** [comment]
@peastman changes you requested ready, see if good to merge now?

I've added the new code to RBTorsion and CustomTorsion, but unfortunately extracting it out into a function is not possible - the existing code differs in its treatment of no-wildcard cases between PeriodicTorsion, which will reshuffle the order based on the coded rules, even if there was no wildcards:

```python
                if type1 in types1:
                    for (t2, t3, t4) in itertools.permutations(((type2, 1), (type3, 2), (type4, 3))):
                        if t2[0] in types2 and t3[0] in types3 and t4[0] in types4:
                            # Workaround to be more consistent with AMBER.  It uses wildcards to define most of its
                            # impropers, which leaves the ordering ambiguous.  It then follows some bizarre rules
                            # to pick the order.
                            a1 = torsion[t2[1]]
                            a2 = torsion[t3[1]]
                            e1 = data.atoms[a1].element
                            e2 = data.atoms[a2].element
                            if e1 == e2 and a1 > a2:
                                (a1, a2) = (a2, a1)
                            elif e1 != elem.carbon and (e2 == elem.carbon or e1.mass < e2.mass):
                                (a1, a2) = (a2, a1)
                            match = (a1, a2, torsion[0], torsion[t4[1]], tordef)
                            break

while RBTorsion and CustomTorsion do not do anything if there was no wildcards:

```python
                if type1 in types1:
                    for (t2, t3, t4) in itertools.permutations(((type2, 1), (type3, 2), (type4, 3))):
                        if t2[0] in types2 and t3[0] in types3 and t4[0] in types4:
                            if hasWildcard:
                                # Workaround to be more consistent with AMBER.  It uses wildcards to define most of its
                                # impropers, which leaves the ordering ambiguous.  It then follows some bizarre rules
                                # to pick the order.
                                a1 = torsion[t2[1]]
                                a2 = torsion[t3[1]]
                                e1 = data.atoms[a1].element
                                e2 = data.atoms[a2].element
                                if e1 == e2 and a1 > a2:
                                    (a1, a2) = (a2, a1)
                                elif e1 != elem.carbon and (e2 == elem.carbon or e1.mass < e2.mass):
                                    (a1, a2) = (a2, a1)
                                match = (a1, a2, torsion[0], torsion[t4[1]], tordef)
                            else:
                                # There are no wildcards, so the order is unambiguous.
                                match = (torsion[0], torsion[t2[1]], torsion[t3[1]], torsion[t4[1]], tordef)
                            break
```

Unless you want to just extract the new `ordering='amber'` code into a function?

peastman:
I'm inclined to make that same change in RBTorsionForce and CustomTorsionForce too.  If the order is unambiguous it really should use it, and I think that case only got left out by accident.  In theory this could change existing results, but only in ways that I think make them more correct.  Also, none of the standard existing force fields use RBTorsionForce, and the only one that uses CustomTorsionForce is the CHARMM polarizable force field.

rafwiewiora:
> I'm inclined to make that same change in RBTorsionForce and CustomTorsionForce too. If the order is unambiguous it really should use it, and I think that case only got left out by accident. In theory this could change existing results, but only in ways that I think make them more correct

I'm a bit confused here. The more correct ones now are RBTorsionForce and CustomTorsionForce (i.e. in the case of unambiguous order they don't change it), while the less correct one is PeriodicTorsionForce (i.e in the case of unambiguous order it will still change it). So the change would be in PeriodicTorsionForce and will affect interpretation of all existing FFs. This is what you want to do?

peastman:
Sorry, I misunderstood.  Yeah, changing the behavior PeriodicTorsionForce wouldn't be a good idea, since that would affect almost all the current force fields.  Let's still extract out the `ordering='amber'` code though.  That's about 50 lines that's currently duplicated in three in different places.

rafwiewiora:
Ok, on it! 

I would also like to point out here for @ChayaSt @jchodera  - we are leaving the improper assignments code as it was, BUT it will re-shuffle your improper order even if there were no wildcards. It seems to me you will want to introduce an `ordering='charmm'` code that does not do any reshuffling. This is exactly what it will need to be:

```python
if type1 in types1:
    for (t2, t3, t4) in itertools.permutations(((type2, 1), (type3, 2), (type4, 3))):
        if t2[0] in types2 and t3[0] in types3 and t4[0] in types4:
            if tordef.ordering = 'charmm':
                # There are no wildcards, so the order is unambiguous.
                match = (torsion[0], torsion[t2[1]], torsion[t3[1]], torsion[t4[1]], tordef)
                break
```

peastman:
Ack.  That's a good point.  In all the discussions of CHARMM impropers I've been assuming it would preserve ordering, because that's how I thought it worked.  But it looks like that somehow missed getting implemented in PeriodicTorsionForce, only RBTorsionForce and CustomTorsionForce.

So really, ick.

Let me run a few tests to see how it would affect the existing Amber force fields if we change that behavior.

peastman:
Actually, that's not quite it either.  That *did* get implemented for PeriodicTorsionForce at the same time as the others: https://github.com/pandegroup/openmm/commit/774a257e38e71d041d672d7e768b4e80aac22e46.  I'm now trying to track down when it got removed again.

rafwiewiora:
>Let me run a few tests to see how it would affect the existing Amber force fields if we change that behavior.

Quick comment to this - when I was writing the test for this PR, I discovered that the improper ordering that comes out of the current code depends on the topology, i.e look at the PDB I have in the test here - the 'problematic' improper I'm testing for is in residue 4 (GLU), but if you only load in the res3-res4 fragment the 'problematic' improper gets assigned the same way in the old and the new code, it is only if you load in the whole 4-AA fragment that you get a wrong behavior from the old code and a right from the new code. I didn't quite understand how this was happening, but just a warning that there might be more factors influencing your tests than you think.

peastman:
Right.  Here's the commit where I made the change: https://github.com/pandegroup/openmm/commit/9128e9efe15e01b723ff0c06fa5e41d7dfadb752.  Notice the commit description.

And here's my comment describing why I made that change: https://github.com/pandegroup/openmm/pull/531#issuecomment-49073661.

Given all that, adding an `ordering='charmm'` option is probably the best solution.  At least we can then fully unify the handling of this between all three classes.  The only difference will be that if you omit the `ordering` attribute, PeriodicTorsionForce will default to `'default'` but RBTorsionForce and CustomTorsionForce will default to `'charmm'`.

rafwiewiora:
>This is a really ugly hack. It makes me sad when I have to do things like this.

Yeah...

Alright - so we're saying extract everything to a function with `default`, `amber` and `charmm` ordering options, and as you said: "The only difference will be that if you omit the ordering attribute, PeriodicTorsionForce will default to 'default' but RBTorsionForce and CustomTorsionForce will default to 'charmm'.", right? Ok I'm adding that in a sec.

peastman:
Thanks!  Yeah, ugly hacks indeed.  Even when they're necessary, you still usually come to regret them.

rafwiewiora:
@peastman this is all done as we discussed and passes tests

peastman:
Great!

PR #1561, opened 2016-08-15
Empty interaction group should produce no interactions

### LINKED ISSUE #1554: CustomNonbondedForce.addInteractionGroup has wrong behavior when empty list is used
For `CustomNonbondedForce`, the `addInteractionGroup` behavior is described thusly:
> By default, it computes the interaction of every particle in the System with every other particle. Alternatively, you can restrict it to only a subset of particle pairs. To do this, specify one or more “interaction groups”. An interaction group consists of two sets of particles that should interact with each other. Every particle in the first set interacts with every particle in the second set. For example, you might use this feature to compute a solute-solvent interaction energy, while omitting all interactions between two solute atoms or two solvent atoms.

By this logic, one might expect that adding interaction groups like `[]` and `[0,1,2,3]` would lead to no interactions. Instead, adding an interaction group specification where one group is empty does nothing to the energy:
```python
from simtk import unit, openmm
# Create a box of CustomNonbondedForce particles
testsystem = testsystems.WCAFluid()
[system, positions] = [testsystem.system, testsystem.positions]
# Compute potential energy
integrator = openmm.VerletIntegrator(1.0 * unit.femtoseconds)
context = openmm.Context(testsystem.system, integrator)
print(context.getState(getEnergy=True).getPotentialEnergy())
del context, integrator
# Now exclude all particles by putting them in separate groups
    if hasattr(force, 'addInteractionGroup'):
        print('Adding interaction group with one empty group')
        force.addInteractionGroup([], range(system.getNumParticles()))
integrator = openmm.VerletIntegrator(1.0 * unit.femtoseconds)
context = openmm.Context(testsystem.system, integrator)
print(context.getState(getEnergy=True).getPotentialEnergy())
del context, integrator
```
yields
```
[LSKI1497:choderalab/alchemy/alchemy.jchodera] choderaj% python interaction-groups-bug.py
322.710700257 kJ/mol
Adding interaction group with one empty group
322.710700257 kJ/mol
```

peastman:
Fixes #1554.

### DIFF
--- a/platforms/cpu/include/CpuCustomNonbondedForce.h
+++ b/platforms/cpu/include/CpuCustomNonbondedForce.h
@@ -129,6 +129,7 @@ private:
     bool useSwitch;
     bool periodic;
     bool triclinic;
+    bool useInteractionGroups;
     const CpuNeighborList* neighborList;
     float recipBoxSize[3];
     RealVec periodicBoxVectors[3];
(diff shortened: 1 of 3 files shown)


jchodera:
Thanks!

PR #1570, opened 2016-08-25
Amber FF conversion - AmberTools 16

rafwiewiora:


rafwiewiora:
@peastman @jchodera 

Nearly done here! All here in the folder: https://github.com/choderalab/openmm/tree/forcefield_conversion_bonanza_amber16/devtools/forcefield-scripts/amber

Description in the README.MD

Remaining TODO:
* need to clarify which ions we should be using for the standard (default) sets - in the previous conversion I used the JC +1/-1 ions, but Amber 16 manual recommends the Li/Merz HFE set
* the ipq forcefields still have some bug to sort out - 14ipq had a problem with a torsion in Amber Tools 15, which meant we couldn't convert it (not reproducible Leap vs OpenMM) - that bug might still be there, but the ff15ipq is also coming up with a torsion problem I need to debug
* phosphoaa10 has a new bug - was working on in Amber Tools 15, here not
* add separate DNA and RNA testing arguments to the command line (only have DNA and RNA as 'nucleic' now - Amber Tools 16 started separating the two).

peastman:
Could you explain what all the files in here are?  It looks like most of it consists of converted force fields, which for some reason are in  devtools/forcefield-scripts, and PDB files, which I'm not sure what they're doing there.

rafwiewiora:
> some reason

We have not decided how to organize these files. First, as I said above:

>  One thing that comes to mind now we should decide is the naming of the ffxml's - right now you do amberXXX.xml, I did here for now ffXXX.xml as the leaprc's are called in Amber.

Second how to you want to organize the directories - in Amber they have the main directory, and the `oldff` directory - do you want to create an `oldff` directory here? do you want to put the old FFs close to the main forcefield directory anyway or keep them in devtools? Where do you want to put the script? 

> and PDB files, which I'm not sure what they're doing there.

They are the systems used in the script for the LeAP vs OpenMM energy validations. Have you had time to look through the script code yet?

peastman:
The first question we need to decide is which of these files we actually want.  A large part of what's included here is either obsolete force fields that we definitely don't want to add (like ff94), or reconversions of force fields that we already have (like all the ff99 variants).  So we need to narrow it down to only the force fields we actually want.

When we talked about this before, we discussed having a small number of "top level" files that automatically include a common set of files.  So a user can just specify `'amber14.xml'` and automatically get a consistent set of protein, DNA, and RNA parameters.

Now that they're including the ions in the same file with the water parameters, we need to decide how to handle this for other water models like TIP3P-FB.  @leeping what are your thoughts on that?

Regardless of how we organize them, all force fields need to be in the data directory, not in devtools.  They don't do users any good there.

rafwiewiora:
> A large part of what's included here is either obsolete force fields that we definitely don't want to add (like ff94)
I thought we had decided to make the old forcefields available somehow (whether in a distribution or downloadable somewhere?) for reproducibility of older works etc. - tagging @jchodera on this point as it was his suggestion

> or reconversions of force fields that we already have (like all the ff99 variants)
However the files you already have are of the 'old format' - no provenance info, the thousands of atom types etc. Wasn't one of the points of this work to have cleanly, in 'new format' converted all FFs? 

> When we talked about this before, we discussed having a small number of "top level" files that automatically include a common set of files. So a user can just specify 'amber14.xml' and automatically get a consistent set of protein, DNA, and RNA parameters.

And that's what we have. Just note that Amber 16 has decided to separate protein, DNA and RNA (contrary to earlier Ambers where those were together) - no we have a larger number of only protein forcefields (with `protein` in the leaprc name, and hence also in my ffxmls), and only a couple for DNA and RNA. The combined forcefields have been made obsolete by Amber (e.g. `ff14SB` is now obsolete and placed in `oldff` and has been replaced by `protein.ff14SB`), recommended DNA and RNA forcefield, should you want to combine those with protein, are `DNA.OL15` and `RNA.OL3` - it is not immediately clear to me if this results in exactly the same parameters for DNA and RNA as having loaded `ff14SB` in Amber Tools 15 would have given. Do we want to go with this separation that Amber 16 employs or recombine for our files? 

> Now that they're including the ions in the same file with the water parameters, we need to decide how to handle this for other water models like TIP3P-FB. @leeping what are your thoughts on that?

The water parameters are not converted by my script from Amber files, rather there are ffxml's from OpenMM 7.0 in the `files` that get combined with converted ion params by the script. No problem adding another model - just want to make sure it is compatible with particular ion parameters - just need to add the xml to `files` and add another entry to the `master.yaml`. 

> Regardless of how we organize them, all force fields need to be in the data directory, not in devtools. They don't do users any good there.

This PR is just to show you what is going on and for further discussion. Not quite sure what the problem is, obviously they need to go to the data directory. The old conversion scripts were here so I temporarily put the stuff here too. Let's decide what we're doing further and organize all that in a new PR?

jchodera:
> The first question we need to decide is which of these files we actually want. A large part of what's included here is either obsolete force fields that we definitely don't want to add (like ff94), or reconversions of force fields that we already have (like all the ff99 variants). So we need to narrow it down to only the force fields we actually want.

We have three classes of forcefields we want to include:
* legacy versions of the AMBER forcefields provided with earlier versions of OpenMM, which should still work but are not recommended (since more recent conversions have improvements)
* "recommended" versions of the AMBER forcefields that are easy to import in a single command
* deprecated versions of AMBER forcefields provided for compatibility and reproducibility, which require the user preface the files with `oldff` or something to access them

> When we talked about this before, we discussed having a small number of "top level" files that automatically include a common set of files. So a user can just specify 'amber14.xml' and automatically get a consistent set of protein, DNA, and RNA parameters.

Correct. We wanted to make the new, recommended AMBER forcefields asy to use.

> Now that they're including the ions in the same file with the water parameters, we need to decide how to handle this for other water models like TIP3P-FB. @leeping what are your thoughts on that?

I think the best approach is to include both a "bundled" version of the XML file and the "unbundled" versions for all of these. "bundled" versions would be easy to import at the top level, like `amber14-tip3p-ions.xml`, while "unbundled" versions would be segregated in an `amber/` forcefield subdirectory and require the user preface them with that directory name.

> Regardless of how we organize them, all force fields need to be in the data directory, not in devtools. They don't do users any good there.

Agreed---we should have the forcefield output to `openmm/wrappers/python/simtk/openmm/app/data` and `data/amber`.

rafwiewiora:
> Agreed---we should have the forcefield output to openmm/wrappers/python/simtk/openmm/app/data and data/amber.

So hold on - do we want the conversion script to output into those directories? My plan here was - the script outputs wherever you run it, I will take those ffxml's and copy them into the right directories for the PR. That is because the script was meant to be able to convert from any yaml for any need, not just for this particular 'mega conversion'.

peastman:
> (e.g. ff14SB is now obsolete and placed in oldff and has been replaced by protein.ff14SB)

In that case do we even want to include ff14?  Or just skip it and add only ff16?

> This PR is just to show you what is going on and for further discussion.

Ok, that's fine.  I thought this was actually supposed to get merged.  If this is just for discussion, and you'll send a clean PR once we figure out exactly what we want, you can ignore my comments above about that.

> I think the best approach is to include both a "bundled" version of the XML file and the "unbundled" versions for all of these.

Sounds good to me.

jchodera:
> In that case do we even want to include ff14? Or just skip it and add only ff16?

I think we should include all the forcefields we can reasonably support without a lot of technical debt (for enabling the reproduction of earlier studies) but make it harder for the user to access the deprecated or non-recommended ones. If @rafwiewiora's YAML conversion makes it easy to also include ff14, we should include it. If it's a pain or causes a big increase in technical debt, the argument for keeping it is much less strong.

> So hold on - do we want the conversion script to output into those directories? My plan here was - the script outputs wherever you run it, I will take those ffxml's and copy them into the right directories for the PR. That is because the script was meant to be able to convert from any yaml for any need, not just for this particular 'mega conversion'.

Why not just add an output directory to the YAML file, and have your conversion YAML files point to the `simtk/openmm/app/data` directory?

peastman:
> If it's a pain or causes a big increase in technical debt, the argument for keeping it is much less strong.

It's less a technical issue and more a usability one.  We have never added a force field that we didn't actively recommend to users.  That's why there's no "old forcefields" folder.  Because every one of them was, at the time we added it, a reasonable choice.  And once it's added, we can't move it, because that would break all existing scripts that use it.

So what you're proposing is that we add a force field that, at the time we add it, we already consider to be out of date and don't recommend to users.  And we put it in a special folder to say, "Yeah we're adding this, but you really shouldn't use it."  And then we somehow have to explain to users why ff14 is in the "old forcefields" folder by ff96 isn't.  It's going to create confusion.

What about only including ff16, but posting the ff14 files on a website somewhere for people who really want them and know what they're doing?

> Why not just add an output directory to the YAML file, and have your conversion YAML files point to the simtk/openmm/app/data directory?

It doesn't matter where the script outputs them to.  The contents of the data directory are manually curated.  The script can create whatever files it creates wherever it creates them, and then we'll decide which ones to move into the data directory.

rafwiewiora:
> Why not just add an output directory to the YAML file, and have your conversion YAML files point to the simtk/openmm/app/data directory?

Ok, good idea! 

> In that case do we even want to include ff14? Or just skip it and add only ff16?

> I think we should include all the forcefields we can reasonably support without a lot of technical debt (for enabling the reproduction of earlier studies) but make it harder for the user to access the deprecated or non-recommended ones. If @rafwiewiora's YAML conversion makes it easy to also include ff14, we should include it. If it's a pain or causes a big increase in technical debt, the argument for keeping it is much less strong.

Just a bit confused about using a 'ff16' name - you mean the Amber Tools 16 version of ff14SB, i.e. `protein.ff14SB` right? Well this is the same situation as with other deprecated / old FFs (i.e. all included in the `oldff` folder in Amber) - my YAML already converts all those, so no problem - just gotta decide how you want to include all these deprecated / old FFs ??

> I think the best approach is to include both a "bundled" version of the XML file and the "unbundled" versions for all of these.
> Sounds good to me.

What do we mean by "unbundled" exactly? Would those be the ions combined in the same way as in the "bundled" files (see the README for how this is done), but without the waters, or we want individual conversion per each ion frcmod file from Amber?

jchodera:
We've already discussed that precise issue at great length.

> We have never added a force field that we didn't actively recommend to users.

This is not true. We wouldn't have added OBC GBSA or AMBER parm96 if this were the case, as they were already deprecated by the time they were added.

The fundamental issue is that we need to be able to reproduce some of the >10,000 papers using these forcefields in order to do comparative studies. At first, the thought was that we could facilitate the conversion of forcefields for users who really wanted those old forcefields, but the conversion machinery is so complex that this is simply infeasible.

Instead, we will convert the files for them, and just make the user make a conscious choice in order to get at those forcefields. They have to know what they are doing---that they aren't recommended---but that they need to replicate the work to, say, benchmark their algorithm against someone else's algorithm.

There is a much, much bigger movement afoot for advancing quantitative predictive modeling for drug discovery and materials design, and a pillar of this is the development of standard benchmark sets that will allow us to compare algorithms on the same test systems. There is no requirement that the forcefield used for this effort is recommended for new predictive studies, but we require reproducibility to make progress as a field.

jchodera:
> What do we mean by "unbundled" exactly? Would those be the ions combined in the same way as in the "bundled" files (see the README for how this is done), but without the waters, or we want individual conversion per each ion frcmod file from Amber?

Ideally, this is close to a one-to-one mapping between AMBER and ffxml files, but really "unbundled" should be whatever granularity level is reasonably easy to achieve.

peastman:
> This is not true. We wouldn't have added OBC GBSA or AMBER parm96 if this were the case, as they were already deprecated by the time they were added.

At the time we added those files, they absolutely were recommended (ff96 was considered to work better than ff99 for implicit solvent), and we were using them in production simulations.  If that hadn't been the case, we wouldn't have added them!

What about my suggestion of only including recommended force fields with OpenMM, but posting the other ones online so people who know what they're doing can get them?

rafwiewiora:
> At the time we added those files, they absolutely were recommended (ff96 was considered to work better than ff99 for implicit solvent), and we were using them in production simulations. If that hadn't been the case, we wouldn't have added them!

So the policy is to keep any forcefields that have ever been included in OpenMM in all future versions? ff96 was already in `oldff` in Amber 14 (don't know about before) - are we going to create an `oldff` folder at some point then in OpenMM? 

> What about my suggestion of only including recommended force fields with OpenMM, but posting the other ones online so people who know what they're doing can get them?

Or they could also go into the `devtools`. Are we going to keep the conversions script itself in the `devtools`? The full conversion, including all FFs, could go together with it.

jchodera:
It's a good idea, but we've seen that model doesn't work well. Consider the
case with the stuff in pyopenmm and the online script repository: Anything
that doesn't end up in the distribution is orders of magnitude harder to
use and hence barely used.

We want people to feel that they SHOULD use openmm to implement their
algorithms and compare to equivalent work done with other algorithms or
different packages, but we want to make sure they don't accidentally use
those parameters without knowing what they're doing. The combination of a
warning in the manual and the requirement that they specify a special
directory prefix to get at those files accomplishes that goal without
throwing wholly unnecessary roadblocks in their way.

peastman:
> So the policy is to keep any forcefields that have ever been included in OpenMM in all future versions?

Correct.  Otherwise, we break compatibility with existing scripts.

> ff96 was already in oldff in Amber 14 (don't know about before) - are we going to create an oldff folder at some point then in OpenMM?

No.  That again would break compatibility.

> Anything that doesn't end up in the distribution is orders of magnitude harder to use and hence barely used.

In this case, that's kind of the point. :)  Most users *shouldn't* use it.  People who really know what they're doing will be able to find it (as in, "know how to do a google search for 'openmm amber14'"), and people who don't know what they're doing are unlikely to stumble across it.

swails:
To play devil's advocate here:

> Correct. Otherwise, we break compatibility with existing scripts.

From a pragmatic point of view, this seems to be at odds with the assertion that you should not bundle force fields you know to be bad.  To me, it's hard to make the argument that "old scripts must *never* break" [1] justifies including old, decrepit force fields more than "users should be able to reproduce any study ever done with whatever force field it was done with".  One benefits only the inexperienced users of today, yielding the same drawbacks to those of tomorrow that are being used to reject the addition of old force fields now.

But there are ways of making these force fields easily accessible without bundling them with OpenMM proper -- release them as a separate conda package.  Furthermore, you can also easily implement behavior where importing the "extra force field" package automatically adds those force fields to the default search path (this may require a trivial change to OpenMM's code here):

```bash
$ conda install -c omnia obsoleteff
```

```python
import simtk.openmm as mm
import oldopenmmforcefieldsyoushouldneveruse # monkey-patches XML data directory search path

ff = mm.ForceField('amber82.xml') # works!
```

[1] My gut reaction to purposely breaking old scripts is strongly against it.  But when I push past that initial emotional reaction, I feel the benefit -- even to experienced users who may not realize the old script they pulled out used an old force field -- ultimately outweighs the cost.  Especially if old force fields are just a conda package + import away.

peastman:
We still need to make decisions about this.  Which force fields do we want to add?  Definitely ff16.  What about ff14?  Please vote yes or no.  Any others we should include?

Also, do we have the features we need to create a "main" XML file for a force field that automatically includes the separate files making it up?  I remember we discussed this before, but I'm not sure if it was resolved.

jchodera:
What do you think about the suggestion from @swails?

I strongly feel that the effort of my lab is wasted if we do not have a route to easily using old forcefields vital for comparison with existing published data.

rafwiewiora:
Ok, so a few points:

1) What about the forcefields that already are included? Do you want to update them to the new format?

2) What are we deciding to do with the old forcefields? Include in the distribution or somewhere online? @swails suggestion? As @jchodera also has asked while I was writing this

3) >Definitely ff16. What about ff14?

I'm not clear what we're referring here to. Can you use the names as in Amber? Is ff14 == ff14SB? There's no forcefield with 16 in the name. 

4) >Also, do we have the features we need to create a "main" XML file for a force field that >automatically includes the separate files making it up? I remember we discussed this before, but >I'm not sure if it was resolved.

I'm confused here also. There are two things that you might be referring to:

a) an old discussion where we wondered whether we should convert all `frcmod` files separately (so that the user could load `frcmod1` over `frcmod2` and make mix-match forcefields) - that idea was killed and we settled on `xml` to `leaprc` correspondence - i.e. Amber has `leaprc.ff99SBildn` and we have `ff99SBildn.xml`, not individual files that are called within that `leaprc`. 

b) AmberTools16, in a change from AmberTools15, introduced a separation of forcefields into `protein` and `nucleic` components - do we want to keep protein together with nucleic or separate them also? The separation seems to have achieved a greater simplicity in Amber - previously it wasn't exactly clear what exact parameters for nucleic acids you were getting with that particular forcefield - there is now only 2 or 3 parameter sets for DNA or RNA and the allowed combination are nicely laid out in the manual. We could of course just apply these combinations and make 'master' files (but that will increase the number of files substantially - you'll have `ff14SBprotein-DNA1', `ff14SBprotein-DNA2` etc.

swails:
Summary of my suggestion:

* Create a new conda package that, upon import, provides any and all FFs desired and inserts them into the default OpenMM search path (and do *not* bundle these with OpenMM)
* Move the old FFs we no longer recommend for *any* new study (like ff96) to this repo and out of OpenMM proper (this can be done via some kind of deprecation timeline policy where FFs have an expiration date that allows a couple releases to stop releasing a FF.)
* Only bundle FFs we would recommend for new studies with OpenMM (+ those in the middle of their phase-out process).

Rationale: neither argument "scripts must *never* break" or "everyone should be able to reproduce any study ever done" seem to be pragmatically superior.

jchodera:
@peastman: Can we schedule a follow-up skype to finalize these decisions?

peastman:
> 1) What about the forcefields that already are included? Do you want to update them to the new format?

No, that would risk breaking backward compatibility.

> I'm confused here also. There are two things that you might be referring to:

See https://github.com/pandegroup/openmm/pull/1570#issuecomment-245977616.

peastman:
> Create a new conda package that, upon import, provides any and all FFs desired

Fine with me as long as it's a manual install.  That is, you shouldn't get it automatically if you just type `conda install omnia`.  The user should clearly indicate that they specifically want that package.

rafwiewiora:
> No, that would risk breaking backward compatibility.

I just realized - the forcefields currently in OpenMM have ions in them. I don't see how we can make this work with the new water-ions ffxml's - it's just going to be a big mess. How do you see doing this? Why not just deprecate all current ffxml's into the new conda package and start fresh?

> So a user can just specify 'amber14.xml' and automatically get a consistent set of protein, DNA, and RNA parameters.

Quote here from https://github.com/pandegroup/openmm/pull/1570#issuecomment-245977616 - so you're talking about case 2 as I wrote above. So - do you want to use the AmberTools15 forcefields then + any new ones from AmberTools16, as AmberTools15 has them bundled in exactly this way, or you do want to 'rebundle' the ones from AmberTools16?

peastman:
> I just realized - the forcefields currently in OpenMM have ions in them. I don't see how we can make this work with the new water-ions ffxml's

We wouldn't try to.  The new ion files are for use with the new force fields, not with the old ones.

> So - do you want to use the AmberTools15 forcefields then + any new ones from AmberTools16, as AmberTools15 has them bundled in exactly this way, or you do want to 'rebundle' the ones from AmberTools16?

I don't think we want to mix and match them like that.  We'll get all the files from AmberTools16.  Then we'll make "top level" files for a handful of recommended combinations.

rafwiewiora:
> We wouldn't try to. The new ion files are for use with the new force fields, not with the old ones.

Ok, there's many ions that are missing from those older files I think, what if one wants to use one of those FFs with an ion parameter that is in the new files, but not in the old files? Do we do a separate ion conversion to work with those older files? Or do we want to test the new ion files for whether they work with the older ffxmls? I mean the ones that will have the `overload` tags (so the non-standard divalent and up sets) should work ok. 

> I don't think we want to mix and match them like that. We'll get all the files from AmberTools16. Then we'll make "top level" files for a handful of recommended combinations.

Ok got it. 

BTW I'll be back to finishing this off next week - in a fellowship writing craze until Saturday.

jchodera:
> Let's schedule a time for Skype about all this to be more effective communication-wise?

Agreed. We need to discuss in person. 

My jury duty for the week has ended, so we're essentially free anytime before 6.00P EST this week. @peastman: When are you free?

peastman:
> Ok, there's many ions that are missing from those older files I think

They include all the ions that were defined in those force fields.  The early ones had fewer ions than the newer ones, but that's just the nature of the force fields.  If you look at the current files, you'll see amber10.xml has more ions than amber99sb.xml.

> My jury duty for the week has ended, so we're essentially free anytime before 6.00P EST this week.

Just about any time is fine with me.  @swails do you have time to join too?  Or if not, can you give us your suggestions?  You probably know the amber force fields the best of all of us.

swails:
I can't join during working hours EST.  Best bet to get me on a call would be 8:15-8:30 pm EST (after the baby goes to sleep).

The major split in Amber FFs occurred as I was transitioning out of Dave's lab, so I don't have as intimate of knowledge as I've had in previous releases.  But the AmberTools 16 files define exactly the same force field as AmberTools 15, I believe (at least that is the intent, and I think @drroe invested significant time validating this for numerous protein and nucleic acid systems).

If that's the case, then just go with what is in AmberTools 16 (since the current OMM repo doesn't have anything newer than ff99SBildn as far as I can tell).  ff10 == ff99SB for proteins and ff99 for DNA/RNA + some other modifications (bsc0 and something else;  I don't remember).

The other force fields maybe worth considering are:

1. Lipid 14
2. GLYCAM 06h (latest one)
3. FF15ipq (or whatever the latest # is)

There are problems with each of these:

1. The Lipid14 residue templates break up single lipid residues into multiple "pieces", requiring the PDB file be preprocessed (via `charmmlipid2amber.py`) to split up single residues to match the template names in the .lib files via a hand-compiled mapping.
2. Similar to 1, the residue templates and conventions used in these files are tightly integrated with how tleap builds the system out, and GLYCAM has a glycam builder that's used to do a lot of the modelling/building.
3. Jury is still out on this FF, and it is protein-only at this point.  It *looks* like a good force field, but is still what I would call 'in development'.  And it uses a different charge model so it's incompatible with other RNA FFs.

I think 1. and 2. would both require changes to `Modeller` and `ForceField` to actually get working.

drroe:
> But the AmberTools 16 files define exactly the same force field as AmberTools 15, I believe (at least that is the intent, and I think @drroe invested significant time validating this for numerous protein and nucleic acid systems).

I will try to clarify this a bit. What @swails says is true depending on which `leaprc` file you are loading. For example, AT16 `leaprc.protein.ff14SB` and `leaprc.protein.ff14SBonlysc` load the same FF parameters and _protein_ units as AT15 `leaprc.ff14SB` and `leaprc.ff14SBonlysc`; the difference is that now no nucleic acid, water, or ion units are pulled in. We made a conscious decision to separate out parameters for the different biological units so it would be more clear to users what parameters they were using for each. For example, having an all-encompassing `ff14SB` ended up obfuscating what parameters were being used for nucleic acids. The parameters previously used with FF14 (BSC0) are in fact no longer available by themselves for DNA - the currently available DNA parameters are `leaprc.DNA.bsc1` (parameters from M. Orozco et al) or `leaprc.DNA.OL15` (parameters from J. Sponer et al). The previous RNA parameters (which were comprised of the BSC0 + OL3 corrections) are still available as `leaprc.RNA.OL3`, along with `leaprc.RNA.YIL` which was `oldff/leaprc.parmCHI_YIL.bsc` but fixed so that it uses parm10, not the older parm99 parameter set.

Parameters for the `leaprc.water.*` files are the same as in AT15 except now they pull in the recommended ion parameters for that solvent (monovalents only I think). `leaprc.gaff` is the same, but `leaprc.gaff2` is now available.

Let me know if I can clarify anything else.

peastman:
It sounds to me like supporting Lipid 14 and GLYCAM  06h is beyond the scope of this release.  It would take significant additional design and coding.

The most difficult question is probably how to handle the ions, since Amber now combines those with the water models.  The problem is that water models aren't specific to Amber.  Things like TIP3P, TIP4P-Ew, SPC/E, etc. were all published independently of any particular force field, and can be used with many different force fields.  But Amber uses different ion parameters depending on the water model, and stores water and ion parameters in a single file.  That creates two problems for us:

1. You should be able to use the water models with non-Amber force fields.
2. You should be able to use other water models (like TIP4P-FB) with Amber force fields.

If the water and ion parameters are in a single file, both of those become impossible.  How can we handle this?

rafwiewiora:
> It sounds to me like supporting Lipid 14 and GLYCAM 06h is beyond the scope of this release. It would take significant additional design and coding.

Lipids - I was very close to having this converted, don't remember off the top of my head what the problem with them was there, will have to look through the history of our Issues later - but as far as I remember this should be solvable not too hard - but not for this release, we can include them moving forward. 

GLYCAM - @swails I thought Parmed doesn't support reading .PREP files right? That's why I couldn't do the GLYCAM conversions. So that will need quite a lot of work maybe. 

> But Amber uses different ion parameters depending on the water model, and stores water and ion parameters in a single file. That creates two problems for us:
> 
> You should be able to use the water models with non-Amber force fields.
> You should be able to use other water models (like TIP4P-FB) with Amber force fields.
> If the water and ion parameters are in a single file, both of those become impossible. How can we handle this?

So Amber doesn't actually store them in the same files, it just tells you in the names of the ion files which water you should use them with. We decided at some point to bundle water and ions together in this conversion to make sure people 'always did the right thing' and didn't mix the wrong water with wrong ions up. Very easy to separate them. In fact, the waters are NOT converted by the script, but they are fed to it from a manual conversion (i.e. I just took the water files from OpenMM 7.0) - that's because dealing with dummy atom --> virtual site conversions just seemed like too much work for something that was already converted. So those waters are just pasted in the ion parameters that the script converts. We can separate them (but then how do you enforce use of the right parameters with the right water?) or have bundles but also separate files. 

> For example, having an all-encompassing ff14SB ended up obfuscating what parameters were being used for nucleic acids. The parameters previously used with FF14 (BSC0) are in fact no longer available by themselves for DNA

Thanks for the clarification @drroe ! So @peastman in this case we need to decide, if you want protein-nucleic bundles in OpenMM:

a) go with the AmberTools15 parameters
b) go with the AmberTools16 parameters and bundle all allowed protein-nucleic combinations ourselves
c) do not bundle, as is currently the case in this PR

peastman:
> So Amber doesn't actually store them in the same files, it just tells you in the names of the ion files which water you should use them with.

Ok, that's good to know.

> So @peastman in this case we need to decide, if you want protein-nucleic bundles in OpenMM:

Let's take everything from AmberTools 16.  No strong opinion about whether they should be bundled or separate.

jchodera:
We'll migrate this to https://github.com/choderalab/openmm-forcefields

rafwiewiora:
Closing this, stuff migrating to https://github.com/choderalab/openmm-forcefields .

PR #1584, opened 2016-09-08
support DESTDIR in PythonInstall

wesbarnett:


peastman:
Could you explain what this is meant to accomplish?

wesbarnett:
The main issue for me is DESTDIR in staged installs. I build binary packages for my cluster and distribute them using my system's package manager (PKGBUILD's on Arch Linux if you are familiar). Without DESTDIR python tries to install to the root dir even though I'm attempting to stage it in another folder and compress into my package. So the goal would be able to do:

    make DESTDIR=$pkgdir PythonInstall

Just like I can do for the regular `make install`.

wesbarnett:
I removed `CMAKE_INSTALL_PREFIX` since my main concern with this was `DESTDIR` and I don't think it make sense to go with the `--root` flag like `DESTDIR`.

peastman:
This seems plausible, but I'm not familiar with the DESTDIR variable, and I can't find any documentation on the --root flag.  Anyone else have comments on this?

peastman:
If no one else has any comments on this, I'm going to go ahead and merge it.  Anyone?

PR #1591, opened 2016-09-21
updateParametersInContext() causes stateChanged() to be called

peastman:
See https://github.com/choderalab/openmmtools/issues/118.  @gregoryross or @jchodera can one of you verify this fixes the problem for you?

### DIFF
--- a/openmmapi/include/openmm/internal/ContextImpl.h
+++ b/openmmapi/include/openmm/internal/ContextImpl.h
@@ -253,6 +253,10 @@ public:
         integratorIsDeleted = true;
     }
     /**
+     * Notify the integrator that some aspect of the system has changed, and cached information should be discarded.
+     */
+    void systemChanged();
+    /**
      * This is the routine that actually computes the list of molecules returned by getMolecules().  Normally
      * you should never call it.  It is exposed here because the same logic is useful to other classes too.
      */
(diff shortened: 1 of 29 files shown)


gregoryross:
@peastman Thanks a lot, I'm happy to test this out. Could you recommend a way to test it? A good suggestion I've had is to create a separate conda environment for this PR. Do you have any other recommendations?

peastman:
Are you not set up to build from source?  The instructions are at http://docs.openmm.org/7.0.0/userguide/library.html#compiling-openmm-from-source-code.

gregoryross:
I'm trying to install it from source but I'm getting many errors with `make PythonInstall` and can't figure out the problem. I'm using a Mac, so I'm going to switch and try installing on a Linux machine.

jchodera:
Is it essential to test if this solves the problem before merging? Once we merge, the nightly builds will allow us to test the `dev` package in the morning without problems.

peastman:
No, I assumed it would be easy for you to test.  But if it isn't I can go ahead and merge.

gregoryross:
Okay great, thanks. That certainly simplifies things for me!

jchodera:
I've retriggered the `conda-dev-recipes` build, so the new package should appear [here](https://anaconda.org/omnia/openmm/files) in a few minutes.

jchodera:
Something is broken with `conda-dev-recipes`---will have a look in the morning.

jchodera:
Fixed:

You should now be able to do
```
conda install -c https://conda.anaconda.org/omnia/label/dev
```
to test.

PR #1592, opened 2016-09-22
Fixed a potential nan

### LINKED ISSUE #1586: Particle coordinate is nan error related to GB forces.
I've been experiencing some issues running implicit solvent simulations. In our particular test case, we observe crashes for a `ACE-TYR-NME` system. They occur often, but it is hard to reproduce the exact moment when it's going to happen. 

@jchodera discovered that the errors disappear when we eliminate `GBSAOBC` Forces from the system. 

Here is the minimal example, which contains the script, and serialized systems with and without GB forces. This is using openmm 7.0.1. I included the conda environment specifications in the zip file.

[debug.zip](https://github.com/pandegroup/openmm/files/464730/debug.zip)

When we run the script included, we get these errors, which happen in some iteration of the loop.

```pytb
  File "minimal.py", line 26, in <module>
    integrator.step(10)  # MD
  File "/home/rustenburg/anaconda/envs/gbforces/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 5330, in step
    return _openmm.CustomIntegrator_step(self, steps)
Exception: Particle coordinate is nan
```

It is hard to reproduce exactly the moment where it fails, but it is bound to happen if we run the loop for a few iterations. It does not occur when we remove GB forces.  We think there could be a numerical issue with the GB forces that is causing this behavior.

peastman:
Fixes #1586.

### DIFF
--- a/platforms/cpu/src/CpuGBSAOBCForce.cpp
+++ b/platforms/cpu/src/CpuGBSAOBCForce.cpp
@@ -88,7 +88,7 @@ void CpuGBSAOBCForce::setParticleParameters(const std::vector<std::pair<float, f
     particleParams = params;
     bornRadii.resize(params.size()+3);
     obcChain.resize(params.size()+3);
-    for (int i = bornRadii.size()-3; i < bornRadii.size()-1; i++) {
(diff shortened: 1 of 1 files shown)


jchodera:
Thanks!

PR #1601, opened 2016-09-26
CompoundIntegrator::stateChanged() calls through to component integrators

peastman:
Fixes https://github.com/choderalab/openmmtools/issues/118.


jchodera:
Thanks!

jchodera:
We can test again once this is merged and the `dev` build rebuilds.

gregoryross:
Thanks so much for your fast response! You're fix looks good. @peastman, would you mind merging this so I can test this on the `dev` build as I'm still having trouble compiling `openmm` from source.

jchodera:
I've retriggered the nighly build, so it should update shortly:

gregoryross:
Thanks!

PR #1608, opened 2016-10-04
Add support for customIntegrator serialization and test

kyleabeauchamp:


kyleabeauchamp:
Not sure what the issue is with the test, AFAIK it runs on my machine.

jchodera:
The whole testing behavior looks really wonky right now, with several bizarre failures and a timeout.

peastman:
TestSerializeIntegrator is failing everywhere, on both travis and appveyor, so it's clearly a real error.  But it's not rerunning it or printing any output, so I'm not sure what's going on.

jchodera:
We are having an issue with yank where conda-build updated to 2.0.3 and that seems to have broken things. Try pinning to 2.0.2 or 2.0.1 in devtools?

peastman:
Something totally bizarre is going on here.  I'm debugging right now.  It's... really strange.

peastman:
Here's what I've found so far.

The test passes when run directly, but fails when run by ctest.  Ok, that's strange.

Here's the line where it fails:

```C++
    XmlSerializer::serialize<Integrator>(intg, "CustomIntegrator", ss);
```

But when I tried printing messages from inside `serialize()` to identify exactly where it's failing, I found that it gets all the way through the method without problem.  A print command at the very end of `serialize()` gets executed, but one in the calling method immediately after it doesn't.  That sounds like memory corruption.

I started commenting out code in the test case to see which of your changes was triggering the error.  It's this line:

```C++
    intg->addGlobalVariable("conditional_v1", 0);
```

Simply adding more global variables causes the failure.  But only for certain variable names.  After some experimenting, I figured out that it works if all the global variables are in alphabetical order, but fails if they aren't.  But remember, only if it's run through ctest.  If run directly, it works.

That's what I've found so far.  Really strange.

peastman:
I figured out what's going on.  It's not producing any output because TestSerializeIntegrator.cpp (which was not written by me!) doesn't report errors.  It just swallows them and exits with code 1!  Sigh.

The actual error is that serialization isn't preserving the order of the global variables.  They get stored as attributes of the XML tag, and end up in alphabetical order.  It looks like other custom classes may have the same problem.  But that's separate from what you're doing here.  So for the moment, just move the two calls to `addGlobalVariable()` to the start of the list, so all global variables are added in alphabetical order.  That will make the test pass, and we can merge this.  And then I can deal with the ordering separately.

While you're fixing that, could you also add a few lines to the `main()` function?  The end of the file should look like

```C++
    catch(const exception& e) {
        cout << "exception: " << e.what() << endl;
        return 1;
    }
    cout << "Done" << endl;
    return 0;
}
```

That is, it should print the exception if there is one, or "Done" if there isn't one.

kyleabeauchamp:
Done.  Instead of re-ordering, I renamed.  I also confirmed that switching the "v1" and "v2" tags now cause a test failure to be printed.

peastman:
Thanks!

Issue #1616, opened 2016-10-07
Energy Conservation of MTSIntegrator


kyleabeauchamp:
Is there a canonical way to test the energy conservation performance of our integrators?  I'm finding reduced conservation as compared to Verlet, as measured by the below script.

```
import simtk.openmm.app as app
import simtk.openmm as mm
from simtk import unit as u
from openmmtools import testsystems

f = lambda x: (x.getPotentialEnergy(), x.getKineticEnergy(), x.getKineticEnergy() + x.getPotentialEnergy())

temperature = 300. * u.kelvin
testsystem = testsystems.LennardJonesFluid(nparticles=2048)
timestep = 0.1 * u.femtoseconds

integrators = [mm.VerletIntegrator(timestep), mm.MTSIntegrator(timestep, groups=[(0, 1)])]

for integrator in integrators:
    print(integrator)

    simulation = app.Simulation(testsystem.topology, testsystem.system, integrator)
    simulation.context.setPositions(testsystem.positions)
    simulation.context.setVelocitiesToTemperature(temperature)

    integrator.step(1)  # Do a single step first to apply constraints before we run tests.

    state1a = simulation.context.getState(getPositions=True, getParameters=True, getEnergy=True, getVelocities=True)
    old = f(state1a)
    print("PE, KE, TOTAL")
    print(old)
    integrator.step(1)
    state1b = simulation.context.getState(getPositions=True, getParameters=True, getEnergy=True, getVelocities=True)
    new = f(state1b)
    print(new)
    print("delta = %s" % (new[-1] - old[-1]))


<simtk.openmm.openmm.VerletIntegrator; proxy of <Swig Object of type 'OpenMM::VerletIntegrator *' at 0x7fbcdd422f60> >
PE, KE, TOTAL
(Quantity(value=-220.9337554729007, unit=kilojoule/mole), Quantity(value=7704.49522032775, unit=kilojoule/mole), Quantity(value=7483.561464854849, unit=kilojoule/mole))
(Quantity(value=-220.95013973622264, unit=kilojoule/mole), Quantity(value=7704.512487734668, unit=kilojoule/mole), Quantity(value=7483.562347998445, unit=kilojoule/mole))
delta = 0.0008831435961837997 kJ/mol
<simtk.openmm.mtsintegrator.MTSIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fbcb9059fc0> >
PE, KE, TOTAL
[...]

PR #1617, opened 2016-10-07
[WIP] Fix precision issue in MTSIntegrator

jchodera:
This should fix the precision issue with `MTSIntegrator` noted by @kyleabeauchamp in #1616.


jchodera:
D'oh. I'm not sure why it pulled in those other changes.

PR #1618, opened 2016-10-07
Fix precision issue in MTSIntegrator

jchodera:
This should fix the precision issue with `MTSIntegrator` noted by @kyleabeauchamp in #1616.

### DIFF
--- a/wrappers/python/tests/TestIntegrators.py
+++ b/wrappers/python/tests/TestIntegrators.py
@@ -1,0 +1,103 @@
+import unittest
+import tempfile
+from datetime import datetime, timedelta
+from simtk.openmm import *
+from simtk.openmm.app import *
+from simtk.unit import *
+import math, random
+
+class TestIntegrators(unittest.TestCase):
+    """Test Python Integrator classes"""
+
+    def testMTSIntegratorExplicit(self):
+        """Test the MTS integrator on an explicit solvent system"""
+        # Create a periodic solvated system with PME
+        pdb = PDBFile('systems/alanine-dipeptide-explicit.pdb')
+        ff = ForceField('amber99sbildn.xml', 'tip3p.xml')
+        system = ff.createSystem(pdb.topology, cutoffMethod=PME)
+
+        # Split forces into groups
+        for force in system.getForces():
+            if force.__class__.__name__ == 'NonbondedForce':
+                force.setForceGroup(1)
+                force.setReciprocalSpaceForceGroup(2)
+            else:
+                force.setForceGroup(0)
+
+        # Create an integrator
+        integrator = MTSIntegrator(4*femtoseconds, [(2,1), (1,2), (0,8)])
+
+        # Run a few steps of dynamics
+        context = Context(system, integrator)
+        context.setPositions(pdb.positions)
+        integrator.step(10)
+
+        # Ensure energy is well-behaved.
+        state = context.getState(getEnergy=True)
+        if not (state.getPotentialEnergy() / kilojoules_per_mole < 0.0):
+            raise Exception('Potential energy of alanine dipeptide system with MTS integrator is blowing up: %s' % str(state.getPotentialEnergy()))
+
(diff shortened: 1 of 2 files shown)


jchodera:
Hm, there don't appear to be any tests at all for `MTSIntegrator`...

jchodera:
What kind of test should I add for this? 

I could load one of the [systems in here](https://github.com/pandegroup/openmm/tree/master/wrappers/python/tests/systems) and take a few steps, or perhaps test integration for a small harmonic system.

kyleabeauchamp:
The MTS integrator with `groups=[(0, 1)]` is *almost* identical to VV, with the caveat being the split of the velocity update into two parts separated by a constraint evaluation.  For an unconstrained system, then, you might be able to check exact positional agreement after 100 steps of both VV and MTS, starting from the same velocities and positions.

kyleabeauchamp:
Thanks!

jchodera:
Does this provide a useful precision improvement, @peastman? Or does it have no real effect?

jchodera:
And if it's worth merging, what test(s) should I add? There is currently no test coverage for `MTSIntegrator`.

peastman:
It seems quite plausible this change could meaningfully improve the results.  We just have to test it and see.

The obvious test would be energy conservation, but of course, the difference will only be detectable with CUDA or OpenCL running in single precision mode.  The reference and CPU platforms always do integration in double precision.  And test cases for the Python layer can't assume CUDA or OpenCL is available.  So adding a test to make sure MTSIntegrator works (by checking energy conservation with the reference platform) would be useful.  It just won't demonstrate that this change made a difference.

jchodera:
We have zero test coverage for MTSIntegrator right now, so literally
anything would be an improvement.

What kind of energy convergence test are you thinking? Can you point to a
relevant one for the C++ integrators I should replicate?

peastman:
Something like this is a good example of a test that checks energy conservation: https://github.com/pandegroup/openmm/blob/master/tests/TestVerletIntegrator.h#L83-L134

kyleabeauchamp:
The problem that that example is that it's in the C++ API, but the MTS integrator is defined only in python space.

peastman:
Right, so we'll want a Python test case for it.  That example was just meant as an example to use as inspiration.

jchodera:
Whoops, forgot to add that test!

jchodera:
Tests added. They're more verbose than I would have liked because I seem to recall we're still avoiding the use of `numpy` (is that right?).

wrappers/python/tests/TestIntegrators.py
```diff
@@ -0,0 +1,103 @@
...
+        context.setPositions(pdb.positions)
+        integrator.step(10)
+
+        # Ensure energy is well-behaved.
+        state = context.getState(getEnergy=True)
+        if not (state.getPotentialEnergy() / kilojoules_per_mole < 0.0):
+            raise Exception('Potential energy of alanine dipeptide system with MTS integrator is blowing up: %s' % str(state.getPotentialEnergy()))
+
+        # Clean up
+        del context, integrator
```
peastman: This line is a noop.  It's the end of the method, so those variables are about to go out of scope anyway.

peastman:
> I seem to recall we're still avoiding the use of numpy (is that right?).

Correct.  Tests look good.

wrappers/python/tests/TestIntegrators.py
jchodera: Done.

peastman:
I'm not sure why AppVeyor failed, but it's clearly not related to this, since the failure was in a C++ test and this only modifies Python files.

Is this ready to merge?

jchodera:
> Is this ready to merge?

If you are OK with the tests, it's ready to merge---but remember that any tests are certainly better than no tests!

peastman:
> but remember that any tests are certainly better than no tests!

Agreed!

PR #1619, opened 2016-10-07
[WIP] Allow PDBFile to automatically add bonds for all standard monomers in RCSB

jchodera:
As we prepare to add forcefields that contain more than just 20 amino acids or other components, the limitation that`PDBFile` can only recognize bonds for these residues is becoming an increasing headache.

Instead, we should allow any bonds in the [standard RCSB monomer library](http://ligand-expo.rcsb.org/ld-download.html) to be automatically recognized.

This PR is just the starting point for a discussion to demonstrate how this might work. I started with the [`findBonds.py` script from the SimTK `pyopenmm` repo](https://github.com/pandegroup/openmm/tree/master/devtools/forcefield-scripts) and modified it to use the [PDBML format dictionary](http://ligand-expo.rcsb.org/dictionaries/components-pub-xml.tar.gz), automatically identifying amino acids or nucleic acids by atom content. The final XML file was tens of MB, so I instead used `GzipFile` to produce a 3MB `.xml.gz` file, modifying the `Topology` object to instead load this internally.

The `residues.xml` generation script can eventually be migrated to the [`openmm-forcefields` repo](https://github.com/choderalab/openmm-forcefields), but I think we would want to keep the ability to read `residues.xml.gz` (potentially falling back to `residues.xml` if it doesn't exist).


jchodera:
Wow. There's a whole host of travis failures that are completely unrelated to this PR.

A lot of these were a bunch of stochastic tests that failed at the same time. That would seem to be fairly improbable.

There *is* an API building step that seemed to fail:
```
reading sources... [  6%] generated/OpenMM.AmoebaMultipoleForce                 
Exception occurred:
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/sphinxcontrib/autodoc_doxygen/autosummary/__init__.py", line 41, in import_by_name
    raise ImportError('no module named %s' % ' or '.join(tried))
ImportError: no module named NonbondedMethod
The full traceback has been saved in /tmp/sphinx-err-Di6yPw.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make[3]: *** [api-c++/index.html] Error 1
make[2]: *** [docs-source/api-c++/CMakeFiles/C++ApiDocs.dir/all] Error 2
make[1]: *** [docs-source/api-c++/CMakeFiles/C++ApiDocs.dir/rule] Error 2
make: *** [C++ApiDocs] Error 2
```
Not sure why.

peastman:
> the limitation thatPDBFile can only recognize bonds for these residues is becoming an increasing headache.

Note that it does have the ability to handle other residues.  You just need to provide the bond definitions by calling `Topology.loadBondDefinitions()`.

jchodera:
So the idiom here would be something like:
```python
# Load additional bond definitions
Topology.loadBondDefinitions('my-residues.xml')
# Load my PDB file
pdb = PDBFile('my-pdb.pdb')
```
It looks like this logic causes `my-residues.xml` to be read first, and then triggers the standard `residues.xml` to be read afterwards when `PDBFile` invokes `createStandardBonds()` (presuming `createStandardBonds()` hasn't previously been invoked). Is that correct?

This is useful to know! But since we are already supporting *some* standard RCSB monomers, we might as well just support them all without having to make the user go through backflips just to read residues that exist in the forcefields we are adding support for.

peastman:
> But since we are already supporting some standard RCSB monomers, we might as well just support them all without having to make the user go through backflips just to read residues that exist in the forcefields we are adding support for.

I'd definitely agree as far as including all residues that are found in forcefields we provide.  But there are many thousands of monomers defined in the chemical components dictionary, and for most of them, the only way they'll ever appear in a user's PDB file is by mistake, if they're using nonstandard residue names.  This is actually a pretty common source of bug reports from users.  Someone will ask, "Why can't OpenMM build a system from my PDB file?"  And I'll explain to them, "You've labeled your ions as CAL, which is the code for 5-AMINO-6-CYCLOHEXYL-4-HYDROXY-2-ISOBUTYL-HEXANOIC ACID.  The correct name for a calcium ion is CA."

So I'm not sure there's a benefit to including things like that, and it might just produce more confusion.

jchodera:
Hand curating a list of residues is going to be crazy hard---have you
looked at CHARMM?

What about using an optional flag to specify whether PDBFile should
automatically add bonds for "minimal" residue sets (current behavior), all
monomers in the PDB, or nothing? Current behavior could be maintained by
default.

On Oct 7, 2016 1:10 PM, "peastman" <notifications@github.com> wrote:

But since we are already supporting some standard RCSB monomers, we might
as well just support them all without having to make the user go through
backflips just to read residues that exist in the forcefields we are adding
support for.

I'd definitely agree as far as including all residues that are found in
forcefields we provide. But there are many thousands of monomers defined in
the chemical components dictionary, and for most of them, the only way
they'll ever appear in a user's PDB file is by mistake, if they're using
nonstandard residue names. This is actually a pretty common source of bug
reports from users. Someone will ask, "Why can't OpenMM build a system from
my PDB file?" And I'll explain to them, "You've labeled your ions as CAL,
which is the code for 5-AMINO-6-CYCLOHEXYL-4-HYDROXY-2-ISOBUTYL-HEXANOIC
ACID. The correct name for a calcium ion is CA."

So I'm not sure there's a benefit to including things like that, and it
might just produce more confusion.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/pandegroup/openmm/pull/1619#issuecomment-252308283>,
or mute
the thread

peastman:
> have you looked at CHARMM? 

I haven't.  But presumably there's a list somewhere of the residues they include?  If so, we don't need to hand curate it.  It's already there.

jchodera:
> But presumably there's a list somewhere of the residues they include? If so, we don't need to hand curate it. It's already there.

Nope! That would be too easy.

Also, they don't use standard residue names, so we can't automatically extract them.

I suppose we (by which I mean "someone besides me") could write some sort of complicated graph-matching thing to map CHARMM residues to residues in the PDB, but that doesn't seem like the right approach for something like `PDBFile`.

To me, the "right" approach seems to be to give `PDBFile` a few different operating modes:
* Read the PDB file as-is ("strict" mode)
* Read the PDB file with some help from the PDB monomer library to perceive bonds
* Read the PDB file with some help from a `ForceField` object to perceive bonds in residues defined by the forcefield

I'd love to see all three modes supported.  What do you think?

jchodera:
These test failures are so weird. Why would all the stochastic tests suddenly be failing at a high rate?

peastman:
Technically speaking, the "correct" thing to do is exactly what we're currently doing (add bonds for the ones we currently add bonds for, and no others), with the one exception that we shouldn't automatically add bonds for disulfides.  The documentation for the CONECT record (http://www.wwpdb.org/documentation/file-format-content/format33/sect10.html#CONECT) specifies:

> CONECT records are present for:
>
> - Intra-residue connectivity within  non-standard (HET) residues (excluding water).
> - Inter-residue connectivity of HET  groups to standard groups (including water) or to other HET groups. 
> - Disulfide bridges specified in the  SSBOND records have corresponding records.

The section on heterogens (http://www.wwpdb.org/documentation/file-format-content/format33/sect4.html) defines them as anything other than:

> standard amino acids, or
> standard nucleic acids (C, G, A, U, I, DC, DG, DA, DU, DT and DI), or
> unknown amino acid (UNK) or nucleic acid (N) where UNK and N are used to indicate the unknown residue name.

So if we assume input files conform to the spec, we're expected to fill in bonds for standard amino acids, standard nucleotides, and water.  Anything else is expected to have CONECT records.

But of course, non-conforming PDB files are very common. :(  The second mode you mentioned (using the monomer library for filling in bonds) is simple as long as the non-conforming PDB file still uses standard names for all residues and atoms, and is non-conforming only in the absence of CONECT records.  But if any names are nonstandard, that won't work.  For the third mode you mentioned (using a force field), I don't see how that would work.  We can't match on names, because force fields aren't expected to use any sort of standard naming system.  They match based on bonds (which we've already assumed are missing) and elements (which also are frequently missing).

jchodera:
> So if we assume input files conform to the spec, we're expected to fill in bonds for standard amino acids, standard nucleotides, and water. Anything else is expected to have CONECT records.

Aha! Excellent point.

> But of course, non-conforming PDB files are very common. :( The second mode you mentioned (using the monomer library for filling in bonds) is simple as long as the non-conforming PDB file still uses standard names for all residues and atoms, and is non-conforming only in the absence of CONECT records. But if any names are nonstandard, that won't work. 

Do you think there is value in adding a mode where it will try to fill in bonds for nonstandard residues that might be missing `CONECT` residues? We unfortunately run into this all the time, and if the philosophy is to be permissive on reading and strict on writing, adding an optional mode to help users read in nonconforming files (many of which are generated by other simulation packages) may be helpful. This is potentially functionality that could be incorporated into `pdbfixer` instead.

> For the third mode you mentioned (using a force field), I don't see how that would work. We can't match on names, because force fields aren't expected to use any sort of standard naming system. They match based on bonds (which we've already assumed are missing) and elements (which also are frequently missing).

It would have to match on the `ffxml` `Residue` names. They're arbitrary, but there is really no other information to go on.

peastman:
I think the second mode could potentially be useful.  I'm much more doubtful about the third mode.

jchodera:
@peastman : Just wanted to see what you thought of this functionality.

On the one hand, OpenMM already supports adding bonds automatically for `ATOM` records for standard residues in the PDB spec. Everything else should be a `HETATM` with bonds explicitly specified by `CONECT`. (Or more modern mmCIF/PDBx files, but I'm not sure how those handle this issue.)

On the other hand, it was a nightmare for either of us to even generate files that conformed to that spec.
Users might be immensely frustrated by this as well and just give up.

What's the right solution here?

peastman:
If you can suggest a solution to the problems discussed above, I'd love to hear it.  The issue is how to identify what bonds to add.  Since we don't know the bonds in advance, we can't rely on any sort of template matching.  The only thing we have to go by are the names of residues and atoms.  If the file came from RCSB, we can count on it having standard names, but we can also count on it having CONECT records.  If it's a nonstandard file from some other source, it probably won't use standard names, so how do we match it up to a standard residue definition?

wrappers/python/simtk/openmm/app/topology.py
```diff
@@ -256,8 +257,12 @@ def createStandardBonds(self):
         """
         if not Topology._hasLoadedStandardBonds:
             # Load the standard bond definitions.
+            try:
+                infile = GzipFile(os.path.join(os.path.dirname(__file__), 'data', 'residues.xml.gz'), 'r:gz')
+            exception Exception as e:
```
swails: Well this won't work.

Also, you should catch a more general exception.

peastman:
Closing abandoned PRs.

PR #1621, opened 2016-10-09
README updates

jchodera:
* Remove outdated `openmm-dev` anaconda and travis badges
* Replace `openmm` anaconda badge (previously read 0.0.0) with one that lists total number of downloads
* Move badges to top
* Eliminate outdated links to OpenMM 6.3 Python and C++ references while maintaining links to overall documentation 

Here is a [preview](https://github.com/jchodera/openmm/blob/6516827afc01b885abb45dec847b1d5f8bd974ca/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+[![Build Status](https://travis-ci.org/pandegroup/openmm.svg?branch=master)](https://travis-ci.org/pandegroup/openmm)
+[![Anaconda Cloud Badge](https://anaconda.org/omnia/openmm/badges/downloads.svg)](https://anaconda.org/omnia/openmm)
+
 ## OpenMM: A High Performance Molecular Dynamics Library
 
 Introduction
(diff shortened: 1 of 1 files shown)


peastman:
Is this ready to merge?

I'm not sure what caused the travis failure, but it doesn't seem related to this.

jchodera:
> Is this ready to merge?

Yes, if you like the format in the [preview](https://github.com/jchodera/openmm/blob/6516827afc01b885abb45dec847b1d5f8bd974ca/README.md)!

> I'm not sure what caused the travis failure, but it doesn't seem related to this.

Related tests have been failing repeatedly, like [here](https://travis-ci.org/pandegroup/openmm/jobs/165869666#L1342-L1344), which is why I worry we might to see if we can enable [all tests to run](https://github.com/pandegroup/openmm/issues/1620#issuecomment-252679630).

PR #1629, opened 2016-10-13
Remove warning about runtime compiler on CUDA 8.0

peastman:


peastman:
The OpenCL tests failed with this error:

```
ERROR: cannot verify jenkins.choderalab.org's certificate, issued by ‘/C=US/ST=Arizona/L=Scottsdale/O=GoDaddy.com, Inc./OU=http://certs.godaddy.com/repository//CN=Go Daddy Secure Certificate Authority - G2’:

  Issued certificate has expired.
```

@jchodera it looks like you need to update a certificate.  I'll go ahead and merge this, since it involves only changes to the CUDA platform.

jchodera:
Sorting out the SSL certificate issues now.

jchodera:
I'll move this to an amazon S3 bucket in the meantime.

PR #1630, opened 2016-10-13
Fixed bug computing energies without forces

### LINKED ISSUE #1625: Force Group Funniness (CUDA)
I've been testing out some force group related integrators and have run into some issues where the CUDA platform behaves erratically when using force groups.  So far, the minimal example is constructed by:

1.  Create a simple testsystem (LJ Box)
2.  Create a Custom Integrator that is essentially the HMC integrator from OpenMMTools but with the force term replaced by either "f" or "f0". 
3.  The system has only a single force group (0), so both choices should be identical.
4.  Run the integrator
5.  When the force group is "f", all platforms work fine.
6.  When the force group is "f0", the CUDA platform rejects all moves due to energy being NAN, while the other platforms have reasonable accept rates.

Any ideas?  Perhaps there's something fishy going on with energy recalculation when using force groups?

Example code is at the following gist, with output shown below:

https://gist.github.com/kyleabeauchamp/706413cba0da1fb22a0c7cf409a80e96

```
********************************************************************************
force groups:
<simtk.openmm.openmm.NonbondedForce; proxy of <Swig Object of type 'OpenMM::NonbondedForce *' at 0x7fc4dfcd0f00> > 0
********************************************************************************
Platform, integrator / acceptance rate, Enew
CUDA <__main__.HMCIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fc4dfcd0f90> >
0.095 -301.8164672889289
CUDA <__main__.HMCIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fc4dfcd0a20> >
0.0 nan
OpenCL <__main__.HMCIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fc4e57d2bd0> >
0.105 -304.9968477691166
OpenCL <__main__.HMCIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fc4df77f090> >
0.09 -300.0922760765787
CPU <__main__.HMCIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fc4df77f060> >
0.145 -314.9459738356866
CPU <__main__.HMCIntegrator; proxy of <Swig Object of type 'OpenMM::CustomIntegrator *' at 0x7fc4df77f0c0> >
0.195 -323.46419506994835
```

peastman:
Fixes #1625.

This was really hard to track down!  The error occurs when `ContextImpl::calcForcesAndEnergy()` gets called with `includeEnergy` true but `includeForces` false.  That doesn't happen when you call `getState()`, since even if you don't ask for forces, it still needs them to correct the kinetic energy.  And it doesn't happen within standard integrators, because they don't use energy.  The only place it happens is in a CustomIntegrator that requests energy but not forces.  That happened in #1625 because it specifies the forces as `f0` (only group 0), but the energy as `energy` (all groups).

### DIFF
--- a/platforms/cuda/src/kernels/nonbonded.cu
+++ b/platforms/cuda/src/kernels/nonbonded.cu
@@ -278,9 +278,9 @@ extern "C" __global__ void computeNonbonded(
                 localData[tbx+tj].fz += dEdR2.z;
 #endif 
 #endif // end USE_SYMMETRIC
+#endif
 #ifdef ENABLE_SHUFFLE
                 SHUFFLE_WARP_DATA
-#endif
 #endif
 #ifdef USE_EXCLUSIONS
                 excl >>= 1;
@@ -484,9 +484,9 @@ extern "C" __global__ void computeNonbonded(
                     localData[tbx+tj].fz += dEdR2.z;
 #endif 
 #endif // end USE_SYMMETRIC
+#endif
 #ifdef ENABLE_SHUFFLE
                     SHUFFLE_WARP_DATA
-#endif
 #endif
                     tj = (tj + 1) & (TILE_SIZE - 1);
                 }
@@ -555,9 +555,9 @@ extern "C" __global__ void computeNonbonded(
                     localData[tbx+tj].fz += dEdR2.z;
 #endif 
 #endif // end USE_SYMMETRIC
+#endif
 #ifdef ENABLE_SHUFFLE
                     SHUFFLE_WARP_DATA
-#endif
 #endif
                     tj = (tj + 1) & (TILE_SIZE - 1);
                 }


jchodera:
Could this have had effects beyond the specific use of force groups, where the `energy` is requested but not the forces in `CustomIntegrator`? We've still been seeing weird behavior in some of our `CustomIntegrators` that use GHMC (where energies are requested separately from forces) but have been having a hard time coming up with really simple bug demonstration code.

peastman:
It's possible.  Try this and see if it fixes them.  Another easy test: this bug existed in the CUDA platform but not OpenCL, so the behavior would have been different.

jchodera:
> It's possible. Try this and see if it fixes them. Another easy test: this bug existed in the CUDA platform but not OpenCL, so the behavior would have been different.

Thanks! We'll test it out after this is merged.

PR #1639, opened 2016-10-17
Update docstrings to specify units and domain for torsions

### LINKED ISSUE #1637: What is the domain for the `dihedral()` function in CustomCompoundBoundForce?
To use the `dihedral` function defined in [`CustomCompoundBondForce`](http://docs.openmm.org/7.0.0/api-python/generated/simtk.openmm.openmm.CustomCompoundBondForce.html#simtk.openmm.openmm.CustomCompoundBondForce), it's necessary to know what domain this function uses. The docs don't seem to specify.

jchodera:
Fixes #1637

Updates docstrings for `CustomTorsionForce` and `CustomCompoundBondForce` to specify units and domain for computed torsion/dihedral.

### DIFF
--- a/openmmapi/include/openmm/CustomTorsionForce.h
+++ b/openmmapi/include/openmm/CustomTorsionForce.h
@@ -52,6 +52,7 @@ namespace OpenMM {
  * part of the system definition, while values of global parameters may be modified during a simulation by calling Context::setParameter().
  * Finally, call addTorsion() once for each torsion.  After an torsion has been added, you can modify its parameters by calling setTorsionParameters().
  * This will have no effect on Contexts that already exist unless you call updateParametersInContext().
+ * theta is guaranteed to be in the range [-pi,+pi]
  *
  * As an example, the following code creates a CustomTorsionForce that implements a harmonic potential:
  *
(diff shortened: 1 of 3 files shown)


openmmapi/include/openmm/CustomTorsionForce.h
```diff
@@ -63,14 +63,15 @@ namespace OpenMM {
...
+ *
  * This class also has the ability to compute derivatives of the potential energy with respect to global parameters.
  * Call addEnergyParameterDerivative() to request that the derivative with respect to a particular parameter be
  * computed.  You can then query its value in a Context by calling getState() on it.
  *
  * Expressions may involve the operators + (add), - (subtract), * (multiply), / (divide), and ^ (power), and the following
- * functions: sqrt, exp, log, sin, cos, sec, csc, tan, cot, asin, acos, atan, sinh, cosh, tanh, erf, erfc, min, max, abs, floor, ceil, step, delta, select.  All trigonometric functions
- * are defined in radians, and log is the natural logarithm.  step(x) = 0 if x is less than 0, 1 otherwise.  delta(x) = 1 if x is 0, 0 otherwise.
+ * functions: sqrt, exp, log, sin, cos, sec, csc, tan, cot, asin, acos, atan, sinh, cosh, tanh, erf, erfc, min, max, abs, floor, ceil, step, delta, select.
+ * All trigonometric functions are defined in radians, with the torsion `theta` guranteed to be in the range [-pi,+pi].
```
peastman: Could you move this to an earlier point in the description?  This paragraph appears verbatim in the comments of lots of custom forces, and when I anything in it changes, I just copy and paste it.

jchodera:
Why not use an #include there then?

jchodera:
Oh, you couldn't include into a single docstring.

jchodera:
Fixed, thanks!

Should the API docs note the units? They're universally omitted, so it presumably requires reading the User guide to figure out the unit system? Or would it make sense to add a two-line summary of the units to all of these headers?

peastman:
Let's leave it as is for the moment.  The same text gets used for both the C++ and Python API docs, and those handle units in very different ways.

jchodera:
Thanks!

PR #1641, opened 2016-10-18
Change variable 'q' to 'charge' in custom gb models

jchodera:
We just realized we will have to automatically modify energy expressions of `CustomGBForce` forces in YANK to support alchemically-modified GB models that aren't `GBSAOBCForce`.

Previously, the charge variable was `q`, which I fear is going to cause too much trouble due to its non-uniqueness. Changing all of these to `charge` should not affect anything (since these are internal) and will make it easier to recognize and modify these parameters in code that manipulates them (both our `alchemy` and the constant-pH `proton-drive` will need to do this).

More generally, if we could standardize naming of common parameters in `Custom*Force` forces (such as `charge`, `sigma`, and `epsilon`), the tools we build will better be able to support custom forces.

### DIFF
--- a/wrappers/python/simtk/openmm/app/internal/customgbforces.py
+++ b/wrappers/python/simtk/openmm/app/internal/customgbforces.py
@@ -353,13 +353,13 @@ def _createEnergyTerms(force, solventDielectric, soluteDielectric, SA, cutoff, k
     if cutoff is not None:
         params += "; cutoff=%.16g" % cutoff
     if kappa > 0:
-        force.addEnergyTerm("-0.5*138.935485*(1/soluteDielectric-exp(-kappa*B)/solventDielectric)*q^2/B"+params,
+        force.addEnergyTerm("-0.5*138.935485*(1/soluteDielectric-exp(-kappa*B)/solventDielectric)*charge^2/B"+params,
                 CustomGBForce.SingleParticle)
     elif kappa < 0:
         # Do kappa check here to avoid repeating code everywhere
         raise ValueError('kappa/ionic strength must be >= 0')
     else:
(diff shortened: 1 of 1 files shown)


jchodera:
More context on this: https://github.com/choderalab/alchemy/issues/22

jchodera:
Test failures are unrelated to this PR:
```
The following tests FAILED:
	 13 - TestOpenCLCustomExternalForceSingle (SEGFAULT)
	 14 - TestOpenCLCustomExternalForceMixed (OTHER_FAULT)
	 55 - TestOpenCLBrownianIntegratorSingle (OTHER_FAULT)
	119 - TestOpenCLRpmdMixed (Timeout)
Errors while running CTest
Rerunning failing tests...
    Start 14: TestOpenCLCustomExternalForceMixed
1/3 Test #14: TestOpenCLCustomExternalForceMixed ....   Passed    4.86 sec
    Start 55: TestOpenCLBrownianIntegratorSingle
2/3 Test #55: TestOpenCLBrownianIntegratorSingle ....***Exception: SegFault  0.86 sec
    Start 13: TestOpenCLCustomExternalForceSingle
3/3 Test #13: TestOpenCLCustomExternalForceSingle ...***Exception: SegFault  1.87 sec
33% tests passed, 2 tests failed out of 3
Total Test time (real) =   7.59 sec
The following tests FAILED:
	 13 - TestOpenCLCustomExternalForceSingle (SEGFAULT)
	 55 - TestOpenCLBrownianIntegratorSingle (SEGFAULT)
Errors while running CTest
```

peastman:
I think the travis OpenCL is just trying to torment us.

PR #1642, opened 2016-10-18
Fixed bug computing neighbor list

peastman:
This fixes a bug that was introduced in #1564 (about two months ago).


kyleabeauchamp:
Any thoughts about the effects?  E.g. a one-liner description of possible outcome?

peastman:
It occasionally caused interactions to be missed.

PR #1657, opened 2016-11-01
Add Bohrs to units

rmcgibbo:


peastman:
Where did you get the value from?  I just checked several online sources, and all of them report a slightly different value.  Wikipedia (https://en.wikipedia.org/wiki/Bohr_radius) lists it as 0.52917721067.

rmcgibbo:
I don't remember. I'll change it to the value from CODATA.

rmcgibbo:
Do you mind if I add `hartree`s too?

jchodera:
> Do you mind if I add hartrees too?

Sounds like a good idea.

Are there other QM units that would be good to add at this point? Or constants?

rmcgibbo:
There could be, but those were the two things that came to mind for my purposes.

swails:
Whatever happened to this PR?

jchodera:
It is among the forgotten...

peastman:
@rmcgibbo looks like we were waiting for you.  Just we go ahead and merge this?

rmcgibbo:
yeah, +1 to just merge as is.

peastman:
Ok, thanks.  We should go through old PRs and clean them up.

PR #1658, opened 2016-11-03
Fix overflow in dcdfile

rmcgibbo:
For simulations with more than 2**32-1 steps that use a DCDReporter (using a 2fs timestep, this is around 2.2 microseconds), line 125 of dcdfile.py breaks because of an overflow.

```
 File "simtk/openmm/app/dcdreporter.py", line 92, in report
   self._dcd.writeModel(state.getPositions(), mm.Vec3(a[0].value_in_unit(nanometer), b[1].value_in_unit(nanometer), c[2].value_in_unit(nanometer))*nanometer)
 File "/simtk/openmm/app/dcdfile.py", line 125, in writeModel
   file.write(struct.pack('<i', self._firstStep+self._modelCount*self._interval))
struct.error: 'i' format requires -2147483648 <= number <= 2147483647
```

In a shell, you can reproduce this with something like

```
In [17]: import struct

In [18]: struct.pack('<i', 2**32)     
error                                     Traceback (most recent call last)
<ipython-input-18-06a6912b051f> in <module>()
----> 1 struct.pack('<i', 2**32)

error: 'i' format requires -2147483648 <= number <= 2147483647
```

It seems to be that the least invasive way to fix this is just to wrap around.

### DIFF
--- a/wrappers/python/simtk/openmm/app/dcdfile.py
+++ b/wrappers/python/simtk/openmm/app/dcdfile.py
@@ -122,7 +122,7 @@ class DCDFile(object):
         file.seek(8, os.SEEK_SET)
         file.write(struct.pack('<i', self._modelCount))
         file.seek(20, os.SEEK_SET)
-        file.write(struct.pack('<i', self._firstStep+self._modelCount*self._interval))
+        file.write(struct.pack('<i', (self._firstStep+self._modelCount*self._interval) % (2**32-1)))
 
         # Write the data.
 


peastman:
That will totally break the file!  It isn't writing a running counter or anything like that.  It's updating the header to record the amount of data stored in the file.  We can't just wrap that to a smaller value and expect it to work correctly.

This is a kind of complicated issue, and I think it needs some discussion.  Could you open an issue describing the problem?  I can see several possible ways of addressing it, but none of them is obviously right.

rmcgibbo:
> That will totally break the file! It isn't writing a running counter or anything like that. It's updating the header to record the amount of data stored in the file. 

This isn't quite true. The amount of data stored in the file is stored at the offset 8 bytes in. `cd` to `mdtraj/testing/reference` in the MDTraj source tree and try corrupting the bytes at offset 20 for yourself:

```
mcgibbon@Roberts-MacBook ~/projects/mdtraj/mdtraj/testing/reference
$ python -c "b = open('frame0.dcd', 'rb').read(); f = open('frame0-corrupted.dcd', 'wb'); f.write(b[:20]); f.write(b'\x00\x00\x00\x00'); f.write(b[24:])"

mcgibbon@Roberts-MacBook ~/projects/mdtraj/mdtraj/testing/reference
$ python -c 'import mdtraj; print(mdtraj.load("frame0-corrupted.dcd", top="native.pdb"))'
<mdtraj.Trajectory with 501 frames, 22 atoms, 3 residues, and unitcells>
```

PR #1668, opened 2016-11-10
Created Bond class that can store bond type and order

### LINKED ISSUE #1443: Add optional bond order to Topology object?
In working toward making OpenMM more friendly toward small molecules and nonstandard residues, we've run into some nontrivial difficulties with the representation in `Topology` that relies only on connectivity and elemental atomic identities. Given this information alone, it makes it difficult to assign bond orders for parameterizing small molecules or nonstandard residues on the fly. In fact, if some protons are missing, it makes it impossible to do so without knowing the total charge of the residue.

I wonder what people think about adding an (optional) bond order to the `Topology` object bonds. If this information is provided (e.g. when loading through `mol2` or other formats that explicitly provide bond order), this can be used to infer missing protons more reliably even for nonstandard residues, or avoid the use of the Antechamber `bondtype` program to guess bond orders for generating small molecule parameters. Again, this would be fully optional, but could be used by a few file readers (such as `mol2`, `SDF`, `PDBML`, etc.) or by looking up in the [chemical component dictionary](http://ligand-expo.rcsb.org/ld-download.html) to fill in bond orders.

peastman:
Fixes #1443

### DIFF
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -32,35 +32,37 @@ from __future__ import absolute_import
 __author__ = "Peter Eastman"
 __version__ = "1.0"
 
+from collections import namedtuple
 import os
 import xml.etree.ElementTree as etree
 from simtk.openmm.vec3 import Vec3
+from simtk.openmm.app.internal.singleton import Singleton
 from simtk.unit import nanometers, sqrt, is_quantity
 from copy import deepcopy
 
 # Enumerated values for bond type
 
-class Single(object):
+class Single(Singleton):
     def __repr__(self):
         return 'Single'
 Single = Single()
 
-class Double(object):
+class Double(Singleton):
     def __repr__(self):
         return 'Double'
 Double = Double()
 
-class Triple(object):
+class Triple(Singleton):
     def __repr__(self):
         return 'Triple'
 Triple = Triple()
 
-class Aromatic(object):
+class Aromatic(Singleton):
     def __repr__(self):
         return 'Aromatic'
 Aromatic = Aromatic()
 
-class Amide(object):
+class Amide(Singleton):
     def __repr__(self):
         return 'Amide'
 Amide = Amide()
@@ -437,15 +439,15 @@ class Atom(object):
     def __repr__(self):
         return "<Atom %d (%s) of chain %d residue %d (%s)>" % (self.index, self.name, self.residue.chain.index, self.residue.index, self.residue.name)
 
-class Bond(tuple):
+class Bond(namedtuple('Bond', ['atom1', 'atom2'])):
     """A Bond object represents a bond between two Atoms within a Topology.
-    
+
     This class extends tuple, and may be interpreted as a 2 element tuple of Atom objects.
     It also has fields that can optionally be used to describe the bond order and type of bond."""
 
     def __new__(cls, atom1, atom2, type=None, order=None):
         """Create a new Bond.  You should call addBond() on the Topology instead of calling this directly."""
-        bond = tuple.__new__(cls, (atom1, atom2))
+        bond = super(Bond, cls).__new__(cls, atom1, atom2)
         bond.type = type
         bond.order = order
         return bond
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -6,7 +6,7 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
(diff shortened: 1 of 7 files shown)


wrappers/python/simtk/openmm/app/topology.py
```diff
@@ -155,7 +182,7 @@ def addAtom(self, name, element, residue, id=None):
         residue._atoms.append(atom)
         return atom
 
-    def addBond(self, atom1, atom2):
+    def addBond(self, atom1, atom2, type=None, order=None):
```
jchodera: Should *both* `type` and `order` be acceptable?
What if `type` and `order` are both specified, but conflict with each other? Like 
```python
topology.addBond(atom1, atom2, type=Double, order=1)
```
?

wrappers/python/simtk/openmm/app/topology.py
```diff
@@ -404,3 +436,32 @@ def __init__(self, name, element, index, residue, id):
 
     def __repr__(self):
         return "<Atom %d (%s) of chain %d residue %d (%s)>" % (self.index, self.name, self.residue.chain.index, self.residue.index, self.residue.name)
+
+class Bond(tuple):
```
jchodera: Wow, will this really work? So you can still say
```python
for (atom1, atom2) in topology.bonds():
   # do stuff
```
but you could also say
```python
for bond in topology.bonds():
   [atom1, atom2] = bond
   bond_order = bond.order
   bond_type = bond.type
```
Is that right?

peastman: Right.  I tested both those cases to be sure.

swails: What about making it a namedtuple subclass so you can do attribute access on it?

peastman: namedtuple requires that all members are actually part of the tuple.  For backward compatibility, this has to be a two element tuple containing only the two atoms.  And namedtuple doesn't have a per-instance dict, so you can't then add extra fields to it.

jchodera: > Right. I tested both those cases to be sure.
 
I imagine you plan to add the tests before merging, but just wanted to propose the API through this change?

swails: I gave some example code in #1443 that used a namedtuple. You can use it the same way you used tuple here, except it also lets you access the two members via attributes like atom1 and atom2

peastman: It's already being tested.  Both those idioms get used all over the place in Modeller and ForceField.  If either of them broke, there would be tons of test failures.

swails:
One other thing I saw looking through this is that the enumerated bond type options are not singletons when you start using pickle.  Since pickle is used in at least some of the parallel computing libraries to "copy" objects from one memory in one process to memory in another (if memory serves either multiprocessing,  mpi4py, or both), it is probably worth making these instances pickleable -- and making them Singletons under pickle/unpickle, too.

Since this pattern is used in other places as well, I made a `Singleton` subclass and just had the classes that were intended to be singletons inherit from this class.

I also demonstrated what I meant with `namedtuple` there to (hopefully) show that it's the same thing, plus the ability to access atom indices using `atom1` and `atom2` attributes as well.  Just for ideas.

davidlmobley:
I'll try to review this later today. Sorry for the delay -- thanks to anti-Trump protestors pulling a fire alarm during a seminar I was giving and then shutting down freeways, I basically ended up getting no work done yesterday so I'm a little backlogged.

wrappers/python/simtk/openmm/app/topology.py
davidlmobley: @jchodera - see #1443, and specifically https://github.com/pandegroup/openmm/issues/1443#issuecomment-256101964 . The idea was to allow both as it will provide for easier bookkeeping especially in "in between" cases (Chris Bayly's example was formal single and double bonds around pyridine or furan, but associate with the "aromatic" type associated).

I'm not necessarily averse to adding a check that `type` and `order` are compatible when both specified, but this would require a chemistry engine to handle generally (rather than just catching obvious incompatibilities like the one you just noted), so I think it's out for the near term.

davidlmobley:
@peastman : this looks great to me, actually. Thanks for doing it! 

@swails - this would enable the ParmEd `.mol2` readers/writers to preserve bond order when passing through OpenMM, I presume? 

Probably, we should put a test of this type of functionality somewhere (round trip mol2 -> Topology -> mol2, or molecule -> Topology -> molecule), though it likely it does not belong here since it requires either (a) a general chemistry toolkit, or (b) mol2 reading/writing support (as in ParmEd).

peastman:
Thanks!  Is everyone happy with this?  Should I merge it?

swails:
I submitted one more PR to your branch with more Singleton assignments.  I'm not sure if there are more places where that should be done or not...

PR #1669, opened 2016-11-11
Created Bond class that can store bond type and order

### LINKED ISSUE #1443: Add optional bond order to Topology object?

swails:
Fixes #1443 

This builds off of #1668 with a few other features:

1. `Bond` now inherits from `namedtuple` as an example of what I meant (and passing tests will demonstrate that it still behaves the same, hopefully)
1. The bond type classes are now *really* singletons and work with pickling.  This is particularly important to make `Topology` play nicely with multiprocessing and/or mpi4py, if I recall correctly.  It also adds a test for that.

The second point in particular is important, IMO, since it allows you to use the `is` binary operator to compare the `bond.type` to the enumerated options.  This is the natural operator to use (over, for instance, `==`, which wouldn't work correctly upon pickling, either).

The first change here can be scrapped.

### DIFF
--- a/wrappers/python/setup.py
+++ b/wrappers/python/setup.py
@@ -229,8 +229,8 @@ def buildKeywordDictionary(major_version_num=MAJOR_VERSION_NUM,
 
 
 def main():
-    if sys.version_info < (2, 6):
-        reportError("OpenMM requires Python 2.6 or better.")
+    if sys.version_info < (2, 7):
+        reportError("OpenMM requires Python 2.7 or better.")
     if platform.system() == 'Darwin':
         macVersion = [int(x) for x in platform.mac_ver()[0].split('.')]
         if tuple(macVersion) < (10, 5):
--- a/wrappers/python/simtk/openmm/app/__init__.py
+++ b/wrappers/python/simtk/openmm/app/__init__.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import
 __docformat__ = "epytext en"
 
 __author__ = "Peter Eastman"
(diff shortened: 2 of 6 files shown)


swails:
Why did the appveyor tests fail? :frowning_face:

peastman:
I appreciate the help, but please, let's keep just one PR for this!  Right now I'm waiting for feedback from @davidlmobley to find out whether the whole approach I used is even correct.  Also, I'm in a conference all day so I won't have time to do much on this until Monday.

jchodera:
Ouch.

swails:
I submitted this as a PR on your branch (which you could merge to automatically add to your PR).  I almost did that first, but wanted to solicit others' input.

You can see it at peastman/openmm#4

PR #1671, opened 2016-11-17
User-defined atom types for the ForceField class

mzott:
Added keyword arguments to pass in user-defined atom types and atom charges in the ForceField createSystem method. This allows systems to be made that do not have templates as assigning atom types and atom charges from dictionaries eliminates the necessity of a template. 

This is useful for small molecules where templates would have to be created; with these changes, simulations can be made on the fly. Also added these changes in the Modeller class addSolvent method.

One issue I have ran across is that the Modeller class will create a system during the addition of solvent before adding solvent, but it does not return the system. In the case of passing in atom types and charges, using dictionaries for atom types and charges with the new modeller positions and topology will fail as the new water molecules will not be in the dictionaries. Perhaps the system that is created during the addition of solvent molecules could be returned if the atom types and charges are passed in as arguments.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -962,7 +962,7 @@ class ForceField(object):
 
     def createSystem(self, topology, nonbondedMethod=NoCutoff, nonbondedCutoff=1.0*unit.nanometer,
                      constraints=None, rigidWater=True, removeCMMotion=True, hydrogenMass=None, residueTemplates=dict(),
-                     ignoreExternalBonds=False, **args):
+                     ignoreExternalBonds=False, atomTypes=None, atomCharges=None, **args):
         """Construct an OpenMM System representing a Topology with this force field.
 
         Parameters
@@ -998,6 +998,15 @@ class ForceField(object):
             not terminated properly.  This option can create ambiguities where multiple
             templates match the same residue.  If that happens, use the residueTemplates
             argument to specify which one to use.
+        atomTypes : dict
+            A dictionary of atom type strings (e.g. 'ca') where the key is the index of an
+            atom in the input geometry. This allows atom types to be assigned by an external 
+            program rather than relying on template matching. Further, it can allow simulations
+            to be made on the fly with residues that do not have templates.
+        atomCharges : dict
+            A dictionary of point charge values where the key is the index of an atom in the
+            input geometry. This is necessary when a residue does not have a template and thus does
+            not have charges.
         args
             Arbitrary additional keyword arguments may also be specified.
             This allows extra parameters to be specified that are specific to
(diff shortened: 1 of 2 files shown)


peastman:
I was going through PRs and realized no one ever replied to this.  Sorry about that!

We already have one mechanism for dealing with small molecules: template generators.  Does that not meet your needs?  I like to avoid a proliferation of mechanisms, so I don't want to add another if an existing one works just as well.  If there are problems with using a template generator for your purpose, could you describe what they are?  Then we can consider whether this design or something else is the best way of addressing them.

mzott:
I looked into the template generators somewhat, but for what I want to do they seemed a bit bulky. One of the implementations of it (the implementation?) also seemed to need OpenEye which was not ideal. Essentially all I want to do is skip the template matching by passing dictionary or list objects of atom types and charges directly into the system in order to pass information back and forth between a QM package and OpenMM with a program that assigns atom types in between. The change to the Modeller class is  no longer needed as I have been adding solvent molecules through a different script now. Let me know if my understanding of the template generation is incorrect.

jchodera:
I'm in favor of adding ways to make parameter assignment flexible, especially if it has minimal impact on the code. Some questions:
* I'd love to understand your use case a bit more. How are you generating the atom type lists and charges you pass in?
* Why specify the atom types and charges as a `dict` and not a `list` if every particle needs to have defined an atom type and charge?
* Is there utility in being able to specify atom types and charges for some molecules, but use template matches for others?

mzott:
The atom type lists are generated either by parsing an extant atom typing tool's output (Antechamber), or using another atom typing software. Currently, I'm working on making a more general atom typing tool for more than GAFF. I already fairly well replicate the Antechamber assignments, but there is still a lot of work to be done on it in regards to generalization for more than just GAFF.

The charges that are passed in are from Psi4, but the RESP plugin does not yet work with the new release. In regards to object passing as an idea, the primary motivator is that OpenMM is easily called as a library, and with the new Psi4 release, it can also be called as a library. Thus, calling both of them from Python and passing objects rather than text feels like a nice progression. Generating charges and atom types allows QM input (xyz) to be transformed into MM input. 

As far as the use of a `dict` vs `list`, I think the third bullet is the main reason - if you want to create a system using both templates and atom type/charge generation, a `dict` could be more convenient to keep track of the two subsets of systems.

jchodera:
Does that mean you also pass in an entirely new list of atom types, bond definitions, angles, torsions, etc. each time, or are those fixed? I can understand why passing in new charges after a Psi4 call would be useful, but I'm trying to understand your workflow a bit better so as to see why the atom types would need to be dynamically passed in rather than templated. 

If they don't have to be dynamic, what about an easier way of defining a template? Or is the difficulty that you need to update or override the charges for some atoms?

mzott:
Perhaps an easier way to define a template would be helpful, but the primary reason for dynamically passed in atom types is that many of the molecules won't be reused, so the creation of a template seems unnecessary. However, I'm sure having a template would be desirable in some situations. Still, the atom types and charges are assigned quickly, so they could be recomputed without much difficulty. 

I agree that atom types should only be needed to be passed in once per MM simulation. The workflow would be

1. Start with a geometry in xyz format
2. Run charges using Psi4
3. Generate atom types and bonding information
4. Pass these objects in appropriate format to OpenMM and create a simulation
5. Do any calculations you want with OpenMM, then go back and forth between OpenMM and Psi4 using the objects in 1-3 as appropriate

jchodera:
It sounds like you might actually instead want to update charges inside a Context on the fly. There is actually a mechanism to do this via NonbondedForce.updateParametersInContext(). You can create your system, create a Context, do some calculations of energies and forces or integrating positions, then after calling Psi4, update the charges inside NonbondedForce (one of the Force objects in your System) directly, then push those changes to the Context.

If that works for you, we just need to find a better way to get your molecule parameterized through Antechamber and into a template automatically. How are you currently doing this? Are you starting with a Tripos mol2 file in addition to your XYZ file?

mzott:
Currently, I'm not really using Antechamber, but instead the tool I mentioned that I'm working on. As part of the atom type generation, a bonding map is made primarily using covalent radii. This is where the bonding information comes from once it is translated into OpenMM appropriate format. This information is then passed into a script that will make the topology object.

tristanc1:
Sorry to resurrect a somewhat old thread, but I'm interested in somewhat similar functionality, for slightly different reasons. In my case, I'm typically working with one large "master" structure, from which I/the user will create simulations of fragments based on arbitrary selections. At the moment I'm generating the topology afresh for each simulation, but since this is an interactive application it's very important to me that startup time is kept to a minimum - and I'm particularly worried that topology generation will slow down a bit once the selection of available residues grows to the full CHARMM set, for example. A separate issue is that I absolutely have to maintain strict correspondence to PDB residue and atom nomenclature, so to make it bombproof I'll have to be applying some checks to make sure the topology applied to each residue actually matches the one expected from the Chemical Components Dictionary. I think the best approach for me would be to create a persistent master topology of the entire structure so the most time-consuming tasks only have to be done once, then cut out the bits matching the current selection to feed straight to OpenMM. What I'm envisioning is an object that behaves rather like a Numpy array, such that:

```
new_topology = topology[atom_indices]
```
... yields a simulation-ready topology covering the selected atoms.

peastman:
@tristanc1 could you open a new issue on that?  I don't think it's closely related to this PR.

We also need to decide what to do with this PR, either merge it or close it.  @jchodera and @mzott are you agreed on what should be done?  (Not obvious to me from the discussion above.)  Mostly I think these changes look fine, except that it should support arbitrary per-atom attributes instead of just charge.  The question, though, is whether this is actually necessary to address important use cases.  If it is, I'm fine with adding it.  But if we can already do what's needed by other means, I generally prefer not to add new mechanisms.

tristanc1:
@peastman will do.

mzott:
@peastman I would still like to see some sort of functionality similar to what I implemented, although I agree with you that generalizing it to support per-atom attributes would be better. I looked into the method proposed by @jchodera , but I stuck with using my locally altered version of OpenMM with the ability to change the atom types/charges directly. Of course @peastman you have a much better understanding of OpenMM's organization and needs than I, but having more direct control over the simulation properties seems to add utility from my perspective.

mzott:
I am going to close this pull request. I am currently using templates to resolve the issue as recommended above, but I would still like to see a similar change in the future where the atom properties can be changed in an easier fashion.

PR #1679, opened 2016-11-30
Fixes to CustomIntegrator

### LINKED ISSUE #1677: Precision in CustomIntegrator kernels
In GHMC (Metropolized Langevin dynamics) implementations for OpenCL using mixed precision, the following forms of the inner velocity Verlet step have very different 

Version 1 (based on the "velocity Verlet" example code):
```python
        self.addComputePerDof("v", "v + 0.5*dt*f/m")
        self.addComputePerDof("x", "x + v*dt")
        self.addComputePerDof("x1", "x")
        self.addConstrainPositions()
        self.addComputePerDof("v", "v + (x-x1)/dt + 0.5*dt*f/m")
        self.addConstrainVelocities()
```
Version 2:
```python
        self.addComputePerDof("v", "v + 0.5*dt*f/m")
        self.addComputePerDof("x", "x + v*dt")
        self.addComputePerDof("x1", "x")
        self.addConstrainPositions()
        self.addComputePerDof("v", "v + (x-x1)/dt")
        self.addComputePerDof("v", "v + 0.5*dt*f/m")
        self.addConstrainVelocities()
```
Surprisingly, Version 2 has much better acceptance rates than version 1.

Looking through the [code for `CustomIntegrator](https://github.com/pandegroup/openmm/blob/master/platforms/cuda/src/CudaKernels.cpp#L7076-L7097), it looks like there is some cleverness going on with how positions are represented---even in mixed precision.

Any chance we could get a more elaborate explanation of how this works, and what we might be able to do to improve numerical conditioning in these Metropolized integrators?

peastman:
Fixes #1677.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -7109,6 +7109,7 @@ void CudaIntegrateCustomStepKernel::prepareForComputation(ContextImpl& context, 
                 needsForces[step-1] = (needsForces[step] || needsForces[step-1]);
                 needsEnergy[step-1] = (needsEnergy[step] || needsEnergy[step-1]);
                 needsGlobals[step-1] = (needsGlobals[step] || needsGlobals[step-1]);
+                computeBothForceAndEnergy[step-1] = (computeBothForceAndEnergy[step] || computeBothForceAndEnergy[step-1]);
             }
         
         // Loop over all steps and create the kernels for them.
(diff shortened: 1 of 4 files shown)


jchodera:
Good catch!

Can we also change `energy` from `real` to `mixed` as you note in https://github.com/pandegroup/openmm/issues/1677#issuecomment-263327501
and clean up any other minor precision issues there?

peastman:
I already did.

jchodera:
Whoops, thanks!

jchodera:
Travis is [having issues with `osx` builds](https://www.traviscistatus.com/incidents/nq0sf0srvx8s), but it looks like the linux tests pass (except for a stochastic test failure).

PR #1682, opened 2016-12-02
Fixed error building neighbor list with triclinic boxes

### LINKED ISSUE #1666: NaN issue
I tried running 25 ns md with the script from the attached zip ([opennm_nan.zip](https://github.com/pandegroup/openmm/files/583926/opennm_nan.zip)) on several frames from a gromacs md run (which have triclinic box vectors). However, a lot of runs (7 out of 9) stop because of ```Particle coordinate is nan```, while gromacs runs them just fine. 

This seems to happen at random times (ranging from 0.28 ns to 24.68 ns).

I checked the forces between using gromacs directly and using the gromacs forcefields in OpenMM, and the total force per atom matches pretty well between them.

I also tried running it with the Langevin integrator and with some changes that were suggested to me by @jchodera :
- Setting ```integrator.setConstraintTolerance``` to ```(1.0e-8)```.
- Using ```OpenCL```.
- Running with ```properties ={'CudaDisablePmeStream':'true'}```.

But all of these still generate a bunch of NaNs as seen in this table:
table

I'm running this on:
GPU: NVIDIA GTX Titan X (Maxwell)
Driver Version: 352.79
CUDA Version: 7.5.18
Openmm Version: 7.1.0.dev-04b1a06

Does someone has an idea what might be causing this?

peastman:
Hopefully fixes #1666.  Try this out and see if it makes your nans go away.

### DIFF
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -513,6 +513,8 @@ void CudaNonbondedUtilities::createKernelsForGroups(int groups) {
         defines["NUM_TILES_WITH_EXCLUSIONS"] = context.intToString(exclusionTiles->getSize());
         if (usePeriodic)
             defines["USE_PERIODIC"] = "1";
+        if (context.getBoxIsTriclinic())
+            defines["TRICLINIC"] = "1";
         defines["MAX_EXCLUSIONS"] = context.intToString(maxExclusions);
         defines["MAX_BITS_FOR_PAIRS"] = (canUsePairList ? "2" : "0");
         CUmodule interactingBlocksProgram = context.createModule(CudaKernelSources::vectorOps+CudaKernelSources::findInteractingBlocks, defines);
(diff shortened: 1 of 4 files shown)


sroet:
> Try this out and see if it makes your nans go away.

I don't know if this was directed at me, but I can't install openmm and all the dependencies from source (because I don't have sudo rights on my machine). Would it be possible to merge this into the openmm-dev conda channel, making it possible for me to test it on my system?

jchodera:
@sroet : I've already updated the `dev` conda release for you to try.
```bash
# downgrade
conda install --yes -c omnia openmm==7.0.1
# clean to make sure we pick up fresh dev package
conda clean -plti --yes
# upgrade
conda install --yes -c omnia/labels/dev openmm==7.1.0
```

peastman:
No, openmm-dev isn't a separate repository.  It doesn't pick it up until it gets merged into the master repository.

You don't need sudo rights.  You can install it to any directory you want.

jchodera:
Note that we deprecated the `openmm-dev` repo 7 months ago. I've just deleted those packages to make sure nobody is trying to use them accidentally.

@peastman : I'm temporarily building from your branch to push the `dev` label builds so we can test that fix. The recipe is here:
https://github.com/omnia-md/conda-dev-recipes/blob/master/openmm/meta.yaml#L6-L7

sroet:
@jchodera Thanks! I meant the ```conda install --yes -c omnia/labels/dev openmm==7.1.0``` when I  said ```openmm-dev```. 

It seems to run fine on the NVE simulations (running 6 times 20000) and I will run the original 25 ns NPT simulations tonight.

sroet:
This build, however, doesn't recognizes the CUDA (7.5) platform on my machine, I will run it on the OpenCL platform in the meantime.

jchodera:
Whoops, this version is built with CUDA 8, sorry about that!

sroet:
So far, so good. 6/9 runs have completed the 25 ns runs without NaNs and the last 3 are halfway there. (should be done in 3.5 hours)

sroet:
It seems like this fixes my issue, I ran 9 simulations of 25 ns without issues on OpenCL. Thanks!

peastman:
Great, thanks!

PR #1683, opened 2016-12-02
Try to make checkpoint reporting atomic

mpharrigan:
Previously, re-starting a simulation would overwrite the existing checkpoint file *at CheckpointReporter creation time*. If your simulation was killed before you wrote a new checkpoint, you'd be out of luck.

With this change, passing in a file object still works the same way. If you pass a filename, the file object will be opened and closed at the time of checkpointing. If you pass an optional new parameter `tempFile`, we'll write to this file instead, and use `os.replace` to atomically overwrite the previous checkpoint. (this part requires python 3)

### DIFF
--- a/wrappers/python/simtk/openmm/app/checkpointreporter.py
+++ b/wrappers/python/simtk/openmm/app/checkpointreporter.py
@@ -153,7 +153,3 @@ class CheckpointReporter(object):
             self._reportFilePath(chk)
         else:
             self._reportFileObject(chk)
-
-    def __del__(self):
-        if self._out is not None:
-            self._out.close()
--- a/wrappers/python/simtk/openmm/app/checkpointreporter.py
+++ b/wrappers/python/simtk/openmm/app/checkpointreporter.py
@@ -29,10 +29,12 @@ OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 from __future__ import absolute_import
+
+import os
+
 __author__ = "Robert McGibbon"
 __version__ = "1.0"
 
-import simtk.openmm as mm
 __all__ = ['CheckpointReporter']
 
 
@@ -68,7 +70,8 @@ class CheckpointReporter(object):
     throwing an exception.
 
     """
-    def __init__(self, file, reportInterval):
+
+    def __init__(self, file, reportInterval, tempFile=None):
         """Create a CheckpointReporter.
 
         Parameters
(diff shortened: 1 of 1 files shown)


peastman:
It looks like we were both fixing the same problem at the same time, and sent PRs one minute apart!  See #1684.

mpharrigan:
oh my! I was complaining about this yesterday and you told me to fix it myself :) I must have gotten to you

mpharrigan:
My brief sales pitch for this PR over #1684:

Mine doesn't have to worry about buffering or calling truncate, as it opens and closes the file only during writing. It uses a context manager to safely close the file even if there's an exception. This PR has an extra feature where you can give it a second filename to use during writing, and then use an atomic operation to replace the old checkpoint in case somethings goes wrong during checkpoint writing.

End sales pitch. I won't be (too) crushed if you pick your fix instead :)

peastman:
I agree that doing a safe save is a good idea.  But there's no need to make the user specify a second filename (we can just add '.backup' to the existing name or something like that), and we can also do it in a way that works on Python 2.  Let me see what I can do.

mpharrigan:
>  But there's no need to make the user specify a second filename (we can just add '.backup' to the existing name or something like that)

Sure

> we can also do it in a way that works on Python 2

You can just use `os.rename` on POSIX, but things get weird on windows

>  Let me see what I can do.

Great, I'll let you take over

peastman:
> You can just use os.rename on POSIX, but things get weird on windows

The trick is to do a series of renames:

1. Write the new file as `.backup1`.
2. If there's an existing file, rename it to `.backup2`.
3. Rename `.backup1` to the correct name.
4. Delete `.backup2`.

mpharrigan:
That sounds like it will work. I was curious about how the python 3 did it and found the patch http://bugs.python.org/issue8828

swails:
You can also use the tempfile module in the stdlib to write a temporary file. It's pretty good.

mpharrigan:
> You can also use the tempfile module in the stdlib to write a temporary file. It's pretty good.

Wouldn't it be worse if you're writing to `/tmp` and renaming to a path that's stored on a different filesystem (scratch or networked on clusters)

swails:
> Wouldn't it be worse if you're writing to /tmp and renaming to a path that's stored on a different filesystem (scratch or networked on clusters)

I don't know -- do you have any way of testing one way vs. the other in situations like that?

``tempfile`` is portable and (when used properly) guarantees that you won't be overwriting an existing temporary file, which was the primary reason I suggested it.  You can also use environment variables (like `TMP` or `TEMP`) to determine which directory to use as the temporary storage location (which could help get around filesystem issues).

peastman:
`tempfile` doesn't do what we want.  It creates a file that disappears as soon as you close it.  It also isn't visible to the rest of the file system unless you use `NamedTemporaryFile`, and even then it doesn't guarantee that any other process can successfully read it.  The whole point here is to create a real file, not a temporary one, so that if anything goes wrong you'll be left with a real file on your filesystem.

swails:
> It creates a file that disappears as soon as you close it.

I've never used `tempfile` this way (i.e., I've never used `TemporaryFile`). All I've ever used is `tempfile.mkstemp`, which explicitly says the user is required to destroy the temporary file afterwards.

I've used this on many occasions, and it's always created a temporary file that I could access outside of the Python session that created it.

peastman:
This was implemented in #1684.

Issue #1645, opened 2016-10-20
Checkpoints


mpharrigan:
This goes along with the discussion in #1610 . 

A common use case for a *user* of openmm is to run a big ol' simulation on a cluster. When the job runs out of time, it writes a checkpoint from whence you can continue simulating. I am finding the UX for this very bad:

(1) The checkpoint only stores positions and velocities. You have to make sure you have your topology (pdb) and an unmodified copy of your python driver script to re-create the system, integrators, and reporters each time. This is especially annoying because loading from the checkpoint involves mucking with your python driver script. Don't mess up any of the system building!

(2) DCDReporter clobbers previous instances of the trajectory! It should append. 

(3) StateDataReporter and the `Simulation` object has no idea what step it's on. If you want a 100ns trajectory, you need to keep track of all the previous fragments to determine how much more simulation to do. More logic in the driver script! StateDataReporter can't give meaningful %done estimates

Amber and gromacs have better UX with regards to checkpointing. `gmx mdrun -cpi`  does everything.

jchodera:
This certainly sounds useful!

What about using something like the HDF5 trajectory format, augmented with serialized versions of the other things that OpenMM needs (System, Integrator)?

mpharrigan:
Additional gripe:

(4) OpenMM overwrites the checkpoint on construction of the checkpoint reporter, *not* immediately prior to writing a checkpoint. This means when you restart a simulation, you better hope your job makes it to the checkpoint writing stage or you've lost your ability to restart

epretti:
There are currently at least two remaining issues with checkpoints in OpenMM from a user experience standpoint:

- `CheckpointReporter` overwrites checkpoints atomically, but `simulation.saveCheckpoint()` does not, so when using it to accomplish custom logic beyond what `CheckpointReporter` does, or when it is called by `runForClockTime()`, there is a possibility that the checkpoint will be corrupted or lost if the Python process is interrupted while overwriting it.
[...]

PR #1684, opened 2016-12-02
CheckpointReporter does not create output file until first report

peastman:
Fixes one part of #1645.


peastman:
This now does a safe save, as suggested in #1683.

mpharrigan:
Somewhat general question: what's the difference between mode `w+b` vs `wb`? Is `w+b` truly what we want here?

mpharrigan:
After reading http://stackoverflow.com/questions/16208206/confused-by-python-file-mode-w I think I understand (`w+b` is if you also want to read), and I think we could safely use either here (maybe `wb` is more idiomatic)

PR #1686, opened 2016-12-08
Automate Windows builds

peastman:
I'm making progress on this.  Slowly.  And very painfully.  And why does nothing ever work?  And did I mention how much I hate Windows?  It's bad for my health.

It's not yet fully automated, but it's a lot closer than before.  I'm using Vagrant to create a VM.  Then I log into it, clone the OpenMM repository inside it, and run the build script to do the build.

I've never managed to get sphinx to build pdfs on Windows.  See https://github.com/pandegroup/openmm/issues/1362#issuecomment-187351460.  And I don't think it's worth worrying about that right now.  So for this release I'll probably just disable building documentation, then copy over a docs directory from one of the other builds.

Can we drop support for Python 3.4 on Windows?  To build for Python 3.5, we use Visual Studio 2015 which I can get with a simple `choco install vcbuildtools`.  But Python 3.4 needs Visual Studio 2010, which is much harder to install.  *Much* harder.

I've discovered that choco is awesome when it works, which unfortunately isn't anywhere close to 100% of the time.  It fails unpredictably and erratically in various ways.  Something will be working perfectly, and then stop working for no apparent reason.  Why does nothing ever work?

Did I mention how much I hate Windows?

### DIFF
--- a/devtools/packaging/scripts/windows/Vagrantfile
+++ b/devtools/packaging/scripts/windows/Vagrantfile
@@ -1,0 +1,10 @@
+Vagrant.configure("2") do |config|
+  config.vm.box = "gusztavvargadr/windows10ee"
+  config.vm.provision :shell, path: "prepare.ps1"
+
+  config.vm.provider :virtualbox do |vb|
+    vb.customize ["modifyvm", :id, "--memory", "2048"]
+    vb.customize ["modifyvm", :id, "--cpus", "2"]
+    vb.customize ["modifyvm", :id, "--ioapic", "on"]
+  end  
+end
--- a/devtools/packaging/scripts/windows/build.bat
+++ b/devtools/packaging/scripts/windows/build.bat
@@ -1,16 +1,13 @@
-REM "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
-
-cd openmm
 mkdir build
 cd build
 
-set FFTW=C:\Program Files\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
+set FFTW=C:\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
 "C:\Program Files\CMake\bin\cmake.exe" .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DOPENMM_GENERATE_API_DOCS=ON -DFFTW_INCLUDES="%FFTW%/include" -DFFTW_LIBRARY="%FFTW%/lib/libfftw3f-3.lib"
 
 jom
 jom PythonInstall
 jom C++ApiDocs
 jom PythonApiDocs
-# jom sphinxpdf
+REM jom sphinxpdf
 jom install
 jom PythonBdist
--- a/devtools/packaging/scripts/windows/build.bat
+++ b/devtools/packaging/scripts/windows/build.bat
@@ -1,0 +1,16 @@
+REM "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
+
+cd openmm
+mkdir build
+cd build
+
+set FFTW=C:\Program Files\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
+"C:\Program Files\CMake\bin\cmake.exe" .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DOPENMM_GENERATE_API_DOCS=ON -DFFTW_INCLUDES="%FFTW%/include" -DFFTW_LIBRARY="%FFTW%/lib/libfftw3f-3.lib"
+
+jom
+jom PythonInstall
+jom C++ApiDocs
+jom PythonApiDocs
+# jom sphinxpdf
+jom install
+jom PythonBdist
--- a/devtools/packaging/scripts/windows/build.bat
+++ b/devtools/packaging/scripts/windows/build.bat
@@ -2,7 +2,10 @@ mkdir build
 cd build
 
 set FFTW=C:\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
-"C:\Program Files\CMake\bin\cmake.exe" .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DOPENMM_GENERATE_API_DOCS=ON -DFFTW_INCLUDES="%FFTW%/include" -DFFTW_LIBRARY="%FFTW%/lib/libfftw3f-3.lib"
+set APPSDK=C:\Program Files (x86)\AMD APP SDK\2.9-1
+"C:\Program Files\CMake\bin\cmake.exe" .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DOPENMM_GENERATE_API_DOCS=ON ^
+    -DOPENCL_INCLUDE_DIR="%APPSDK%\include" -DOPENCL_LIBRARY="%APPSDK%\lib\x86_64\OpenCL.lib" ^
+    -DFFTW_INCLUDES="%FFTW%/include" -DFFTW_LIBRARY="%FFTW%/lib/libfftw3f-3.lib"
 
 jom
 jom PythonInstall
--- a/devtools/packaging/scripts/windows/prepare.ps1
+++ b/devtools/packaging/scripts/windows/prepare.ps1
@@ -9,6 +9,11 @@ choco install -y doxygen.portable swig cmake doxygen.install vcbuildtools git jo
 wget https://developer.nvidia.com/compute/cuda/8.0/prod/network_installers/cuda_8.0.44_win10_network-exe -UseBasicParsing -OutFile cuda_8.0.44_win10_network.exe
 .\cuda_8.0.44_win10_network.exe -s compiler_8.0 cudart_8.0 cufft_8.0 cufft_dev_8.0 nvrtc_8.0 nvrtc_dev_8.0 | Out-Null
 
+# Install AMD APP SDK.
+
+wget https://s3.amazonaws.com/omnia-ci/AMD-APP-SDK-v2.9-1.599.381-GA-Full-windows-64.exe -UseBasicParsing -OutFile AMD-APP-SDK-v2.9-1.599.381-GA-Full-windows-64.exe
+.\AMD-APP-SDK-v2.9-1.599.381-GA-Full-windows-64.exe /S /v/qn | Out-Null
+
 # Install Miniconda.
 
 wget https://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86_64.exe -UseBasicParsing -OutFile Miniconda3-latest-Windows-x86_64.exe
--- a/devtools/packaging/scripts/windows/prepare.ps1
+++ b/devtools/packaging/scripts/windows/prepare.ps1
@@ -1,0 +1,20 @@
+cd C:\Users\vagrant
+
+# Install everything we can with choco.
+
+choco install -y doxygen.portable swig cmake doxygen.install vcbuildtools git jom
+
+# Install CUDA.
+
+wget https://developer.nvidia.com/compute/cuda/8.0/prod/network_installers/cuda_8.0.44_win10_network-exe -UseBasicParsing -OutFile cuda_8.0.44_win10_network.exe
+.\cuda_8.0.44_win10_network.exe -s compiler_8.0 cudart_8.0 cufft_8.0 cufft_dev_8.0 nvrtc_8.0 nvrtc_dev_8.0 | Out-Null
+
+# Install Miniconda.
+
+wget https://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86_64.exe -UseBasicParsing -OutFile Miniconda3-latest-Windows-x86_64.exe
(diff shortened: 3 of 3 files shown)


mpharrigan:
I think we already agreed to do python 3.5 only for windows (?)

peastman:
For 7.0.1 we had versions for Python 3.3, 3.4, and 3.5.

peastman:
I'm trying to model the build script after the existing ones for other platforms, but I'm a bit confused.  We seem to have three independent sets of nearly identical build scripts in different places:  `omnia-md/conda-recipes`, `omnia-md/conda-dev-recipes`, and in this repo under `devtools/packaging/scripts`.  Do we really need all of those?  Are they for different things?

I was also wondering why we bother building documentation in the conda recipe?  It's just going to get buried somewhere inside `anaconda/pkgs/openmm-8.0-py35_0` or somewhere equally hidden where no one will ever think to look for it.

jchodera:
Sorry for the lack of prompt replies! Visiting Bernie Brooks's group at the NIH today.


> I was also wondering why we bother building documentation in the conda recipe? It's just going to get buried somewhere inside anaconda/pkgs/openmm-8.0-py35_0 or somewhere equally hidden where no one will ever think to look for it.

You're right that it's not essential, but I thought it would be useful to have the local docs if someone wanted them! They're installed in `$MINICONDA/docs/openmm/`:
```
[LSKI1497:~] choderaj% ls -tlr ~/miniconda/docs/openmm/
total 3192
-rw-rw-r--  2 choderaj  9696963      172 Nov 15 23:34 Python API Reference.html
-rw-rw-r--  2 choderaj  9696963      169 Nov 15 23:34 C++ API Reference.html
-rw-rw-r--  2 choderaj  9696963   252861 Nov 15 23:41 OpenMMDeveloperGuide.pdf
-rw-rw-r--  2 choderaj  9696963  1369691 Nov 15 23:41 OpenMMUsersGuide.pdf
```

jchodera:
> I'm trying to model the build script after the existing ones for other platforms, but I'm a bit confused. We seem to have three independent sets of nearly identical build scripts in different places: omnia-md/conda-recipes, omnia-md/conda-dev-recipes, and in this repo under devtools/packaging/scripts. Do we really need all of those? Are they for different things?

They are all different things! I'm happy to document these if you can tell me where it would be best to put a detailed description. 

Summary:
* `omnia-md/conda-recipes/openmm/` - This is the production recipe for the official `omnia` release build. The recipe is only built once and pushed to the `pre` label, then manually moved to `main` manually. This is where the 7.0.1 build script lives.
* `omnia-md/conda-dev-recipes/` - These builds are done nightly, and pushed to the `dev` label. This is where we have been experimenting with the 7.1.0 build.
* `devtools/packaging/scripts/` in the openmm repo: These are the script to automate building the ZIP installers. These builds used to happen on Jenkins, but now we've updated these scripts to build within the `jchodera/omnia-build-box` docker image so that the compiler and library dependencies will be the same across conda and ZIP installers.

I realize this is pretty complicated, and I'd love to find ways to simplify it, but it's been a bit difficult to keep it simpler with multiple mandated build targets and the desire to retool the build environment requirements each release.

peastman:
Is the intention that after each new release, the script from `conda-dev-recipes` will be copied over to `conda-recipes`, which then serves just as a record of the most recent version?  Or are they maintained independently?

```
[LSKI1497:~] choderaj% ls -tlr ~/miniconda/docs/openmm/
total 3192
-rw-rw-r--  2 choderaj  9696963      172 Nov 15 23:34 Python API Reference.html
-rw-rw-r--  2 choderaj  9696963      169 Nov 15 23:34 C++ API Reference.html
-rw-rw-r--  2 choderaj  9696963   252861 Nov 15 23:41 OpenMMDeveloperGuide.pdf
-rw-rw-r--  2 choderaj  9696963  1369691 Nov 15 23:41 OpenMMUsersGuide.pdf
```

It looks like you're missing most of the html files?

jchodera:
> Is the intention that after each new release, the script from conda-dev-recipes will be copied over to conda-recipes, which then serves just as a record of the most recent version? Or are they maintained independently?

They're maintained independently. The `conda-dev-recipes` one continues to be built nightly from `master` most of the time. We copy over the relevant material to `conda-recipes/openmm/` when we want to generate a release candidate build in the `pre` label and then move it manually to the `main` label.

> It looks like you're missing most of the html files?

Whoops! We should fix that.

jchodera:
Or just remove the docs if we decide we don't want them anyway.

peastman:
Did you know that on Windows, `conda build` will fail if you installed miniconda into a path that has a space in it?  Such as `C:\Program Files`?  Which happens to be the default location it installs to?

Because *nothing EVER works*.

I've got it to a point where `conda build` runs for a while, doing things that look reasonable.  It installs a bunch of dependencies, then checks out OpenMM from github.  But then I get the following:

```
Package: openmm-7.1.0-py35_0
source tree in: C:\Miniconda3\conda-bld\vagrant_1481246146330\work
number of files: 0
Fixing permissions
Fixing permissions
updating index in: C:\Miniconda3\conda-bld\win-64
updating: openmm-7.1.0-py35_0.tar.bz2
TEST START: openmm-7.1.0-py35_0
updating index in: C:\Miniconda3\conda-bld\win-64
updating index in: C:\Miniconda3\conda-bld\noarch

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    openmm-7.1.0               |           py35_0           4 KB  local

The following NEW packages will be INSTALLED:

    fftw3f:         3.3.4-vc14_2  omnia [vc14]
    openmm:         7.1.0-py35_0  local
    pip:            9.0.1-py35_1
    python:         3.5.2-0
    setuptools:     27.2.0-py35_1
    vs2015_runtime: 14.0.25123-0
    wheel:          0.29.0-py35_0


C:\Miniconda3\conda-bld\vagrant_1481246146330\test_tmp>call "C:\Miniconda3\Scripts\activate.bat" "C:\Miniconda3\conda-bld\vagrant_1481246146330\_t_env"  1>nul 2>&1

(C:\Miniconda3\conda-bld\vagrant_1481246146330\_t_env) C:\Miniconda3\conda-bld\vagrant_1481246146330\test_tmp>if errorlevel 1 exit 1

(C:\Miniconda3\conda-bld\vagrant_1481246146330\_t_env) C:\Miniconda3\conda-bld\vagrant_1481246146330\test_tmp>C:\Miniconda3\conda-bld\vagrant_1481246146330\_t_env\python.exe -s C:\Miniconda3\conda-bld\vagrant_1481246146330\test_tmp\run_test.py
===== testing package: openmm-7.1.0-py35_0 =====
import: 'simtk'
  File "C:\Miniconda3\conda-bld\vagrant_1481246146330\test_tmp\run_test.py", line 27, in <module>
    import simtk
ImportError: No module named 'simtk'
```

Why is it trying to install OpenMM from conda?  I'm *building* OpenMM.  It's certainly not listed as a dependency in meta.yaml, and it shouldn't be a dependency for anything else that's listed there.

jchodera:
> Why is it trying to install OpenMM from conda? I'm building OpenMM. It's certainly not listed as a dependency in meta.yaml, and it shouldn't be a dependency for anything else that's listed there.

Wow, that's weird. It's built everything, and has started the test phase. It also has figured out it wants to install the local version you just built. But somehow it also wants to download an `openmm` package? I don't understand that.

peastman:
It hasn't actually built anything.  All it's done is check out the source code.  So perhaps that indicates a problem with the build script.

Do I need to do anything other than put a file called `build.bat` in the same directory with `meta.yaml`?

jchodera:
That should do the trick. Can you build any other recipes?

peastman:
Here's the complete content of my meta.yaml.  Does it look ok to you?

```yaml
package:
  name: openmm
  version: 7.1.0

source:
  git_url: https://github.com/pandegroup/openmm.git
  git_tag: master

build:
  number: 0
  skip: False # [win]

requirements:
  build:
    # on windows, need to install cmake manually
    - cmake     [not win]
    - jom       [win]
    - python
    # Pin fftw3f to 3.3.3 to work around OSX py34 issues
    - fftw3f ==3.3.3 [not win]
    - fftw3f ==3.3.4 [win]
    # swig is pinned to use omnia swig 3.0.8
    - swig ==3.0.8
    # on osx, need to install doxygen manually
    - doxygen   [not osx]
    # for building docs
    - sphinx
    - sphinxcontrib-bibtex
    - sphinxcontrib-lunrsearch
    - sphinxcontrib-autodoc_doxygen
    - lxml

  run:
    - python
    - fftw3f

test:
  requires:
    - python
  imports:
    - simtk
    - simtk.openmm
  commands:
    - python -m simtk.testInstallation

about:
  home: http://openmm.org
  license: LGPL and MIT
  summary: A high performance toolkit for molecular simulation.
```

I tried putting `echo "Running the build script!"` as the first line of the build script, and nothing gets printed.  That suggests it's not getting executed.

mpharrigan:
It's bld.bat

Also note that it's not downloading openmm 7.1, it's installing it from the "local" channel, aka the package it thinks it just built (but didn't)

peastman:
Brilliant!

Yes, that was the problem. :)

mpharrigan:
I've heard tell that ms people named it "program files" to force devs to make their programs handle spaces in paths. [On the whole, it has worked. spaces in filenames in linux is a headache]

peastman:
Well, it went all the way through building this time, and created the bdist.  And then it failed with the same error as before.

mpharrigan:
What does your bld.bat look like?

jchodera:
@peastman : Maybe you can check in your changes so we can see what you're working with and what is failing?

peastman:
I've checked in my latest updates.  If you want to try it, here are the steps I'm currently following.

The build script I've checked in is for building the zip libraries.  The conda version is almost but not quite identical.  In `devtools/packaging/scripts/windows` create a file called `meta.yaml` with the content I posted above, and one called `bld.bat` with this content:

```bat
mkdir build
cd build

set FFTW=C:\Miniconda3\pkgs\fftw3f-3.3.4-vc14_2\Library
"C:\Program Files\CMake\bin\cmake.exe" .. -G "NMake Makefiles JOM" -DCMAKE_INSTALL_PREFIX=%PREFIX% -DCMAKE_BUILD_TYPE=Release -DOPENMM_GENERATE_API_DOCS=ON -DFFTW_INCLUDES="%FFTW%/include" -DFFTW_LIBRARY="%FFTW%/lib/libfftw3f-3.lib"

jom
jom PythonInstall
jom C++ApiDocs
jom PythonApiDocs
REM jom sphinxpdf
jom install
jom PythonBdist
```

Now follow these steps:

1. cd to `devtools/packaging/scripts/windows`.
2. Type `vagrant up` to start up a VM.
3. When it finishes, run VirtualBox, select the running VM, and click "Show".
4. For some reason that I haven't figured out, neither conda nor git is initially in your path.  They're *supposed* to have been added automatically, but they aren't.  The workaround I've found is to open a command window, `choco uninstall git`, then immediately `choco install git`, then close the window.  From that point on, both conda and git will be in your path in any windows you open after that.
5. Select Start->Visual Studio 2015->VS2015 x64 Native Tools Command Prompt.
6. `cd C:\Users\vagrant`
7. `conda build -c omnia C:\vagrant`

peastman:
Found the problem.  I needed to do a `jom install` before `jom PythonInstall`.

peastman:
For Linux, you put a copy of the AMD APP SDK on a server where we can download it.  Is there a Windows one there too?

jchodera:
>
> For Linux, you put a copy of the AMD APP SDK on a server where we can
> download it. Is there a Windows one there too?
>

Oh, not yet. I can do that if it would be helpful, but would have to be
tonight---on a plane now with poor bandwidth.

Just let me know!

peastman:
That would be great.

peastman:
I think that's the last thing holding back this PR.  As soon as you get it posted, I can commit the final changes to the script and start building binaries.  Perhaps we can actually get the beta out this week.

peastman:
@jchodera have you had a chance to post it?  I think that's all that's holding up the beta now.

jchodera:
Sorry about this. There hasn't been time with good internet access, but let me see if I can do this now!

jchodera:
Oh, it's only 4MB! Give me just a moment---I can do this from the plane now.

jchodera:
It's now up at [this URL](https://s3.amazonaws.com/omnia-ci/AMD-SDK-InstallManager-v1.4.87-1.exe). 

Sorry again---all my travel and grant obligations end after Fri, so I will have much more spare bandwidth to clean things up then.

peastman:
I don't think that's the right file.  For 2.9.1 the installer is 118 MB.  For 3.0 instead of posting that, they post a 4 MB "Install Manager".  You can then use that to download the actual installer, which is 162 MB.  But there's no way to do that from a script.  It's strictly a GUI application.

I think the plan was to stick with 2.9.1, right?

jchodera:
Whoops, sorry about that. Fixing now.

jchodera:
OK, I won't be able to upload this from the airplane. Modern technology is not yet that advanced.

I've made a note to do this as soon as I get home this evening.

jchodera:
Try [this URL](https://s3.amazonaws.com/omnia-ci/AMD-APP-SDK-v2.9-1.599.381-GA-Full-windows-64.exe).

peastman:
Thanks!  Some day we'll have post-modern technology, and then this won't be a problem.

Issue #1700, opened 2016-12-26
Deployment of latest docs to S3 failing


jchodera:
I just noticed that the documentation for the current `master` branch, which is supposed to be built and pushed to S3 by travis (and accessible at the bottom of http://openmm.org/documentation.html under "The Bleeding Edge"), seems to not be working.

This is currently built by the [`2.7_with_system_python_packages` travis branch](https://github.com/pandegroup/openmm/blob/master/.travis.yml#L84-L92), which is [currently failing](https://travis-ci.org/pandegroup/openmm/jobs/186584272#L5963-L5973) because it is unable to build the C++ API docs---something I've run into earlier with OS X ZIP installer builds. Not sure what's wrong here, but it's probably something we'll need to fix.

peastman:
This is because of https://github.com/rmcgibbo/sphinxcontrib-autodoc_doxygen/pull/17.  @rmcgibbo we really need you to create a new pypi release.

rmcgibbo:
I just cut a new version to pypi. I'll add a conda-forge this evening.

jchodera:
Thanks! I think the build system uses the `pypi` version, so we should be good!

jchodera:
@peastman : Can you retrigger the build of the master branch on travis?

peastman:
Thanks!  It's running now.

peastman:
It seems to have fixed that problem, but it's still reporting an error building documentation.  I'm not sure why.  https://travis-ci.org/pandegroup/openmm/jobs/186584272#L6406

jchodera:
It looks like the docs built fine, but the S3 access key is not valid:

Deploying application
uploading "userguide/application.html" with {:content_type=>"text/html"}
Invalid S3 Access Key Id, Stopping Deploy
failed to deploy

Who is in charge of the OpenMM.org S3 bucket?

rmcgibbo:
It's in a bit of limbo...

jchodera:
I think we can issue multiple sets of credentials, but *someone* must have the keys right now, right?

I'm happy to take over if we need someone to do so (or to take over billing).

pandelab:
Hi Everyone,
What are you looking for access to? Is it the AWS account? I do have password information to account?
Theresa

mpharrigan:
cc https://github.com/pandegroup/openmm-org/pull/13

jchodera:
Resolved! See #1835

PR #1704, opened 2017-01-03
Update travis s3 credentials

mpharrigan:
cc #1700


jchodera:
Thanks!

jchodera:
Can we merge this, @peastman? It looks like the failure is #1714

PR #1711, opened 2017-01-06
Added more error checking to constraints

### LINKED ISSUE #1699: Context.setVelocitiesToTemperature() produces  large or`nan` velocities
I've been testing the `7.1.0` `beta` conda package (and `7.2.0` `dev` build) with `miniconda` python, and have run into a case where simply calling `context.setVelocitiesToTemperature(300 * unit.kelvin)` produces velocities of extremely large magnitude or `nan`. I've checked to make sure the masses were reasonable. A test case is attached.

I worry that this may be due to the `clang`-compiled packages, so I'll switch the `dev` builds back to `gcc` temporarily (and leave the `beta` untouched) to be able to compare.

Here is some truncated example output showing what the large velocities look like:
```
Reference
velocities before setVelocitiesToTemperature:
[[ 0.  0.  0.]
 [ 0.  0.  0.]
 ...
[ 0.  0.  0.]] nm/ps
velocities after setVelocitiesToTemperature:
[[  6.29332748e+45   2.95952189e+46   6.99333855e+44]
 [  8.88686047e+44  -3.26026826e+45   4.30704692e+45]
...
 [  1.87835349e+00  -3.08385810e-01  -9.40770089e-01]] nm/ps
CPU
velocities before setVelocitiesToTemperature:
[[ 0.  0.  0.]
 [ 0.  0.  0.]
...
 [ 0.  0.  0.]] nm/ps
velocities after setVelocitiesToTemperature:
[[  5.87550066e+45   2.76303321e+46   6.52903656e+44]
 [  8.25422226e+44  -3.04926672e+45   4.36184621e+45]
...
 [  7.21309559e-01  -6.82054253e-01   1.04775292e+00]] nm/ps
CUDA
velocities before setVelocitiesToTemperature:
[[ 0.  0.  0.]
 [ 0.  0.  0.]
...
 [ 0.  0.  0.]] nm/ps
velocities after setVelocitiesToTemperature:
[[        inf         inf         inf]
 [        nan         nan         nan]
...
 [-0.42207414 -1.61906338  0.88288629]] nm/ps
OpenCL
velocities before setVelocitiesToTemperature:
[[ 0.  0.  0.]
 [ 0.  0.  0.]
...
 [ 0.  0.  0.]] nm/ps
velocities after setVelocitiesToTemperature:
[[        nan         nan         nan]
 [        nan         nan         nan]
...
 [-1.45576859 -0.17318904 -3.09035373]] nm/ps
```
[velocities-bug.zip](https://github.com/pandegroup/openmm/files/672492/velocities-bug.zip)

peastman:
Fixes #1699.

### DIFF
--- a/openmmapi/src/ContextImpl.cpp
+++ b/openmmapi/src/ContextImpl.cpp
@@ -56,12 +56,13 @@ const static char CHECKPOINT_MAGIC_BYTES[] = "OpenMM Binary Checkpoint\n";
 ContextImpl::ContextImpl(Context& owner, const System& system, Integrator& integrator, Platform* platform, const map<string, string>& properties) :
         owner(owner), system(system), integrator(integrator), hasInitializedForces(false), hasSetPositions(false), integratorIsDeleted(false),
         lastForceGroups(-1), platform(platform), platformData(NULL) {
-    if (system.getNumParticles() == 0)
+    int numParticles = system.getNumParticles();
+    if (numParticles == 0)
         throw OpenMMException("Cannot create a Context for a System with no particles");
     
     // Check for errors in virtual sites and massless particles.
     
-    for (int i = 0; i < system.getNumParticles(); i++) {
+    for (int i = 0; i < numParticles; i++) {
         if (system.isVirtualSite(i)) {
             if (system.getParticleMass(i) != 0.0)
                 throw OpenMMException("Virtual site has nonzero mass");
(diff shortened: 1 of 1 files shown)


jchodera:
Is throwing an `Exception` the right behavior? Perhaps we could throw an exception during `Context` initialization (which is annoying, but prevents problems) *and* during `System.addConstraint()` (to actually help users catch the problem when it happens)?

peastman:
The principle I follow in cases like this is not to throw an exception until we're certain there really is an error.  Until they create a Context, it's possible they're planning to call `removeConstraint()` to eliminate the duplicates, or `setConstraintParameters()` to change one of them.  For example, they might have two different routines that add different types of constraints, so they call both of them, and then go through and remove any redundant ones.  They haven't definitely done anything wrong until they create a Context.

jchodera:
Don't we already violate this principle when, say, we add an exception in NonbondedForce for a pair of molecules that already have an exception defined?

I think we have to do SOMETHING to help the user track down. The problem. Otherwise, it will be frustratingly difficult to track down the origin of the problem. For example, listing the duplicate entries might help.

jchodera:
@peastman : You have no comment here?

jchodera:
I'll add this as a separate issue since this PR was merged.

PR #1721, opened 2017-01-13
Fix handling of scientific notation for field 7 in torsions in certain GROMACS topologies

davidlmobley:
Certain GROMACS topologies (see https://github.com/choderalab/yank/issues/623) fail to process correctly when the final term in a torsion uses scientific notation, as it is handled as an integer. Specifically, terms like this are problematic:

```
      1       2       3       4     1    1.80000080e+02    1.51670000e+01    2.00000000e+00
```
and produce an error `ValueError: invalid literal for int() with base 10: '2.00000000e+00'` due to the fact that int('2.00000000e+00') raises a value error. Seems like the simplest fix is to process as a float (which works correctly) and if, indeed, this is supposed to be an int (can someone confirm?) then cast to an int after reading.

### DIFF
--- a/wrappers/python/simtk/openmm/app/gromacstopfile.py
+++ b/wrappers/python/simtk/openmm/app/gromacstopfile.py
@@ -772,7 +772,7 @@ class GromacsTopFile(object):
                                 if periodic is None:
                                     periodic = mm.PeriodicTorsionForce()
                                     sys.addForce(periodic)
-                                periodic.addTorsion(baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], baseAtomIndex+atoms[2], baseAtomIndex+atoms[3], int(params[7]), float(params[5])*degToRad, k)
+                                periodic.addTorsion(baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], baseAtomIndex+atoms[2], baseAtomIndex+atoms[3], int(float(params[7])), float(params[5])*degToRad, k)
                         elif dihedralType == '2':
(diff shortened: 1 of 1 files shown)


peastman:
The last value is multiplicity.  I'm pretty sure that should always be an integer, since otherwise the potential would be discontinuous.  Your fix looks reasonable.

davidlmobley:
OK, then I'd be delighted to have this merged once you're ready, @peastman . Thanks.

jchodera:
Thanks!

@davidlmobley : The updated openmm master is conda-installable via
```bash
conda install --yes -c omnia/label/dev openmm==7.2.0
```
If you already have this installed, you have to first force a downgrade and clean:
```bash
conda install --yes -c omnia openmm==7.0.1
conda clean -plti --yes
conda install --yes -c omnia/label/dev openmm==7.2.0
```

davidlmobley:
Hmm, @jchodera - I've done this twice now and I'm still getting the older openmm prior to the fix. Maybe I'll try and install from source.

davidlmobley:
(Or there is somehow still an old openmm in my tree which is not getting cleanedup...)

davidlmobley:
OK, I have this running now. Thanks, @jchodera . I ended up just going in and manually editing the Python files in my anaconda3 directory. I'm guessing that the 7.2.0 version being pulled down by conda didn't update yet or something.

jchodera:
You're right---there was an issue with the build infrastructure such that the last update was [three days ago](https://anaconda.org/omnia/openmm/files).

Issue #1688, opened 2016-12-09
A bug in amoeba multipole force?


yumipxd:
In line 339 and 348 of “openmm/plugins/amoeba/platforms/cuda/src/kernels/multipoleElectrostatics.cu” (and in 1512 and 1521 of  openmm/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp ), whether "fIZ" should be replaced by "fJZ" ?

peastman:
It does look like it.  @andysim this is in your code.  Could you take a look at it?

peastman:
It looks to me like this is a mistake, but one that doesn't actually change the results.  That's because the only place `fIZ` or `fJZ` ever gets used is in the sum `fJZ+fIZ`.  So if you add something to the wrong one, it doesn't matter.  The sum of the two is still the same.

andysim:
You're correct, @peastman.  Because both variables are initialized [here](https://github.com/pandegroup/openmm/blob/master/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L1472-L1473), the code is correct, but misleading, as written.  I think I was part way through just removing the "I" and "J" from those variable names (that term is equal and opposite), but simply didn't finish it up.  The clearest way to write it is to replace the offending "fIZ" with "fJZ", as suggested by @yumipxd; that would be most consistent with the equivalent permanent multipole terms [here](https://github.com/pandegroup/openmm/blob/master/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.cpp#L1472-L1473).  Sorry for the confusion - I'll open a PR soon to clear this up.

peastman:
Great, thanks.

yumipxd:
I see what you mean，thank you，@peastman and @andysim

PR #1724, opened 2017-01-19
Trivial changes to AMOEBA plugin.

andysim:
# Description

Just a couple of small changes to the AMOEBA code, to make the CUDA tests pass cleanly, and to make the code more readable.

# Status
- [x] Rename some variables, to avoid the confusion raised in #1688.
- [x] Avoid finite difference test in single precision, which fails due to roundoff error.
- [x] Ready to go!

### DIFF
--- a/plugins/amoeba/platforms/cuda/src/kernels/multipoleElectrostatics.cu
+++ b/plugins/amoeba/platforms/cuda/src/kernels/multipoleElectrostatics.cu
@@ -336,7 +336,7 @@ __device__ void computeOneInteraction(AtomData& atom1, AtomData& atom2, bool has
     iEIY -= eCoef*(qiUinpI.y*qiUindJ.x + qiUindI.y*qiUinpJ.x);
     iEJY -= eCoef*(qiUinpJ.y*qiUindI.x + qiUindJ.y*qiUinpI.x);
     fIZ += dCoef*(qiUinpI.x*qiUindJ.x + qiUindI.x*qiUinpJ.x);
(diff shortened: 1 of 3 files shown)


peastman:
What test was failing?  I haven't had any problems with that function.  The step size is big enough that it ought to work in single precision.

andysim:
The Cuda AMOEBA extrapolated test's finite difference test was failing (by a tiny amount) on our K20 and K40 cards.  I'm not sure why it works for you, but feel free to ignore this change is you think it's not needed.

peastman:
If that's the case, how about just increasing the tolerance slightly?  In the last line of `check_finite_differences()`, try changing the 1e-4 to 2e-4.  Does that make it pass?  We could even make it use a higher tolerance in double precision.  Something like:

```C++
double tol = 2e-4;
if (Platform::getPlatformByName("CUDA").getPropertyValue(context, "Precision") == "double")
    tol = 1e-6;
```

andysim:
I did a little testing, and found that the step size is pretty good:

```
1e-4: Expected 379.946, found 379.665
5e-4: Expected 379.946, found 379.972
1e-3: Expected 379.946, found 379.957
5e-3: Expected 379.946, found 379.984
1e-2: Expected 379.946, found 380.069 
```

It looks like the precision argument was being ignored previously so I fixed that, and implemented the tolerance check as you suggested.  Again, feel free to merge / ignore / squash this PR as you see fit.

peastman:
Thanks!  This looks good now.  Since we're just about to build the release candidate for 7.1, I'll wait to merge this until after that is released.

PR #1743, opened 2017-02-16
Cleanup to cmake scripts

peastman:
The CMake scripts have accumulated a lot of obsolete bits over the years (and they were never too clean to begin with).  This is a first pass at cleaning up some of the most obvious parts.


peastman:
I'm not sure what's going on here.  The OpenCL version failed because it didn't build any of the OpenCL test cases (except ones associated with plugins).  It did build the OpenCL library itself.  It just for some reason skipped building the tests.  But `CMAKE_FLAGS` specified `-DOPENMM_BUILD_OPENCL_TESTS=ON`.  And none of my changes are related to either OpenCL or to building tests!

PR #1746, opened 2017-02-22
Improved performance of Python State objects

### LINKED ISSUE #1689: Poor performance converting vec3 lists to numpy arrays
Hi,

I've been running some timings on a live simulation, looking for bottlenecks. For my test structure (1438 residues, 50 timesteps between GUI updates) I'm currently seeing 2-3 frames per second - so roughly 300-500 milliseconds per frame. At each GUI update I get the positions and the forces (the latter to do a quick check for overly large forces indicating instability), and between the two of these, about 150-180 milliseconds is going simply into converting the two 21,400*Vec3 lists into numpy arrays. 

```
c = sim.context
%timeit state = c.getState(getPositions = True, getForces = True)

> 10 loops, best of 3: 35.1 ms per loop

state = c.getState(getPositions = True, getForces = True)
%timeit state.getPositions(asNumpy=False)

> The slowest run took 29.31 times longer than the fastest. This could mean that an intermediate result is being cached.
> 1000000 loops, best of 3: 1.53 µs per loop

%timeit state.getPositions(asNumpy=True)

> The slowest run took 74189.11 times longer than the fastest. This could mean that an intermediate result is being cached.
> 1000000 loops, best of 3: 1.54 µs per loop

p = state.getPositions(asNumpy=False)
import numpy
%timeit numpy.array(p._value)

> 10 loops, best of 3: 86.2 ms per loop
```

Filling the values directly in C++ into a n*3 numpy float64 array should bring the timing of this step down at least 100-fold. Would it be possible to add this as an option?

peastman:
Fixes #1689.  I redesigned how State objects are represented in Python.  Instead of implementing them from scratch, it now has SWIG wrap them directly, then replaces only the methods whose signatures are different.  In addition to being much simpler, this also gives much better performance.  I tested it with the following code:

```Python
t1 = datetime.now()
for i in range(100):
    state = simulation.context.getState(getPositions=True, getVelocities=True, getParameters=True)
    pos = state.getPositions(asNumpy=False)
t2 = datetime.now()
for i in range(100):
    state = simulation.context.getState(getPositions=True, getVelocities=True, getParameters=True)
    pos = state.getPositions(asNumpy=True)
t3 = datetime.now()
elapsed1 = t2-t1
elapsed2 = t3-t2
print('Vec3: ', elapsed1.seconds+elapsed1.microseconds*1e-6)
print('Numpy: ', elapsed2.seconds+elapsed2.microseconds*1e-6)
```

Using the old version it produce these times:

```
Vec3:  4.55014
Numpy:  11.018711
```

With the new version, here are the times:

```
Vec3:  2.221929
Numpy:  0.111581
```

The 2x improvement to the Vec3 version comes from the fact that it doesn't convert each array until the first time you request it, rather than converting them all when the State is created.  The test code requests both positions and velocities, but then never queries the velocities, so they never have to get converted.  Most of the time you probably aren't requesting information you won't use, so the speed of getting Vec3s should be about the same as before.

With numpy, though, the new version is dramatically faster.  The entire process of copying over the values is now basically just a single call to `memcpy()`, so it's very fast.

To get a pointer to the numpy array's data, I'm using an inline function defined in one of the numpy header files.  That means it now requires numpy to be installed at compile time.  We'll need to update the conda recipe accordingly.  I think it should still work if numpy isn't installed at runtime, though I haven't tested that to be sure.


jchodera:
Niiiiiiice!

Tagging: @pgrinaway @gregoryross @bas-rustenburg

tristanic:
That's fantastic. Thanks very much!

PR #1755, opened 2017-02-28
Add script to build on my Jenkins machine

swails:
test

### DIFF
--- a/devtools/ci/jenkins/install.sh
+++ b/devtools/ci/jenkins/install.sh
@@ -36,7 +36,7 @@ make PythonInstall
 # Now run the tests
 python -m simtk.testInstallation
 cd python/tests && py.test -v && cd ../..
-python devtools/run-ctest.py --start-time $(date +%s)
+python devtools/run-ctest.py --job-duration=120 --timeout 300
 
 # Now remove the conda environment
 source deactivate
--- a/devtools/ci/jenkins/install.sh
+++ b/devtools/ci/jenkins/install.sh
@@ -15,7 +15,6 @@ module load cuda conda/jenkins
 
 # Constants
 CONDAENV=openmm-test-3.5
-BUILD_DIRECTORY="${WORKSPACE}/openmm-build"
 INSTALL_DIRECTORY="${WORKSPACE}/openmm-install"
 SRC_DIRECTORY="${WORKSPACE}/openmm-src" # set in the Jenkins configuration
 
@@ -29,10 +28,8 @@ source activate ${CONDAENV} # enter our new environment
 
 # Build OpenMM
 CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=\"${INSTALL_DIRECTORY}\" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang"
-test -d "${BUILD_DIRECTORY}" && rm -fr "${BUILD_DIRECTORY}"
-mkdir "${BUILD_DIRECTORY}"
-cd "$BUILD_DIRECTORY"
-cmake $CMAKE_FLAGS "${SRC_DIRECTORY}"
+cd "${SRC_DIRECTORY}"
+cmake $CMAKE_FLAGS .
 make -j4 install
 make PythonInstall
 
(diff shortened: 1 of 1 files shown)


swails:
run jenkins

swails:
All of the tests seem to pass, with a caveat.  Just a few comments here:

1. Jenkins will not build *all* pull requests automatically -- only those initiated from a whitelist of users.  Currently, that is every user in the omnia-md, Amber-MD, and pandegroup groups.  This is primarily as a protection for my computer.  Although the Jenkins user has a strictly limited permission set, I'm still wary of letting *any* pull request with any arbitrary code execute on my machine.  If someone outside those groups submits a PR, Jenkins will post a message (posing as @peastman), with something like "Will an admin please verify this patch?".  Anyone in one of those groups can then trigger a build with the phrase "r u n  j e n k i n s" after they check that the diff isn't clearly malicious (without the extra spaces -- I didn't want to trigger another build here).
1. @peastman or I can add (trustworthy :smiley:) users to the whitelist if I missed anybody
1. You can use the trigger phrase I mentioned above to rebuild if necessary.
1. The script `devtools/ci/jenkins/install.sh` runs the install and test, and the script in the Jenkins setup is just sourcing this file.  Any command in the build process that ends in a non-zero exit status kills the whole build and marks it as failed.

[It also appears that timeout tests do not get rerun](http://www.jasonswails.com:8888/job/Omnia/job/OpenMM/7/default/console) -- is this known? Intentional?

peastman:
Excellent!  I think the whitelist makes sense, and that will cover most PRs.

> Jenkins will post a message (posing as @peastman)

Wait—it has the ability to post messages pretending to be me?  How does it manage that?

> It also appears that timeout tests do not get rerun -- is this known? Intentional?

That's because of the way we run OpenCL tests.  Travis doesn't give us enough time for it to get through all of them.  So we run them in a random order, and let it go for a fixed amount of time.  Whichever test it's in the middle of when time runs out will get a timeout, but that doesn't indicate any problem with it, so we don't rerun it.

Is this ready to merge?

swails:
> Wait—it has the ability to post messages pretending to be me? How does it manage that?

The token you gave me to post status notifications was derived from your account.

> That's because of the way we run OpenCL tests. Travis doesn't give us enough time for it to get through all of them. So we run them in a random order, and let it go for a fixed amount of time. Whichever test it's in the middle of when time runs out will get a timeout, but that doesn't indicate any problem with it, so we don't rerun it.

We don't have any such time constraint on Jenkins, so should we change how the tests are run to get all of the test cases?

On the one hand, it's irritating to have a super long build set up as a merge gate, but on the other hand it would probably be nice to ensure all tests are being run.

Another option is to set up SCM polling so that the master branch is built nightly or weekly and runs through a more comprehensive test suite and opens an issue if the tests fail.  That way we get earl(ier) failure notifications without introducing huge latency in the test step.

Assuming we are happy with this Jenkins build as a merge gate right now, this is good to merge.

peastman:
By default, the `run-ctest.py` script limits each test to 180 seconds and the whole suite to 30 minutes.  But you can pass arguments to it to increase those, which I think would be a good idea.  As you say, you don't have any strict time limit there.  Right now, the main limiter on merging builds is the travis OS X build, which often takes a few hours to start running.  So we could easily let jenkins run for a couple of hours without slowing anything down.  We can reconsider that in the future if things change.

swails:
Done -- although is something wrong with Travis?

peastman:
I think travis is having problems related to the AWS outage.

peastman:
Jenkins is reporting failure, but I'm not sure what to make of the console output.  It's very short, and I don't see anything that looks like it actually tried to build.

swails:
> It's very short, and I don't see anything that looks like it actually tried to build.

You're looking at the parent matrix build.  Click on the link to the "default" build in that console output to see the actual build. [See here](http://www.jasonswails.com:8888/job/Omnia/job/OpenMM/8/default/console)

I figured at some point we may want to include some more tests, or use the matrix build to split up the tasks.  For instance, we can have one build just do the OpenCL tests and one do just the CUDA tests to speed things up.  We can also test the GCC compilers (I have multiple versions to choose from) and/or the Intel compilers (ditto).

There's a lot we can do with Jenkins.  We can also set up a polling job that periodically runs some benchmark tests to start generating a report of performance vs. time, we can set up a pipeline job that checks the stability of the Omnia suite (PDBFixer + OpenMM, etc.), and so on.

peastman:
Ok, I see.  The error is a failure in TestOpenCLVirtualSitesSingle.  Not sure why it's failing (possibly just exceeding the tolerance), but I'll look into that later.  Anyway, I think this is ready to merge!

PR #1757, opened 2017-03-03
Allow docs to build in Sphinx 1.3.1 and 1.5.1

Lnaden:
Sphinx's `DUspan` function was improved and renamed to `DUrole` between 1.3.1 and 1.5.1. Since OpenMM's docs extend the custom text role to support `code`, this caused the docs to not compile on both versions. 

This PR allows the old rewrite to `DUspan` (will continue to break for anything other than `code`) while also allowing the new `DUrole` (with proper extension to allow `code`). Both versions yield the same typesetting output for this function. 

Related issues:
omnia-md/conda-recipes#723
omnia-md/conda-dev-recipes#81

### DIFF
--- a/docs-source/developerguide/conf.py
+++ b/docs-source/developerguide/conf.py
@@ -185,14 +185,14 @@ latex_elements = {
     \\usepackage{caption}
     \\setcounter{tocdepth}{3}
     \\captionsetup[figure]{labelformat=empty}
-    % Backwards behavior for sphinx < 1.5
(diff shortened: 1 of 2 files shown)


Lnaden:
The failed windows tests do not appear related to this PR's changes

peastman:
It's caused by my recent changes to DCD files (just a problem in the test case though, not the DCD classes themselves).  I'm working on fixing it now.

Lnaden:
Awesome, thanks!

PR #1761, opened 2017-03-15
Fix bug where CHARMM impropers did not respect toroidal boundary conditions

### LINKED ISSUE #1760: CHARMM reader produces incorrect improper energies
[`CharmmPsfFile`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py) uses [this code](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py#L910) to encode CHARMM improper torsions:
```python
        # Ick. OpenMM does not have an improper torsion class. Need to
        # construct one from CustomTorsionForce
        force = mm.CustomTorsionForce('k*(theta-theta0)^2')
```
Given that OpenMM computes [the torsion `theta` on the domain `[-pi,pi]`](http://docs.openmm.org/7.1.0/api-python/generated/simtk.openmm.openmm.CustomTorsionForce.html#simtk.openmm.openmm.CustomTorsionForce), guess what happens when `theta0` is not 0?

jchodera:
Fixes #1760.

This is a major bug that affects the reading of CHARMM files, and can cause anything from subtle incorrect simulations to major explosions. Recommend bugfix point release after merging.

### DIFF
--- a/devtools/packaging/README.md
+++ b/devtools/packaging/README.md
@@ -1,10 +1,16 @@
 # Packaging OpenMM into ZIP installers
 
+Set your environment variable `TAG` to the git tag for the release:
+```bash
+# OpenMM 7.1.0rc1
+export TAG="9567ddb"
+```
+
 ## Source
 
 Start the docker container:
 ```bash
-docker run -i -t --rm -v `pwd`:/io jchodera/omnia-build-box:cuda80-amd30-clang38 bash
+docker run -i -t --rm -e TAG -v `pwd`:/io jchodera/omnia-build-box:cuda80-amd30-clang38 bash
 ```
 Inside the docker container:
 ```bash
@@ -23,7 +29,7 @@ cp packaging/compressed/* /io
 
 Start the docker container:
 ```bash
-docker run -i -t --rm -v `pwd`:/io jchodera/omnia-build-box:cuda80-amd30-clang38 bash
+docker run -i -t --rm -e TAG -v `pwd`:/io jchodera/omnia-build-box:cuda80-amd30-clang38 bash
 ```
 Inside the docker container:
 ```bash
@@ -50,4 +56,3 @@ source openmm/devtools/packaging/scripts/osx/prepare.sh
 source openmm/devtools/packaging/scripts/osx/build.sh
 source openmm/devtools/packaging/scripts/osx/package.sh
 ```
-
--- a/openmmapi/include/openmm/CustomTorsionForce.h
+++ b/openmmapi/include/openmm/CustomTorsionForce.h
@@ -65,6 +65,10 @@ namespace OpenMM {
  * force->addPerTorsionParameter("theta0");
  * </pre></tt>
  *
+ * If a harmonic restraint is desired, it is important to be careful of the domain for theta, using an idiom like this:
+ *
+ * <tt>CustomTorsionForce* force = new CustomTorsionForce("0.5*k*min(dtheta, 2*pi-dtheta)^2; dtheta = abs(theta-theta0); pi = 3.1415926535");</tt>
+ *
  * This class also has the ability to compute derivatives of the potential energy with respect to global parameters.
  * Call addEnergyParameterDerivative() to request that the derivative with respect to a particular parameter be
  * computed.  You can then query its value in a Context by calling getState() on it.
--- a/openmmapi/include/openmm/CustomTorsionForce.h
+++ b/openmmapi/include/openmm/CustomTorsionForce.h
@@ -54,9 +54,9 @@ namespace OpenMM {
  * This will have no effect on Contexts that already exist unless you call updateParametersInContext().
  * theta is guaranteed to be in the range [-pi,+pi]
  *
- * As an example, the following code creates a CustomTorsionForce that implements a harmonic potential:
+ * As an example, the following code creates a CustomTorsionForce that implements a periodic potential:
  *
- * <tt>CustomTorsionForce* force = new CustomTorsionForce("0.5*k*(theta-theta0)^2");</tt>
+ * <tt>CustomTorsionForce* force = new CustomTorsionForce("0.5*k*(1-cos(theta-theta0))");</tt>
  *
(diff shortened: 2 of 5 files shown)


swails:
Looks like parmed needs the same fix

jchodera:
Whoops, I'm not sure why this pulled in a stray `README.md` change.

jchodera:
Though we actually *want* that README change. It must have been neglected in another PR.

jchodera:
Actually, we should also change the [problematic `CustomTorsionForce` example that led to this issue](http://docs.openmm.org/7.1.0/api-python/generated/simtk.openmm.openmm.CustomTorsionForce.html#simtk.openmm.openmm.CustomTorsionForce) too.

peastman:
Thanks for catching this!  Would it be possible to add a test case?  Your fix looks reasonable to me, but it's just complicated enough that I'd feel more comfortable if something were verifying it.

jchodera:
> Thanks for catching this! Would it be possible to add a test case? Your fix looks reasonable to me, but it's just complicated enough that I'd feel more comfortable if something were verifying it.

I've been trying to think of what a good test case might be. We noticed this because a system we prepared through the CHARMM-GUI containing spin labels had utterly bizarre behavior, and by turning off a force at a time, noticed the `CustomTorsionForce` was responsible and specific impropers in the `CYR1` residue (defined in `toppar_all36_label_spin.str`) with 180-degree reference angles were to blame:
```
IMPROPER
CTS1 CES1 CES1 CTS3    200.00   0   180.00 ! X X X P 2  !from Benoit
CTS1 CES1 CES1 CTS2    200.00   0   180.00 ! - - - - 0
CTS1 CES1 CES1 HES1    200.00   0   180.00 ! X X X P 1
```
What would a minimal test case be? I suppose I could craft an artificial system of four atoms with artificial `IMPROPER` entry, but what's the correct way to test this without introducing a lot of external machinery to perform a torsion drive and compute energies to validate against?

peastman:
How about just manually putting the atoms into a few positions and verifying the energies are correct for them?

peastman:
What's the status on this?

jchodera:
Sorry for letting this slip. There was enough effort required to produce the customized test you suggested (configurations, CHARMM input files, reference energies) that I thought it might actually be *easier* (and much more useful) to add a Lepton expression for computing minimum-image angular differences and their derivatives as an alternative to the current hack. I managed to figure out that this seemed to be straightforward, but haven't actually started that coding yet.

swails:
What functional form does CHARMM use?  Is it not what you implemented here?

jchodera:
CHARMM uses a simple harmonic form, but the harmonic restraint must be centered on the reference improper torsion angle.

Some CHARMM parameter sets (such as EPR spin probe dyes) define the reference torsion angle to be 180 degrees, e.g.:
```
IMPROPER
CTS1 CES1 CES1 CTS3    200.00   0   180.00 ! X X X P 2  !from Benoit
CTS1 CES1 CES1 CTS2    200.00   0   180.00 ! - - - - 0
CTS1 CES1 CES1 HES1    200.00   0   180.00 ! X X X P 1
```
In this case, the [-180, +180] torsion domain used by OpenMM becomes problematic because the naive torsion expression `(K/2)*(theta-theta0)^2` uses a torsion deviation `(theta-theta0)` that is not the minimum-image torsion deviation about `theta0`.

jchodera:
I suppose we could use the simpler expression
```
(K*2)*min(dtheta^2, (2*pi-dtheta)^2); dtheta = abs(theta-theta0);
```

instead?

swails:
> There was enough effort required to produce the customized test you suggested (configurations, CHARMM input files, reference energies) that I thought it might actually be easier (and much more useful) to add a Lepton expression for computing minimum-image angular differences and their derivatives as an alternative to the current hack. I managed to figure out that this seemed to be straightforward, but haven't actually started that coding yet.

How does that eliminate the need for the test?  If this new expression works, why not just use it?

And why can't you just turn the system you found this issue on into the test case?  Or if it's too large, shave it down a little?

jchodera:
> How does that eliminate the need for the test?

It doesn't eliminate the need for *a* test, but it could change what we're testing. If we implement `minimum_image_deviation(theta,theta0)` in lepton, we could test that instead, which is easier.

Arguably, we are missing some key things from our CHARMM validation suite as well, but that probably extends far beyond this particular term and should probably be tackled separately (e.g. by automating the tests we did for the OpenMM 7 paper).

> If this new expression works, why not just use it?

...because spending some time discussing the best approach to fixing this issue and what deeper implications this issue has in the first place is maybe better than merging in the first hacky fix that came to mind? 

For example, our [OpenMM 6.3.0 release also has this bug](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py#L910) but doesn't support some lepton operations that this fix uses, but I think *does* support the alternative:
```
(K*2)*min(dtheta^2, (2*pi-dtheta)^2); dtheta = abs(theta-theta0);
```
If we wanted to backport the fix, we may want to adopt this simpler form.

peastman:
If you still think we need a 7.1.1 patch release to fix this, we really should do the simple change.  That's a small change I'd feel comfortable including in a patch (as long as we have a test case for it).  But adding a new function is a big change involving lots of new code, separate implementations for each platform, etc.  Any release including that would be much more than a patch.

That's not to say we can't *also* add that function in 7.2.  But for a patch, let's do the simplest, safest fix.

peastman:
Simpler is good.  An even simpler version is

```
(K*2)*min(dtheta, 2*pi-dtheta)^2; dtheta = abs(theta-theta0);
```

peastman:
Any progress on this?

jchodera:
We put a local workaround in place, so this had ended up back on the bottom of the pile.

I still don't have a good idea for how to build a good test without burning a whole afternoon. Any ideas?

I can at least substitute the simpler form right now...

peastman:
I've written a test case for this and sent it to you as a PR.

jchodera:
Thank you, @peastman!

PR #1769, opened 2017-03-28
Use C++11 style loops

peastman:
I've been gradually going through the code and converting loops to the newer C++11 style.  This isn't comprehensive, but it gets a lot of them and simplifies a lot of code.

### DIFF
--- a/olla/src/Platform.cpp
+++ b/olla/src/Platform.cpp
@@ -70,10 +70,10 @@ static int platformInitializer = registerPlatforms();
 
 Platform::~Platform() {
     set<KernelFactory*> uniqueKernelFactories;
-    for (map<string, KernelFactory*>::const_iterator iter = kernelFactories.begin(); iter != kernelFactories.end(); ++iter)
-        uniqueKernelFactories.insert(iter->second);
-    for (set<KernelFactory*>::const_iterator iter = uniqueKernelFactories.begin(); iter != uniqueKernelFactories.end(); ++iter)
-        delete *iter;
+    for (auto& factory : kernelFactories)
+        uniqueKernelFactories.insert(factory.second);
+    for (auto factory : uniqueKernelFactories)
+        delete factory;
 }
 
 const vector<string>& Platform::getPropertyNames() const {
@@ -102,8 +102,8 @@ void Platform::setPropertyDefaultValue(const string& property, const string& val
     string propertyName = property;
     if (deprecatedPropertyReplacements.find(property) != deprecatedPropertyReplacements.end())
         propertyName = deprecatedPropertyReplacements.find(property)->second;
-    for (int i = 0; i < (int) platformProperties.size(); i++)
-        if (platformProperties[i] == propertyName) {
+    for (auto& prop : platformProperties)
+        if (prop == propertyName) {
             defaultProperties[propertyName] = value;
             return;
         }
@@ -121,8 +121,8 @@ void Platform::registerKernelFactory(const string& name, KernelFactory* factory)
 }
 
 bool Platform::supportsKernels(const vector<string>& kernelNames) const {
-    for (int i = 0; i < (int) kernelNames.size(); ++i)
-        if (kernelFactories.find(kernelNames[i]) == kernelFactories.end())
+    for (auto& name : kernelNames)
+        if (kernelFactories.find(name) == kernelFactories.end())
             return false;
     return true;
 }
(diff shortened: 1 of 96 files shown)


mpharrigan:
So much nicer!

swails:
One thing I'll point out is that the Jenkins machine seems to have failed every time it's built OpenMM on the same test in the same way:

```
1/3 Test #274: TestOpenCLVirtualSitesSingle .........***Failed    2.16 sec
exception: Assertion failure at TestVirtualSites.h:402.   Expected [0, 0, 0], found [-0.00298543, 0.0200184, 0.00502852]

    Start 160: TestOpenCLBrownianIntegratorSingle
2/3 Test #160: TestOpenCLBrownianIntegratorSingle ...   Passed    9.88 sec
    Start 196: TestOpenCLCustomIntegratorSingle
3/3 Test #196: TestOpenCLCustomIntegratorSingle .....   Passed   29.46 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =  41.51 sec

The following tests FAILED:
	274 - TestOpenCLVirtualSitesSingle (Failed)
Errors while running CTest
Rerunning failing tests...
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

I realize we've discussed failing tests being ok in general for stochastic tests, but I believe this particular one has failed the last 32 times in a row on Jenkins (and seems deterministic).  If we're satisfied with that difference in single precision on the OpenCL platform, perhaps we ought to loosen the tolerance in just that instance (or figure out why it's not close enough).

It'd be nice to see a little green checkmark next to the Jenkins build *sometimes* :smiley:

peastman:
I've been meaning to look into that.  I think it's just a case of the tolerance being a little too low, since it only exceeds it by a little bit, and only in single precision mode.  But I'd like to figure out why the error is higher on your computer than on others I've tried.  What GPU is it running on?

peastman:
Half the travis builds are failing with "Unable to establish SSL connection" to anaconda.org.  Is there something wrong with the site right now?

swails:
The Jenkins machine has 2 GPUs, a GTX 740 and a GTX 680 (the former drives the display):

```
$ nvidia-smi 
Wed Mar 29 13:17:20 2017       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 375.26                 Driver Version: 375.26                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GT 740      Off  | 0000:01:00.0     N/A |                  N/A |
| 21%   29C    P8    N/A /  N/A |     73MiB /   979MiB |     N/A      Default |
+-------------------------------+----------------------+----------------------+
|   1  GeForce GTX 680     Off  | 0000:07:00.0     N/A |                  N/A |
| 35%   25C    P8    N/A /  N/A |      1MiB /  1998MiB |     N/A      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID  Type  Process name                               Usage      |
|=============================================================================|
|    0                  Not Supported                                         |
|    1                  Not Supported                                         |
+-----------------------------------------------------------------------------+
$ eselect opencl list 
Available OpenCL implementations:
  [1]   nvidia *
```

That last command shows what OpenCL provider my computer is using -- in this case it's from the NVidia driver (version 375.26).  It's also using clang 3.7.1.  Device 0 (according to the CUDA API) is the GTX 680, which I presume is the GPU that OpenMM chooses.

peastman:
It should be picking the 680 (more cores, higher clock rate, and same compute capability), though it might be good to be sure.  I don't know of any reason that either of those would have lower accuracy.  It's mostly just embedded GPUs that can have lower accuracy.

swails:
> though it might be good to be sure.

How do we check that on OpenCL?  Is there some kind of platform method that lets us get the default device or something?

peastman:
After you create a Context, you can call `context.getPlatform().getPropertyValue(context, "DeviceName")` on it.

swails:
Then you can add the command

```bash
python -c "from simtk.openmm import Context, System, VerletIntegrator, Platform; system=System(); system.addParticle(6); context = Context(system, VerletIntegrator(0.002), Platform.getPlatformByName('OpenCL')); print(Platform.getPlatformByName('OpenCL').getPropertyValue(context, 'DeviceName'))"
```
to the [Jenkins install script](https://github.com/pandegroup/openmm/blob/master/devtools/ci/jenkins/install.sh#L29) after activating the Conda environment in order to print out the information.

Just as an FYI, here's what I get running on my desktop just now:

```
swails@Batman ~ $ python -c "from simtk.openmm import Context, System, VerletIntegrator, Platform; system=System(); system.addParticle(6); context = Context(system, VerletIntegrator(0.002), Platform.getPlatformByName('OpenCL')); print(Platform.getPlatformByName('OpenCL').getPropertyValue(context, 'DeviceName'))"
GeForce GTX 680
```

So it's quite fair to assume you'd get the same answer in the Jenkins environment.

peastman:
I don't know what's up with travis, but it's clearly unrelated to this PR, so I'm going to go ahead and merge it, then see if I can debug the problem separately.

Issue #1773, opened 2017-04-03
Travis failing to connect to anaconda.org


peastman:
Some of the travis builds recently started failing with this error:

```
$ wget https://anaconda.org/omnia/ccache/3.2.4/download/${TRAVIS_OS_NAME}-64/ccache-3.2.4-0.tar.bz2
--2017-04-01 04:16:26--  https://anaconda.org/omnia/ccache/3.2.4/download/linux-64/ccache-3.2.4-0.tar.bz2
Resolving anaconda.org (anaconda.org)... 52.21.146.187, 54.81.117.184, 107.23.5.139, ...
Connecting to anaconda.org (anaconda.org)|52.21.146.187|:443... connected.
Unable to establish SSL connection.
The command "wget https://anaconda.org/omnia/ccache/3.2.4/download/${TRAVIS_OS_NAME}-64/ccache-3.2.4-0.tar.bz2" failed and exited with 4 during .
```

This error happens on the STATIC_LIB, PYTHON2, and PYTHON3 builds.  On the other hand, the CPU_OPENCL, CUDA_COMPILE, and OSX builds all successfully connect and download that file.  It's not clear to me what's different between the ones that work and the ones that don't.  Of the three that work correctly, two specify

```
sudo: required
dist: trusty
```

and the third one is OS X:

```
language: objective-c
      os: osx
```

The three that fail all specify `sudo: false` and some value for `python:` (but not all the same value), and none of them specify `dist:` or `os:`.  Possibly the problem is related to one of these differences?

jchodera:
Looks like a `wget` SSL failure? [This post](http://stackoverflow.com/questions/28757232/unable-to-establish-ssl-connection-upon-wget-on-ubuntu-14-04-lts) suggests this might be due to the fact that Ubuntu 14.04 uses TLS1.2 by default, which can be worked around by supplying the `--secure-protocol=TLSv1` argument to `wget`. Worth a shot?

peastman:
Or maybe just specify `dist: trusty`?

jchodera:
Whoops, I misread the above and thought it was `trusty` that was failing. 

Specifying `dist: trusty` and `sudo: true` will slow things down a bit since it won't use the docker instances, but is probably worth getting those test branches to run again.

peastman:
[...]

PR #1774, opened 2017-04-03
Fix errors on travis downloading files

peastman:
Attempt at fixing #1773.


peastman:
Well that didn't work.  When I told it to use `http:`, it just redirected to `https:`.  Let's try something different.

PR #1776, opened 2017-04-04
Fix test failure on Jenkins

peastman:


peastman:
It looks like the drift in momentum comes from using SETTLE for the constraints.  That's a long sequence of analytic calculations, and if you do them in single precision, the result at the end is kind of inaccurate.  If I modify the constraint distances slightly to make the molecules non-symmetric, that forces it to use SHAKE instead, and the momentum drift gets much smaller.  Perhaps there's some way the SETTLE calculation could be restructured to be more accurate in single precision, though it's not obvious to me how.

peastman:
Though maybe that's not quite it either.  If I just look at the constrained distances coming out of SETTLE, they match the expected values to excellent accuracy.  Is there a reason SHAKE would be intrinsically better than SETTLE at conserving momentum (given the same accuracy for the constrained distances)?

If I switch to mixed precision, the momentum conservation is quite good.

swails:
🎉

PR #1778, opened 2017-04-07
Improved gromacs

scons:
This patch expands OpenMM support for GROMACS input files.  Specifically, it adds support for combination rule ‘1’ which corresponds to a non bonded potential of ‘q1\*q2/(4\*pi\*e0)/r -sqrt(C1\*C1)/r^6 + sqrt(A1\*A2)/r^12’.  This is required for the use of Gō model input files from SMOG (http://smog-server.org).  There are also various associated changes to support running these models in OpenMM.  After the pull, example files from SMOG (http://smog-server.org/2ci2.smog.tar.gz) should be able to run without modification.


swails:
Cool! Do you have a test case you can add for this?

scons:
Hi @swails.  I have mixed feelings about writing a test for this.  On one hand writing a test would be easy: just grab the sample input files from SMOG and make sure they run.  On the other hand: there currently aren't any other similar tests so it would not be fitting with the rest of the test suite.  Also, this test might be cumbersome (input files tracked in github?) so I might want to write my own test from scratch… which could be time consuming.  As this code is part of a 2 month lab rotation, my advisor and I would rather move on the the next problem we have to solve.

To reiterate, the easiest way to test this is just to attempt to run the example Gromacs input files from SMOG (http://smog-server.org/2ci2.smog.tar.gz).  Before applying my patches, the code will print an error and exit, after incorporating my patches the simulation will run.  I am willing to provide my small python script that executes this simulation, but it is more or less identical to the Gromacs example from Section 4.3 of the OpenMM Users Guide and Theory Manual (version 7.1) and probably won't add much.

Thanks for your understanding.

peastman:
Thanks!  I haven't had time to look over the code yet, but I'll try to do that soon.

A simple test that just loads an input file and checks that the energy matches Gromacs is totally fine.  That's most of what we do now.  Take a look at https://github.com/pandegroup/openmm/blob/master/wrappers/python/tests/TestGromacsTopFile.py#L40-L52.  You can just add another test similar to that one.  Input files go in wrappers/python/tests/systems.

swails:
Most of the Python tests are, as @peastman says, just a simple single-point energy calculation that validates against the "reference" implementation for a single system (preferably chosen to maximize coverage).

Without this test, regular contributors are faced with one of a few options:

1. Write their own tests for your feature to make sure edits don't break that functionality (and unless they have experience with this feature, they're unlikely to do as well without a lot more time investment)
2. Not worry about breaking these additions in the future
3. Consider the GROMACS parsers feature-complete and try to not touch them again to reduce the risk of breakage

There's a lot of stuff OpenMM *could* support in GROMACS topologies that it currently doesn't.  A small test will go a long way toward making this easier for the community to maintain and support even when you move on to the next thing.

scons:
thank you @peastman for the directions on where to find the unit tests for the Gromacs handling code.  I have added a unit test called test_SMOG that uses the above-mentioned example file and calculates the potential energy.  And @swails you're absolutely right about needed to avoid breakage in the future, thank you.

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -591,7 +596,14 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
             sys.setDefaultPeriodicBoxVectors(*boxVectors)
         elif nonbondedMethod in (ff.CutoffPeriodic, ff.Ewald, ff.PME, ff.LJPME):
             raise ValueError('Illegal nonbonded method for a non-periodic system')
-        nb = mm.NonbondedForce()
+        if self._defaults[1] == '1':
+            # factor of 138.935456 for coulomb taken from tests/TestCustomNonbondedForce.h:574
+            nb = mm.CustomNonbondedForce('138.935456*q1*q2/r-sqrt(C1*C2)/r^6+sqrt(A1*A2)/r^12')
```
swails: Is this method supposed to work with PME? If so, PME requires NonbondedForce to handle electrostatics. Then just set the epsilon parameters to 0 and handle the L-J part with a CustomNonbondedForce. 

As it stands now, you can't use PME with this method.

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -857,19 +880,44 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
...
+                            if self._defaults[1] == '1':
+                                exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atom))
+                            elif self._defaults[1] == '2':
+                                exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atom, 0, 0, 0))
 
 
         # Create nonbonded exceptions.
+        if self._defaults[1] == '1':
+            # implement named exceptions (generally used for 1-4 interactions)
+            pair_bond = mm.CustomBondForce('138.935456*q/r-C/r^6+A/r^12')
```
swails: Same comment here.

swails: If you use NonbondedForce for the electrostatics you can also have it handle the exception parameters (rather than requiring that they be done with a CustomBondForce).

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -776,13 +790,16 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
                         elif dihedralType == '2':
                             # Harmonic torsion
                             k = float(params[6])
+                            phi0 = float(params[5])
                             if k != 0:
                                 if harmonicTorsion is None:
-                                    harmonicTorsion = mm.CustomTorsionForce('0.5*k*(theta-theta0)^2')
+                                    harmonicTorsion = mm.CustomTorsionForce('0.5*k*(thetap-theta0)^2; thetap = step(-(theta-theta0+pi))*2*pi+theta+step(theta-theta0-pi)*(-2*pi); pi = 3.14159')
```
peastman: Only using 6 significant digits for pi isn't very accurate.  A better way to write this is `'pi = %.15g' % math.pi`.

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -857,19 +880,44 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
                         continue # We'll use the automatically generated parameters
                     atom1params = nb.getParticleParameters(baseAtomIndex+atoms[0])
                     atom2params = nb.getParticleParameters(baseAtomIndex+atoms[1])
-                    exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], atom1params[0]*atom2params[0]*fudgeQQ, params[0], params[1]))
+                    # enable tracking pairs and exceptions separately
+                    if self._defaults[1] == '1':
+                        pairs.append((baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], atom1params[0]*atom2params[0]*fudgeQQ, params[0], params[1]))
+                    elif self._defaults[1] == '2':
+                        exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], atom1params[0]*atom2params[0]*fudgeQQ, params[0], params[1]))
```
peastman: I'm a bit confused by the above lines.  In mode 1, don't you still need to add nonbonded exclusions for these (in addition to adding terms to the CustomBondForce)?

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -857,19 +880,44 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
...
                 for fields in moleculeType.exclusions:
                     atoms = [int(x)-1 for x in fields]
                     for atom in atoms[1:]:
                         if atom > atoms[0]:
-                            exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atom, 0, 0, 0))
+                            if self._defaults[1] == '1':
+                                exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atom))
+                            elif self._defaults[1] == '2':
+                                exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atom, 0, 0, 0))
```
peastman: I think this change is unnecessary.  Just adding the same value to `exceptions` should work equally well in either case.

wrappers/python/simtk/openmm/app/gromacstopfile.py
scons: Hi @swalis.  To be honest, in our models we don't need to include electrostatics (q = 0 everywhere).  I thought this would be nice to attempt to support electrostatics for future users.  I suppose I could implement the split method you suggested.

wrappers/python/simtk/openmm/app/gromacstopfile.py
scons: Not according to our runs with Gromacs.  Specifically, setting some parameters in the [ pairs ] section adds an additional force, but does not exclude from the non bonded force.  it is possible to set [ pairs ] and not set the [ exclusions ] for the same atoms.

peastman: Does that mean it shouldn't be adding them to `exceptions` in mode 2 either?

scons: yes, it does.  But due to the way that the native NonbondedForce API differs from the CustomNonbondedForce API, I couldn't see a clean way to implement this.  I'm currently a first-year PhD student so I don't have a lot of time right now.

peastman: Got it.  I can make the changes and send them in a PR to your branch, and then we can get this merged.

agrossfield:
Thanks -- we appreciate it. Steve is just a rotation student, so I only have him for a few weeks. That said, I'd really like to see this get into mainline openmm, both for us and also for Steve.

Cheers,

Dr. Alan Grossfield
Department of Biochemistry and Biophysics
University of Rochester Medical Center

On Apr 17, 2017, at 7:07 PM, peastman <notifications@github.com<mailto:notifications@github.com>> wrote:


@peastman commented on this pull request.

________________________________

In wrappers/python/simtk/openmm/app/gromacstopfile.py<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_pandegroup_openmm_pull_1778-23discussion-5Fr111841168&d=DQMCaQ&c=4sF48jRmVAe_CH-k9mXYXEGfSnM3bY53YSKuLUQRxhA&r=49qnaP-kgQR_zujl5kbj_PmvQeXyz1NAoiLoIzsc27zuRX32UDM2oX8NQCaAsZzH&m=YO8vjvUwHiGSTVH7xy9z-4xjtUO68rkoH6U7Up021Y0&s=-scVeO9nVSGIb9ealxPUn2-u9seSvVQdqUTnZgQMy58&e=>:

> @@ -857,19 +880,44 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
                         continue # We'll use the automatically generated parameters
                     atom1params = nb.getParticleParameters(baseAtomIndex+atoms[0])
                     atom2params = nb.getParticleParameters(baseAtomIndex+atoms[1])
-                    exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], atom1params[0]*atom2params[0]*fudgeQQ, params[0], params[1]))
+                    # enable tracking pairs and exceptions separately
+                    if self._defaults[1] == '1':
+                        pairs.append((baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], atom1params[0]*atom2params[0]*fudgeQQ, params[0], params[1]))
+                    elif self._defaults[1] == '2':
+                        exceptions.append((baseAtomIndex+atoms[0], baseAtomIndex+atoms[1], atom1params[0]*atom2params[0]*fudgeQQ, params[0], params[1]))


Got it. I can make the changes and send them in a PR to your branch, and then we can get this merged.

-

peastman:
I just sent a PR to your branch with various fixes.  Take a look at it, and if everything looks ok, merge it.  Those changes will then automatically be added to this PR.

scons:
Hi Peter, thank you very much for your work.  I manually reviewed the commit, merged it, and ran my tests that I have here and everything looks fine.

peastman:
Great, thanks!  I can merge this as soon as the tests finish running.

peastman:
Merged.  Thanks so much for doing this!

yongwangCPH:
Hi,

I tried the latest version OpenMM7.1.1 and used the example files from http://smog-server.org/2ci2.smog.tar.gz, but got errors:

  File "OpenMM_SMOG.py", line 39, in <module>
    top = GromacsTopFile(top_file)
  File "/home/wanyong/usr/local/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/gromacstopfile.py", line 482, in __init__
    self._processFile(file)
  File "/home/wanyong/usr/local/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/gromacstopfile.py", line 126, in _processFile
    self._processLine(append+' '+line, file)
  File "/home/wanyong/usr/local/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/gromacstopfile.py", line 219, in _processLine
    self._processDefaults(line)
  File "/home/wanyong/usr/local/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/gromacstopfile.py", line 262, in _processDefaults
    raise ValueError('Too few fields in [ defaults ] line: '+line)
ValueError: Too few fields in [ defaults ] line:             1           1 no

It seems there is a problem as reading the top file. Could anyone help?

Yong

scons:
Hello Yong,

From what I can tell, this error message means that you don't have my code installed.  This code is quite new, so even the latest version of OpenMM (7.1.1) does not include it.  Probably the easiest way to get my code is to download the latest prerelease version of OpenMM from Github.  Go to https://github.com/pandegroup/openmm, make sure Branch is set to master, click 'Clone or Download', and finally click 'Download ZIP'.  This should download a version of OpenMM that will allow you to test the features I've add it, but bear in mind that it is prerelease software so may not be 100% stable.

Thanks,
Steve

yongwangCPH:
@scons Thank you for the reply. Yeah, I am just using the code for some performance test. Looking forward it to be merged in the coming version.

PR #1780, opened 2017-04-07
Fix typo in bibtex entry

Lnaden:
Missing comma causes docs not to compile.

Fixes omnia-md/conda-dev-recipes#84


jchodera:
We need this merged quickly so we can get @leeping's `dev` build of openmm with AMBER-FB15 out when the paper hits.

Lnaden:
Groovy, Thanks!

PR #1783, opened 2017-04-11
Improved logic for guessing elements from atom names

peastman:
I've found some PDB files that put a digit at the start of some atom names.  This tries to deal better with those.


everyday847:
Maybe I'm ignorant of some openmm-specific history here, but can't you also
look for the field specifically dedicated to the element (characters 77-78,
in 1-indexing)? That's at least true for standard PDB formats; it'll of
course fail if some other data is stored there, but it can help corroborate
the decision.

peastman:
Yes, but it's also common to find PDB files where that column is missing.  In that case, we do our best to guess the element from the atom name.  That's what this relates to.

PR #1784, opened 2017-04-14
Fixed bug accessing parameters for CustomHbondForce

### LINKED ISSUE #1771: Discrepancies with CustomHbondForce between Reference and CUDA platforms
CustomHbondForce is supposed to accept expressions which "may depend on arbitrary distances, angles, and dihedral angles of the particles involved in computing the force." When using the Reference platform, I have no problems including these distance(), angle(), or dihedral() functions in the .xml file for my force field, and the results agree with numbers I can obtain from other programs.

Upon switching to the CUDA platform, however, I get very different results for the distance() and dihedral() functions in particular. 

I've tried to provide an example of this behavior in the zip file below,

[dihedral_bug.zip](https://github.com/pandegroup/openmm/files/877329/dihedral_bug.zip)

which can be run with the command

```
$ python run_co2.py
```
and by changing the 'energy' expression in sapt_co2.xml and the platform in run_co2.py. For different energy expressions (listed below), which use the distance(), angle(), and dihedral() functions, I've summarized the resulting energies between the Reference and CUDA platforms. 

```
Energy Expression     Reference Platform      CUDA Platform
-------------------------------------------------------------------------
distance(a1,d1)               158244.998031         256944.557419
angle(d2,d1,a1)               332305.565796         332607.309815
dihedral(d3,d1,d2,a1)         16890.3384786         nan
```
 
For reference, I'm running these calculations on Tesla M2070 GPUs with CUDA 8.0 and version 7.1.0.dev-9567ddb of OpenMM.

These discrepancies between the two platforms are rather disconcerting, so any help diagnosing the underlying problem/bug would be most appreciated!

peastman:
Fixes #1771.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -4521,12 +4521,12 @@ void CudaCalcCustomHbondForceKernel::initialize(const System& system, const Cust
     for (int i = 0; i < (int) donorParams->getBuffers().size(); i++) {
         CudaNonbondedUtilities::ParameterInfo& buffer = donorParams->getBuffers()[i];
         extraArgs << ", const "+buffer.getType()+"* __restrict__ donor"+buffer.getName();
-        addDonorAndAcceptorCode(computeDonor, computeAcceptor, buffer.getType()+" donorParams"+cu.intToString(i+1)+" = donor"+buffer.getName()+"[index];\n");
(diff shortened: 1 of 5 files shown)


peastman:
@swails Jenkins doesn't seem to be able to access the GPU.  All the CUDA and OpenCL tests are failing with CUDA_ERROR_NO_DEVICE.  I'm pretty sure it's not related to this PR, so I'll go ahead and merge it.

swails:
Ack, I updated a lot of the packages on my system, which included the NVidia drivers.  I just need to reboot the machine (which I'll do shortly when it's done with the other thing it's doing).

PR #1789, opened 2017-04-17
Defined entry point so other packages can add force field directories

peastman:
The group is called `openmm.forcefielddir`.  The entry point should be a function that returns a string containing the path to a directory.

See https://github.com/choderalab/openmm-forcefields/issues/6.  Also see http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins for information on creating entry points.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -46,13 +46,16 @@ import simtk.unit as unit
 from . import element as elem
 from simtk.openmm.app import Topology
 from simtk.openmm.app.internal.singleton import Singleton
-from pkg_resources import iter_entry_points
 
 # Directories from which to load built in force fields.
 
 _dataDirectories = [os.path.join(os.path.dirname(__file__), 'data')]
-for entry in iter_entry_points(group='openmm.forcefielddir'):
-    _dataDirectories.append(entry.load()())
+try:
+    from pkg_resources import iter_entry_points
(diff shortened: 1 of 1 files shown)


jchodera:
Thanks! I'll give this a try!

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -47,6 +47,16 @@
 from simtk.openmm.app import Topology
 from simtk.openmm.app.internal.singleton import Singleton
 
+# Directories from which to load built in force fields.
+
+_dataDirectories = [os.path.join(os.path.dirname(__file__), 'data')]
+try:
+    from pkg_resources import iter_entry_points
+    for entry in iter_entry_points(group='openmm.forcefielddir'):
```
jchodera: What about a less awkward name, like `openmm.forcefield_directory`?

jchodera:
Actually, while we're at it, what about:
* Also having it search `$MINICONDA/share/openmm/forcefields/`? I think we can use `pkg_resources.resource_filename()` to get the `$MINICONDA` path.
* Allowing the user to supplement (or override?) the search path with an environment variable, like `$OPENMM_FORCEFIELD_PATH`?

peastman:
I'd rather not.  This isn't meant to be a free for all where users indiscriminately add files to the default search path.  That could easily get out of hand and create problems where people aren't sure where their files are loaded from.

PR #1802, opened 2017-05-03
Added option for CPU PME to be deterministic

### LINKED ISSUE #1799: Is there a way to make PME energies on the CPU deterministic?
We just realized (or were reminded) today that PME energies on the CPU platform is non-deterministic when more than one thread is in use, which causes some problems for specific applications (such as measuring nonequilibrium work). Setting `OPENMM_CPU_THREADS=1` appears to eliminate the non-determinism.

Is there a way to enforce determinism (possibly with some performance loss, but not forcing only one compute thread) as a platform property?

peastman:
Fixes #1799.  The interface is the same as for CUDA: specify `'DeterministicForces':'true'` when you create your context.  This *only* affects the behavior of PME, so unlike with CUDA, the results are still not fully deterministic.  But PME is often the largest source of non-determinism, so the variation should at least be reduced.

### DIFF
--- a/olla/include/openmm/kernels.h
+++ b/olla/include/openmm/kernels.h
@@ -1295,8 +1295,9 @@ public:
      * @param gridz        the z size of the PME grid
      * @param numParticles the number of particles in the system
      * @param alpha        the Ewald blending parameter
+     * @param deterministic whether it should attempt to make the resulting forces deterministic
      */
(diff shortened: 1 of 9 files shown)


jchodera:
Awesome, thanks!

Tagging @gregoryross, @bas-rustenburg, and @andrrizzi.

PR #1805, opened 2017-05-05
Make the nonpolar part of GBSA available via an argument to createSystem for both Amber and CHARMM files.

### LINKED ISSUE #1803: ACE term not calculated with OBC2 and CHARMM force field
I've been running some MD using the CHARMM force field and the OBC2 implicit solvent model. When the system is created by CharmmPsfFile.createSystem, the ACE term of the implicit solvent model is not computed. You can check this around line 1200 of file 
```wrappers/python/simtk/openmm/app/charmmpsffile.py```
which calls the ```GBSAOBC2Force``` function using ```None``` as third parameter instead of ```'ACE'```. 

I consider this a bug and I would suggest to set the default to use the ACE term. 
Also, I would like to have the possibility to activate/inactivate the ACE term with a switch in the createSystem function. It would be even nicer to have the surface tension term freely modifiable, not hard coded.

swails:
argument

Fixes #1803


peastman:
Thanks!

Issue #1796, opened 2017-04-26
Features to support metadynamics


jlmaccal:
I've been working to build a metadynamics integrator in OpenMM and am hitting a few snags.

Note, that I'm aware of the plumed plugin for OpenMM, but (a) it's to slow for us to use on GPUs, and (b) I want to define reaction coordinates in other ways, specifically  using MELD's energy function and other OpenMM custom forces.

The approach I'm taking is similar to that for the `AMDForceGroupIntegrator`. The goal is to treat the energy of a specified force group as a collective variable and to make that distribution uniform over a given range. I'm going to build the biasing potential on the CPU side, as it only needs to be updated occasionally. The modifications to the forces are straightforward and can easily be applied during integration. 

I think this should be straightforward to to implement, but I'm hitting a few issues.

## Issue 1: The energy is not correct.

Metadynamics works by applying a biasing potential. Using a custom integrator, it is possible to generate forces that correspond to this biased distribution, but it's not possible to modify the energy itself. The same issue occurs with the `AMD` integrators. The energies reported in `State` objects are not correct, although these could be corrected on the CPU-side after the fact. A more problematic issue is that the `MonteCarloBarostat` would be using the wrong energies.

I'm not sure of the best solution, but it might be nice if custom forces could access energies and forces from another force group. This would enable things like meta-D and AMD to be developed as forces rather than integrators and would enable the correct energies and forces to be reported. I'm not sure of the exact semantics of this, but probably these would be the energies and forces to that point in the calculation, so the order of forces would matter. 

This is just one possible solution, but it would be great if there was some mechanism where the forces and energies from one force could be modified in some way.

## Issue 2: There is no direct access to the derivative of tabulated functions

Generally, when we use tabulated functions, we use them as part of an energy expression and OpenMM automatically takes care of the correct derivative for us. However, in a custom integrator it is sometimes useful to have access to both the value from a tabulated function and its derivative, but it's currently not possible to access the derivative.

Of course, it's possible to define a tabulated function for both the energy and its derivative, but the derivatives don't actually match due to interpolation---the derivative is a second order spline in the first case and a third order spline in the second.

The proposed fix would be to make both the lookup table, `lookup(r)`, and its derivative `Dlookup(r)` available in custom expressions. For the 2D case, maybe the notation could be `D10lookup(r)` and `D01lookup(r)` or similar.

## Issue 3: Tabulated functions cannot be queried from python

[...]

PR #1816, opened 2017-05-18
CustomIntegrator can use tabulated functions

### LINKED ISSUE #1772: Add tabulated function capabilities to CustomIntegrator?
How hard would it be to add tabulated function capabilities to `CustomIntegrator` to match those found in the `Custom*Force` classes? 

This feature turns out to be surprisingly useful in build nonequilibrium integrators that use optimized nonequilibrium switching protocols, which could be very useful for the constant-pH code (and other tools) we are working on.

peastman:
Fixes #1772.

### DIFF
--- a/openmmapi/include/openmm/CustomIntegrator.h
+++ b/openmmapi/include/openmm/CustomIntegrator.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2011-2016 Stanford University and the Authors.      *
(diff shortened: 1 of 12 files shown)


jchodera:
Woohoo!

jchodera:
Don't forget these will need to get serialized/deserialized (with a test to ensure everything is preserved)!

peastman:
Oops, thanks for pointing that out.

jlmaccal:
Does this address either of the points in #1796, and if not, could it be extended?

Specifically:
- Access to the derivative of the tabulated function from the integrator
- Access to the values and derivatives of the tabulated function from the C++ and python APIs

peastman:
No, this only covers #1772.  I haven't implemented those features yet.

Issue #1820, opened 2017-05-20
Travis master builds have been failing due to sphinx docs issue


jchodera:
This is odd: The travis master builds have been repeatedly failing:

https://travis-ci.org/pandegroup/openmm/branches

![image](https://cloud.githubusercontent.com/assets/3656088/26273145/c8d3c35a-3cf7-11e7-92b4-4861f8f395a5.png)

It looks to be a failure in the sphinx docs:
https://travis-ci.org/pandegroup/openmm/jobs/234088080#L5590-L5602

```
reading sources... [  6%] generated/OpenMM.AmoebaMultipoleForce
Exception occurred:
  File "/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/sphinxcontrib/autodoc_doxygen/autosummary/__init__.py", line 44, in import_by_name
    raise ImportError('no module named %s' % ' or '.join(tried))
ImportError: no module named NonbondedMethod or NonbondedMethod
The full traceback has been saved in /tmp/sphinx-err-NJ__ko.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make[3]: *** [api-c++/index.html] Error 1
make[2]: *** [docs-source/api-c++/CMakeFiles/C++ApiDocs.dir/all] Error 2
make[1]: *** [docs-source/api-c++/CMakeFiles/C++ApiDocs.dir/rule] Error 2
make: *** [C++ApiDocs] Error 2
mv: cannot stat ‘sphinx-docs/developerguide/html’: No such file or directory
```

jchodera:
This may be related to why the [conda-dev-recipe for openmm](https://travis-ci.org/omnia-md/conda-dev-recipes/jobs/234439715#L3675-L3753) is also failing on the sphinx docs.

peastman:
There's an earlier error that might or might not be connected to that:

```
Exception occurred:
  File "/home/travis/build/pandegroup/openmm/docs-source/sphinx/autonumber.py", line 60, in doctree_resolved
    raise ValueError('Unknown target for autonumber reference: '+target)
ValueError: Unknown target for autonumber reference: Example,Gromacs example
```

Do you know what versions of Doxygen and Sphinx it's using?

jchodera:
[...]

PR #1821, opened 2017-05-21
Pin sphinx to 1.5.6 because 1.6.x breaks things

jchodera:
cc: #1820

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -179,7 +179,7 @@ script:
   - python devtools/run-ctest.py --start-time $START_TIME
 
   - if [[ ! -z "${DOCS_DEPLOY}" && "${DOCS_DEPLOY}" = "true" ]]; then
-      pip install sphinx sphinxcontrib-bibtex sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen;
+      pip install sphinx==1.5.6 sphinxcontrib-bibtex sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen;
       make sphinxhtml;
       make C++ApiDocs PythonApiDocs;
       mkdir -p api-docs;


jchodera:
This seems to have at least fixed the travis builds, @peastman.

peastman:
Great!  I'm going to hold off on merging until we've determined exactly what it was that broke.  If it doesn't work with the latest version of sphinx, that will cause problems for users.  I'll look into it today.

jchodera:
@peastman : Do you want to merge this since a solution doesn't seem forthcoming?

@Lnaden : You haven't discovered anything that would resolve this issue without pinning sphinx, have you?

Lnaden:
I have not, I also have not looked at the problem in a while.

peastman:
Sounds reasonable.  I tried to figure out the problem, but wasn't able to.  I then posted on the Sphinx discussion group (https://groups.google.com/forum/#!topic/sphinx-users/6rqewb-SqH4), but didn't get much there either.  At this point, the only way I'm likely to figure it out is by digging through the Sphinx source code.  I still plan to do that eventually, but in the mean time pinning the version seems like a reasonable worksaround.

peastman:
It looks like this still failed: https://travis-ci.org/pandegroup/openmm/jobs/242962112

jchodera:
It seems to be failing due to errors in the docstrings now:
https://travis-ci.org/pandegroup/openmm/jobs/242962112#L6028-L6045
```
/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/simtk/openmm/app/charmmpsffile.py:docstring of simtk.openmm.app.charmmpsffile.CharmmPsfFile:5: ERROR: Unexpected indentation.
...
/home/travis/virtualenv/python2.7_with_system_site_packages/local/lib/python2.7/site-packages/simtk/openmm/app/forcefield.py:docstring of simtk.openmm.app.forcefield.ForceField.registerTemplateGenerator:15: ERROR: Unexpected indentation.
```
It looks like it doesn't like indentation in lists, like [here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/charmmpsffile.py#L119-L130):
```python
class CharmmPsfFile(object):
    """A chemical structure instantiated from CHARMM files.
    This structure has numerous attributes that are lists of the elements of
    this structure, including atoms, bonds, torsions, etc. The attributes are
        - residue_list
        - atom_list
        - bond_list
...
```
This seems to be a different error than before, where we were running into `ImportError` errors:
https://github.com/pandegroup/openmm/issues/1820

I'm not sure why we are only seeing this now, though, rather than in the PR.

peastman:
I don't think we build the documentation in PRs.  That doesn't happen until they get merged.

jchodera:
OK, I've captured this in a new issue: https://github.com/pandegroup/openmm/issues/1835

PR #1823, opened 2017-05-24
Flexible constraints

swails:
Add support for flexibleConstraints to ForceField.createSystem

There are times when it is helpful to have the energy terms actually
added to the System even if that degree of freedom is constrained. Other
parsers (notably the Amber parsers) already support this keyword, so add
that support to ForceField as well

This also replaces some string-based element detection algorithms with
whatever the PDB parser does (by comparing directly to atom.element)

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -223,7 +223,7 @@ class ForceField(object):
             trees.append(tree)
 
         # Process includes.
-        
+
         for parentFile, tree in zip(files, trees):
             if isinstance(parentFile, str):
                 parentDir = os.path.dirname(parentFile)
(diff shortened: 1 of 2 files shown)


swails:
run jenkins

swails:
@peastman -- did you turn off Jenkins integration with OpenMM?

peastman:
> @peastman -- did you turn off Jenkins integration with OpenMM?

No, I don't know why it didn't run.  I'm pretty sure it's been running on other recent PRs.

It looks like the new test case you added is failing.

swails:
> It looks like the new test case you added is failing.

Because `validateConstraints` checks that a "bond" is either in the list of constraints or the list of bonds, but not both.  I didn't realize that :)

peastman:
It's still failing, just on a different line now:

```
>       self.assertGreater(af2.getNumAngles(), af2.getNumAngles())
E       AssertionError: 785 not greater than 785
```

swails:
Fixed. Appveyor is still broken but has been for some time it appears.

peastman:
Looks good, thanks.

PR #1837, opened 2017-06-16
Created CustomCVForce

### LINKED ISSUE #1796: Features to support metadynamics

peastman:
Fixes #1796.

This is functional, but needs a lot of testing.  Please try it out and see how it works for you.  I'm not sure how fast it will be.  I might be able to further reduce the overhead, but that requires some good tests I can profile to see what happens in realistic use cases.

### DIFF
--- a/docs-source/usersguide/theory.rst
+++ b/docs-source/usersguide/theory.rst
@@ -1158,6 +1158,21 @@ specified in three ways:
 * Per-donor parameters are defined by specifying a value for each donor group.
 * Per-acceptor parameters are defined by specifying a value for each acceptor group.
 
+CustomCVForce
+*************
+
+CustomCVForce computes an energy as a function of "collective variables".  A
+collective variable may be any scalar valued function of the particle positions
+and other parameters.  Each one is defined by a :code:`Force` object, so any
+function that can be defined via any force class (either standard or custom) can
+be used as a collective variable.  The energy is then computed as
+
+.. math::
+   E=f(...)
+
+where *f*\ (...) is a user supplied mathematical expression of the collective
+variables.  It also may depend on user defined global parameters.
+
 
 .. _writing-custom-expressions:
 
--- a/olla/include/openmm/Platform.h
+++ b/olla/include/openmm/Platform.h
@@ -128,6 +128,14 @@ public:
      */
     virtual void contextCreated(ContextImpl& context, const std::map<std::string, std::string>& properties) const;
     /**
+     * This is called whenever a new Context is created using ContextImpl::createLinkedContext().  It gives the
+     * Platform a chance to initialize the context and store platform-specific data in it.
+     *
+     * @param context          the newly created context
+     * @param originalContext  the original context it is linked to
+     */
+    virtual void linkedContextCreated(ContextImpl& context, ContextImpl& originalContext) const;
+    /**
      * This is called whenever a Context is deleted.  It gives the Platform a chance to clean up
      * any platform-specific data that was stored in it.
      */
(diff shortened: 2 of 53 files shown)


peastman:
Here's what I'm seeing for the overhead using the DHFR benchmark.  Without the CustomCVForce, it gets 301 ns/day.  Adding a CustomCVForce that doesn't actually compute any CVs, it's 297 ns/day, so fairly negligible cost.  If I then add two CVs to it (each one is just the distance between two atoms), that goes down to 264 ns/day.  So about a 10% slowdown, which isn't bad.  I think that could be further improved by evaluating the CVs on a separate thread, but I really want to get feedback before I do more optimizing.

jchodera:
Awesome!

Let me have our `openmm-dev` conda packaging framework build an `openmm 7.3` dev package for this branch. That'll make it easier for us to play with.

jchodera:
OK, some of the `openmm` 7.3.0 dev packages are up here:

peastman:
Is anyone testing this?  If not, I'm going to go ahead and merge it.

jlmaccal:
I haven't had a chance to play with this yet, but looking forward to it.

jchodera:
I am! But I won't have results until Monday.

peastman:
Shall I go ahead and merge this?

jchodera:
Go for it---still behind in testing. I'll switch the `openmm` dev package back to building the master after you merge so that others can easily test too.

PR #1838, opened 2017-06-19
Fix a typo that was preventing an error from being raised

mohebifar:


peastman:
Looks good.  Thanks for catching this!

Issue #1832, opened 2017-06-09
Long-range correction computation times


jchodera:
We're running into an odd problem with a `CustomIntegrator` and an alchemical system containing `Custom*Force` terms where changing a context parameter seems to result in a sudden increase in the per-timestep time by 60x.

On a GTX-TITAN, for example, the per-timestep time remains low until the context parameter `lambda_sterics` is modified from 1.0, at which point the time-per-timestep rockets from 10 ms to 600 ms:
```
     90 : -319121.711 kJ/mol :   11.909 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.09000000
      91 : -318910.208 kJ/mol :   11.130 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.08000000
      92 : -318920.677 kJ/mol :   11.298 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.07000000
      93 : -319024.438 kJ/mol :    8.587 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.06000000
      94 : -319066.106 kJ/mol :   11.298 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.05000000
      95 : -319027.394 kJ/mol :    8.534 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.04000000
      96 : -318954.743 kJ/mol :   12.134 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.03000000
      97 : -318848.571 kJ/mol :    9.846 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.02000000
      98 : -318685.161 kJ/mol :   11.206 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.01000000
      99 : -318529.539 kJ/mol :   10.777 ms : lambda_sterics   1.00000000 lambda_electrostatics   0.00000000
     100 : -318504.142 kJ/mol :  620.636 ms : lambda_sterics   0.99333333 lambda_electrostatics   0.00000000
     101 : -318609.827 kJ/mol :  617.799 ms : lambda_sterics   0.98666667 lambda_electrostatics   0.00000000
     102 : -318698.663 kJ/mol :  624.226 ms : lambda_sterics   0.98000000 lambda_electrostatics   0.00000000
     103 : -318668.663 kJ/mol :  618.002 ms : lambda_sterics   0.97333333 lambda_electrostatics   0.00000000
[...]

PR #1841, opened 2017-06-22
Optimized computing long range correction for CustomNonbondedForce

peastman:
This tries to address #1832, but don't expect anything dramatic.  It's a bit faster than before, but not hugely faster.

### DIFF
--- a/openmmapi/include/openmm/internal/CustomNonbondedForceImpl.h
+++ b/openmmapi/include/openmm/internal/CustomNonbondedForceImpl.h
@@ -69,7 +69,7 @@ public:
     static void calcLongRangeCorrection(const CustomNonbondedForce& force, const Context& context, double& coefficient, std::vector<double>& derivatives);
 private:
     static double integrateInteraction(Lepton::CompiledExpression& expression, const std::vector<double>& params1, const std::vector<double>& params2,
-            const CustomNonbondedForce& force, const Context& context);
+            const CustomNonbondedForce& force, const Context& context, const std::vector<std::string>& paramNames);
     const CustomNonbondedForce& owner;
     Kernel kernel;
 };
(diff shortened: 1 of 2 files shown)


peastman:
Has something changed with the travis configuration?  The CUDA and OpenCL builds are both failing with the error

```
In file included from /home/travis/build/pandegroup/openmm/platforms/opencl/./include/OpenCLContext.h:40:0,
                 from /home/travis/build/pandegroup/openmm/platforms/opencl/./include/OpenCLArray.h:30,
                 from /home/travis/build/pandegroup/openmm/platforms/opencl/src/OpenCLArray.cpp:27:
/home/travis/build/pandegroup/openmm/platforms/opencl/src/cl.hpp:177:19: fatal error: GL/gl.h: No such file or directory
 #include <GL/gl.h>
```

It looks like the OpenGL headers aren't being installed anymore.

jchodera:
The [travis image update history](https://docs.travis-ci.com/user/build-environment-updates/) doesn't show any updates to `trusty` images (Ubuntu 14.04).

peastman:
I'm going to merge this anyway, since the error is unrelated to this PR and everything passed on Jenkins.

jchodera:
The [travis issue log](https://www.traviscistatus.com/incidents/rnlm5ffmk7dm) doesn't suggest anything either.

Where does `GL/gl.h` come from? Is it installed in the `trusty` build image, do we install it, or is it something in the ccache (which may have been corrupted or cleared)?

peastman:
No idea where it was coming from in this particular case.  There are various packages that can install OpenGL headers, such as `mesa-dev`.

jchodera:
It looks like you can add `mesa-common-dev` from the [`trusty` whitelist](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty#L3859) to [this block in `.travis.yml`](https://github.com/pandegroup/openmm/blob/master/.travis.yml#L5-L10) and that might ensure `GL/gl.h` is available. But I'm not sure why this changed.

peastman:
They did in fact just change their images:

https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch

And the new versions removed the `libgl1-mesa-dev` package:

https://github.com/travis-ci/build-environment-updates/blob/master/2017Q2/gce/garnet-trusty/dpkg-manifest.diff#L402

I'll add it back.

PR #1842, opened 2017-06-22
Fixed compilation error on travis

peastman:
See #1841.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
       - python-numpy
       - python-scipy
       - libfftw3-dev
+      - libgl1-mesa-dev
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


peastman:
It's still failing.  How do I tell if it's actually installing it?  There's no mention of mesa anywhere in the build log.

jchodera:
Hm, I don't see anything in the raw logs about `apt-get`ting those packages either. 

There's a [mysterious line](https://travis-ci.org/pandegroup/openmm/jobs/245959410#L618) in the output that says:
```
Could not locate requirements.txt. Override the install: key in your .travis.yml to install dependencies.
```

jchodera:
The failing travis branches have `sudo: required`, so we can just install things directly with `sudo apt-get install -y libgl1-mesa-dev` in those branches in [here](https://github.com/peastman/openmm/blob/49c90bb3d35d341a9b9b4242c771fad8143a66b4/.travis.yml#L121-L132). Maybe that will work?

Lnaden:
Do you think this change will need to be made on `conda-recipes` as well?

jchodera:
We use our own docker image there, so this shouldn't be necessary.

PR #1846, opened 2017-06-24
Master

mharger:
changes to match pandeGroup github.


jchodera:
Whoa, it looks like this totally breaks the public API for a bunch of Forces.

mharger:
did I push to the main? I am sorry. I am only intending to keep local. This
is my mistake. Please reject.

everyday847:
You may have meant to have mharger:master as the base for the comparison,
rather than pandegroup:master (i.e., to merge pandegroup:master INTO your
branch).

mharger:
I think this was the problem. I am sorry for the confusion.

PR #1853, opened 2017-07-03
CustomIntegrator avoids extra force computations when UpdateContextState doesn't change them

peastman:
This is an optimization to CustomIntegrator.  Before, it assumed that whenever it encountered an UpdateContextState step, that invalidated all forces and they would need to be recalculated.  But usually that isn't the case.  Of all the standard forces, only barostats actually invalidate the forces, and even those only do it occasionally (when they accept a move).  This change makes it smarter about that, so it only recalculates forces when it's actually needed.


jchodera:
Awesome! I hadn't realized this was the case! Would this have required an extra force update per timestep for essentially *all* `CustomIntegrator` implementations that included a call to `integrator.addUpdateContextState()` per timestep? If so, this sounds like it will make all the [integrators in `openmmtools` much faster](http://openmmtools.readthedocs.io/en/latest/integrators.html).

peastman:
Not necessarily.  It depends where a particular integrator needs forces, and whether something else is invalidating them as well.

jchodera:
If anyone wants to play with this branch, I've built a copy in the `omnia/label/dev` channel as `openmm==7.3.0` and [pushed it to `omnia`](https://anaconda.org/omnia/openmm/files):
```bash
conda install -c omnia/label/dev openmm==7.3.0
```

Tagging @maxentile @pgrinaway

jchodera:
I'm not sure I'm noticing any practical difference here for our `LangevinIntegrator` custom integrators:

openmm 7.1.1
```
[chodera@lilac:chodera]$ python benchmark-integrators.py
Timing: openmm.LangevinIntegrator
     10.069 s :   85.807 ns/day
Timing: openmmtools.integrators.LangevinIntegrator with VRORV
     11.719 s :   73.725 ns/day
Timing: openmmtools.integrators.LangevinIntegrator with OVRVO
     11.205 s :   77.107 ns/day
```
openmm 7.3 (built from the @peastman `updatecontextstate` branch):
```
[chodera@lilac:chodera]$ python benchmark-integrators.py
Timing: openmm.LangevinIntegrator
     10.247 s :   84.320 ns/day
Timing: openmmtools.integrators.LangevinIntegrator with VRORV
     11.959 s :   72.245 ns/day
Timing: openmmtools.integrators.LangevinIntegrator with OVRVO
     11.872 s :   72.774 ns/day
```
Note that we are trying to track down a slowdown introduced in the `dev` channel builds that might be due to which compiler flags are selected by `clang` during the build:
https://github.com/omnia-md/conda-dev-recipes/issues/98

jchodera:
Here's the script I used:
```python
#!/bin/env python

"""
Benchmark openmmtools integrators.
"""

import time
from openmmtools import testsystems, integrators

testsystem_name = 'DHFRExplicit'

testsystem = getattr(testsystems, testsystem_name)()
[system, positions] = testsystem.system, testsystem.positions

nsteps = 5000
temperature = 300 * unit.kelvin
pressure = 1.0 * unit.atmospheres
timestep = 2.0 * unit.femtoseconds
collision_rate = 90 / unit.picoseconds

# Remove CMMotionRemover
system.removeForce(system.getNumForces() - 1)

# Add barostat
barostat = openmm.MonteCarloBarostat(pressure, temperature)
system.addForce(barostat)

    print(force.__class__.__name__)

integrators = [
    ( 'openmm.LangevinIntegrator', openmm.LangevinIntegrator(temperature, collision_rate, timestep) ),
    ( 'openmmtools.integrators.LangevinIntegrator with VRORV', integrators.LangevinIntegrator(temperature, collision_rate, timestep, splitting='V R O R V') ),
    ( 'openmmtools.integrators.LangevinIntegrator with OVRVO', integrators.LangevinIntegrator(temperature, collision_rate, timestep, splitting='O V R V O') ),
    ]

for (name, integrator) in integrators:
    integrator.step(1)
    print('Timing: %s' % name)
    initial_time = time.time()
    integrator.step(nsteps)
    final_time = time.time()
    elapsed_time = final_time - initial_time
    ns_per_day = (nsteps * timestep) / (elapsed_time * unit.seconds) / (unit.nanoseconds / unit.day)
    print('   %8.3f s : %8.3f ns/day' % (elapsed_time, ns_per_day))
    del context
```

peastman:
Were you expecting this to affect those integrators?  What's the series of operations they use?

jchodera:
I *think* so. These systems have a barostat, and they call `updateContext` at the beginning of the integration cycle. Our integrators have a `pretty_print()` method that produces the following output:
```
Timing: openmmtools.integrators.LangevinIntegrator with VRORV
step      0 : allow forces to update the context state
step      1 : if(has_kT_changed = 1):
step      2 :    sigma <- sqrt(kT/m)
step      3 :    has_kT_changed <- 0
step      4 : end
step      5 : v <- v + (dt / 2) * f / m
step      6 : constrain velocities
step      7 : x <- x + ((dt / 2) * v)
step      8 : x1 <- x
step      9 : constrain positions
step     10 : v <- v + ((x - x1) / (dt / 2))
step     11 : constrain velocities
step     12 : v <- (a * v) + (b * sigma * gaussian)
step     13 : constrain velocities
step     14 : x <- x + ((dt / 2) * v)
step     15 : x1 <- x
step     16 : constrain positions
step     17 : v <- v + ((x - x1) / (dt / 2))
step     18 : constrain velocities
step     19 : v <- v + (dt / 2) * f / m
step     20 : constrain velocities
     11.596 s :   74.509 ns/day
Timing: openmmtools.integrators.LangevinIntegrator with OVRVO
step      0 : allow forces to update the context state
step      1 : if(has_kT_changed = 1):
step      2 :    sigma <- sqrt(kT/m)
step      3 :    has_kT_changed <- 0
step      4 : end
step      5 : v <- (a * v) + (b * sigma * gaussian)
step      6 : constrain velocities
step      7 : v <- v + (dt / 2) * f / m
step      8 : constrain velocities
step      9 : x <- x + ((dt / 1) * v)
step     10 : x1 <- x
step     11 : constrain positions
step     12 : v <- v + ((x - x1) / (dt / 1))
step     13 : constrain velocities
step     14 : v <- v + (dt / 2) * f / m
step     15 : constrain velocities
step     16 : v <- (a * v) + (b * sigma * gaussian)
step     17 : constrain velocities
     11.624 s :   74.330 ns/day
```
VRORV has a force computation right after the context state update.

jchodera:
Also, feel free to let us know if that's not the most readable format for pretty-printing `CustomIntegrator programs`---we can easily change it!

peastman:
I ran your script and verified that it's only computing the force once per step.  So I think it's working correctly.

Issue #1306, opened 2015-12-18
Nose-Hoover thermostat?


swails:
Does anybody have a `CustomIntegrator` implementation of the Nose-Hoover thermostat that they would be willing to share?

jchodera:
I haven't seen one, but I also wouldn't recommend Nosé-Hoover for new production simulations not intended to replicate earlier results due to its ergodicity problems.

jchodera:
[This paper from Scheraga](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC2577381/) describes a useful discrete-timestep implementation (Eqs 6-17) that may aid your implementation.  It looks like it should be straightforward to implement.

swails:
The iso-kinetic integrators proposed out of Tuckerman's lab are built from NHC integrators.  So I wanted that as a starting-point for building those integrators (which, as far as I've been able to tell from the articles, don't have those ergodicity problems).

jchodera:
You mean [these isokinetic thermostats](http://scitation.aip.org/content/aip/journal/jcp/118/6/10.1063/1.1534582)?

jchodera:
An interesting [analysis of these thermostats for a simple system](http://arxiv.org/abs/1004.4294).  

Behavior looks pretty good, though I am always worried about recommending these in general for cases in which we simulate two uncoupled subsystems (as in alchemical free energy calculations, where the ligand is partially or fully decoupled from the rest of the system) since there is no way for energy exchange to occur between uncoupled modes unless stochastic thermostats are coupled to each degree of freedom.

swails:
Thanks!

PR #1865, opened 2017-08-01
Nosé-Hoover thermostat

andysim:
I was chatting with somebody at a conference, who wanted a Nosé-Hoover thermostat to study transport properties, so I decided to code one up using the super-slick CustomIntegrator machinery.  Give the discussion on issue #1306 this may not be something you want, so feel free to close this PR.  The implementation uses the [reversible, higher-order MTS integration method](http://dx.doi.org/10.1080/00268979600100761) from Martyna and co-workers.  I arrived at the default parameters by testing on a couple of systems, but they may be overkill.  The only thing I'm not sure about is how to obtain the number of degrees of freedom; I have a janky solution implemented now, but it fails to account for translations and constraints.  Perhaps that could be provided as one of the automatic variables in `CustomIntegrator`?

### DIFF
--- a/wrappers/python/simtk/openmm/__init__.py
+++ b/wrappers/python/simtk/openmm/__init__.py
@@ -19,6 +19,7 @@ if sys.platform == 'win32':
 from simtk.openmm.openmm import *
 from simtk.openmm.vec3 import Vec3
 from simtk.openmm.mtsintegrator import MTSIntegrator
+from simtk.openmm.nhcintegrator import NHCIntegrator
 from simtk.openmm.amd import AMDIntegrator, AMDForceGroupIntegrator, DualAMDIntegrator
 
 if os.getenv('OPENMM_PLUGIN_DIR') is None and os.path.isdir(version.openmm_library_path):
--- a/wrappers/python/simtk/openmm/nhcintegrator.py
+++ b/wrappers/python/simtk/openmm/nhcintegrator.py
@@ -52,29 +52,29 @@ class NHCIntegrator(CustomIntegrator):
         Parameters:
         -----------
 
-        timestep: unit.Quantity compatible with femtoseconds, default=1*simtk.unit.femtoseconds
+        timeStep: unit.Quantity compatible with femtoseconds, default=1*simtk.unit.femtoseconds
             The integration timestep for particles.
 
         temperature: unit.Quantity compatible with kelvin, default=298*simtk.unit.kelvin
             The target temperature for the thermostat.
 
-        chainlength: integer, default=5
+        chainLength: integer, default=5
             The number of thermostat particles in the Nosé-Hoover chain.  Increasing
             this parameter will affect computational cost, but will make the simulation
             less sensitive to thermostat parameters, particularly in stiff systems; see
             the 1992 paper referenced above for more details.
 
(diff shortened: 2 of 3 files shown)


jchodera:
We've also been collecting integrators in [`openmmtools.integrators`](http://openmmtools.readthedocs.io/en/latest/integrators.html), though our focus has been on Langevin integrators that don't have all the pathologies that continuous thermostats share.

andysim:
I agree that for most uses the Langevin integrators are ideal; this implementation came from the need to study transport properties, which Langevin integrators can't do.  If you're interested in adding NHC to openmmtools, you're free to copy/paste it, or I can slap an MIT license on the code, if that's what's needed legally.

peastman:
Thanks!  @jchodera you were the one most negative about Nose-Hoover.  Would this be a good thing to put in OpenMM itself, or would `openmmtools.integrators` be a better place for it?  I agree NH is probably not a good choice for most people most of the time, but it does also have its uses.

jchodera:
I think there are enough people interested in Nosé-Hoover that it's worth merging into OpenMM directly since the PR is already set up that way.

peastman:
Ok, sounds reasonable.  Based on a quick scan the code looks pretty good.  I'll try to go through it in more detail shortly.

I notice you're using underscore separated names, or in some cases all lower case names with no separator, which isn't consistent with the rest of OpenMM.  For example, it should be `getCurrentTemperature()` instead of `get_current_temperature()`, `chainLength` instead of `chainlength`, `numYoshidaSuzuki` instead of `n_yoshidasuzuki`, etc.  I don't care about local variables or private fields, only things that are part of the public API.

andysim:
Thanks for the feedback, guys.  I think we're all on the same page; this is designed to help with corner cases, rather than being a go-to tool for everyday use.  Therefore, if it's better suited to openmmtools, I'm happy to close this PR and launch a new one to that project.  I'm glad you noticed the camel case problem, @peastman; I was up late last night cleaning it up and changed everything from the correct convention to the underscores, for some unknown reason.  Will fix that now.

jchodera:
How about we put it in `openmmtools.integrators`, then? I think there's utility in collecting them there until we decide how best to migrate a batch of stable `CustomIntegrator`s into OpenMM proper.

Notably, there's an issue we are still trying to find the best way to resolve: Subclasses of `CustomIntegrator` work very well, but swig-wrapped functions that retrieve the integrator (like `System.getIntegrator()`) return only the base `CustomIntegrator` class, losing any of the subclass methods. This can be surprising behavior and, at minimum, needs to be handled with some documentation. We can hopefully figure out a way to improve this behavior by improving the Swig interface at some point (or maybe migrating to pybind11).

peastman:
That sounds like a reasonable plan.  I think I can fix `Context.getIntegrator()` pretty easily.  It already stores a reference to the Python integrator object to keep it from getting garbage collected, so I just need to replace `getIntegrator()` to make it return that object instead of creating a new Python object from the C++ integrator.

bas-rustenburg:
Would a similar improvement be possible for `CompountIntegrator.getIntegrator()`? If I recall correctly, that also returns only a `CustomIntegrator`.

peastman:
Probably.  I'll look into that too.

andysim:
Closing this one, because it's been ported to [this PR](https://github.com/choderalab/openmmtools/pull/265) to openmmtools.

PR #1870, opened 2017-08-08
Context.getIntegrator() returns the original Python object

peastman:
See https://github.com/pandegroup/openmm/pull/1865#issuecomment-320298921.  I tried to do the same with `CustomIntegrator.getIntegrator()`, but ran into trouble getting it to work.  We need to set `STEAL_OWNERSHIP` for `addIntegrator()`, which defines both `pythonappend` and `pythonprepend` macros for it, and that prevents me from using either of those to store the Python object.  But that's a much more obscure case than `Context.getIntegrator()`, so I don't think it's as important.


peastman:
I figured out how to get this working with CompoundIntegrator too.

jchodera:
Huzzah!

Tagging @andrrizzi and @lnaden, who can make use of this feature in Context caching in openmmtools.

Issue #1835, opened 2017-06-15
Fix documentation issues and add a branch that tests documentation build


jchodera:
We've run into an issue where we have accidentally introduced errors into the docstrings of some of the Python bits of OpenMM without noticing.

We should 
* correct the docstring errors (here is an [example](https://travis-ci.org/pandegroup/openmm/jobs/242962112#L6028-L6045))
* ensure the sphinx docs build is tested in some travis branch

cc #1821

jchodera:
Deployment of the [development documentation build](http://docs.openmm.org/development/api-python/index.html) [is failing](https://travis-ci.org/pandegroup/openmm/jobs/263314139#L6539-L6546) due to what appears to be S3 bucket permissions issues:
```
Preparing deploy
Logging in with Access Key: ****************ICHQ
dpl.3
Deploying application
uploading "userguide/application.html" with {:content_type=>"text/html"}
Oops, It looks like you tried to write to a bucket that isn't yours or doesn't exist yet. Please create the bucket before trying to write to it.
failed to deploy
```
It would be great to get this fixed ASAP.

jchodera:
Just a reminder that the documentation deployment is failing, but it would be great to get this fixed.

@rmcgibbo : Any chance you can help?

jchodera:
Deployment of the dev docs is [still failing](https://travis-ci.org/pandegroup/openmm/jobs/268497277#L6552-L6559):
```
Preparing deploy
Logging in with Access Key: ****************ICHQ
dpl.3
Deploying application
uploading "userguide/application.html" with {:content_type=>"text/html"}
Oops, It looks like you tried to write to a bucket that isn't yours or doesn't exist yet. Please create the bucket before trying to write to it.
failed to deploy
```
We should really get this fixed ASAP.

mpharrigan:
that access key can only push to openmm.org. I'll add docs.openmm.org as well

mpharrigan:
try it now

jchodera:
> that access key can only push to openmm.org. I'll add docs.openmm.org as well

I think @peastman may have to retrigger travis on `master` to see if this works. I don't have access to do so:

https://travis-ci.org/pandegroup/openmm/builds/268497272

peastman:
It's running now.

jchodera:
It worked! Thanks!

peastman:
Yay!  Thanks!

peastman:
[...]

PR #1876, opened 2017-08-15
Fix sphinx errors

jchodera:
Addresses some of #1835


jchodera:
Woohoo! Looks like this worked!

peastman:
Thanks!

PR #1881, opened 2017-08-24
Bug fix to finding identical molecules

peastman:
This fixes an obscure bug in the logic for identifying identical molecules that can be reordered internally.  It happens when two molecules are identical except for their bonds.  Furthermore, each molecule has the same number of bonds of each type, and corresponding bonds even have identical parameters, but don't connect corresponding atoms.  In that case, it will incorrectly conclude they are identical.

This came up when someone was simulating graphite sheets, so all atoms and all bonds had identical parameters.  See https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=8138&p=0&start=0&view=&sid=4f4b9091e65d616b177f72d2fd0739ac.


peastman:
The jenkins error is a stochastic failure.

PR #1882, opened 2017-08-25
Fix segmentation fault

### LINKED ISSUE #1878: Segfault in CompoundIntegrator
The `CompoundIntegrator` throws a segfault when it is garbage collected or deleted. 

Minimal test case (segfaults on termination):
```python
import simtk.openmm as mm
compound = mm.CompoundIntegrator()
compound.addIntegrator(mm.VerletIntegrator(0.002)) # Will not segfault w/o at least 1 sub-integrator.
```
You can force the segfault early by trying to `del` the integrator, which will throw it before the end of script. This appears to have started after #1870 was merged. The GDB backtrace is as follows (although I think our cluster is missing symbols):
```
(gdb) bt
#0  0x0000000000000041 in ?? ()
#1  0x00007fffef6265ed in OpenMM::CompoundIntegrator::~CompoundIntegrator (this=0x8ad6a0)
   from /home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/../../../../libOpenMM.so
#2  0x00007fffef626679 in OpenMM::CompoundIntegrator::~CompoundIntegrator (this=0x8ad6a0)
   from /home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/../../../../libOpenMM.so
#3  0x00007fffefc74a47 in _wrap_delete_CompoundIntegrator (args=<optimized out>)
   from /home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/_openmm.cpython-35m-x86_64-linux-gnu.so
#4  0x00007ffff79770f9 in PyCFunction_Call (func=0x7ffff00b69d8, args=0x7ffff7e6a0b8, kwds=<optimized out>) at Objects/methodobject.c:109
#5  0x00007ffff792233a in PyObject_Call (func=func@entry=0x7ffff00b69d8, arg=arg@entry=0x7ffff7e6a0b8, kw=kw@entry=0x0) at Objects/abstract.c:2166
#6  0x00007ffff7923429 in PyObject_CallFunctionObjArgs (callable=0x7ffff00b69d8) at Objects/abstract.c:2446
#7  0x00007fffefc6eef8 in SwigPyObject_dealloc (v=0x7ffff7e66300)
   from /home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/_openmm.cpython-35m-x86_64-linux-gnu.so
#8  0x00007ffff79691b5 in dict_dealloc (mp=0x7ffff7f038c8) at Objects/dictobject.c:1644
#9  0x00007ffff798a171 in subtype_dealloc (self=0x7ffff7e8a320) at Objects/typeobject.c:1194
#10 0x00007ffff7969317 in free_keys_object (keys=0x67eaa0) at Objects/dictobject.c:351
#11 dict_dealloc (mp=0x7ffff7f4e288) at Objects/dictobject.c:1652
#12 0x00007ffff79779e7 in module_dealloc (m=0x7ffff7f49778) at Objects/moduleobject.c:644
#13 0x00007ffff7969e77 in insertdict (value=0x7ffff7d63130 <_Py_NoneStruct>, hash=9218776295280080043, key=0x7ffff7f57470, mp=0x7ffff7fcdb48)
    at Objects/dictobject.c:807
#14 PyDict_SetItem (op=op@entry=0x7ffff7fcdb48, key=0x7ffff7f57470, value=0x7ffff7d63130 <_Py_NoneStruct>) at Objects/dictobject.c:1232
#15 0x00007ffff7a2dacb in PyImport_Cleanup () at Python/import.c:410
#16 0x00007ffff7a3fcd5 in Py_Finalize () at Python/pylifecycle.c:576
#17 0x00007ffff7a613c3 in Py_Main (argc=argc@entry=2, argv=argv@entry=0x602010) at Modules/main.c:798
#18 0x0000000000400b54 in main (argc=2, argv=<optimized out>) at ./Programs/python.c:65
```

There is a second issue which is probably related since it started happening at the same time:

Deep copy of a `CompoundIntegrator` corrupts the object's SWIG methods.

Minimal Test Case
```python
import simtk.openmm as mm
from copy import deepcopy

compound = mm.CompoundIntegrator()
compound.addIntegrator(mm.VerletIntegrator(0.002))

comp_copy = deepcopy(compound)
compound.getIntegrator(0) # Line pass
comp_copy.getIntegrator(0) # Line fails
# Segfault happens on termination, independent of copy error
```
Python backtrace is as follows:
```
  File "compintfail.py", line 9, in <module>
    comp_copy.getIntegrator(0) # Line fails
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 14968, in getIntegrator
    return self._integrators[index]
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 14842, in <lambda>
    __getattr__ = lambda self, name: _swig_getattr(self, CompoundIntegrator, name)
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 74, in _swig_getattr
    return _swig_getattr_nondynamic(self, class_type, name, 0)
  File "/home/nadenl/miniconda3/envs/ommt/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 69, in _swig_getattr_nondynamic
    return object.__getattr__(self, name)
AttributeError: type object 'object' has no attribute '__getattr__'
```
You also cannot copy a `CompoundIntegrator` with no other integrators in it, but that is a very fringe case which raises a `Exception: Illegal index for setCurrentIntegrator()` error.

cc choderalab/openmmtools#270, choderalab/openmmtools#265, @jchodera, @andrrizzi

peastman:
Fixes #1878

### DIFF
--- a/wrappers/python/src/swig_doxygen/swig_lib/python/pythoncode.i
+++ b/wrappers/python/src/swig_doxygen/swig_lib/python/pythoncode.i
@@ -33,7 +33,12 @@ from simtk.openmm.vec3 import Vec3
 %}
 
 %pythonprepend OpenMM::CompoundIntegrator::addIntegrator %{
+    if not integrator.thisown:
+        s = ("the %s object does not own its corresponding OpenMM object"
+             % integrator.__class__.__name__)
+        raise Exception(s)
     self._integrators.append(integrator)
+    integrator.thisown=0
 %}
 
 %pythonprepend OpenMM::AmoebaAngleForce::addAngle %{


jchodera:
Awesome, thanks!

Want to add the minimal python test that segfaults (on termination) to the Python tests?
```python
import simtk.openmm as mm
compound = mm.CompoundIntegrator()
compound.addIntegrator(mm.VerletIntegrator(0.002)) 
```
and maybe the quick deepcopy test?
```
import simtk.openmm as mm
from copy import deepcopy
compound = mm.CompoundIntegrator()
compound.addIntegrator(mm.VerletIntegrator(0.002))
comp_copy = deepcopy(compound)
compound.getIntegrator(0) # Line pass
comp_copy.getIntegrator(0) # Line fails
```

Lnaden:
Woo! I tested this locally and it does fix the segfault. The `deepcopy` problem still exists though, but its a new error:

```python
import simtk.openmm as mm
from copy import deepcopy
compound = compound = mm.CompoundIntegrator()
compound.addIntegrator(mm.VerletIntegrator(0.002))

comp_copy = deepcopy(compound)
compound.getIntegrator(0) #Line passes
comp_copy.getIntegrator(0) # Line fails
```
Throws this error:
```python
  File "/home/nadenl/miniconda3/envs/oms/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 5917, in getIntegrator
    return self._integrators[index]
  File "/home/nadenl/miniconda3/envs/oms/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 5786, in <lambda>
    __getattr__ = lambda self, name: _swig_getattr(self, CompoundIntegrator, name)
  File "/home/nadenl/miniconda3/envs/oms/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 80, in _swig_getattr
    raise AttributeError("'%s' object has no attribute '%s'" % (class_type.__name__, name))
AttributeError: 'CompoundIntegrator' object has no attribute '_integrators'
```
I checked the `dir()` for both `compound` and `comp_copy` and the only method missing from `comp_copy` was the `_integrators`; all other methods were present.

Lnaden:
As that might be a separate issue, I can open a new issue for that if you would prefer to get this one in to fix the more critical segfault

peastman:
No, I'll fix that here too.

peastman:
I think I may need to just revert the original fix, and let CompoundIntegrator go back to returning generic CustomIntegrator objects.  I'm realizing that there are nasty edge case that this whole approach isn't going to work for.  What if you serialize your CompoundIntegrator?  The extra information I'm trying to preserve will be completely lost at that point anyway.  And that's the reason `deepcopy()` is failing: because we implement it by serializing and deserializing.

jchodera:
Thanks, @peastman! Good point about serialization. I'll have to think about whether there is a more generic way to address this kind of issue---allowing Python subclasses to be reconstituted if retrieving, copying, or deserializing from Python---in the future. My guess is that this will likely require either moving from Swig to a more modern [pybind11 wrapping](https://github.com/pybind/pybind11) or a fully Python implementation of the container classes that manipulate the Swig objects under the hood.

PR #1895, opened 2017-09-08
Reporters can request wrapped or unwrapped coordinates

### LINKED ISSUE #1390: Keeping solute at the centre in MD simulation
I am trying to simulate Aniline in water. Minimisation in water goes well. **But right in the first step of simulation solute moves away from the center and goes to the corner**. Am I missing any option while running MD simulation ? or is this a bug. Here is the code for OpenMM simulation. Force field files for [aniline](https://gist.github.com/leelasd/e6138765544df44f2c73) and [water](https://gist.github.com/leelasd/08ba3592fe1829fdec2a) are here. Aniline is [PDB](https://gist.github.com/leelasd/0acabe62a135927c8964) is also attached.  

```
import numpy as np
from simtk.openmm import *
from simtk.unit import *
from sys import stdout,exit

def OPLS_LJ(system):
        forces = { system.getForce(index).__class__.__name__ : system.getForce(index) for index in range(system.getNumForces()) }
        nonbonded_force = forces['NonbondedForce']
        lorentz=CustomNonbondedForce('4*epsilon*((sigma/r)^12-(sigma/r)^6); sigma=sqrt(sigma1*sigma2); epsilon=sqrt(epsilon1*epsilon2)')
        lorentz.setNonbondedMethod(nonbonded_force.getNonbondedMethod())
        lorentz.addPerParticleParameter('sigma')
        lorentz.addPerParticleParameter('epsilon')
        lorentz.setCutoffDistance(nonbonded_force.getCutoffDistance())
        system.addForce(lorentz)
        LJset={}
        for index in range(nonbonded_force.getNumParticles()):
                charge,sigma,epsilon=nonbonded_force.getParticleParameters(index)
                LJset[index]=(sigma,epsilon)
                lorentz.addParticle([sigma,epsilon])
                nonbonded_force.setParticleParameters(index, charge, sigma, epsilon*0)
        for i in range(nonbonded_force.getNumExceptions()):
                (p1, p2, q, sig, eps) = nonbonded_force.getExceptionParameters(i)
                lorentz.addExclusion(p1,p2) ## ALL THE 12,13 and 14 interactions are EXCLUDED FROM CUSTOM NONBONDED FORCE
                if eps._value != 0.0:
            print p1,p2,sig,eps
                        sig14=sqrt(LJset[p1][0]*LJset[p2][0])
                        eps14=sqrt(LJset[p1][1]*LJset[p2][1])
                        nonbonded_force.setExceptionParameters(i, p1, p2, q, sig14, eps)
        return system

pdb = PDBFile('anilin.pdb')

modeller = Modeller(pdb.topology, pdb.positions)
forcefield = ForceField('anilin.xml','tip3p.xml')
modeller.addSolvent(forcefield,boxSize=Vec3(2.0, 2.0, 2.0)*nanometers)

system = forcefield.createSystem(modeller.topology, nonbondedMethod=NoCutoff,  constraints=HBonds)
system = OPLS_LJ(system)

integrator = LangevinIntegrator(300*kelvin, 1/picosecond,  0.001*picoseconds)

simulation = Simulation(modeller.topology, system, integrator)
simulation.context.setPositions(modeller.positions)
## MINIMIZING 
simulation.minimizeEnergy(maxIterations=300)
position = simulation.context.getState(getPositions=True).getPositions()
PDBFile.writeFile(simulation.topology, position, open('boxmin.pdb', 'w'))
## Setting up simulation for 10ps
simulation.reporters.append(PDBReporter('boxoutput.pdb', 1))
simulation.reporters.append(StateDataReporter(stdout, 1, step=True,potentialEnergy=True, temperature=True))
simulation.step(1)
```
## Minimized structure

min_box
## First step of MD simulation

step1

Any suggestions are highly appreciated.

peastman:
Fixes #1390.

### DIFF
--- a/wrappers/python/simtk/openmm/app/simulation.py
+++ b/wrappers/python/simtk/openmm/app/simulation.py
@@ -183,6 +183,9 @@ class Simulation(object):
         nextReport = [None]*len(self.reporters)
         while self.currentStep < endStep and (endTime is None or datetime.now() < endTime):
             nextSteps = endStep-self.currentStep
+            
+            # Find when the next report will happen.
+            
             anyReport = False
             for i, reporter in enumerate(self.reporters):
                 nextReport[i] = reporter.describeNextReport(self)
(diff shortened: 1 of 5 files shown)


mpharrigan:
Is there a way to do this without breaking the API. For example, mdtraj contains some custom reporters. I don't know if anyone uses them, but they would no longer work in this case. I'm not immediately sure how to write custom reporters that will work with either api besides querying the openmm version number before returning

peastman:
They'll continue to work exactly as they did.  If a reporter returns a five element tuple, it treats that as equivalent to returning `None` for the sixth element.

At some point I might replace that whole reporter API with something that's more extensible.  But even then, I'd make sure existing reporters continue to work.

mpharrigan:
Sorry! I see the logic for distinguishing between 5 or 6 return values now :)

wrappers/python/simtk/openmm/app/simulation.py
```diff
@@ -198,25 +201,57 @@ def _simulate(self, endStep=None, endTime=None):
...
+                        if not report[1]:
+                            either.append((reporter, report))
+                        elif wantWrap:
+                            wrapped.append((reporter, report))
+                        else:
+                            unwrapped.append((reporter, report))
+                if len(wrapped) > len(unwrapped):
+                    wrapped += either
+                else:
+                    unwrapped += either
```
stefdoerr: From an implementation perspective it seems redundant to recalculate the wrapped/unwrapped reporter lists at each time step, no? I would just move it out of the loop.

stefdoerr: Oh right, I see.

PR #1906, opened 2017-09-21
Support for Volta

peastman:


peastman:
run jenkins

peastman:
@swails I think something's up with Jenkins.  It seemed to have hung before.  I told it to rerun the build, and now it claims to be building, but the "details" link goes to a local network address (192.168.1.3) rather than a real address.

peastman:
I don't know what's up with Jenkins, but I tested the CUDA platform offline, so I'll go ahead and merge this.

swails:
I changed the published URL of Jenkins so that it could find the Windows VM I was trying to attach as a slave (so I can do Windows builds on it).

Can you go back into the webhooks that are registered on this repo and change "192.168.1.3:8080" to "www.jasonswails.com:8888" wherever you see it?

peastman:
Looks like you may have already fixed the problem without me needing to change anything.  It lists a couple of failed deliveries on 9/23, but that it started working again on 9/24.  And it already has the correct URL listed there.  Thanks!

swails:
The change only had to be temporary since the windows vm used the published address of the Jenkins server to initiate the connection. 

It obviously couldn't connect over the internet since whatever port it tried to use is blocked by my router, so I had to change it to the local address to initiate the connection.

I wasn't expecting that change to get propagated out to repos that were already linked with it, but I'm glad they changed back.

Issue #1907, opened 2017-09-22
Labeling of chains in mmCIF files: entity_id vs asym_id


JoaoRodrigues:
Hi there,

Reading/writing mmCIF files with multiple chains is a bit troublesome with OpenMM. I have a 16-chain PDB file that I converted to a mmCIF file using `PDBxFile.writeFile`. When I read this file I correctly get a topology with 16 chains, and the resulting `.cif` file labels the residues correctly. However, re-reading this mmCIF file produces a topology (and subsequent files) containing one chain only. 

I narrowed this down, I think, to a problem in the `PDBxFile` class, specifically in the mmCIF dictionary field it reads the chain information from to create `Chains` in the topology. While the PDB format chains are read to `asym_id`, the reader expects to have `entity_id` fields for chains. I am not super knowledgeable of the mmCIF format, but this should be somewhat consistent at least within OpenMM. I'd expect writing a file from one `Topology` to yield the same upon re-reading.

The fix could be as simple as using `asymIdCol` in the `Chain` creation if-clause, but I'm not sure how this would play with the format specifications... In Biopython we changed our mmCIF parser to use `auth.asym_id` since that matches the chains in PDB files, but again, my experience with these files is limited and this was based on a few observations. What is your opinion?

P.S. I can reproduce this easily, @peastman I can e-mail you a test file if necessary ..

peastman:
Thanks!  I agree with your analysis.  And I see that http://mmcif.wwpdb.org/docs/pdb_to_pdbx_correspondences.html#ATOMP now recommends `auth_asym_id` as the equivalent for chain ID.  I think that may have changed since I originally wrote `PDBxFile`.

If you feel inclined to write a PR fixing this, that would be great.  If not, I'll try to fix it soon.  We should also add a test to `TestPdbxFile.py` that tests saving and reloading a structure with multiple chains.

JoaoRodrigues:
Will do, perfect, thanks!

peastman:
Fixed by #1908.

PR #1908, opened 2017-09-24
Fixed chain labeling in mmCIF files IO.

JoaoRodrigues:
Addresses issue #1907 

multiple chain mmCIF files are now written with the same chain composition as they were read. I tested in on some of my mmCIF files and the results are correct. The headers also look good. This fix looks super simple, so please let me know if there is some other place where this should be changed/checked.


peastman:
It identifies each atom by the key `(residue, chain, atom)`.  You switched it to get the chain ID from `auth_asym_id`, but for constructing that key it's still using `label_asym_id` as the chain identifier:

```python
atomKey = ((row[resIdCol], row[asymIdCol], row[atomNameCol]))
```

We should change it there too (and in fact just get rid of the `asymIdCol` variable).  We also should add a test case to verify this works correctly: take a topology containing multiple chains, write it to a file, read it back in, and verify the topology is still correct.  If you prefer, I can write that test and send it in a PR to your branch.

JoaoRodrigues:
Hi @peastman 

I removed references to `asymIdCol` and replaced them by `chainIdCol` where appropriate. This led to a problem when two consecutive residues have the same name and chain and residue id (as read by the mmCIF parser, from label_seq_id). I fixed this by changing all 'label_*' reads (label_seq_id, label_atom_id, label_asym_id) to 'auth_*'. 

This passes all tests (I added one for multiple chain parsing/writing) and ensures better compatibility of you read a PDB file and write a CIF. I ended up removing variables `asymIdCol` and `resIdCol`, since they are replaced by `chainIdCol` and `resNumCol`, respectively.

peastman:
> I fixed this by changing all 'label_' reads (label_seq_id, label_atom_id, label_asym_id) to 'auth_'.

Are you sure that's valid?  According to http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v40.dic/Categories/atom_site.html, `auth_atom_id` and `auth_seq_id` are not mandatory fields, so we can't assume they'll always exist.  But then why is `auth_asym_id` mandatory when the others aren't?  That doesn't make a lot of sense to me, so maybe it's just an error in the data dictionary...

http://mmcif.wwpdb.org/docs/pdb_to_pdbx_correspondences.html#ATOMP does list `auth_atom_id` and `auth_seq_id` as the recommended equivalents for atom name and residue number.  I wish the documentation were clearer...

JoaoRodrigues:
I followed the correspondences table, assuming that it's better to preserve PDB-mmCIF matching (otherwise you read a PDB, write a CIF, read the CIF and have different chains/residues than the original PDB). If these values are missing, indeed there is a problem. The logic of defining new residues must be changed to rely on some more information. I had the issue that parsing the `triclinic.pdbx` file in the test suite of the Python wrappers after removing `asymIdCol`, since it is used to check for a new residue (it's set to None on a new chain). Changing to read the residue numbers from the auth column made it work, because the unique residue id is built using those (which are otherwise '.').

Also, some tests failed because the test file I created wasn't there, while others pass.. ?

peastman:
Yeah, it's probably reasonable to go by what they say in the correspondence table.  If we start encountering files that don't have those fields, we can reconsider in the future.  Hopefully it won't be an issue.

Files used for tests need to get copied over to the build directory.  It automatically copies anything with a long list of filename extensions, but apparently `.cif` wasn't one of them.  Add a line to this block: https://github.com/pandegroup/openmm/blob/master/wrappers/python/CMakeLists.txt#L49-L73

JoaoRodrigues:
I'll be running some simulations so I can keep track if something is going wrong. 

As for the test suite, maybe I can just rename it to `.pdbx` to be consistent with everything else?

peastman:
That would be fine too.  Both extensions seem to get used for the format.

JoaoRodrigues:
Did both. The unit tests already used `pdbx`, which is also the name of the class (PDBxFile). However, since `.cif` is the format the PDB uses in their files, added it to the `CMakeLists` file just in case in the future someone adds a cif file..

Thanks a lot @peastman !

peastman:
The tests are still failing.  You renamed the file but didn't change the code that refers to it, so now they don't match.

peastman:
Looks good.  Thanks!

PR #1919, opened 2017-10-22
Add documentation for <Include> ffxml tags.

jchodera:
Document the newly-added ffxml `<Include>` tag added in #1785.


peastman:
We already have documentation on it.  See lines 2943-2966 in that file.

jchodera:
Huh, I had totally missed that when searching through the docs. Thanks!

PR #1924, opened 2017-11-03
Fixed a segfault

### LINKED ISSUE #1920: Context seems to be left in an inconsistent state after NaN
In `openmmtools` we have implemented a system to recover from NaNs. Briefly, if after integrating the energy is NaN, we restore positions/box vectors in the last working state, we randomize the velocities and we try again for few times before giving up and interrupt the simulation. This works most of the time, but I've noticed few weird things.

I'm running a protein-ligand system in a dodecahedral box with `openmm 7.1.1` and CUDA platform on a GTX1080. After 4 consecutive failed recovery attempts (using a `LangevinIntegrator`), I tried these:

1. If I delete the `Context` object and create a new one in the last working state (i.e., with last working positions/velocities/box vectors), the integration does not result in a NaN anymore.
2. If I call [`context.reinitialize()`](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.Context.html#simtk.openmm.openmm.Context.reinitialize) and set the last working state, the integration does not result in a NaN.
3. I serialized `System`, `Integrator` and `State` objects and created a checkpoint in the last working state. I can reproduce the persistent NaN if I load the checkpoint, but I can't if I recreate the `Context` using simply the serialized `State`.

This makes me think that the first NaN can sometimes trigger an inconsistent internal state in the `Context` object or on the GPU. Is this known/to be expected? And are there workarounds that avoid recreating the `Context`?

peastman:
Fixes #1920.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -8021,6 +8021,7 @@ void CudaIntegrateCustomStepKernel::getGlobalVariables(ContextImpl& context, vec
         // The data structures haven't been created yet, so just return the list of values that was given earlier.
         
         values = initialGlobalVariables;
+        return;
     }
     values.resize(numGlobalVariables);
     for (int i = 0; i < numGlobalVariables; i++)
--- a/platforms/opencl/src/OpenCLKernels.cpp
+++ b/platforms/opencl/src/OpenCLKernels.cpp
@@ -8380,6 +8380,7 @@ void OpenCLIntegrateCustomStepKernel::getGlobalVariables(ContextImpl& context, v
         // The data structures haven't been created yet, so just return the list of values that was given earlier.
         
         values = initialGlobalVariables;
+        return;
     }
     values.resize(numGlobalVariables);
     for (int i = 0; i < numGlobalVariables; i++)


swails:
@peastman - have you looked at the Jenkins failures at all?  The CUDA drivers are working fine, so this is a real failure of some kind on the Jenkins machine

peastman:
No, thanks for pointing that out.  It had been failing long enough that I stopped paying attention to it.

The error it's getting is

```
E       Exception: Error launching CUDA compiler: 512
E       /tmp/openmmTempKernel0x30542b0_27583.cu(775): error: identifier "__ballot_sync" is undefined
```

`__ballot_sync` was added in CUDA 9.0, and we only use it if the driver version is >= 9000.  So it looks like perhaps you have the CUDA 9.0 driver installed, but you have it using the CUDA 8.0 version of nvcc?

swails:
My package manager handles both the driver and nvidia toolkit.  My driver version is at 384.90 right now, but the nvcc is from CUDA 8.0.44.

The 9.0 toolkit hasn't been added at all to the package manager (and Gentoo usually adds new versions very quickly and masks them for general use, but that option is not even available to me at the moment).

But even before I used Gentoo I always treated the driver and toolkits separately.

swails:
Also, all but one of the failures were the same as this one.  The only recent "false" failure from a delayed restart was build 73.

jchodera:
Shouldn't we upgrade to CUDA 9.0 since OpenMM 7.2 will be built against 9.0?

PR #1932, opened 2017-11-10
Require atoms and residues to be contiguous

### LINKED ISSUE #1931: Bug or operator error in adding atoms/virtual sites to topology/system "out of order"
For reasons I probably don't need to go into here (but relating to adding off-site charges in SMIRNOFF), I'm working through a test example where I try to turn TIP3P water into TIP4P water without using `Modeller`, and I ran into an issue where atoms get reordered in a way which was unexpected to me. **Specifically, it seems that atoms must be numerically in order within the residues in which they occur** (e.g. I can't have atoms 0, 1, 2, and 648 in residue 0, and 3, 4, 5, and 649 in residue 1, etc.), and if I deviate from this, atoms end up having the wrong coordinates.

I'm trying to figure out whether this is a bug or just a "naive user" problem. If it may be a bug, I'm happy to create and post a minimal example to reproduce the problem, though I suspect it's just a "naive user" problem and I need to work around the issue.

Here's what I did, roughly (as it relates to virtual sites and my problem):
- Load a PDB file for TIP3P water (216 residues/648 atoms)
- Create a system using tip3p.xml and  the standard OpenMM `forcefield.createSystem`
- Loop through the forces in the system, and for the `NonbondedForce` add a new virtual site in each residue (when I do so) a new particle at the location of this site; naturally these new particles have indices at the end of the system
- Add relevant exceptions
- Create dummy positions at the end of the position array (entries 648 and onwards)
- Modify Topology to add a new virtual site atom (with indices 648 and onwards) in each residue, corresponding to the changes I made in the System. At this point, I can confirm that atoms in my Topology are in the same order as those in my System and have the intended coordinates
- Set up a simulation and compute virtual sites; output coordinates and check that they are as intended
- Find that they are clearly incorrect

What I had created in this process was, for the first water molecule (for example), a residue consisting of atoms 0, 1, and 2, and particle 648 which was a virtual site. But what I find when I output the results of the procedure (e.g. by writing to a PDB) is that the virtual site ends up with the coordinates which had been assigned to the oxygen of the second water molecule (and the oxygen gets the coordinates of the next hydrogen, etc.) -- basically the virtual acts as if it is atom 4 (and indeed this is what gets output to the PDB) and bumps the indices of all subsequent atoms.

I must admit, what I'm doing is a bit funky (normally presumably one would have atoms that are in numerical order within their residues) which makes me suspect I'm doing something that shouldn't work. However, I haven't run across anywhere in the docs that indicates that atom indices must be consecutive within their residues, so I figured I should at least ask. Also, the current behavior (where this is a silent failure) seems less than ideal.

peastman:
Fixes #1931.  If you try to add atoms to a residue or residues to a chain out of order, it throws an exception.

### DIFF
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -154,6 +154,8 @@ class Topology(object):
         Residue
              the newly created Residue
         """
+        if len(chain._residues) > 0 and self._numResidues != chain._residues[-1].index+1:
+            raise ValueError('All residues within a chain must be contiguous')
         if id is None:
             id = str(self._numResidues+1)
         residue = Residue(name, self._numResidues, chain, id)
@@ -181,6 +183,8 @@ class Topology(object):
         Atom
              the newly created Atom
         """
+        if len(residue._atoms) > 0 and self._numAtoms != residue._atoms[-1].index+1:
+            raise ValueError('All atoms within a residue must be contiguous')
         if id is None:
             id = str(self._numAtoms+1)
         atom = Atom(name, element, self._numAtoms, residue, id)


peastman:
Something strange happened with jenkins.  It failed in just a few seconds with an error in the yaml parser.

PR #1933, opened 2017-11-12
Test both CUDA 8 and CUDA 9 (and have it work concurrently).

swails:
Right now the Jenkins recipe is configured to test both CUDA 8 and CUDA 9.  However, without this PR both builds will just use CUDA 8.0


peastman:
Thanks!

PR #1942, opened 2017-11-28
Added new Amber and CHARMM force fields

peastman:
At long last!  Look this over and see if everything looks OK: the files I've chosen to include, the way I've named and organized them, and the documentation.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -314,7 +314,7 @@ A First Example
 ***************
 
 Let’s begin with our first example of an OpenMM script. It loads a PDB file
-called :file:`input.pdb` that defines a biomolecular system, parameterizes it using the Amber99SB force field and TIP3P water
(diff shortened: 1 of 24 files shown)


peastman:
@swails Jenkins builds seem to be broken again (but only the CUDA 9.0 ones).

jchodera:
Note there are just placeholder references for the ipolq forcefields. @rafwiewiora will fix this, but you'll need to copy over the new files.

cc: https://github.com/pandegroup/openmm/issues/1611#issuecomment-347668514

peastman:
No problem, I'll do that when they're ready.

peastman:
Is it ok if I merge this?  It's blocking other things.  Once @rafwiewiora fixes the references we can change them in a separate PR.

Does anyone see any other problems with this?

jchodera:
Go for it.

Issue #1543, opened 2016-07-20
Serialize topology object


andrrizzi:
As far as I understand, the `Topology` object cannot be serialized through `XMLSerializer` like `System`s. Would it be possible/make sense to add this?

Currently, the only "safe" way I could find to serialize was to go through `mdtraj`, transform the `Topology` into a `pandas` `DataFrame` object, and serialize that as JSON string. This way I'll be able to read old files as long as the API of `mdtraj.Topology.to_dataframe()` doesn't change, and I don't have to rely on `pickle` which comes with poor future compatibility once the internal implementation changes. It would be nice to have a more robust way to do this though.

proteneer:
The difference is that the System class is a C++ object, and the Topology is a python-specific class. The core XML serializers are all written in C++ for C++ objects. No such mechanism currently exists for Python objects. But maybe @peastman has better ideas :)

mpharrigan:
You're right that `pickle` has poor future (forwards) compatibility. But it does have backwards compatibility, which is probably more important here. The use case for forwards compatibility is serializing something from python3 and loading it in python2. This is still a solvable problem: on py3 use
```python
pickle.dump(obj, f, protocol=2)
```

andrrizzi:
Thank you all for your answers!

Sorry if I wasn't clear. I'm actually more worried about future changes in `Topology` internal implementation. To make a minimal example

```python
class TopologyMini(object):

    def __init__(self):
        self._chains = []

    def getNumChains(self):
        return len(self._chains)

top_mini = TopologyMini()
top_mini.getNumChains()
with open('top.pickle', 'w') as f:
    pickle.dump(top_mini, f)
```
if in the future we change the internal representation of `TopologyMini` to
```python
class TopologyMini(object):

    def __init__(self):
        self._chains = []
        self._numChains = 0

    def getNumChains(self):
        return self._numChains

with open('top.pickle', 'r') as f:
    top_mini = pickle.load(f)
top_mini.getNumChains()  # Fail even if the API is the same!
```
the last line raises `AttributeError: 'TopologyMini' object has no attribute '_numChains'`.

This is why I think it would be nice to have a representation of the `Topology` object state that is independent of its implementation. For simple `Python` objects I believe this is as simple as overriding the `__getstate__` and `__setstate__` methods.

mpharrigan:
Sorry! Yes, that is a good point

jchodera:
[...]

PR #1948, opened 2017-12-01
Fix Bond pickling in Python 2.7

Lnaden:
The pickle round trip for the Bond object does not preserve the `type` and `order` parameters in Python 2.7.

Near as I can tell, its because the parent `namedtuple` already has a __getstate__ method, so the default of returning the `__dict__` is never given, so `__setstate__` is not called (as default). This leads to pickle restoring the state, but never invoking the `__new__` statement, resulting in `type` and `order` being lost.

 This PR fixes this problem in python 2.7, without breaking python 3.x versions.

xref: mdtraj/mdtraj#1308 and #1543

### DIFF
--- a/wrappers/python/simtk/openmm/app/topology.py
+++ b/wrappers/python/simtk/openmm/app/topology.py
@@ -465,6 +465,14 @@ class Bond(namedtuple('Bond', ['atom1', 'atom2'])):
         "Support for pickle protocol 2: http://docs.python.org/2/library/pickle.html#pickling-and-unpickling-normal-class-instances"
         return self[0], self[1], self.type, self.order
 
+    def __getstate__(self):
+        """
+        Additional support for pickle since parent class implements its own __getstate__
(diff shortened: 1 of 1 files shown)


jchodera:
@peastman : Shouldn't we squeeze this in?

peastman:
Squeeze it into what?  We should fix it in 7.2, but there's no need for it to be in the beta.

Lnaden:
I agree it should be in 7.2, but not forced into the beta build from today. This is just a bugfix, not a feature.

jchodera:
@peastman : Then how about we label the issues/PRs that will get rolled into the final release candidate using the built-in "Milestone" feature so we don't need to read your mind about what you think will be merged/fixed by then?  :)

image

peastman:
Or even better: tell me which ones you think should be included!  I currently have none at all on my list.  That's not because I don't intend to include any.  It's because I've barely started the project of going through the tracker to come up with a list.

PR #1951, opened 2017-12-04
Fix Quantity.__setitem__ bug with slice assignment

### LINKED ISSUE #1940: Bug in Quantity.__setitem__ when assigning slices
I believe there may be a bug in `Quantity.__setitem__(key, value)` when the `key` is a slice. This code
```python
from simtk import unit as u
import numpy as np

a = np.array(range(10)) * u.kelvin
b = np.array([3, 4]) * u.kelvin
a[4:6] = b
```
results in
```
  File "prova.py", line 6, in <module>
    a[4:6] = b
  File "/Users/andrea/miniconda/envs/yankdev/lib/python3.5/site-packages/simtk/unit/quantity.py", line 741, in __setitem__
    self[i] = value[i]
  File "/Users/andrea/miniconda/envs/yankdev/lib/python3.5/site-packages/simtk/unit/quantity.py", line 733, in __getitem__
    assert not is_quantity(self._value[key])
IndexError: index 4 is out of bounds for axis 0 with size 2
```
I think the problem is [here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/unit/quantity.py#L740-L741)
```python
for i in range(*indices):
    self[i] = value[i]
```
Should it be something like
```python
for value_idx, self_idx in enumerate(range(*indices)):
    self[self_idx] = value[value_idx]
```

andrrizzi:
Fix #1940: Bug in `Quantity.__setitem__` when assigning slices. I've also added a test case.

### DIFF
--- a/wrappers/python/simtk/unit/quantity.py
+++ b/wrappers/python/simtk/unit/quantity.py
@@ -737,8 +737,8 @@ class Quantity(object):
         # Delegate slices to one-at-a time ___setitem___
         if isinstance(key, slice): # slice
             indices = key.indices(len(self))
-            for i in range(*indices):
-                self[i] = value[i]
+            for value_idx, self_idx in enumerate(range(*indices)):
+                self[self_idx] = value[value_idx]
         else: # single index
             # Check unit compatibility
             if self.unit.is_dimensionless() and is_dimensionless(value):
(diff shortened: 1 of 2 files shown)


peastman:
Looks good.  I'll merge it once the tests finish.

PR #1967, opened 2018-01-08
Added link to openmm-forcefields in the manual

peastman:
@jchodera I just linked to the main page for the github repository.  You'll probably want to update the README to provide more information to users arriving there, such as a list of force fields and installation instructions.


jchodera:
Thanks! Will do!

PR #1996, opened 2018-02-18
Add virtual get/setTemperature functions to Integrator base class.

tristanic:
I believe this would also help with #1440.

My reason for wanting this is that I've written a lightweight Python/C++ wrapper designed to take the pointer to an existing `Context`, and push Integrator stepping and minimisation off to a std::thread leaving Python free to do other things. The basics of this work nicely, but after minimising it would be nice to be able to run `Context.setVelocitiesToTemperature()` within the wrapper itself. That requires getting the temperature in the first place, so it would be great to be able to simply do `context->getIntegrator().getTemperature()`. The suggested modification adds the necessary virtual functions to Integrator.h.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -1859,8 +1859,10 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
                     if (useCudaFFT) {
                         cufftSetStream(fftForward, pmeStream);
                         cufftSetStream(fftBackward, pmeStream);
-                        cufftSetStream(dispersionFftForward, pmeStream);
-                        cufftSetStream(dispersionFftBackward, pmeStream);
+                        if (doLJPME) {
+                            cufftSetStream(dispersionFftForward, pmeStream);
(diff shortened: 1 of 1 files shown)


peastman:
I don't think this is quite the right approach.  For one thing, some integrators don't have a temperature, so it would be confusing for them to have `getTemperature()` and `setTemperature()` methods.  This also isn't going to work correctly with CustomIntegrators.  Many of those do involve a temperature, but they won't interface with this method, which will be even more confusing.

We had a discussion about this issue a while back: https://github.com/pandegroup/openmm/issues/1440.  This might be a good time to revive it.

tristanic:
I guess my question is: is there any feasible situation where it makes sense to have different temperatures in the integrator and barostat? Would it make more sense to define it in a single location (i.e. the Context object) and have the former fetch it from there?

 
 
Tristan Croll
Research Fellow

jchodera:
Not that I can think of. I'm not sure if the architecture supports this, but having the temperature and pressure centrally defined would certainly simplify things. But it also might break a lot of code in the wild.

PR #1998, opened 2018-02-21
Added switchDistance option to ForceField.createSystem()

### LINKED ISSUE #1994: Have ForceField.createSystem() accept switchDistance argument, and raise exception if some arguments are unused
cc: https://github.com/choderalab/openmm-forcefields/pull/61#issuecomment-366043877

It would be useful to have `ForceField.createSystem()` accept a `switchDistance` argument (as `CharmmPsfFile.createSystem()` does) to pass on to [`LennardJonesGenerator`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L2342) and [`NonbondedGenerator`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L2282).

Also, it seems like we should really have a mechanism for force generators to flag which options they actually used, and raise an exception if some option was provided that isn't actually used anywhere. That will prevent programming errors.

peastman:
Fixes #1994.

I debated about whether this should apply to `<CustomNonbondedForce>` tags.  It supports a switching function, but we have no way of knowing what kind of interaction it's being used for.  In the end I decided it was safest not to.  Just because you want a switching function on LJ interactions, that doesn't necessarily mean you also want it on all CustomNonbondedForces, no matter what purpose they're being used for.  So I made this option only apply to `<NonbondedForce>` and `<LennardJonesForce>` tags.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -792,6 +792,16 @@ The error tolerance is roughly equal to the fractional error in the forces due
 to truncating the Ewald summation.  If you do not specify it, a default value of
 0.0005 is used.
 
+Another optional parameter when using a cutoff is :code:`switchDistance`.  This
+causes Lennard-Jones interactions to smoothly go to zero over some finite range,
+rather than being sharply truncated at the cutoff distance.  This can improve
+energy conservation.  To use it, specify a distance at which the interactions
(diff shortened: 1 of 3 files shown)


jchodera:
Awesome, thanks!

Issue #2003, opened 2018-03-02
Disulfide bonds are broken in Amber14


peastman:
There's a critical bug in the Amber14 force field in OpenMM 7.2: disulfide bonds don't work.  The CYX template is missing the `<ExternalBond>` tag for SG.

Given the seriousness of this and the fact that 7.2 is already out, I'm inclined to add the missing tag by hand and push out 7.2.1 ASAP.

jchodera:
We didn't have any disulfide containing proteins in our tests with @rafwiewiora, so we must have missed this. 

I agree that adding the tag by hand is a good idea for now. Can you provide an example disulfide containing PDB we can add to the tests so we can get this fixed in openmm-forcefields too?

I'm stuck in meetings this afternoon and unfortunately won't be able to join for the call, but could look into this ASAP.

peastman:
Here's the forum thread where the problem came up:

https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=8704&p=24008

The user there posted a file.  But really, absolutely anything containing a disulfide will work equally well.  Since we ran into exactly the same problem in CHARMM36, I assume you added something with disulfides to the tests for that?

By the way, it would be really useful if you monitored the forum.  A lot of problems first get reported there, not here.

peastman:
Fixed by #2004.

jchodera:
Thanks! I found the problem (see https://github.com/ParmEd/ParmEd/issues/964). 

There may be other residues with this issue, so I'll regenerate these tonight and we can `diff` the files to see if another update is needed.

jchodera:
@peastman : can we keep this open until we're sure there are no other residues missing connections?

For example, I bet the same problem exists in CHARMM.

jchodera:
Correction: CHARMM handles disulfides through patches, so I don't think that's a worry.

peastman:
Reopening until we're sure everything is fixed.

jchodera:
@peastman : See this PR: https://github.com/choderalab/openmm-forcefields/pull/62

peastman:
A release candidate with the changes is now posted.

peastman:
7.2.1 is now released.

PR #2005, opened 2018-03-03
Regenerated files to fix missing external bonds

peastman:
This is a more complete fix for #2003.  This adds the files that were regenerated in https://github.com/choderalab/openmm-forcefields/pull/62.  The changes to protein.ff14SB.xml are simple, but nearly every line of protein.ff15ipq.xml has changed.  However, this seems to be nothing more than reordering the attributes within each tag.

### DIFF
--- a/wrappers/python/simtk/openmm/app/data/amber14/protein.ff14SB.xml
+++ b/wrappers/python/simtk/openmm/app/data/amber14/protein.ff14SB.xml
@@ -1,7 +1,7 @@
 <ForceField>
   <Info>
-    <DateGenerated>2017-11-25</DateGenerated>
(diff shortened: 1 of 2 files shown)


peastman:
AppVeyor has been failing recently while trying to configure the VM.  It looks like it fails on the line

```
choco install -y doxygen.portable swig > null
```

It's not relevant for this PR, but we should look into it.

jchodera:
Good catch about the attribute reordering. I can fix that too.

jchodera:
(At least it future iterations. It's too late to fix it for this iteration.)

jchodera:
@peastman : Don't forget to merge these fixes into `master` too!

PR #2006, opened 2018-03-03
Updated version number to 7.2.1

peastman:


peastman:
The test failure is just a stochastic one, not related to this change.

PR #2008, opened 2018-03-04
handles virtual_sites2 and nbfix in gromacs topology

sunhwan:


wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -840,18 +882,25 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
...
+
                     if len(fields) > 6:
                         q = float(fields[6])
                     else:
                         q = float(params[4])
 
-                    if self._defaults[1] == '1':
+                    if has_nbfix_terms:
+                        if self._defaults[1] != '2':
+                            raise NotImplemented
```
peastman: Could you add a descriptive string to the exception so the user will know what aspect of their file isn't implemented?

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -932,6 +1052,88 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
             lj.setNonbondedMethod(methodMap[nonbondedMethod])
             lj.setCutoffDistance(nonbondedCutoff)
 
+        if has_nbfix_terms:
+            if self._defaults[1] != '2':
+                raise NotImplemented
```
peastman: Same as above.

peastman:
This looks great!  I made one minor suggestion, but other than that I think this should be ready to merge.

wrappers/python/tests/TestGromacsTopFile.py
```diff
@@ -151,6 +151,23 @@ def test_HydrogenMass(self):
...
+        top = GromacsTopFile('systems/bnz.top')
+        gro = GromacsGroFile('systems/bnz.gro')
+        system = top.createSystem()
+
+        self.assertEqual(26, system.getNumParticles())
+        self.assertEqual(1, len(top._moleculeTypes['BENX'].vsites2))
+
+        context = Context(system, VerletIntegrator(1*femtosecond),
+                          Platform.getPlatformByName('Reference'))
+        context.setPositions(gro.positions)
```
peastman: I just realized this doesn't actually test the virtual site, because it's just using the positions loaded from the file.  After calling `setPositions()`, add the line

```python
context.computeVirtualSites()
```

wrappers/python/simtk/openmm/app/gromacstopfile.py
peastman: I didn't mean to add comments in the code.  I meant add a description to the exception so the user will know what the problem is.

```python
raise NotImplemented('Files with ... are not currently supported.')
```

sunhwan: Thanks for the comment. I have updated the description.

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -840,18 +882,25 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
...
+
                     if len(fields) > 6:
                         q = float(fields[6])
                     else:
                         q = float(params[4])
 
-                    if self._defaults[1] == '1':
+                    if has_nbfix_terms:
+                        if self._defaults[1] != '2':
+                            raise NotImplemented('NBFIX terms with LB combination rule is not yet supported')
```
swails: Don't raise `NotImplemented` -- it's for returning from operator overloads when they're not implemented.

```python
def __eq__(self, other):
    if not hasattr(other, 'some_attribute'):
        return NotImplemented
    return self.some_attribute == other.some_attribute
```

Raise `NotImplementedError()` instead.

wrappers/python/simtk/openmm/app/gromacstopfile.py
```diff
@@ -932,6 +1052,88 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*unit.nan
             lj.setNonbondedMethod(methodMap[nonbondedMethod])
             lj.setCutoffDistance(nonbondedCutoff)
 
+        if has_nbfix_terms:
+            if self._defaults[1] != '2':
+                raise NotImplemented('NBFIX terms with LB combination rule is not yet supported')
```
swails: `raise NotImplementedError()` instead

swails:
Actually, `NotImplemented` isn't even an exception:

```python
>>> raise NotImplementedError('This works')
NotImplementedError: This works
>>> raise NotImplemented('This doesnt')
TypeError: 'NotImplementedType' object is not callable
>>> isinstance(NotImplemented, Exception)
False
>>> isinstance(NotImplemented, BaseException)
False
```

So that code would fail, anyway.

peastman:
This looks good now, so I'll go ahead and merge it.  Thanks for implementing it!

PR #2014, opened 2018-03-12
CustomNonbondedForce with interaction groups uses neighbor lists

peastman:
This is an attempt at addressing #1765.  @jchodera could you try it and see how it works for you?

There are a couple of limitations to be aware of.  I've only implemented it for the CUDA and OpenCL platforms.  I could potentially do a similar optimization for CPU, but it would take a different algorithm.

This optimization is only enabled if the system also includes a standard nonbonded force (either a NonbondedForce or a CustomNonbondedForce without interaction groups), such that CudaNonbondedUtilities/OpenCLNonbondedUtilities is building the standard neighbor list.  CustomNonbondedForce is actually building its own neighbor list, but it relies on CudaNonbondedUtilities to figure out when the neighbor list needs to be rebuilt.

PR #2022, opened 2018-03-16
Bug fix

peastman:
Fixes a bug introduced by #2014.

### DIFF
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -266,7 +266,7 @@ void CudaNonbondedUtilities::initialize(const System& system) {
         blockSorter = new CudaSort(context, new BlockSortTrait(context.getUseDoublePrecision()), numAtomBlocks);
         vector<unsigned int> count(2, 0);
         interactionCount.upload(count);
-        rebuildNeighborList.upload(count);
+        rebuildNeighborList.upload(&count[0]);
     }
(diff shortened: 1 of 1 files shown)


jchodera:
All tests pass! But do you want to add a test that would have caught this issue? I'm surprised it slipped by.

peastman:
That's because jenkins is down, so CUDA tests aren't getting run.  @swails any idea what's happened to it?

swails:
Sorry about missing this!  For the future, assume I won't see mentions in GH at all, or at least not for awhile.  Please feel free to email me directly -- I'll get those and respond.

The hard drive where all the Jenkins files were hosted crapped the bed a few months ago.  I brought it back up, but everything that was set up there was wiped away.

Fortunately the script in ``devtools/ci/jenkins/install.sh`` contains the vast majority of the work that went into getting it to work.  So @peastman, if you wanted to get Jenkins back up and working again on this repository I'd be happy to help.  I'll need another API key from an account with write permission on this repo so it can post comments on PRs and statuses.

Email me if you want to do this and we can get it back up pretty quickly.

swails:
Scratch that, @peastman -- I found the last API key you sent me and just reused that.  I need to reboot my computer following a driver update (that's why all the CUDA and OpenCL tests are crapping out in the meaningless PR I just opened), but after that the CUDA and OpenCL tests should be back.

peastman:
Wonderful, thanks!

Issue #2043, opened 2018-04-10
Error when setting switchDistance in createSystem


JoaoRodrigues:
There seems to be a small typo in the `forcefield.py` file around line 2451 ([link](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L2451)). When checking for `switchDistance`, it tries to set parameters on a non-existing variable `force`, instead of `self.force`. This breaks usage of switching functions, e.g. in CHARMM.

peastman:
Fixed by #2044.

PR #2044, opened 2018-04-10
Fixed reference to self.force when using a switch function for LJ

JoaoRodrigues:
Addresses issue #2043


peastman:
Looks good, thanks!

PR #2051, opened 2018-04-20
Fixed incorrect synchronization on Volta 

### LINKED ISSUE #2046: TestCudaNonbondedForce error on Nvidia Volta architecture
Hi Peter,

The following error occurs with the current master HEAD if running on V100/GV100 or Titan V (on Pascal P100 and previous are successful)
./TestCudaNonbondedForce "single"
exception: Assertion failure at TestCustomNonbondedForce.h:1039.  Expected 4280.26, found 4289.31
(Same error for the openCL version on Nvidia GPUs)

The commit 2305f05 from PR 2013 is  the last commit that can pass the test on Volta GPUs.
What is introduced since then? Related to the neighbourlist ?

A simple fix is to modify the CudaContex.cpp file:
Adding:
>    if(major == 7){
>       major = 6;
>       minor = 0;
>    }

just before:
>     gpuArchitecture = intToString(major)+intToString(minor);

This strongly indicates that the issue is related to the independent thread scheduling on Volta GPUs.
Do you recall any code that is not ported to Volta/Cuda9 yet? Normally this happens when code is doing shfl or shared memory based reduction. But I saw you have already ported shfl to shfl_sync.

Ke Li

peastman:
Fixes #2046.

### DIFF
--- a/platforms/cuda/src/kernels/customNonbondedGroups.cu
+++ b/platforms/cuda/src/kernels/customNonbondedGroups.cu
@@ -156,6 +156,7 @@ extern "C" __global__  void buildNeighborList(int* __restrict__ rebuildNeighborL
         int rangeStop = rangeStart + reduceMax(rangeEnd-rangeStart);
         SYNC_WARPS;
         for (int j = rangeStart; j < rangeStop && !anyInteraction[local_warp]; j++) {
+            SYNC_WARPS;
             if (j < rangeEnd && tj < rangeEnd) {
                 bool isExcluded = (((exclusions>>tj)&1) == 0);
                 int localIndex = tbx+tj;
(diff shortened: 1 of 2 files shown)


peastman:
Trying to debug this test failure is driving me crazy.  Using OpenCL on a CPU, the test sometimes (but not always) fails.  I've never seen it fail when running on a GPU.  The problem is clearly related to the neighbor list for interaction groups: if I disable the neighbor list, I've never seen it fail.  It fails more often in mixed and double precision modes, though it also occasionally fails in single precision.  Naturally, the kernel to build the neighbor list has no variables of type "mixed", so it compiles to completely identical code in single and mixed precision modes.  Yet somehow that affects it anyway.  When it fails, it often (though not always) is immediately after reordering atoms.  But if I disable reordering, that often (but not always) causes it to fail after just a few steps.

Maybe it's just out to get me.

PR #2053, opened 2018-04-24
Remove duplicate patches

### LINKED ISSUE #2047: CHARMM36 termini patches are not recognized/handled
Hi all,

I have been trying to setup a simulation of a peptide in water using CHARMM36 and capped termini (`ACE` and `CT3`) but I cannot get past loading the structure. I prepared the PDB using VMD (psfgen) and everything seems fine - check [here](https://gist.github.com/JoaoRodrigues/425b2960fb9d4aad0f9de875d5699a51). The atom names match those on the `charmm36.xml` forcefield file for the `ACE` and `CT3` patches. Any clues? Is there a way of specifying patches that I am not following?

Here's a self-contained example:
```
import simtk.openmm.app as app
pdb = app.PDBFile('peptide.pdb')
ff = app.ForceField('charmm36.xml')
model = app.Modeller(pdb.topology, pdb.positions)

# This is just a trick to cause the error
# Keeping hydrogens and trying to create a system results in the same issue
hydrogens = [a for a in model.topology.atoms() if a.element == app.element.hydrogen]
model.delete(hydrogens)
model.addHydrogens(forcefield=ff, pH=7.0)
```

This gives the following error:
```
ValueError                                Traceback (most recent call last)
<ipython-input-8-9889d0fd1625> in <module>()
----> 1 model.addHydrogens(forcefield=ff, pH=7.0)

/usr/local/lib/python2.7/dist-packages/simtk/openmm/app/modeller.pyc in addHydrogens(self, forcefield, pH, variants, platform)
    840             # Use the ForceField the user specified.
    841
--> 842             system = forcefield.createSystem(newTopology, rigidWater=False, nonbondedMethod=CutoffNonPeriodic)
    843             atoms = list(newTopology.atoms())
    844             for i in range(system.getNumParticles()):

/usr/local/lib/python2.7/dist-packages/simtk/openmm/app/forcefield.py in createSystem(self, topology, nonbondedMethod, nonbondedCutoff, constraints, rigidWater, removeCMMotion, hydrogenMass, residueTemplates, ignoreExternalBonds, switchDistance, flexibleConstraints, **args)
   1131                             break
   1132             if matches is None:
-> 1133                 raise ValueError('No template found for residue %d (%s).  %s' % (res.index+1, res.name, _findMatchErrors(self, res)))
   1134             else:
   1135                 data.recordMatchedAtomParameters(res, template, matches)

ValueError: No template found for residue 1 (MET).  The set of atoms is similar to AOBT, but it is missing 3 atoms.
```

peastman:
Fixes #2047.  @JoaoRodrigues could you try this out and see if it fixes your problem?


peastman:
Any update on this?  I'm just waiting for your confirmation that this fixed the problem before moving ahead with the release.

JoaoRodrigues:
Hi @peastman,

Just tested it on the previous cases that I had originally found the problem in (referenced in #2047) and I am running into the same error as before. To clarify, this is not the 'Multiple matching templates' but the mismatching of a capped MET (ACE+MET). See the first error I reported in #2047.

Sorry ...

peastman:
@jchodera was that supposed to have been fixed by this change, or is it something different?

JoaoRodrigues:
Actually. Small correction to my previous comment. Running with the `psf` and the `pdb` file and this new `charmm36.xml` file does build the system correctly. If I try to add hydrogen atoms to the system, I run into trouble. So, basically, I *have* to provide OpenMM with a fully protonated file and adequate topology, otherwise it cannot match the right patch to the file.

jchodera:
> So, basically, I have to provide OpenMM with a fully protonated file and adequate topology, otherwise it cannot match the right patch to the file.

From what I understand, that is the exact behavior @peastman intends.

> If I try to add hydrogen atoms to the system, I run into trouble.

I'm a bit confused here. Did you mean to say OpenMM will not add the hydrogens for you if you don't provide the fully protonated topology yourself? The [`hydrogens.xml`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/data/hydrogens.xml) file does contain `ACE` and `MET` residues, but not patched forms of every residue, so it wouldn't know how to add protons in your case. 

A potential workaround would be to generate a `charmm36-hydrogens.xml` file that would contain proton definitions for all residues, or to allow `Modeller` to obtain hydrogen definitions from a `ForceField` object. Thoughts, @Peastman?

PR #2054, opened 2018-04-25
Improved behavior of adding force field directories

### LINKED ISSUE #2028: openmm.forcefielddir entry point must not require openmm import
We came across an issue that could use clearer documentation: If you want your Python module to register a callable `entry_point` function for `openmm.forcefielddir` to install new XML files for `simtk.openmm.app.ForceField`, that function must not require the import of any `openmm` modules to import and run or else we end up with circular dependencies. The recommended practice is to have this function defined in a separate `__init__.py` at the top-level of your package, before importing anything else.

cc: @pgrinaway @bas-rustenburg

peastman:
Fixes #2028.  @jchodera could you verify this fixes it for you?

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -6,7 +6,7 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
-Portions copyright (c) 2012-2017 Stanford University and the Authors.
+Portions copyright (c) 2012-2018 Stanford University and the Authors.
 Authors: Peter Eastman, Mark Friedrichs
 Contributors:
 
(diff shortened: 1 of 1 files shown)


jchodera:
cc: @pgrinaway and @bas-rustenburg who observed this problem in the first place.

jchodera:
To clarify the new behavior for @pgrinaway and @bas-rustenburg: `app.ForceField` now queries the registered ffxml plugin directories only when it is asked to load an XML file.

jchodera:
There's a weird SSL error on `osx` travis:
```
Download error on https://pypi.python.org/simple/pytest/: [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:590) -- Some packages may not be found!
```

peastman:
Apparently that's caused by this problem: http://pyfound.blogspot.com/2017/01/time-to-upgrade-your-python-tls-v12.html.  PyPI has stopped supporting TLS 1.0 and 1.1, but the system python on Macs doesn't support TLS 1.2.  I've found lots of posts from people suggesting various workarounds.  At least some people claim that it works on macOS 10.13, so the simplest solution might be to just tell travis to use that (instead of 10.12, which it does by default).  https://docs.travis-ci.com/user/reference/osx/

peastman:
Can you confirm if this fixes your problem?

jchodera:
@bas-rustenburg and @pgrinaway have been stuck in a conference, but I'll ping them to check on Monday!

bas-rustenburg:
This works.

I tested by importing openmm in the module that holds my function that produces the data directory. This normally leads to it failing to find the files included. With the fix, it works just fine.

peastman:
Thanks!

Issue #2029, opened 2018-03-24
Does CustomIntegrator have a way to compute sums across the (x,y,z) coordinates for each particle?


jchodera:
I've been fiddling with the [FIRE minimizer](https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.97.170201) ideas again, but I realize I think we need a way to compute the sum of a per-dof variable across the (x,y,z) components of individual particles, an dI don't think we have a way to do that. Is there a clever workaround that doesn't involve bringing back the per-dof vectors from the GPU?

peastman:
Nothing I can think of.  It doesn't have any feature for combining values across the components of each particle.

peastman:
This is possible with the vector functions added in 7.3.

Issue #1445, opened 2016-04-01
[Feature request] Additions to CustomIntegrator syntax


pgrinaway:
This is a set of feature requests--I'm not sure how difficult they are to implement, so I understand if they'd have to be on the back burner, but basically, these would make it easier to do various Monte Carlo moves inside a `CustomIntegrator`. I'm thinking of methods like [Configurational Bias Monte Carlo](http://towhee.sourceforge.net/algorithm/cbmc.html) and the Pruned-Enriched Rosenbluth Method. For those unfamiliar, these techniques involve atom-by-atom insertion, where new atom insertion involves proposal of internal degrees of freedom as fast "trial" moves, then accepts one of those based on the entire potential. 

More specifically, I was thinking:
- Ability to address individual particle coordinates:
  e.g. `self.addComputeGlobal('x[3]', 'x[3]+gaussian')`
- Conveniences for rotating a particle about a vector:
  - exposed cross and dot product, as well as rotation convenience:
    e.g,. suppose there is a torsion defined by {x[0], x[1], x[2], x[3]} and we want to rotate
    x[0] about the torsion axis.
  
  ``` python
      self.addComputeGlobal('axis', 'x[2]-x[1]')
      self.addComputeGlobal('axis_u', 'a/dot(a, a)')
      self.addComputeGlobal('x[0]', 'rotate(x[0], axis, pi/2')
  ```
  
  (the cross product might be needed, for example, to set a bond angle).
- Array-valued variables, along with some kind of loop construct:
  This example is for generating a torsion scan, which is necessary to normalize the torsion probabilities:

``` python
    #assuming we have already calculated the above axis
    self.addGlobalVariable('torsion_energies[n_trials]')
    generate_torsion_scan = """
    for(int i=0; i<n_trials, i++)
    {
    x[0] = rotate(x[0], axis, 2*pi/n_trials);
    torsion_energies[i] = energy();
    }
    """
    self.addComputeGlobal(generate_torsion_scan)
```
- Convenience to draw from a categorical distribution--this would arise in the CBMC example where you have generated a set of trial distributions, and want to choose one based on the total potential energies. Assuming we have normalized `torsion_energies` above:

``` python
    self.addComputeGlobal('torsion_idx', 'choose(torsion_energies)')
```
- Exposing the Lagrangian of the constraint algorithm to make Metropolis-Hastings a bit easier.

[...]

PR #2057, opened 2018-04-27
CustomIntegrator can use vector functions

### LINKED ISSUE #2050: Is it possible to distinguish between vector components in CustomIntegrator?
Consider, for instance, the implementation of the basic velocity verlet integrator, which contains code such as:
```
self.addComputePerDof("v", "v+0.5*dt*f/m")
self.addComputePerDof("x", "x+dt*v")
self.addComputePerDof("x1", "x")
self.addConstrainPositions()
self.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt")
```
I would like to extend the per DOF expressions with terms that depend on different components of the positions and velocities, for instance, scale the `y` and `z` components of the velocity by some global variable (while leaving the `x` component untouched), etc. Is this possible?

The motivation for this question comes from trying to implement external forces that depend on the velocities as well as the positions, which means I cannot use the normal API for adding custom forces.

peastman:
Fixes #2050 and #2029.  Also addresses parts of #1445.

### DIFF
--- a/libraries/lepton/include/lepton/CustomFunction.h
+++ b/libraries/lepton/include/lepton/CustomFunction.h
@@ -72,6 +72,38 @@ public:
     virtual CustomFunction* clone() const = 0;
 };
 
+/**
+ * This class is an implementation of CustomFunction that does no computation.  It just returns
+ * 0 for the value and derivatives.  This is useful when using the parser to analyze expressions
+ * rather than to evaluate them.  You can just create PlaceholderFunctions to represent any custom
+ * functions that may appear in expressions.
+ */
+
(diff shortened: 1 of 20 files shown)


peastman:
CUDA and OpenCL are both implemented now.  One thing I'd particularly like to see tested is if you could try out your existing CustomIntegrators and check whether the speed has changed at all.  I don't expect any significant changes one way or the other, but it would be good to check just to be sure.

peastman:
I tested a bunch of the integrators from openmmtools with both CUDA and OpenCL.  In all cases, they were slightly faster than before with this change.

Issue #2041, opened 2018-04-06
Changes for 7.2.2


peastman:
We need to decide what changes to include in the 7.2.2 patch release.  Definitely #2040, of course.  Are there other safe, reasonably high priority bug fixes we should include?  I suggest #2038, #2023, #2016, and https://github.com/pandegroup/openmm/pull/2014/commits/5cb31cfa1a2b74f60d7d260b1f51ae66cc6837c0.  (That last one was included in #2014.)

peastman:
Let's include #2044 too.  It's not very high priority, but it should be quite safe.

jchodera:
@peastman: There's one more very minor change we'd love to sneak in here to make it easier to structure python packages that install additional forcefields via entry points. My group ended up with circular imports when they tried the naive way to do this---I think this can be avoided with either some small doc additions or a small restructuring of where the entry points are scanned. More details in a few hours as I compile info from the group.

Sorry for the delay here.

jchodera:
Also, anaconda updates have broken the build system again, so we are stalled in testing/releasing while @lnaden addresses that today.

jchodera:
Conda builds have been temporarily fixed by https://github.com/omnia-md/conda-recipes/pull/875

@peastman: Can we address #2028 in this as well?

jchodera:
https://github.com/pandegroup/openmm/issues/2047#issuecomment-381472426 brings up the point that we should also expand the documentation on how to use the CHARMM forcefield conversion. For example, it's not obvious how patches are treated, or why PDB files you might want to use from CHARMM-GUI or VMD will generally fail (because they don't contain `CONECT` records).

peastman:
> @peastman: Can we address #2028 in this as well?

I'd rather not change that in a patch.  Your group is probably the only people affected by it and you have an easy workaround.  So it's not very high priority.  It's also a change in behavior rather than just a bug fix, which is another argument against doing it in a patch release.

jchodera:
Totally fine. This means we'll be sticking to CUDA 9.0 for this build as well?

peastman:
Unless we can figure out a way to support multiple CUDA versions?

Lnaden:
We have had some ideas on that involving conda features, but I don't think we will have a fix for that for this release. The reason we ask is I have the `dev` builds on CUDA 9.1 right now so its just to make sure we use the correct Docker image.

jchodera:
[...]

PR #2061, opened 2018-05-07
Changes for 7.2.2

peastman:
I believe these are all the changes that are supposed to go into 7.2.2.  See #2041.  If you think anything is incorrect or missing, please let me know.


jchodera:
I don't believe we've yet addressed documentation issues with the CHARMM forcefield and its limitations within 7.2.2.

peastman:
If you're still planning to do that, please do so ASAP!

peastman:
Remember how I said #2044 was low priority but should be very safe?  That's the one that caused the build to fail.  Serves me right for thinking any change, no matter how minor, is ever "safe".  I've taken it back out.

PR #2068, opened 2018-05-14
Preserve PDB insertion codes

### LINKED ISSUE #2064: support for pdb insertion codes?
Related to https://github.com/pandegroup/pdbfixer/issues/153
Current situation is, in my opinion, a bug, since residues are being merged when keepIds=True with no warning to users (i'd expect an exception).

I went digging into the code, and it appears to be a relatively simple fix:
1) just save the insertion_code in the topology.Residue pdbfile.py
2) then if keepIds=True, grab it back out in writeModel and put it in the atom line.
3) (optional) display it in Residue.__repr__

Thoughts?

```
> diff -C1 pdbfile.py pdbfile.py.orig 
*** pdbfile.py	2018-05-10 11:47:41.437670338 -0400
--- pdbfile.py.orig	2018-05-10 11:47:34.863996364 -0400
***************
*** 109,110 ****
--- 109,111 ----
                  r = top.addResidue(resName, c, str(residue.number))
+                 r.insertion_code = residue.insertion_code
                  if resName in PDBFile._atomNameReplacements:
***************
*** 352,355 ****
--- 353,358 ----
                      resName = res.name
+                 ic = ' '
                  if keepIds and len(res.id) < 5:
                      resId = res.id
+                     ic = res.insertion_code
                  else:
***************
*** 372,375 ****
                      coords = positions[posIndex]
!                     line = "%s%5d %-4s %3s %s%4s    %s%s%s  1.00  0.00          %2s  " % (
!                         recordName, atomIndex%100000, atomName, resName, chainName, resId, _format_83(coords[0]),
                          _format_83(coords[1]), _format_83(coords[2]), symbol)
--- 375,378 ----
                      coords = positions[posIndex]
!                     line = "%s%5d %-4s %3s %s%4s%s   %s%s%s  1.00  0.00          %2s  " % (
!                         recordName, atomIndex%100000, atomName, resName, chainName, resId, ic, _format_83(coords[0]),
                          _format_83(coords[1]), _format_83(coords[2]), symbol)
```

peastman:
Fixes #2064.

### DIFF
--- a/wrappers/python/simtk/openmm/app/pdbfile.py
+++ b/wrappers/python/simtk/openmm/app/pdbfile.py
@@ -6,7 +6,7 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
-Portions copyright (c) 2012-2016 Stanford University and the Authors.
+Portions copyright (c) 2012-2018 Stanford University and the Authors.
 Authors: Peter Eastman
 Contributors:
 
@@ -106,7 +106,7 @@ class PDBFile(object):
                 resName = residue.get_name()
                 if resName in PDBFile._residueNameReplacements:
                     resName = PDBFile._residueNameReplacements[resName]
-                r = top.addResidue(resName, c, str(residue.number))
+                r = top.addResidue(resName, c, str(residue.number), residue.insertion_code)
                 if resName in PDBFile._atomNameReplacements:
                     atomReplacements = PDBFile._atomNameReplacements[resName]
                 else:
@@ -352,8 +352,10 @@ class PDBFile(object):
                     resName = res.name
                 if keepIds and len(res.id) < 5:
                     resId = res.id
+                    resIC = res.insertionCode
                 else:
                     resId = "%4d" % ((resIndex+1)%10000)
+                    resIC = " "
                 if res.name in nonHeterogens:
                     recordName = "ATOM  "
                 else:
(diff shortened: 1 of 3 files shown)


wrappers/python/simtk/openmm/app/pdbfile.py
```diff
@@ -370,8 +372,8 @@ def writeModel(topology, positions, file=sys.stdout, modelIndex=None, keepIds=Fa
                     else:
                         atomName = atom.name
                     coords = positions[posIndex]
-                    line = "%s%5d %-4s %3s %s%4s    %s%s%s  1.00  0.00          %2s  " % (
-                        recordName, atomIndex%100000, atomName, resName, chainName, resId, _format_83(coords[0]),
+                    line = "%s%5d %-4s %3s %s%4s%s   %s%s%s  1.00  0.00          %2s  " % (
+                        recordName, atomIndex%100000, atomName, resName, chainName, resId, resIC, _format_83(coords[0]),
```
swails: If `resIC` is more than 1 character long, this poops out a broken PDB file :(

Same is true if chainName is more than 1 character, resName is more than 3 characters, atomName is more than 4 characters, etc.

Might be worth adding some protection in here so it properly truncates all strings to at most the right length.

peastman: In either of those cases, it throws an exception on the next line:

```python
assert len(line) == 80, 'Fixed width overflow detected'
```

This only happens if the caller specifies `keepIds=True`, and we do warn them it's their responsibility to make sure the IDs are valid!

I should probably change the format spec to say `%1s` for that field, though, so if `resIC` is an empty string it will work correctly.

swails: Another word of warning, then:  Python compiles pure-Python modules into bytecode (.pyc files now buried in `__pycache__` folders).  And it has 3 optimization levels (including no optimizations, the default)!  But they're all kinda... lame.

* `-O` goes through and trashes all `assert` statements
* `-OO` goes through and trashes all docstrings in addition to `assert` statements

The intent here according to email threads I've seen in Python user and developer forums is for `assert` to be strictly a debugging-like tool.  Spread them around liberally to catch bugs early.  You can then eliminate their performance hit in production by optimizing all `assert` statements out.

The resulting conventions for how `assert` should be used is:

1. *Never* tuck a side-effect into an `assert` statement.  Very bad:
```python
assert my_obj.initialize_and_verify(), "Verification failed"
```
2. As a general rule, don't use `assert` to check that user input is valid

After all, you never know when somebody will install your library with optimizations and wipe out all of your asserts.

peastman: Good point.  I'll change it to throw an exception more properly.

swails:
run jenkins

PR #2076, opened 2018-05-17
Improved speed of applying patches

peastman:
When building a system with the CHARMM36 force field, this will often save about ten seconds.

It first tries to match every residue to a template without using patches.  If it succeeds, that's fine.  But if even one residue can't be matched, in then needs to build a list of all possible patched templates.  (And it really does need to build all of them, since it needs to throw an exception if there's more than one way to match it.)  So for every template, it needs to apply all possible combinations of patches.  In some cases, that led to thousands of patched versions of a single template.

To do that, it needs to decide which patches are compatible with each other.  Before it did that in a naive way: just try to apply each patch on top of the others, and if it throws an exception (for example, because a patch wants to modify an atom that another patch has already deleted), ignore that patch and continue.  But that approach was too forgiving and allowed combinations that really shouldn't be allowed.  For example, if two patches both modify the same atom, they should be considered incompatible since it's ambiguous what the parameters of that atom should be.

So now it records a list of every atom that a patch alters in any way (adds, changes, or deletes).  Two patches are considered incompatible if there's any atom they both alter in any way.  This reduces the set of allowed combinations and makes it a lot faster to build the patched templates.

Now that it's doing that more rigorous check, it really shouldn't catch exceptions any more.  Any remaining exceptions indicate genuine mistakes in the force field definition.  But when I tried removing the exception handling, CHARMM36 produced an error!

```
ValueError: Patch 'HS2' adds an atom with the same name as an existing atom: HE2
```

And sure enough, it's right.  The patch `HS2` adds an atom called `HE2`.  The residue templates `HSP` and `EIMM` each specify `<AllowPatch name="HS2"/>` but already have an atom called `HE2`.  It looks like the logic for selecting allowed patches is including ones it shouldn't.  So we should fix that, but for the moment, I'll have ForceField continue to catch the exceptions so I don't break CHARMM36.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -307,16 +307,15 @@ class ForceField(object):
                     else:
                         numResidues = 1
                     patchData = ForceField._PatchData(patchName, numResidues)
-                    allAtomNames = set()
                     for atom in patch.findall('AddAtom'):
                         params = {}
                         for key in atom.attrib:
                             if key not in ('name', 'type'):
                                 params[key] = _convertParameterToNumber(atom.attrib[key])
                         atomName = atom.attrib['name']
-                        if atomName in allAtomNames:
+                        if atomName in patchData.allAtomNames:
                             raise ValueError('Patch '+patchName+' contains multiple atoms named '+atomName)
-                        allAtomNames.add(atomName)
+                        patchData.allAtomNames.add(atomName)
                         atomDescription = ForceField._PatchAtomData(atomName)
                         typeName = atom.attrib['type']
                         patchData.addedAtoms[atomDescription.residue].append(ForceField._TemplateAtomData(atomDescription.name, typeName, self._atomTypes[typeName].element, params))
@@ -326,17 +325,17 @@ class ForceField(object):
                             if key not in ('name', 'type'):
                                 params[key] = _convertParameterToNumber(atom.attrib[key])
                         atomName = atom.attrib['name']
-                        if atomName in allAtomNames:
+                        if atomName in patchData.allAtomNames:
                             raise ValueError('Patch '+patchName+' contains multiple atoms named '+atomName)
-                        allAtomNames.add(atomName)
+                        patchData.allAtomNames.add(atomName)
                         atomDescription = ForceField._PatchAtomData(atomName)
                         typeName = atom.attrib['type']
                         patchData.changedAtoms[atomDescription.residue].append(ForceField._TemplateAtomData(atomDescription.name, typeName, self._atomTypes[typeName].element, params))
                     for atom in patch.findall('RemoveAtom'):
                         atomName = atom.attrib['name']
-                        if atomName in allAtomNames:
+                        if atomName in patchData.allAtomNames:
                             raise ValueError('Patch '+patchName+' contains multiple atoms named '+atomName)
-                        allAtomNames.add(atomName)
+                        patchData.allAtomNames.add(atomName)
                         atomDescription = ForceField._PatchAtomData(atomName)
                         patchData.deletedAtoms.append(atomDescription)
                     for bond in patch.findall('AddBond'):
@@ -695,6 +694,7 @@ class ForceField(object):
             self.deletedBonds = []
             self.addedExternalBonds = []
             self.deletedExternalBonds = []
+            self.allAtomNames = set()
 
         def createPatchedTemplates(self, templates):
             """Apply this patch to a set of templates, creating new modified ones."""
@@ -1509,7 +1509,7 @@ def _applyPatchesToMatchResidues(forcefield, data, residues, bondedToAtom, ignor
             if len(patches) > 0:
                 newTemplates = []
                 patchedTemplates[name] = newTemplates
-                _generatePatchedSingleResidueTemplates(template, patches, 0, newTemplates)
+                _generatePatchedSingleResidueTemplates(template, patches, 0, newTemplates, set())
                 for patchedTemplate in newTemplates:
                     signature = _createResidueSignature([atom.element for atom in patchedTemplate.atoms])
                     if signature in patchedTemplateSignatures:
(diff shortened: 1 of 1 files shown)


jchodera:
> And sure enough, it's right. The patch HS2 adds an atom called HE2. The residue templates HSP and EIMM each specify <AllowPatch name="HS2"/> but already have an atom called HE2. It looks like the logic for selecting allowed patches is including ones it shouldn't. So we should fix that, but for the moment, I'll have ForceField continue to catch the exceptions so I don't break CHARMM36.

This is precisely why I asked that we expose the patch compatibility assessment functionality externally so that we could use the *same* logic from ParmEd instead of creating and maintaining two completely separate implementations that operated on different data models that somehow were magically supposed to always produce identical results.

jchodera:
> And sure enough, it's right. The patch HS2 adds an atom called HE2. The residue templates HSP and EIMM each specify <AllowPatch name="HS2"/> but already have an atom called HE2. It looks like the logic for selecting allowed patches is including ones it shouldn't. So we should fix that, but for the moment, I'll have ForceField continue to catch the exceptions so I don't break CHARMM36.

Can you open an issue in ParmEd?

jchodera:
Actually, this seems to reveal a much deeper flaw with our approach to patches than I previously appreciated: The issue seems to be that we currently assume that a patch applied to a compatible residue will always produce the same set of atoms that are added, changed, or deleted, and that we can check this by applying the patch to the first compatible residue and looking at what changed. See [this line](https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/parameters.py#L644-L647).

However, this example seems to suggest that this assumption is violated. 

What should the solution be? It's not even clear what the right approach is here:
1. Apply the patch to the first residue, determine which atoms are added/changed/deleted to create an OpenMM `<Patch>`, and reject any residues as compatible where the patch would not work
2. Apply the patch to *all* compatible residues, selecting the variant of add/change/delete that maximizes the number of compatible patches
3. Apply the patch to *all* compatible residues, writing all possible variants of the patch that have different sets of add/change/delete under different names
4. Something else?

jchodera:
> When building a system with the CHARMM36 force field, this will often save about ten seconds.

Also, this part is awesome.

peastman:
There's a "correct" solution here, but I'm afraid it's a labor intensive one.  Looking at top_all36_prot.rtf, here's the beginning of the definition of the HS2 patch:

```
PRES HS2          0.00 ! Patch for neutral His, move proton from ND1 to NE2
```

So this patch is only ever intended to be used with one specific residue (neutral HIS).  By relying on an automated "compatibility check", we have incorrectly concluded that it can also be used with a few others.  There are other similar cases.  For example,

```
PRES PROP         1.00 ! Proline N-Terminal
```

So that's only meant to be used with N-terminal prolines, but we've marked several other molecules as being compatible with it.

The CHARMM files do contain information about what residues each patch should be used with, but only in comments and not in any machine readable form.  So the "right" solution is to go through all those comments and create a table of which residues each patch is actually meant to be used with.

jchodera:
> The CHARMM files do contain information about what residues each patch should be used with, but only in comments and not in any machine readable form. So the "right" solution is to go through all those comments and create a table of which residues each patch is actually meant to be used with.

These comments are also incomplete. @ChayaSt had looked through these and noted the comments were not exhaustive.

It is not practical for us to process all files manually and identify which patches we should add from the comments, and then also figure out which patch compatibility is missing. This is why @ChayaSt and I initially suggested that users be asked to specify which patches they want to use. Instead, we were directed to implement an automated algorithm.

Of the practical ways forward, which do you prefer?
1. Apply the patch to the first residue, determine which atoms are added/changed/deleted to create an OpenMM `<Patch>`, and reject any residues as compatible where the patch would not work
2. Apply the patch to *all* compatible residues, selecting the variant of add/change/delete that maximizes the number of compatible patches
3. Apply the patch to *all* compatible residues, writing all possible variants of the patch that have different sets of add/change/delete under different names
4. Something else?

jchodera:
Alternatively, we can have a minimal whitelist and have users to list their own patches, but it may be easiest to just let users specify their own patches.

peastman:
I suggest the following:

1. Create a mechanism where you can manually specify the residues a patch should be applied to.
2. If that mechanism is not used for a particular patch, use the current mechanism to identify compatible residues.
3. Apply the patch to all compatible residues and check to see whether they produce identical lists of atoms to add/change/delete.
4. If they don't, throw an exception.  This patch needs manual curation!  Someone needs to check the comments in the file and figure out which residues it should actually be applied to.

Hopefully the number of patches needing manual curation will be small, but we won't know until we try and see.  It's unfortunate that parts of the force field definition are only given in a human readable (not machine readable) form, but that's how it is.

jchodera:
This is a proposal for the forcefield conversion, not for ForceField.createSystem()?

Why not just have users specify which patches they want for specific residues during createSystem()? If we get ANY part of the proposed pipeline wrong and exclude a patch accidentally, we make it impossible for the user to use that patch.

peastman:
It's our job as software developers to do hard work so that our users don't have to! 😀

This particular problem is part of the forcefield conversion, and it's our job to do that right.  Once it's done correctly, everything will "just work", even if a user knows little about the CHARMM force field has no clue what a patch is.

> Why not just have users specify which patches they want for specific residues during createSystem()? 

By doing that, we'd be requiring users to make a decision they're probably not qualified to make, and that can lead to incorrect results if they make it wrong.

> If we get ANY part of the proposed pipeline wrong and exclude a patch accidentally, we make it impossible for the user to use that patch.

So we'd better get all of it right! 😀  But remember, there are worse things than a user being unable to use a patch.  There are workarounds in that case (e.g. build their system with CHARMM-GUI and load the PSF file directly).  That's much better than having them run hundreds of simulations and publish a paper, only to discover their results were wrong because they chose the wrong patch and used incorrect parameters.

jchodera:
> This particular problem is part of the forcefield conversion, and it's our job to do that right. 

I think the problem here is that none of us have the knowledge to make these decisions with 100% accuracy because the information we need is not contained in the parameter files at all. Even though there are other routes to loading in CHARMM systems, it seems problematic that we might accidentally omit a patch someone will want to use (or add). Can we at least provide some mechanism to allow people to specify patch overrides if they need to?

peastman:
There's already a way they can do that: copy the patch definition into a new XML file, add an `<ApplyToResidue>` tag, and load it as part of the force field.  It's not entirely convenient and certainly "for advanced users only", but then again, hopefully no one will ever have to do it.  The only reason they would need to is if we messed something up and they need a workaround.

jchodera:
Ok, that greatly reduces the risk with your proposed plan. We can try to put that into action from our end, but I think myself, @chayast, and @rafwiewiora are backlogged by about four weeks. Is there a way you would be able to take a stab at this before then? If not, I can take a stab at it after our R01 is due June 5.

peastman:
That's fine.  I don't think this is urgent.

jchodera:
Great. It's on my calendar for June 6 then!

Issue #2099, opened 2018-06-12
Ions are placed too close to membrane leading to very large energies


JoaoRodrigues:
Hi there,

Using the `neutralize=True` option in `modeller.addMembrane()` (sometimes) leads to very large potential energies in the system that cannot be solved by the minimizer. The cause (at least from what we gathered) is that ions are placed at the position of randomly picked water molecules. Sometimes, these are waters very close to the lipid bilayer and since ions are substantially larger than the water oxygens they replace, there is essentially an overlap. @peastman you can verify this by trying to minimize the system I sent you privately last week that was causing the membrane water addition to fail.

I solved this problem in my system by adding an extra check in `addMembrane()` to ignore waters that are within a certain distance (in Z) from the min/max lipid Z coordinates, when selecting the random subset to be replaced by ions. It does not add that much of an overhead and seems to be a pretty generalizable solution for any lipid membrane. I am running some tests and cleaning the code before I create a pull request.

EDIT: sometimes this still fails, because two neighboring waters are selected randomly to be replaced, leading to the same issue (overlapping particles, in this case two ions).

peastman:
That sounds like a good solution.  It looks like nearly all positive ions have radii that are similar to or smaller than the water oxygen.  But negative ions tend to be a bit larger, so that's probably where the problem is coming from.

JoaoRodrigues:
New implementation of the fix in #2274

PR #2100, opened 2018-06-12
Improved ion addition in addMembrane()

JoaoRodrigues:
Addresses #2099 

Adds a couple of checks to `addMembrane`:
1. Ensures ions are not placed within a certain cutoff of the lipid bilayer.
2. Ensures ions are not placed next to each other (to avoid overlaps)

### DIFF
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -1451,12 +1451,72 @@ class Modeller(object):
             numPositive += numPairs
             numNegative += numPairs
         totalIons = numPositive+numNegative
-        
-        # Now add the ions.  We do this by randomly selecting a set of waters that were just added and replacing
-        # them with ions.
 
+        # Now add the ions. We do this by randomly selecting a set of waters a certain
+        # distance away from either leaflet of the membrane and replacing them with ions,
+        # while ensuring ions are not placed too close to each other.
         if totalIons > 0:
-            toReplace = random.sample(list(waterPos), totalIons)
+
+            # Redo waterPos because previous modeller.delete() changes Chain indexes
+            waterPos = {}
+            for chain in list(modeller.topology.chains())[-2:]:
+                for residue in chain.residues():
+                    if residue.name == 'HOH':
+                        for atom in residue.atoms():
+                            if atom.element == elem.oxygen:
+                                waterPos[residue] = modeller.positions[atom.index].value_in_unit(nanometer)
+
+            # Calculate lipid Z boundaries
+            lipidNames = {res.name for res in patch.topology.residues() if res.name != 'HOH'}
+            lipidZMax = 0.0
+            lipidZMin = 99999.0
+            for res in modeller.topology.residues():
+                if res.name in lipidNames:
+                    for atom in res.atoms():
+                        atomZ = modeller.positions[atom.index][2].value_in_unit(nanometer)
+                        lipidZMax = max(lipidZMax, atomZ)
+                        lipidZMin = min(lipidZMin, atomZ)
+
+            # Ignore waters that are within a certain distance of the membrane
+            lipidOffset = 0.25
+            upperZBoundary = (lipidZMax + lipidOffset)
+            lowerZBoundary = (lipidZMin - lipidOffset)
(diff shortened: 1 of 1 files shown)


wrappers/python/simtk/openmm/app/modeller.py
```diff
@@ -1451,20 +1451,80 @@ def addMembrane(self, forcefield, lipidType='POPC', membraneCenterZ=0*nanometer,
...
+                for residue in chain.residues():
+                    if residue.name == 'HOH':
+                        for atom in residue.atoms():
+                            if atom.element == elem.oxygen:
+                                waterPos[residue] = modeller.positions[atom.index].value_in_unit(nanometer)
+
+            # Calculate lipid Z boundaries
+            lipidNames = {res.name for res in patch.topology.residues() if res.name != 'HOH'}
+            lipidZMax = 0.0
+            lipidZMin = 99999.0
```
peastman: Use `sys.float_info.min` and `sys.float_info.max`.  Users are free to pass in their own patch, and we can't assume what range the positions will be in.  (Ok, assuming they'll be less than 99999 is pretty safe.  But we shouldn't assume they'll be positive.)

wrappers/python/simtk/openmm/app/modeller.py
```diff
@@ -1451,20 +1451,80 @@ def addMembrane(self, forcefield, lipidType='POPC', membraneCenterZ=0*nanometer,
...
+                        break
+                else:
+                    toReplace.append(pickedWater)
+                    ionPositions.append(waterPos[pickedWater])
+                    addedIons += 1
+
+                    n_trials = 10
+
+                if n_trials == 0:
+                    raise Exception('Could not add more than {} ions to the system'.format(addedIons))
```
peastman: Limiting it to 10 tries seems unnecessarily restrictive.  Every time through the loop you remove one water, so eventually it will either succeed or run out of waters.

peastman: Also, it's probably better to use `ValueError` instead of `Exception`.  It's still kind of generic, but it basically means, "There was a problem with the value you specified."

JoaoRodrigues:
Thanks! Reviewing these changes. On second thought, would you consider a good idea to write a generic function `addIons` that is used in both `addSolvent` and `addMembrane`? Not to have duplicated code and inconsistent rules (no filtering in `addSolvent`)?

wrappers/python/simtk/openmm/app/modeller.py
JoaoRodrigues: I removed the `n_trials` altogether and just catch and re-throw an exception when the `random.choice` hits an empty list.

wrappers/python/simtk/openmm/app/modeller.py
```diff
@@ -1451,20 +1451,80 @@ def addMembrane(self, forcefield, lipidType='POPC', membraneCenterZ=0*nanometer,
...
+
+            # Ignore waters that are within a certain distance of the membrane
+            lipidOffset = 0.25
+            upperZBoundary = (lipidZMax + lipidOffset)
+            lowerZBoundary = (lipidZMin - lipidOffset)
+            waterResidues = list(waterPos)
+            for wRes in waterResidues:
+                waterZ = waterPos[wRes][2]
+                if lowerZBoundary < waterZ < upperZBoundary:
+                    del waterPos[wRes]
```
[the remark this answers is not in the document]
JoaoRodrigues: Did not see many list comprehensions in the code so I was avoiding those :)

wrappers/python/simtk/openmm/app/modeller.py
JoaoRodrigues: Indeed, changed!

peastman:
That would be very reasonable.  Though I'd make it a private method `_addIons()`, since it's not really meant to be used externally.

wrappers/python/simtk/openmm/app/modeller.py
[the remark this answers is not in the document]
peastman: I'm a big fan of list comprehensions!

JoaoRodrigues:
Closing this in favor of a new PR that will include all the above.

PR #2105, opened 2018-06-20
NonbondedForce can use context parameters to modify particles

### LINKED ISSUE #1702: Proposed API for CHARMM-like BLOCK nonbonded facility
@peastman and I discussed the potential for implementing something like the CHARMM [BLOCK](https://www.charmm.org/charmm/documentation/by-version/c40b1/params/doc/block/) facility for OpenMM to make it easier to implement things like Jana's [constant-pH lambda dynamics](10.1021/acs.jctc.6b00552) or the Brooks [multisite lambda dynamics](http://dx.doi.org/10.1021/ct200444f). While energy derivatives of global parameters are now available, the complexity of implementing the kind of potentials commonly used in lambda dynamics with the existing `Custom*Force` methods suggests it would be incredibly helpful to have a new way of specifying these global parameter-dependent interactions.

The simplest approach would be to define something like `NonbondedForce` that allows the `sigma`, `epsilon`, and `charge` parameters to optionally be defined in terms of linear combinations of groups of parameters. This way, we could support the same `NonbondedMethod` choices as `NonbondedForce`.

The parameters on a given atom would be computed as:
```
charge  = charge_0  + \sum_{n=1}^{Nblocks} parameter_n * charge_n
sigma   = sigma_0   + \sum_{n=1}^{Nblocks} parameter_n * sigma_n
epsilon = epsilon_0 + \sum_{n=1}^{Nblocks} parameter_n * epsilon_n
```
where `charge_0` is the constant component of the charge, `Nblocks` is the number of blocks that the particle appears in, `parameter_n` is the value of the global parameter for block `n`, and `charge_n` is the additive component of the charge defined for this particle in block `n`. Often, the `parameter_n` will defined in [0,1] and sum to 1, but there's no reason to require this.

The derivatives of the energy are straightforward to compute except for the PME reciprocal space energies, which Jana has conveniently derived in her recent [constant-pH lambda dynamics](10.1021/acs.jctc.6b00552) paper.

Here's an example of what the new elements of the API could look like:
```python
# Create a new force object
force = NonbondedBlockForce()
# Add constant components of charge, sigma, epsilon (which may all be zero)
particle_index = force.addParticle(charge, sigma, epsilon)
...
# Add constant components to exceptions (which may all be zero)
exception_index = force.addException(particle1, particle2, chargeProd, sigma, epsilon)
...
# Now define a BLOCK of additive parameter contributions we will control with a global context parameter
# Add a new block, defining its global context parameter name and default value
global_parameter_name = 'block1_global_parameter_name'
default_value = 0.0
block_index = force.addBlock(global_parameter_name, default_value)
# Add a particle to the block, along with its additive contribution to charges
force.addParticleToBlock(block_index, particle_index, charge, sigma, epsilon) 
...
# Define an additive contribution this block makes to exceptions
force.addExceptionToBlock(block_index, exception_index, chargeProd, sigma, epsilon)
```
We'd also probably want some convenience accessors for modifying default global parameter values, modifying block particle parameters, etc.

It's useful to allow `sigma = 0` in this case, though I imagine at the time of building the actual force, any `sigma = 0` could simply be redefined to `sigma = 1` if `epsilon = 0`.

Tagging @lnaden who also has expertise in this area.

peastman:
Fixes #1702.  So far I've created the API and implemented it on the reference platform.  Give it a try.

### DIFF
--- a/openmmapi/include/openmm/NonbondedForce.h
+++ b/openmmapi/include/openmm/NonbondedForce.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2008-2014 Stanford University and the Authors.      *
+ * Portions copyright (c) 2008-2018 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -76,6 +76,32 @@ namespace OpenMM {
  * the effect of all Lennard-Jones interactions beyond the cutoff in a periodic system.  When running a simulation
  * at constant pressure, this can improve the quality of the result.  Call setUseDispersionCorrection() to set whether
  * this should be used.
+ * 
+ * In some applications, it is useful to be able to inexpensively change the parameters of small groups of particles.
+ * Usually this is done to interpolate between two sets of parameters.  For example, a titratable group might have
+ * two states it can exist in, each described by a different set of parameters for the atoms that make up the
+ * group.  You might then want to smoothly interpolate between the two states.  This is done by calling
+ * addParticleParameterOffset() to create a "parameter offset".  Each offset defines the following:
+ * 
+ * <ul>
+ * <li>A Context parameter used to interpolate between the states.</li>
+ * <li>A single particle whose parameters are influenced by the Context parameter.</li>
+ * <li>Three scale factors (chargeScale, sigmaScale, and epsilonScale) that specify how the Context parameter
+ * affects the particle.</li>
+ * </ul>
+ * 
+ * The "effective" parameters for a particle (those used to compute forces) are given by
+ * 
+ * <tt><pre>
+ * charge = baseCharge + param*chargeScale
+ * sigma = baseSigma + param*sigmaScale
+ * epsilon = baseEpsilon + param*epsilonScale
+ * </pre></tt>
+ * 
+ * where the "base" values are the ones specified by addParticle() and "oaram" is the current value
+ * of the Context parameter.  A single Context parameter can apply offsets to multiple particles,
+ * and multiple parameters can be used to apply offsets to the same particle.  Parameters can also be used
+ * to modify exceptions in exactly the same way by calling addExceptionParameterOffset().
  */
 
 class OPENMM_EXPORT NonbondedForce : public Force {
@@ -131,6 +157,18 @@ public:
      */
     int getNumExceptions() const {
         return exceptions.size();
+    }
+    /**
+     * Get the number of particles parameter offsets that have been added.
+     */
+    int getNumParticleParameterOffsets() const {
+        return particleOffsets.size();
+    }
+    /**
+     * Get the number of exception parameter offsets that have been added.
+     */
+    int getNumExceptionParameterOffsets() const {
+        return exceptionOffsets.size();
     }
     /**
      * Get the method used for handling long range nonbonded interactions.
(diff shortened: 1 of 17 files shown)


peastman:
The CPU platform is implemented too now.

peastman:
@jchodera have you had a chance to look at this or try it out?  I'd appreciate your feedback on it.

jchodera:
> @jchodera have you had a chance to look at this or try it out? I'd appreciate your feedback on it.

Let me build a conda-installable package from this PR to test!

peastman:
This should now work on all platforms.

jchodera:
Awesome!

jchodera:
We're still working on incorporating tests for this feature into our code via https://github.com/choderalab/perses/pull/472, but don't let this hold up merging once this is ready!

peastman:
I don't want to merge it until we're sure it's ready.  That includes getting your feedback on the API, testing the performance, etc.

jchodera:
> I don't want to merge it until we're sure it's ready. That includes getting your feedback on the API, testing the performance, etc.

Sounds good. I'm working on integrating this into our free energy calculations right now. Just doing some debugging of issues on our end.

jchodera:
I think I know what the problem is. For other context parameters, we allow these to be computed from a master `lambda` variable so that they are slaved to it via some functions. It seems we can't take advantage of that here in computing the offset scaling factors via an expression instead of a parameter.

How hard would it be to instead be
```C++
int addParticleParameterOffset(const std::string& lepton_expression, int particleIndex, double chargeScale, double sigmaScale, double epsilonScale)
``` 
where the evaluated value of `lepton_expression` is used to compute the scaling factor? 

Alternatively, is there some way we can have one global parameter computed from other parameters via lepton *outside* of a `CustomIntegrator`?

jchodera:
Quick update here: It turns out the issue I was running into was that I had no way to define the context parameter `lambda_electrostatics` I wanted to use in `NonbondedForce` for the offsets. As a work-around, I can just define this in a `CustomNonbondedForce` and not use the global parameter there, but I think this emphasizes that it would be useful to have a way to define new global parameters, and potentially allow them to depend on other global parameters.

The proposal above for `addComputedGlobalParameter` doesn't even have to belong to `NonbondedForce`. It could belong to `System`, or some other `GlobalParameterForce` that just computes global parameters.

peastman:
I don't understand what you mean.  The NonbondedForce takes care of creating the global parameter.  You don't need to define it anywhere else.  Look at the test cases in TestNonbondedForce.h.

jchodera:
Any idea how I might be able to track down how I ran into [this error](https://github.com/pandegroup/openmm/pull/2105#issuecomment-401511250) then?
```
Exception: Unknown global variable: lambda_electrostatics
```
I'm working in this PR: https://github.com/choderalab/perses/pull/472

peastman:
It's hard to guess from all of that.  Can you reproduce it with a smaller, self contained test?

jchodera:
Huh. I'm not quite sure what I did to fix this---I don't think I did anything but reinstall `miniconda`---but it seems to be working now! Will test further and report back.

jchodera:
If we *did* want to slave the offsets to `1-lambda` or some more complicated function of `lambda` without setting a bunch of separate parameters through `CustomIntegrator`, how would you recommend we do this?

peastman:
Use the same parameter for all of them.  But if you want separate parameters, there's only two ways to set them: have the integrator do it, or call `setParameter()` on the Context each time you want them to change.

jchodera:
Quick question @peastman : Did you end up having the Ewald self-energy terms change with `lambda` as well, or are these fixed at their initial values?

peastman:
Any updates on testing this?  I'd really like to get it merged, since I have another PR that's waiting on it.

jchodera:
We think it's working OK! We've integrated this feature into perses (https://github.com/choderalab/perses/pull/472) and discovered some other issues with our tests, but we think it is behaving well.

peastman:
Thanks!  I'll go ahead and merge this then.

jchodera:
Looks like we neglected to add the XML serialization, so I can't share a simple example here yet.

PR #2119, opened 2018-07-11
Global parameters for NonbondedForce are explicitly defined

### LINKED ISSUE #2117: Extend NonbondedForce API to query/set global parameters
Since #2105 added the ability to specify global parameters to `NonbondedForce` to control the blending of offsets, it is critical we expose some common functionality for querying which parameters are available and setting their defaults. The `Custom*Force` objects do this through the following methods:
```python
force.getNumGlobalParameters()
force.getGlobalParameterName()
force.getGlobalParameterDefaultValue()
force.setGlobalParameterDefaultValue()
force.setGlobalParameterName() # this is less important?
```
I propose we also extend the `NonbondedForce` API with these methods.

Practically, our [`openmmtools`](http://github.com/choderalab/openmmtools) layer depends on [these API methods](https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L746-L768) for inspecting which alchemical parameters are available for control. This API consistency among forces that can be modified with global context parameters turns out to be a very powerful way for writing alchemical (or other) codes with OpenMM, so it would be great to preserve this consistency.

cc: https://github.com/choderalab/openmmtools/issues/359 https://github.com/choderalab/perses/issues/476

peastman:
Fixes #2117.

### DIFF
--- a/openmmapi/include/openmm/NonbondedForce.h
+++ b/openmmapi/include/openmm/NonbondedForce.h
@@ -80,8 +80,9 @@ namespace OpenMM {
  * In some applications, it is useful to be able to inexpensively change the parameters of small groups of particles.
  * Usually this is done to interpolate between two sets of parameters.  For example, a titratable group might have
  * two states it can exist in, each described by a different set of parameters for the atoms that make up the
- * group.  You might then want to smoothly interpolate between the two states.  This is done by calling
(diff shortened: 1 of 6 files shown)


Lnaden:
API looks good! Thank you!

PR #2120, opened 2018-07-11
Implementing Drude polarizable force fields in the CHARMM file parsers

jing-huang:
The Python api for parsing CHARMM psf, parameter and topology files has been extended to handle Drude polarizable simulation systems.

Most importantly, the NBThole function, which is essential for ion interactions, is now supported using a CustomNonbondedForce implementation, and the corresponding force field parameters can be correctly handled by CharmmParameterSet. This completes our efforts in a full implementation of the Drude polarizable force field in OpenMM. 

The modification also allows CharmmPsfFile to handle charmm additive simulation systems with lonepairs, such as in the recent development of CGENFF for halogen atoms.


peastman:
Fantastic!  I'll take a look at the code shortly.

I notice that it doesn't have any test cases.  Without that, we have no idea whether the code works or not (and we won't know if a future change breaks it).  An easy way to test it is to create input files for one or more small systems (a few hundred atoms at most).  The files can go in wrappers/python/tests/systems.  Then a test case can load the files, compute the potential energy (and also forces if you want to check that too), and compare them to the values calculated by CHARMM.  Here's an example of a test that works like that: https://github.com/pandegroup/openmm/blob/master/wrappers/python/tests/TestCharmmFiles.py#L93-L117

peastman:
I'm having a bit of trouble understanding the implementation of nbthole.  As I understand it from the documentation at https://www.charmm.org/charmm/documentation/by-version/c40b1/params/doc/drude/, this is fundamentally a bonded interaction.  It's evaluated for a discrete, fixed set of atom pairs.  Is that correct?  You've implemented it with a CustomNonbondedForce, then added a (possibly large?) number of interaction groups to restrict it to particular pairs.  Can you explain more about how this works?  What do the interaction groups correspond to?

peastman:
Thanks, that makes sense.  Can you explain how the interaction groups are being used?

jing-huang:
Could you please point to me where is test.py so I can have a try? My test using TestCharmmFiles.py passes, although I'm still using python 2.7. In principle the nbt_alpha_list[i] = pow(-1*self.drudeconsts_list[i][0],-1./6.) should not take effects in any of the current openmm tests, because the logical variables has_drude_particle and has_nbthole_terms should both be set to False. 

I'm putting together unit tests for the Drude psf/toppar files.

peastman:
I'll email you the files.  I just used CHARMM-GUI to build files for a protein in solvent with all default settings.

peastman:
Any update on this?  Were you able to reproduce the error?

This is now the last thing preventing us from releasing the beta of 7.3, so I'd really like to get it finished up as soon as possible.  We need a working implementation, along with test cases that show it works.  Once you provide those, I can take things from there to handle optimization, code cleanup, etc.

jing-huang:
I actually haven't managed to reproduce the error you see. After fixing the indentation bug, I tested with both python2.7 and python3.5, and were able to run your test case and obtain the potential energy. One issue is that I have to use either reference or CUDA platform, not the CPU platform, which gives "All Forces must have identical exclusions" error due to the usage of CustomNonbonded class for NBFIX.

This git push contains the indentation fix, as well as a testcase (an alanine 3-mer solvated in cubic water box with additional K+ and Cl- ions). Could you take a look whether it works or not? Thanks a lot!

peastman:
Thanks!  I'll take a look.

jing-huang:
Let me explain NBThole in more details. Each NBThole term defines two atom types and gives a scaling factor. The electrostatics interactions (charge-charge, charge-induced dipole and charge-induced dipole) between any atom pairs within a cutoff distance (5A) matching these two atom types, for example the oxygen atoms in water molecules and the Magnesium ions, are screened using the given Thole screening factor by S(r)=1-(1+ar/2b)exp(-ar/b) where a is the screening factor and b is given by the atomic polarizabilities of these two atoms b=(alpha_i*alpha_j)^(1/6).  

In the Drude force field, any polarizable atom is represented by two charged particles, a parent atom and its Drude oscillator, and the electrostatics interactions between two polarizable atoms is modeled by simple Coulombic interactions between four pairs: atom1 and atom2, atom1 and Drude2, Drude1 and atom2, and Drude 1 and Drude2. Considering by default the full q_i*q_j/r interactions is already being computed, we can set up a CustomNonbonded class computing [S(r)-1]q_i*q_j/r for NBThole-involved atom types to obtain full implementation of all NBThole terms. 

On InteractionGroup, removing it won't affect any result. Adding the same exclusion rules from the other two nonbonded class won't affect any result as well. The reason I use the InteractionGroup feature is because only a small subset of atom type pairs are involved in the NBThole interactions. By setting up interaction groups I hope only those is the groups will be computed, in order to reduce computational cost. Am I misunderstanding how the InteractionGroup works?

peastman:
Let me see if I understand.  There is one interaction group for every NBThole term.  Each of the two particle sets in a group contains the atoms of a particular atom type, plus their associated Drude particles.  The CustomNonbondedForce, when added to the standard NonbondedForce, produces a total interaction equal to the desired screened interaction.  Is that correct?

> Adding the same exclusion rules from the other two nonbonded class won't affect any result as well.

Are you certain of that?  Can you be sure no excluded atom pairs will ever match a NBThole term?  Even if it doesn't happen in the current force field, might it happen in a future one?

In any case, it seems that adding the exclusions is the correct thing to do.  If that situation ever did come up, the interaction should be excluded.  And that will fix the problem in the CPU platform.

> By setting up interaction groups I hope only those is the groups will be computed, in order to reduce computational cost. Am I misunderstanding how the InteractionGroup works?

The performance is complicated.  Interaction groups force it to use a less efficient algorithm than standard nonbonded interactions.  Until recently, interaction groups didn't even use the neighbor list.  I think we just need to test the different possibilities and see how the performance compares.

peastman:
The error on the CPU platform is fixed by adding the following lines starting at line 1415:

```python
        if has_drude_particle and has_nbthole_terms:
            for i in range(force.getNumExceptions()):
                ii, jj, q, eps, sig = force.getExceptionParameters(i)
                nbtforce.addExclusion(ii, jj)
```

Now I'm trying to test performance using the files I sent you before (but increase the box size to 70.5 A, since I made it a little too small before).  This system was created with CHARMM-GUI, so it ought to be correct, but it blows up immediately when I start simulating it.  The initial potential energy is -116228 kJ/mol.  That falls very rapidly to about -400000 kJ/mol, and then everything becomes NaN.  Am I doing something obviously wrong?  Can you get it to work?

peastman:
@jing-huang still waiting for your reply on this.  Can you get it to run an actual protein simulation, such as the one I sent?

jing-huang:
Sorry for my delayed response. The system you sent actually works fine, while the only issue is step2_drude.crd from CHARMM-GUI not being a proper starting coordinates. If you look at this file you will find out that the coordinates of all Drude particles collide with their parent atoms, which make it a very bad initial configuration to start a MD simulation. With Drude force fields one has to allow all Drude particles to relax - by minimizing their positions with all real particles fixed - to create realistic coordinates. With such a corrected crd file (emailed you), I have no trouble running 100ps MD simulations. At 300K the potential energies of this system stably fluctuate around -386000 kJ/mol, and the speed of 6.3 ns/day is obtained with a 1080Ti.   

Please note that in the CHARMM-GUI files there are CHARMM and NAMD inputs to create the Drude-relaxed starting coordinates, although one might have to reduce the number of minimization steps to avoid polarization catastrophe. Putting together a DrudeMinimizer inside OpenMM will be a easy task to do though, which will improve the self-dependence of the Drude/OpenMM implementation.                         

I'm now removing all interaction groups and comparing the performace.

peastman:
Can you post the code you're using to run it?  I tried using the starting coordinates from the file you sent, but the energy still fluctuates wildly, then becomes NaN after only eight steps.

peastman:
Here is the code I'm using.

```python
from simtk.openmm import *
from simtk.unit import *
from sys import stdout

psf = CharmmPsfFile('drude/step2_drude.psf')
psf.setBox(70.5*angstroms, 70.5*angstroms, 70.5*angstroms)
crd = CharmmCrdFile('step3_mini.crd')
params = CharmmParameterSet('drude/toppar_drude/toppar_drude_master_protein_2013e.str')
system = psf.createSystem(params, nonbondedMethod=PME)
integrator = DrudeLangevinIntegrator(300*kelvin, 1.0/picosecond, 1*kelvin, 10/picosecond, 0.001*picoseconds)
simulation = Simulation(psf.topology, system, integrator, Platform.getPlatformByName('CUDA'))
simulation.context.setPositions(crd.positions)
simulation.reporters.append(StateDataReporter(stdout, 1, step=True, potentialEnergy=True))
simulation.step(1000)
```

Here is the output.

```
#"Step","Potential Energy (kJ/mole)"
1,-276842.9894447401
2,44673.01055525988
3,-21900.989444740117
4,-180698.98944474012
5,-4724.989444740117
6,-77172.98944474012
7,-155152.98944474012
  File "test.py", line 16, in <module>
    simulation.step(1000)
  File "/home/peastman/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/simulation.py", line 132, in step
    self._simulate(endStep=self.currentStep+steps)
  File "/home/peastman/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/simulation.py", line 235, in _simulate
    self._generate_reports(unwrapped, False)
  File "/home/peastman/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/simulation.py", line 254, in _generate_reports
    reporter.report(self, state)
  File "/home/peastman/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/statedatareporter.py", line 195, in report
    self._checkForErrors(simulation, state)
  File "/home/peastman/miniconda3/lib/python3.6/site-packages/simtk/openmm/app/statedatareporter.py", line 347, in _checkForErrors
    raise ValueError('Energy is NaN')
ValueError: Energy is NaN
```

jing-huang:
I attached my input below. Please note that 1) the Drude Hardwall is used to prevent Drude particle being too far (0.25 A) from its parent atom. 2) A smaller timestep (0.5fs instead of 1fs) is used. For Drude simulations, it's common that one would either perform heating to gradually increase simulation temperature to the desired one, or start simulation at 300K but using a shorter timestep (0.5fs or even 0.1fs) and then switch to the normal 1 fs timestep. 


from simtk.openmm import *
from simtk.unit import *

psf = CharmmPsfFile('drude/step2_drude.psf')
psf.setBox(70.5*angstroms, 70.5*angstroms, 70.5*angstroms)
crd = CharmmCrdFile('drude/step3_mini.crd')
params = CharmmParameterSet('drude/toppar_drude/toppar_drude_master_protein_2013e.str')
system = psf.createSystem(params, nonbondedMethod=PME)
integrator = DrudeLangevinIntegrator(300*kelvin, 5/picosecond, 1*kelvin, 20/picosecond, 0.0005*picoseconds)
integrator.setMaxDrudeDistance(0.025)
simulation = Simulation(psf.topology, system, integrator, Platform.getPlatformByName('CUDA'))
simulation.context.setPositions(crd.positions)
simulation.context.computeVirtualSites()

nstep=200000 # simulate for 100ps
nprint=200   # report every 0.1 ps 
simulation.reporters.append(StateDataReporter('test.out', nprint, step=True, kineticEnergy=True, potentialEnergy=True, totalEnergy=True, speed=True))
simulation.step(nstep)

peastman:
Thanks!  That does work.  Now I can look into the performance of the interaction groups.  I'll let you know what I find.

peastman:
I'm getting 31.7 ns/day with the interaction groups, and 28.4 ns/day without.  So it looks like they help.  If you can just make the one change noted above to fix the CPU platform, I think this should be ready to merge!

jing-huang:
Thanks. The fix is added.

peastman:
Fantastic.  Thanks so much for doing this!  It's a great addition.

PR #2132, opened 2018-07-24
Attempt at fixing AppVeyor (again...)

peastman:


peastman:
That fixed the problem, although it then timed out trying to run the tests.

PR #2134, opened 2018-07-25
minor fix for a NameError in getAtomIndexByName

stnma7e:
When an atom does not exist in the defined atom types, instead of raising an error with the available atom names, a NameError is raised because atomIndices is not in scope.


peastman:
Thanks, good catch!

PR #2174, opened 2018-09-21
Fix bug in CUDA LJPME energy without Coulomb terms

andysim:
The existing call to the CUDA LJ-PME energy is not idempotent if Coulomb terms are absent; this stems from the energy buffer not being zeroed.  If Coulomb terms are present, the buffer is correctly zeroed each energy call (although I can't see where that happens right now) and the code is correct.  A simple call to zero that buffer when Coulomb terms are absent does the trick, and I've added a test to check that the energy call is idempotent.

Most uses of LJ-PME require Coulomb terms, so I wouldn't expect this to impact too many people; pure Lennard-Jones fluid simulations with PME would be impacted and the use case I encountered was running a separate force object to handle electrostatics, while getting LJPME dispersion forces from the normal NonbondedForce.  Sorry for not getting this right when I first coded up the dispersion PME.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -2192,8 +2192,6 @@ double CudaCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeF
             cu.executeKernel(pmeInterpolateForceKernel, interpolateArgs, cu.getNumAtoms(), 128);
         }
 
-        // As written, we check only the Electrostatic grid pointer to get here.  We could separate them out, but for
-        // now we assume that LJPME can only be used if electrostatic PME is also active.
         if (doLJPME && hasLJ) {
             if (!hasCoulomb) {
                 void* gridIndexArgs[] = {&cu.getPosq().getDevicePointer(), &pmeAtomGridIndex.getDevicePointer(), cu.getPeriodicBoxSizePointer(),
@@ -2202,6 +2200,7 @@ double CudaCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeF
                 cu.executeKernel(pmeDispersionGridIndexKernel, gridIndexArgs, cu.getNumAtoms());
 
                 sort->sort(pmeAtomGridIndex);
+                cu.clearBuffer(pmeEnergyBuffer);
             }
 
             cu.clearBuffer(pmeGrid2);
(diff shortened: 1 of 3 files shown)


peastman:
Thanks!  It took me a bit of work to figure out what was actually going on here.  The key is these lines from the kernel:

```c++
#if defined(USE_PME_STREAM) && !defined(USE_LJPME)
    energyBuffer[blockIdx.x*blockDim.x+threadIdx.x] = 0.5f*energy;
#else
    energyBuffer[blockIdx.x*blockDim.x+threadIdx.x] += 0.5f*energy;
#endif
```

The idea is that the kernel first gets executed for Coulomb, and when it does it directly stores its energy into the array, overwriting whatever is there.  Then it gets executed again for LJ, and it adds its contribution on to what got calculated in the first call.  But if you're not doing Coulomb that doesn't happen, so your fix is exactly right.

There's another call to clearBuffer() inside initialize() at https://github.com/pandegroup/openmm/blob/master/platforms/cuda/src/CudaKernels.cpp#L1804.  With this fix, I believe that line can be removed.

It looks to me like the same problem also exists in the OpenCL platform.  The fix should be almost identical.

andysim:
Ah, that makes a lot of sense.  I'll modify the OpenCL stuff now and push an update.  The OpenCL platform should already have the modified test case because it's the generic version that I modified.  Thanks for the explanation about the buffers - I'd missed that increment operator in the energy call.

peastman:
run jenkins

peastman:
@swails any idea why jenkins isn't running on this PR?

andysim:
I'm pretty sure all the tests passed on my machine, but I'll run them again.  By the way, you may be interested in [this](https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/).  We're considering that for Psi4 because we keep getting timeouts on Appveyor.  Allows macOS testing, too.

swails:
run jenkins

swails:
We may have to change some settings. I’ll message you later tonight with the webhook configuration that works for me on ParmEd.

peastman:
Ok, thanks!  In the mean time, I'm going to go ahead and merge this since Andy tested it manually.

andysim:
The re-run of the tests are still going.  I'll let you know if anything fails so that you can revert if needed.  The only tests that Appveyor complained about were timeouts unrelated to my changes, so I'm not too concerned.

peastman:
Thanks!  Appveyor often times out, so that's not a concern.  The main issue is that Jenkins is the only place where the CUDA tests get run.

andysim:
I have a very silly followup question: when I run
`conda create -n mpid numpy openmm -c omnia/label/devcuda92 -c omnia`
the resulting version of OpenMM that's installed is
`openmm:          7.4.0-py36_cuda92_1     omnia/label/devcuda92`, which seems to be what I want.  However the LJPME patch doesn't appear to be present in that version, based on a test run that I have.  The files on conda show that this version was built just a few hours ago.  Is there some other label I should use to get the master version through Conda?  I can always build from source, so it's not a big deal.

jchodera:
They're built nightly, and that package was built just 4 hours ago:

It's possible `conda` cached the old version. Try `conda remove --force openmm` and then `conda clean -tipsy` to clear out your cache before reinstalling the package?

andysim:
Thanks a lot @jchodera  - that fixed it.  I had no idea that Conda caches packages so aggressively.

PR #2182, opened 2018-10-04
Fix uninitialized variable when Coulomb terms are missing

andysim:
The CUDA NonbondedForce routines check for the presence of Coulomb terms, setting `hasCoulomb` accordingly.  If this boolean is false the variable `usePosqCharges` is undefined, leading to the charge elements of `posq` sometimes being clobbered, sometimes not.  This has no consequence for the NonbondedForce (Coulomb terms are off), but I have a plugin that store some charges in `posq` that is affected when the charges in NonbondedForce are zeroed (I use it for LJPME terms).  I'm sure what I'm doing is bad design, but this simple fix gets things up and running for me.  The full test suite passes for me on a P100.


andysim:
Should've thought to look for that; thanks for the reminder.  I fixed it in the latest commit, and tests still pass.

peastman:
Thanks!

jchodera:
@andysim : The new 7.4.0 `devcudaXY` packages are being built right now from the `master`, but be aware that I also have a 7.5.0 package up for a specific earlier commit since we're trying to debug when everything started NaNing in all our applications.

Issue #2191, opened 2018-10-24
Potential energy inconsistency when creating new contexts with CUDA code


JIMonroe:
Hopefully this is just a fringe case, but I'm posting this to make sure.

When I run OpenMM (7.3.0 for python 2.7) using the CUDA platform (CUDA 8.0 with drivers 367.44) on Pascal GeForce GTX 1080 GPUs, I run into inconsistent potential energies when creating new contexts.  This seems to happen randomly and only when I'm working with alchemical systems (so I have CustomNonbondedForce objects).  It happens more frequently when more of the GPU memory is being used and has never happened with CPU or OpenCL platforms.  In the example code and outputs I've attached, I use test systems from OpenMMTools, but I have also reproduced this with other systems as well.  

As a little background, I ran into this while running alchemical simulations to compute free energies of solvation.  In order to save time at the expense of memory, I was creating a different context for every alchemical state to avoid re-computing long-range corrections.  Occasionally, one of the contexts would have a potential energy that was off right from the start and would stay off for the whole simulation.  This was despite the fact that the system object reference by the context would yield correct potential energies when copied and applied to a new context (i.e. I wrote an energy decomposition function based on those available in OpenMMTools alchemical factory and ParmEd).  I think the code provided is replicating this same issue with a minimal amount of code and some standard test systems.  If more test systems are needed, I can provide additional code and input files.

For now I have a work-around.  I have not tried updating my driver version (I know it's old), but may eventually get to that.

[CUDA_Inconsistency.tar.gz](https://github.com/pandegroup/openmm/files/2508640/CUDA_Inconsistency.tar.gz)

jchodera:
The CUDA platform, when used with PME, produces non-deterministic energies and forces for purposes of performance. This can be disabled with a small performance hit by setting the `DeterministicForces` platform flag to `True`. Perhaps this is the cause of the effect you're seeing, since OpenCL and CPU platforms are deterministic?

@peastman: This seems to be hard to find in the OpenMM docs. Could we maybe make this more prominent in one or two places?

JIMonroe:
Thank you!  That's really good to be aware of, but I'm not sure it's the issue.  I tried setting `DeterministicForces` to `True` and now I randomly get `NaN` instead of a number (which is honestly an improvement because it more clearly tells you something is going wrong).

[...]

PR #2195, opened 2018-10-30
Use separate events for different stream synchronizations

peastman:
Yet another attempt at maybe fixing(?) #2191 .


andrrizzi:
Thanks! I'll test this.

andrrizzi:
Should I go ahead and build this even if not all tests are passing?

peastman:
Yes, the only failure is a random OpenCL failure, not related to this.  And AppVeyor is just timing out, as it often does on PRs.

PR #2199, opened 2018-11-04
Implementation of OPLS through DMS reader

egallicc:
A modified DMSReader class to set up systems with the OPLS force field. Examples are in examples/opls/{test_explicit.py,test_implicit.py}. 

This pull request is in response to this forum [discussion](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=9672&p=0&start=0&view=&sid=61fe4d91a4fee0dcc919b7170999848e). As noted there, while the DMS format is very versatile, there is currently a lack of tools to create and manipulate DMS files. We are considering adding DMS support to ParmEd. Until then, adding this feature to the official OpenMM tree is probably optional. We can continue to maintain this feature "off-site".  Whatever works best. 

Thanks

Emilio


peastman:
Sorry for taking so long to get to this!  We were holding off on all pull requests while waiting for 7.3 to be released.  I really should have still reviewed it though.

Anyway, this looks great and will be a nice addition to the DMS reader.  Just a few changes we need.

First of all, lots of the Python tests are failing with this error:

```
E     File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/simtk/openmm/app/desmonddmsfile.py", line 211
E       length_conv = u.angstrom.conversion_factor_to(u.nanometer)
E                                                                ^
E   TabError: inconsistent use of tabs and spaces in indentation
```

Indentation should be done exclusively with spaces, never tabs.  I suggest doing a search and replace to find all tabs in the file and fix them.

We also need some tests to make sure it's working.  You can add one or more methods to TestDesmondDMSFile.py.  A good sort of test when adding a new force field like this is to load a small file, compute the energy, and compare to the value computed in Desmond.  That way you can make sure that everything's being interpreted correctly.  For an example of this sort of test, see https://github.com/pandegroup/openmm/blob/master/wrappers/python/tests/TestGromacsTopFile.py#L54-L64.  The files you want to load can go in wrappers/python/tests/systems.  For bonus points, you can put each Force into a separate force group and compare their energies individually!

Do try to keep the test files small, preferably less than 1MB total.  

Can you remove all the files from the examples directory?  That isn't really what it's for.

wrappers/python/simtk/openmm/app/desmonddmsfile.py
```diff
@@ -212,17 +344,113 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*nanomete
...
+                        gb = AGBNP3Force()
+                        # add particles
+                        for i in range(len(gb_parms)):
+                            p = gb_parms[i]
+                            gb.addParticle(p[0],p[1],p[2],p[3],p[4],p[5],p[6])
+                        # connection table (from bonds)
+                        self._add_agbnp2_ct(gb)
+                        sys.addForce(gb)
+                else:
+                    print('Warning: AGBNP3 is not supported in this version')
```
peastman: Wrap this in a `if self._verbose`.

wrappers/python/simtk/openmm/app/desmonddmsfile.py
```diff
@@ -212,17 +344,113 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*nanomete
...
+                        # add particles
+                        for i in range(len(gb_parms)):
+                            [radiusN,chargeN,gammaN,alphaN,hbtype,hbwN,ishydrogenN] = gb_parms[i]
+                            h_flag = ishydrogenN > 0
+                            gb.addParticle(radiusN, gammaN, alphaN, chargeN, h_flag)
+                        sys.addForce(gb)
+                        self.gb_parms = gb_parms
+                        self.agbnp = gb
+                else:
+                    print('Warning: AGBNP is not supported in this version')
```
peastman: Wrap this in a `if self._verbose`.

wrappers/python/simtk/openmm/app/desmonddmsfile.py
```diff
@@ -400,21 +657,23 @@ def _addNonbondedForceToSystem(self, sys):
...
+            
         n_total = self._conn.execute("""SELECT COUNT(*) FROM pair_12_6_es_term""").fetchone()
         n_in_exclusions = self._conn.execute("""SELECT COUNT(*)
         FROM exclusion INNER JOIN pair_12_6_es_term
         ON exclusion.p0==pair_12_6_es_term.p0 AND exclusion.p1==pair_12_6_es_term.p1""").fetchone()
         if not n_total == n_in_exclusions:
-            raise NotImplementedError('All pair_12_6_es_terms must have a corresponding exclusion')
-
-        return nb
+            print('Warning: not all pair_12_6_es_terms have a corresponding exclusion')
```
peastman: Are there situations when this isn't an error?

egallicc: Right, sorry, we missed all of the unit testing things. We'll repropose once done.

I think we will use IMPACT to get reference energies rather than Desmond. I am actually not sure how to give .dms files to recent versions of Desmond. And in any case, the implementation of OPLS in Desmond was originally benchmarked against IMPACT ... 

Concerning: 

> Are there situations when this isn't an error?

None that I know of. Thank you for pointing that out. That check needs to be fixed as it does not account for reversed indexes in the two tables (0,1 vs. 1,0 for example). We meant to do it but it fell through the cracks. It will be fixed.

egallicc:
Right, sorry, we missed all of the unit testing things. We'll repropose once done.

I think we will use IMPACT to get reference energies rather than Desmond. I am actually not sure how to give .dms files to recent versions of Desmond. And in any case, the implementation of OPLS in Desmond was originally benchmarked against IMPACT ...

egallicc:
Trying again after implementing the suggestions ...the many diffs for desmonddmsfile.py are mostly changes in indentation to support loading multiple .dms files into one openmm system (used to build complexes for binding free energy calculations).

wrappers/python/simtk/openmm/app/desmonddmsfile.py
```diff
@@ -212,10 +369,107 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*nanomete
...
+                    gb.finalize()
+                    sys.addForce(gb)
+
+            if implicitSolvent is 'AGBNP3':
+                #load AGBNP3 plugin if available
+                try:
+                    from AGBNP3plugin import AGBNP3Force
+                    AGBNP3enabled = True
+                except ImportError:
+                    AGBNP3enabled = False
```
peastman: If the user specifically requested a solvation model that isn't available, that should produce an exception.  Otherwise it's very easy for people to think they're simulating something totally different from what they really are.

wrappers/python/simtk/openmm/app/desmonddmsfile.py
```diff
@@ -212,10 +369,107 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*nanomete
...
+                if self._verbose:
+                    print('Using GVolSA')
+
+            if implicitSolvent is 'AGBNP':
+                #load AGBNP plugin if available
+                try:
+                    from AGBNPplugin import AGBNPForce
+                    AGBNPEnabled = True
+                except ImportError:
+                    AGBNPEnabled = False
```
peastman: Same here.

swails:
run jenkins

swails:
run jenkins

swails:
run jenkins

swails:
run jenkins

swails:
run jenkins

egallicc:
Implemented error checking for implicit solvation settings

peastman:
Looks great.  I think this is ready to merge.  Thanks!

PR #2241, opened 2019-01-08
Fixed bug involving nested cross products

### LINKED ISSUE #2239: Nested cross products don't work in CUDA / OpenCL
While working with the cross product functionality in v7.3, I seem to have stumbled upon a bug. Consider a trivial modification of the standard Velocity Verlet integrator given in the documentation (I know the integrator doesn't do anything useful, this is just an MWE):
```
class TestIntegrator(mm.CustomIntegrator):
    def __init__(self, timestep=1.0 * unit.femtoseconds):
        super(TestIntegrator, self).__init__(timestep)
        self.addPerDofVariable("omega", 0)
        self.addPerDofVariable("OmV", 0)
        self.addPerDofVariable("test1", 0)
        self.addPerDofVariable("test2", 0)
        self.addPerDofVariable("x1", 0);
        self.addUpdateContextState();
        self.addComputePerDof("omega", "vector(0,0,1)")
        self.addComputePerDof("v", "v+0.5*dt*f/m");
        self.addComputePerDof("OmV", "cross(omega,v)")
        self.addComputePerDof("test1", "cross(omega,cross(omega,v))")
        self.addComputePerDof("test2", "cross(omega,OmV)")
        self.addComputePerDof("x", "x+dt*v");
        self.addComputePerDof("x1", "x");
        self.addConstrainPositions();
        self.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt");
        self.addConstrainVelocities();       
```

To test what happens with the cross products, I just run
```
simulation.step(1)
omega = integrator.getPerDofVariableByName("omega")
print("omega =", omega[0:3])
test1 = integrator.getPerDofVariableByName("test1")
print("test1 =", test1[0:3])
test2 = integrator.getPerDofVariableByName("test2")
print("test2 =", test2[0:3])    
```

The output for `test1` and `test2` should be identical. This is indeed the case on the Reference and CPU platforms. On the CUDA and OpenCL platforms, however, `test2` is correct, whereas `test1` is always filled with zeros (no errors are reported). I don't know enough about the GPU infrastructure in OpenMM to pinpoint the cause of the bug, but hopefully this MWE can help or at least let other users know how to avoid a silent cause of incorrect results.

peastman:
Fixes #2239.

### DIFF
--- a/platforms/cuda/src/CudaExpressionUtilities.cpp
+++ b/platforms/cuda/src/CudaExpressionUtilities.cpp
@@ -97,6 +97,7 @@ void CudaExpressionUtilities::processExpression(stringstream& out, const Express
             // at once, so check to see if both are needed.
 
             vector<const ExpressionTreeNode*> nodes;
+            nodes.push_back(&node);
             for (int j = 0; j < (int) allExpressions.size(); j++)
                 findRelatedCustomFunctions(node, allExpressions[j].getRootNode(), nodes);
             vector<string> nodeNames;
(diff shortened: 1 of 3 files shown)


jchodera:
@peastman : Any ideas if this would affect other kinds of operations too? If so, what should we be looking for?

peastman:
It affects functions that are implemented by Lepton custom functions.  That includes tabulated functions, vector functions, and periodicdistance().  It only happens when two such functions are nested.  For example, it could potentially happen if one of the arguments to a tabulated function is itself computed from a tabulated function.

PR #2245, opened 2019-01-22
Fix AmoebaStretchBendForceGenerator

WangXinyan940:
AmoebaStretchBendForceGenerator has a bug when swapping the index of A and C. It can swap the value of bondAB and bondCB, but cannot swap k1 and k2.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -4192,7 +4192,10 @@ class AmoebaStretchBendGenerator(object):
                        raise ValueError(outputString)
 
                     else:
-                        force.addStretchBend(angle[0], angle[1], angle[2], bondAB, bondCB, angleDict['idealAngle']/radian, self.k1[i], self.k2[i])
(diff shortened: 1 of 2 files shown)


wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -4192,7 +4192,10 @@ def createForcePostAmoebaBondForce(self, sys, data, nonbondedMethod, nonbondedCu
                        raise ValueError(outputString)
 
                     else:
-                        force.addStretchBend(angle[0], angle[1], angle[2], bondAB, bondCB, angleDict['idealAngle']/radian, self.k1[i], self.k2[i])
+                        if type1 == types1:
```
peastman: I think this line should be `if type1 in types1`.  `type1` is the type of the first atom.  `types1` is a list of allowed types that the first atom could have to match these parameters.

swails:
Is this still a problem?  And should it be fixed still?

swails:
Is this still relevant?  It's ~3 years old now and conflicts with the master branch.

PR #2246, opened 2019-01-22
Properly define GBSAOBC alpha beta gamma constants to be doubles.

proteneer:
This took me way too long to find. There's no reason why these should be defined as floats. I was comparing against my tensorflow reference implementation and found a discrepancy in double precision.


peastman:
Good catch.  Why did you write them as `long doubles` (with the `L` suffix) instead of `doubles`?  That isn't going to accomplish anything, since they immediately get assigned to variables of type `double`.

peastman:
`f` means single, no suffix means double, and `L` means long double, which depending on the compiler and CPU could be anything from 64 to 128 bits.

Anyway, looks good.  Thanks!

Issue #2250, opened 2019-01-28
OpenCL platform falling back to CPU for minimization


tristanic:
Just updated to OpenMM 7.3 for ChimeraX/ISOLDE, and my built-in demo/test model now runs agonizingly slow on minimization - about 5 minutes where it used to take 1-2 seconds. Each 1,000-round minimization seems to spend most of its time on the Reference platform (one CPU running at 100%), with a brief spike to ~26 cores at the end. Once it _does_ successfully minimize the dynamics runs at the expected speed. The culprit seems to be the MDFF energy term (in this case a `CustomCompoundBondForce` with the energy equation below).

```
'select(step(enabled-0.5), -mdff_global_k_1 * individual_k * map_potential(i,j,k), 0); i = x1 * mdff_rot00_1 + mdff_trn0_1;j = y1 * mdff_rot11_1 + mdff_trn1_1;k = z1 * mdff_rot22_1 + mdff_trn2_1'
```

where `map_potential` is a `Continuous3DFunction`. If I switch to an alternative form which uses a `Discrete3DFunction` instead, the result is the same - but if I leave the map force out entirely then the minimization runs almost instantly. Doesn't seem related to out-of-gamut forces - behaviour is the same whether I start from a model with minor clashes or one that is pre-equilibrated. In the worst case the system energy starts at ~-400,000 kJ/mol and converges to ~-415,000 kJ/mol, which seems quite reasonable.

Not really sure where to go from here in terms of debugging.

peastman:
It only switches to the CPU if it encounters a force that's too big for the OpenCL platform to represent: greater than +/- 2 billion kJ/mol/nm.  That doesn't necessarily mean it's starting from a bad clash, just that it encounters one at some point while doing a line minimization.

Does the CustomCompoundBondForce involve a large number of bonds?  We could potentially create an implementation of it for the CPU platform that would use multiple cores.  That wouldn't completely fix the problem, but it would help.

tristanic:
I'll run a few more experiments to check (e.g. by turning the force 
coupling constant to a very low value or zero) but it looks to me like 
there's some bug here. There's no way the CustomCompoundBondForce should 
be able to generate anything *like* those values - it's tuned so that 
the MDFF forces are substantially weaker than anything coming out of the 
normal forcefield. I also tried setting {"Precision": "double"} when 
setting up the Simulation, and saw the same behaviour (for the same 
construct that started essentially instantly in OpenMM 7.2, where the 
[...]

PR #2253, opened 2019-02-05
Improved behaviour of LocalEnergyMinimizer

tristanic:
Significantly improves the behaviour of the minimizer in my tests as discussed in #2250, particularly on the GPU and in the presence of constraints. Only automatically falls back to the CPU if an inf or NaN force is encountered on the GPU. If a force greater than 2e9 kJ/mol/nm is encountered on the GPU **and** the minimization fails to converge to within constraint tolerance, an `OpenMMException` is raised suggesting to try the CPU platform, but advising to carefully check the model for the source of the problem first. It might be worth extending upon this to catch cases where there are no constraints and it fails to converge to any reasonable energy, and throw a meaningful exception. 

In my test case (the demo model built into ISOLDE - a MDFF simulation of 3io0 in implicit solvent, with all hydrogen bond lengths constrained) on the OpenCL platform it now repeatedly recovers from a random perturbation of all atoms up to +/-0.5A along each axis in about 3 seconds, and recovers from 1A perturbations (i.e. truly crazy geometry) about 4 times out of 5.

### DIFF
--- a/openmmapi/src/LocalEnergyMinimizer.cpp
+++ b/openmmapi/src/LocalEnergyMinimizer.cpp
@@ -39,6 +39,7 @@
 #include <string>
 #include <vector>
 #include <algorithm>
+#include <iostream>
 
 using namespace OpenMM;
 using namespace std;
@@ -47,6 +48,7 @@ struct MinimizerData {
     Context& context;
     double k;
     bool checkLargeForces;
+    bool largeForceEncountered=false;
     VerletIntegrator cpuIntegrator;
     Context* cpuContext;
     MinimizerData(Context& context, double k) : context(context), k(k), cpuIntegrator(1.0), cpuContext(NULL) {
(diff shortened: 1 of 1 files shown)


peastman:
I don't think this is the right behavior.  For one thing, forces greater than 2e9 are the most common sign that clipping is present, *not* nan or inf.  For another, this is a very common situation.  Throwing an exception and telling the user to switch to the CPU platform makes things a lot harder on the user.  It also makes minimization much slower.  That requires the entire minimization to run on the CPU, whereas currently it uses the CPU for just a few steps until the clash is resolved, then goes back to the GPU.

I'm also nervous about your changes to the configuration parameters.  I tested this on various systems and tried to find values that worked reliably on across many cases.  Different values might work better on your particular system but be worse on other systems.  We shouldn't change them without thorough testing to make sure the new values are consistently better.

How about if we start by just making the single change

```c++
double k = 100/workingConstraintTol;
```

Then we can evaluate how things are working across lots of test cases and consider what further changes are appropriate.

jchodera:
> Throwing an exception and telling the user to switch to the CPU platform makes things a lot harder on the user. 

What about having selectable behavior, controlled by an optional flag, with the default behavior as before?

peastman:
Under what circumstances would you want the alternative behavior?

tristanic:
That’s fine with me. My bias is towards not making it too easy for the user to shoot themselves in the foot - people have a tendency to expect minimization and MD to correct for all manner of sins, but the reality is that coordinates causing forces of those magnitudes are often going to be garbage-in, garbage-out whether or not they successfully minimize.

Anyway, I can accept that my needs (in particular, my need to avoid the CPU fall-back) are well outside mainstream use, and I can write my own custom implementation easily enough. 

 
 
Tristan Croll
Research Fellow

jchodera:
> Under what circumstances would you want the alternative behavior?

My understanding was that @tristanic needed the alternate behavior, but I may have misunderstood.

tristanic:
No need to go to that trouble if I’d be the only one to use it. I’m already calling minimize() exclusively through the C++ interface, so now I know where to look it’ll be a fairly trivial task to roll my own. 

 
 
Tristan Croll
Research Fellow

tristanic:
I've gone ahead and implemented a custom minimizer in ISOLDE specific to its needs. For the official one, @peastman's approach is fine by me.

PR #2261, opened 2019-02-13
Copy missing ApoA1 benchmarking file into examples/ on install

jchodera:
@peastman: The `apoa1.pdb` file (required by `benchmark.py` for some benchmarks) was not copied into `examples` on installation. This PR adds that file to the installation.

I notice there are other files in the `examples/` directory that are also not installed, such as `input-charmm36-tip5p.dms`. Should these other files be included as well?

Is there a more robust way to do this with `CMakeLists.txt` than to list all the files explicitly, especially since we seem to forget to modify this file when new example files are added?

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Generate and install examples.
 #
 # This is boilerplate code for generating a set of executables, one per
-# .cpp file in an "examples" subdirectory. 
+# .cpp file in an "examples" subdirectory.
 #
 # For IDEs that can deal with PROJECT_LABEL properties (at least
 # Visual Studio) the projects for building each of these adhoc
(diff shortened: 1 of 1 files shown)


peastman:
We could switch it to use wildcard patterns.  For example, *.pdb rather than listing the PDB files individually.

tobigithub:
tnx,
just a quick solution as reference for others, command on LINUX to get PDB:
```
https://raw.githubusercontent.com/pandegroup/openmm/master/examples/apoa1.pdb
```

jchodera:
> We could switch it to use wildcard patterns. For example, *.pdb rather than listing the PDB files individually.

Sounds great! 

I'd still love to see some organization of the structure within the examples directory, along with a clear `README.md` to describe what is in there, but that would probably have to wait for a more major release.

peastman:
Should I go ahead and merge this, or wait for you to redo it with wildcards?

jchodera:
@peastman: I've switched these to wildcards.

I'd still love to introduce some subdirectory structure and READMEs into the `examples/` directory at some point, if you'd be OK with it. Vomiting a bunch of unrelated files into a single `examples/` directory just seems unkind to our users.

jchodera:
@peastman: Well, it looks like `CMake` doesn't understand wildcards:
```
CMake Error at examples/cmake_install.cmake:68 (file):
  file INSTALL cannot find
  "/Users/travis/build/pandegroup/openmm/examples/*.py".
  cmake_install.cmake:243 (include)
```
How about you take this over and open your own PR with whatever solution you prefer?

peastman:
It doesn't work like that.  You have to use `file(GLOB ...)` to build the list of matching files.  https://cmake.org/cmake/help/v3.0/command/file.html.

> I'd still love to introduce some subdirectory structure

Sounds very reasonable.

peastman:
Closing this since the problem has already been fixed.

PR #2263, opened 2019-02-18
simple optimization to ForceField class

tristanic:
For force field libraries containing large numbers of residue templates (about 15,000 in my current case), `ForceField.__init__()` spends some 20-30% of its time (4 of a total 20 seconds) just on `ForceField._TemplateData.getAtomIndexByName()`. A few simple adjustments cut that by a factor of 10.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -266,13 +266,13 @@ class ForceField(object):
                     atomIndices = template.atomIndices
                     for ia, atom in enumerate(residue.findall('Atom')):
                         params = {}
-                        a_attrib = atom.attrib
-                        atomName = a_attrib.pop('name')
+                        for key in atom.attrib:
+                            if key not in ('name', 'type'):
+                                params[key] = _convertParameterToNumber(atom.attrib[key])
+                        atomName = atom.attrib['name']
                         if atomName in atomIndices:
                             raise ValueError('Residue '+resName+' contains multiple atoms named '+atomName)
-                        typeName = a_attrib.pop('type')
-                        for key in atom.attrib:
-                            params[key] = _convertParameterToNumber(atom.attrib[key])
+                        typeName = atom.attrib['type']
                         atomIndices[atomName] = ia
                         template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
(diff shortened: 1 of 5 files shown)


peastman:
Thanks!  It looks like some tests are failing.

tristanic:
Hmm - will double-check in the morning. I don’t routinely build OpenMM, so just hand-copied my edits over from my working 7.3.0 environment. Probably made a typo somewhere.

 
 
Tristan Croll
Research Fellow

tristanic:
Aargh! Sorry - the `addAtom()` method was a new addition I made while copying over from my working code, and of course I managed to add an out-by-one error. I've now rebuilt locally, and checked that all relevant Python tests are passing.

tristanic:
Sorry for being a complete n00b with GitHub, but that last commit was supposed to be on a different branch for a separate pull request. Noticed a few `ResourceWarning`s while running the Python tests - the ones in the test code are fairly harmless, but `PDBxFile`  and `GromacsGroFile` were both leaving their input files open indefinitely.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -263,17 +263,17 @@ def loadFile(self, files):
...
-                    for atom in residue.findall('Atom'):
+                    atomIndices = template.atomIndices
+                    for ia, atom in enumerate(residue.findall('Atom')):
                         params = {}
-                        for key in atom.attrib:
-                            if key not in ('name', 'type'):
-                                params[key] = _convertParameterToNumber(atom.attrib[key])
-                        atomName = atom.attrib['name']
+                        a_attrib = atom.attrib
+                        atomName = a_attrib.pop('name')
```
peastman: This line makes me a little nervous.  We load the XML into the etree data structure, then copy information from it into our own data structures.  But here you're directly modifying the etree data structures.  That's the sort of thing that could easily lead to bugs later on, for example if someone writes more code that looks at those structures and expects them to still reflect the content of the file.

What about keeping the parsing code as it was before and just adding the single line `template.atomIndices = atomIndices` at the end?

tristanic:
That’s fine by me. In my trials it shaved a couple percent off the load time, but that’s pretty minor in the grand scheme of things. Won’t be able to make the edit until tomorrow morning (it’s 6:45pm here) - if you want to do it, feel free.

peastman:
Look good.

PR #2264, opened 2019-02-19
Ensuring opened files are properly closed

tristanic:
Addressing some `ResourceWarning`s appearing in Python tests. `GromacsGroFile` and `PDBxFile` were both keeping their input file handles open indefinitely. The other cases in the test scripts are fairly harmless, but reducing the number of warnings is never a bad thing.


tristanic:
... although come to think of it, `PDBxFile` should only close the input file when passed a filename rather than an already open handle - which is exactly how `PDBFile` already works.

peastman:
Looks good.  Thanks!

PR #2266, opened 2019-02-22
Fixed bug in vector operations in CustomIntegrator on CUDA

### LINKED ISSUE #2249: Bugs in the CustomIntegrator implementation for CUDA
There are bugs in the `CustomIntegrator` implementation for the CUDA platform. Some expressions cannot be compiled due to missing definitions of some functions and operators. The CUDA compiler throws error messages like:

> error: no operator "+" matches these operands
>            operand types are: float + double3

My guess is that these bugs emerged with the recent implementation of vector operations. I'm using version 7.3.0, installed from omnia/label/cuda100. I could reproduce some of these errors with the piece of code below. Platforms CPU and OpenCL seem to work properly.

```
from simtk import openmm
from simtk import unit

waterbox = testsystems.FlexibleWaterBox()
platform = openmm.Platform.getPlatformByName('CUDA')
properties = dict(Precision='mixed') if platform.getName() == 'CUDA' else dict()

integrator = openmm.CustomIntegrator(1*unit.femtoseconds)
integrator.addComputePerDof('v', 'v+0.5*dt*f/m')
integrator.addComputePerDof('v', 'max(-vlim,min(v,vlim)); vlim=1000')
integrator.addComputePerDof('x', 'x+dt*(v+1)-dt')
integrator.addComputePerDof('v', 'v+0.5*dt*f/m')

simulation = openmm.app.Simulation(waterbox.topology, waterbox.system, integrator, platform, properties)
simulation.context.setPositions(waterbox.positions)
simulation.context.setVelocitiesToTemperature(300*unit.kelvin)
simulation.step(1)
```

peastman:
Fixes #2249.

### DIFF
--- a/platforms/cuda/include/CudaExpressionUtilities.h
+++ b/platforms/cuda/include/CudaExpressionUtilities.h
@@ -123,6 +123,7 @@ private:
     void findRelatedPowers(const Lepton::ExpressionTreeNode& node, const Lepton::ExpressionTreeNode& searchNode,
             std::map<int, const Lepton::ExpressionTreeNode*>& powers);
     void callFunction(std::stringstream& out, std::string singleFn, std::string doubleFn, const std::string& arg, const std::string& tempType);
+    void callFunction2(std::stringstream& out, std::string fn, const std::string& arg1, const std::string& arg2, const std::string& tempType);
     std::vector<std::vector<double> > computeFunctionParameters(const std::vector<const TabulatedFunction*>& functions);
     CudaContext& context;
     FunctionPlaceholder fp1, fp2, fp3, periodicDistance;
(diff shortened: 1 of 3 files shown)


jchodera:
Thanks @peastman! Can you comment on what domain of use cases would be impacted?

peastman:
It's not so much a "domain", just special cases.  Any use of the min() or max() function in a custom integrator with CUDA.  Also any expression that, after parsing and optimization, ended up involving an ADD_CONSTANT operation.  Again, only in a custom integrator with CUDA.

The effect of the error was to produce a failure in compiling the kernel.  If you weren't getting any errors from the kernel compiler, your code wasn't affected.

PR #2269, opened 2019-02-26
Avoid repeated matching of duplicate impropers

tristanic:
Doing some profiling of simulation startup in a large forcefield with 2922 different improper definitions, outside of `new_Context()` the repeated calls to `_matchImproper()` in `PeriodicTorsionGenerator` are far and away the most expensive (3.5 seconds for a 229-residue protein vs. about 200ms for the next-most-expensive call). By caching the improper definitions (or lack thereof) for signatures that have already been seen, the number of calls to `_matchImproper()` is reduced from ~2500 to 92, bringing the time cost down to 125 ms.

Incidentally, the `createForce()` methods in `RBTorsionGenerator` and `CustomTorsionGenerator` are near-identical copies of that in `PeriodicTorsionGenerator`. Looks like these could be quite easily refactored as subclasses of a generic base class?

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -1972,7 +1972,7 @@ class PeriodicTorsionGenerator(object):
         for torsion in data.propers:
             type1, type2, type3, type4 = [data.atomType[data.atoms[torsion[i]]] for i in range(4)]
             sig = (type1, type2, type3, type4)
-            sig = frozenset((sig, sig[:-1]))
+            sig = frozenset((sig, sig[::-1]))
             match = proper_cache.get(sig, None)
             if match == -1:
                 continue
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -1968,24 +1968,31 @@ class PeriodicTorsionGenerator(object):
         else:
             force = existing[0]
         wildcard = self.ff._atomClasses['']
+        proper_cache = {}
         for torsion in data.propers:
-            type1 = data.atomType[data.atoms[torsion[0]]]
-            type2 = data.atomType[data.atoms[torsion[1]]]
-            type3 = data.atomType[data.atoms[torsion[2]]]
-            type4 = data.atomType[data.atoms[torsion[3]]]
-            match = None
-            for index in self.propersForAtomType[type2]:
-                tordef = self.proper[index]
-                types1 = tordef.types1
-                types2 = tordef.types2
-                types3 = tordef.types3
-                types4 = tordef.types4
-                if (type2 in types2 and type3 in types3 and type4 in types4 and type1 in types1) or (type2 in types3 and type3 in types2 and type4 in types1 and type1 in types4):
-                    hasWildcard = (wildcard in (types1, types2, types3, types4))
-                    if match is None or not hasWildcard: # Prefer specific definitions over ones with wildcards
-                        match = tordef
-                    if not hasWildcard:
-                        break
+            type1, type2, type3, type4 = [data.atomType[data.atoms[torsion[i]]] for i in range(4)]
+            sig = (type1, type2, type3, type4)
+            sig = frozenset((sig, sig[:-1]))
+            match = proper_cache.get(sig, None)
+            if match == -1:
+                continue
+            if match is None:
+                for index in self.propersForAtomType[type2]:
+                    tordef = self.proper[index]
+                    types1 = tordef.types1
+                    types2 = tordef.types2
+                    types3 = tordef.types3
+                    types4 = tordef.types4
+                    if (type2 in types2 and type3 in types3 and type4 in types4 and type1 in types1) or (type2 in types3 and type3 in types2 and type4 in types1 and type1 in types4):
+                        hasWildcard = (wildcard in (types1, types2, types3, types4))
+                        if match is None or not hasWildcard: # Prefer specific definitions over ones with wildcards
+                            match = tordef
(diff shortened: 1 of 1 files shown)


tristanic:
May as well do the same for the proper torsion search - the gain is much smaller than for impropers, but not negligible and I can't really see any harm in it. Some real-world timings using `prun` on my standard test case:

No optimisations:
```
823829 function calls (799340	 primitive calls) in 9.835 seconds
								
Ordered by: internal time					
								
ncalls	tottime	percall	cumtime	percall	filename:lineno(function)			
1	4.542	4.542	4.542	4.542	{built-in	method	simtk.openmm._openmm.new_Context}	
3425	3.944	0.001	3.944	0.001	forcefield.py:1650(_matchImproper)			
1	0.239	0.239	4.233	4.233	forcefield.py:1962(createForce)			
1	0.148	0.148	0.148	0.148	{built-in	method	simtk.openmm._openmm.new_Continuous3DFunction}	
```

Cached improper search:
```
823713 function calls (799224 primitive calls) in 5.379 seconds
							
Ordered by: internal time				
							
ncalls	tottime	percall	cumtime	percall	filename:lineno(function)		
1	3.963	3.963	3.963	3.963	{built-in	method	simtk.openmm._openmm.new_Context}
1	0.223	0.223	0.386	0.386	forcefield.py:1962(createForce)		
1	0.152	0.152	0.152	0.152	{built-in	method	simtk.openmm._openmm.new_Continuous3DFunction}
92	0.126	0.001	0.126	0.001	forcefield.py:1650(_matchImproper)		
```

Cached impropers and propers:
```
832593 function calls (808104 primitive calls) in 5.291 seconds
								
Ordered by: internal time					
								
ncalls	tottime	percall	cumtime	percall	filename:lineno(function)			
1	3.904	3.904	3.904	3.904	{built-in	method	simtk.openmm._openmm.new_Context}	
1	0.211	0.211	0.211	0.211	{built-in	method	simtk.openmm._openmm.new_Continuous3DFunction}	
92	0.127	0.001	0.127	0.001	forcefield.py:1650(_matchImproper)			
1	0.062	0.062	0.229	0.229	forcefield.py:1962(createForce)			



```

tristanic:
Hmm... I can see from the tests that I've missed something here. Will look into it.

tristanic:
Ouch - two different errors in my original attempt. One: my attempt at a signature for an improper wasn't sufficient (couldn't distinguish between `(type1, type2, type2, type3)` and `(type1, type2, type3, type3)`), and two: I'd badly misunderstood the contents of the `match` tuple. 

Something I'm still a little surprised about: if I uncomment the alternative signature format:
```
            tcount = defaultdict(int)
            or t in (t2,t3,t4):
                tcount[t] += 1
            sig = (t1, frozenset((atype, count) for atype, count in tcount.items()))
```
... then the test fails. With the current ordered version (`sig = (t1, t2, t3, t4`) I get 333 calls to `_matchImproper()` vs. 94 with the alternative signature.

My question is, is this a bug? `_matchImproper()` works through all permutations of the ordering of positions 2, 3 and 4. If there are wildcard matches but no exact matches, it will return the first wildcard match it hits. If there *are* exact matches, it will return the last exact match, and otherwise it will return None. So in at least some cases the improper definition that gets chosen is going to depend on the order in which the code walks through the atom order permutations. That seems out of step with the intention of the code.

tristanic:
OK, I can see that ordering of the improper atoms *does* matter, albeit in ways I still don't fully understand. With the `frozenset`-based signature, somewhere approaching half the impropers come out with the wrong handedness. Something to do with the "bizarre" ordering rules mentioned in the code? In any case, the version in the last commit should be safe (the same four atom types in order should *always* return the same result), and still gets a ~10-fold speedup for my test case (and I would expect substantially bigger speedups for larger models).

tristanic:
Some timings on a much larger model (6g2j *sans* lipids - 131k non-solvent atoms)

Original:
```
         31682943 function calls (30747723 primitive calls) in 181.350 seconds

   Ordered by: internal time

   133061  147.153    0.001  147.162    0.001 forcefield.py:1650(_matchImproper)
        1    2.623    2.623  160.313  160.313 forcefield.py:1023(createSystem)
        1    2.070    2.070  150.859  150.859 forcefield.py:1962(createForce)
        1    1.804    1.804    1.804    1.804 {built-in method simtk.openmm._openmm.new_Context}
```

New:
```
         32094332 function calls (31159112 primitive calls) in 35.893 seconds

   Ordered by: internal time

        1    2.590    2.590   13.679   13.679 forcefield.py:1023(createSystem)
        1    1.948    1.948    4.232    4.232 forcefield.py:1962(createForce)
1365366/443879    1.760    0.000    4.818    0.000 copy.py:132(deepcopy)
        1    1.611    1.611    1.611    1.611 {built-in method simtk.openmm._openmm.new_Context}
	... 
     562    0.632    0.001    0.632    0.001 forcefield.py:1650(_matchImproper)
```

jchodera:
Nice!

peastman:
Impressive!  I'll go through the code shortly.

peastman:
I'm a bit confused by the line

```python
sig = frozenset((sig, sig[:-1]))
```

Could you explain the logic behind that?

tristanic:
Dihedral definition A-B-C-D is equivalent to D-C-B-A, is it not? The idea behind that line is that once the definition is found for the atoms in one order, it should be applied to all following dihedrals matching either order.

peastman:
If `sig` is A-B-C-D, then `sig[:-1]` is A-B-C.  Did you mean to write `sig[::-1]`?

tristanic:
Ah, damn - yes, I did.

peastman:
Looks good!

PR #2273, opened 2019-02-27
Customgbforces

tristanic:
Still on a bit of an optimisation spree here, and it turned out that preparation of the `GBSAGBn2Force` was my next biggest bottleneck in simulation startup (~13 of 36 seconds for 6g2j). Introducing a little numpy magic brings that down to 4.1 seconds. 

This passes `TestAmberPrmtopFile.py`, but not `TestCharmmFiles.py`, and I'm not sure why yet. Hoping for any insights?

### DIFF
--- a/wrappers/python/simtk/openmm/app/internal/customgbforces.py
+++ b/wrappers/python/simtk/openmm/app/internal/customgbforces.py
@@ -250,7 +250,6 @@ def _bondi_radii(topology):
         E.sulfur:       1.8,
         E.chlorine:     1.5,
     }
-    natoms = topology.getNumAtoms()
     radii = numpy.empty(topology.getNumAtoms(), numpy.double)
     for i, atom in enumerate(topology.atoms()):
         radii[i] = element_to_radius.get(atom.element, default_radius)
@@ -287,7 +286,8 @@ def _mbondi_radii(topology, all_bonds = None):
         elif element is E.carbon:
             radii[i] = 1.7
         # All other elements have fixed radii for all types/partners
-        radii[i] = element_to_const_radius.get(element, default_radius)
+        else:
+            radii[i] = element_to_const_radius.get(element, default_radius)
     return radii  # converted to nanometers above
 
 
@@ -832,45 +832,6 @@ class GBSAGBn2Force(GBSAGBnForce):
 
     def __init__(self, solventDielectric=78.5, soluteDielectric=1, SA=None, cutoff=None, kappa=0.0):
         GBSAGBnForce.__init__(self, solventDielectric, soluteDielectric, SA, cutoff, kappa)
-
-
-    # @staticmethod
-    # def getStandardParameters(topology):
-    #     """ Gets list of standard parameters for this GB model based on an input Topology
-    #
-    #     Parameters
-    #     ----------
-    #     topology : simtk.openmm.app.Topology
-    #         Topology of the system to get parameters for
-    #
-    #     Returns
-    #     -------
-    #     list of float
-    #         List of all parameters needed for this GB model. These can be passed
-    #         to addParticle or setParticleParameters after the charge is inserted
-    #         at the beginning of the list
-    #
-    #     """
-    #     import numpy
-    #     natoms = topology.getNumAtoms()
-    #     radii = numpy.empty((6,natoms), numpy.double)
-    #     radii[:,0] = [[x/10] for x in _mbondi3_radii(topology)]
-    #     for i, atom in enumerate(topology.atoms()):
-    #         radii[i,1] = _screen_parameter(atom)[2]
-    #         e = atom.element
-    #         if e in (E.hydrogen, E.deuterium):
-    #             radii[i, 2:]= [0.788440, 0.798699, 0.437334]
-    #         elif atom.element is E.carbon:
-    #             radii[i].extend([0.733756, 0.506378, 0.205844])
-    #         elif atom.element is E.nitrogen:
-    #             radii[i].extend([0.503364, 0.316828, 0.192915])
-    #         elif atom.element is E.oxygen:
-    #             radii[i].extend([0.867814, 0.876635, 0.387882])
-    #         elif atom.element is E.sulfur:
-    #             radii[i].extend([0.867814, 0.876635, 0.387882])
-    #         else:
-    #             radii[i].extend([0.8, 4.85, 0.5])
-    #     return radii
 
     _atom_params = {
         E.hydrogen:     numpy.array([0.788440, 0.798699, 0.437334]),
--- a/wrappers/python/simtk/openmm/app/internal/customgbforces.py
+++ b/wrappers/python/simtk/openmm/app/internal/customgbforces.py
@@ -40,6 +40,13 @@ from simtk.openmm import CustomGBForce, Discrete2DFunction
 import simtk.unit as u
 from math import floor
 
+_have_numpy = False
+try:
+    import numpy
+    _have_numpy = True
+except ImportError:
+    pass
+
 
 def strip_unit(value, unit):
     """Strip off any units and return value in unit"""
@@ -49,8 +56,7 @@ def strip_unit(value, unit):
 
 
 # GBn and GBn2 lookup tables
-d0= numpy.array(
-   [2.26685,2.32548,2.38397,2.44235,2.50057,2.55867,2.61663,2.67444,
+d0= [2.26685,2.32548,2.38397,2.44235,2.50057,2.55867,2.61663,2.67444,
     2.73212,2.78965,2.84705,2.9043,2.96141,3.0184,3.07524,3.13196,
     3.18854,3.24498,3.30132,3.35752,3.4136,
     2.31191,2.37017,2.4283,2.48632,2.5442,2.60197,2.65961,2.71711,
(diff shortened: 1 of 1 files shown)


tristanic:
Argh! Boneheaded mistake. Fixed now.

peastman:
I've worked very hard to make sure that OpenMM's Python layer has no runtime dependencies on any other libraries, not even numpy.  It's still OK to use numpy for optimizations, but the code needs to work correctly (although more slowly) even if it isn't available.  See, for example, https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L1382-L1397.

tristanic:
An admirable choice! OK, will rearrange with that in mind.

jchodera:
Numpy is _so_ prevalent in scientific computing that I really think we should be able to drop this prohibition against very common scientific python libraries in a future release.

swails:
A sensible compromise may be to say that it's OK for new functionality to depend on numpy (and therefore not work if it's not present), but that numpy unavailability shouldn't make OpenMM crash on import (and functionality that doesn't require numpy will work normally).

I tend to think the value of supporting numpy-less installs is limited given the likely prevalence of numpy among OpenMM user base and ease of numpy installation (via pip, to point out two), but it would be hard to argue there is *no* value to it.

tristanic:
Was knocked out by a cold yesterday, but will try to get a revised commit in today.

Getting a little off the topic, but I personally wouldn't be able to do half of what I'm doing without the use of Numpy. I can see it's nowhere near as important in your typical case of a long-running, "hands-off" simulation, but when you're talking about real-time interaction and/or analysis and feedback from external code, then speed really becomes important. For all my `Custom*Force`s used in ISOLDE I've implemented some little bespoke C++ functions allowing me to set/update their parameters from Numpy arrays in a single call. For example:
C++ side:

```
EXPORT void
custombondforce_add_bonds(void *force, size_t n, int *p_indices, double *params, int *f_indices)
{
    OpenMM::CustomBondForce *f = static_cast<OpenMM::CustomBondForce *>(force);
    try {
        int n_params = f->getNumPerBondParameters();
        std::vector<double> param_vec(n_params);
        int p1, p2;
        for (size_t i=0; i<n; ++i) {
            for (int j=0; j<n_params; ++j)
                param_vec[j] = *params++;
            p1 = *(p_indices++);
            p2 = *(p_indices++);
            *(f_indices++) = f->addBond(p1, p2, param_vec);
        }
    } catch (...) {
        molc_error();
    }
}

EXPORT void
custombondforce_update_bond_parameters(void *force, size_t n, int *indices, double *params)
{
    OpenMM::CustomBondForce *f = static_cast<OpenMM::CustomBondForce *>(force);
    try {
        int n_params = f->getNumPerBondParameters();
        std::vector<double> param_vec(n_params);
        int particle1, particle2;
        for (size_t i=0; i<n; ++i) {
            int index = *(indices++);
            f->getBondParameters(index, particle1, particle2, param_vec);
            for (int j=0; j<n_params; ++j) {
                param_vec[j] = *params++;
            }
            f->setBondParameters(index, particle1, particle2, param_vec);
        }
    } catch (...) {
        molc_error();
    }
}
```

```
class TopOutBondForce(CustomBondForce):
    '''
    A :py:class:`openmm.CustomBondForce` subclass defined as a standard harmonic
    potential with a user-defined fixed maximum cutoff on the applied force. Any
    restraint can be switched on (off) by setting the 'enabled' parameter to 1
    (0). This is designed for steering the simulation into new conformations
    where the starting distance may be far from the target bond length, leading
    to catastrophically large forces with a standard harmonic potential.
    '''

    def add_bonds(self, atom_indices, enableds, spring_constants, targets):
        '''
        Add a set of bonds to the simulation, using a fast C++ function. Fastest
        if all parameters are supplied as NumPy arrays.

        Args:
            * atom_indices:
                - a 2-tuple of integer arrays giving the indices of the bonded
                  atoms in the simulation construct
            * enableds:
                - a Boolean array defining which restraints are to be active
            * spring_constants:
                - a float array of spring constants in :math:`kJ mol^{-1} nm^{-2}`
            * targets:
                - a float array of target distances in nanometers
        '''
        # Concatenates indices and params into suitably-shaped Numpy arrays, 
        # and hands them off to custombondforce_add_bonds()

    def update_targets(self, indices, enableds, spring_constants, targets):
        '''
        Update a set of targets all at once using fast C++ code. Fastest if
        the arguments are provided as Numpy arrays, but any iterable will work.

        Args:
            * indices:
                - the indices of the restraints in the OpenMM force object
            * enableds:
                - a Boolean array defining which restraints are to be enabled
            * spring_constants:
                - The new spring constants in units of :math:`kJ mol^{-1} nm^{-2}`
            * targets:
                - the new target distances in nanometres
        '''
        # Hands off to custombondforce_update_bond_parameters()
```

Taking the large loops out of Python like this brings the typical cost of these calls down into the microseconds, greatly increasing the flexibility of what you can practically do with them. The above example is behind my register-shifting tool as illustrated in this short video: https://youtu.be/SS1d0xSXQZs?t=20 - `TopOutBondForce` restraints on N, C, CA and CB moving along parametric splines from their starting to end positions, updating with every graphics update (50 simulation steps). A fairly trivial example in this case, but it's not that hard to think of bigger examples.

tristanic:
OK, updated to make Numpy optional. Turns out that some reasonable speedup was available even without it. Testing it with and without Numpy (by commenting out lines 44-48 for the latter), I get the following timings for 6g2j.

With Numpy, using this wrapper function (where `GBSAForce` is a very thin wrapper around `GBSAGBn2Force`:
```
    def initialize_implicit_solvent(self, params):
        '''
        Add a Generalised Born Implicit Solvent (GBIS) formulation.

        Args:
            * params:
                - a :py:class:`SimParams` instance
        '''

        gbsa_params = {
            'solventDielectric':    params.gbsa_solvent_dielectric,
            'soluteDielectric':     params.gbsa_solute_dielectric,
            'SA':                   params.gbsa_sa_method,
            'cutoff':               params.nonbonded_cutoff_distance,
            'kappa':                params.gbsa_kappa,
            'nonbonded_method':     params.gbsa_cutoff_method,
            }

        top = self.topology
        system = self._system
        from simtk.openmm.openmm import NonbondedForce
        for f in system.getForces():
            if isinstance(f, NonbondedForce):
                break
        n = f.getNumParticles()
        params = numpy.empty((n, 6))

        for i in range(f.getNumParticles()):
            params[i,0] = f.getParticleParameters(i)[0].value_in_unit(
                defaults.OPENMM_CHARGE_UNIT
            )
        from .custom_forces import GBSAForce
        gbforce = self._gbsa_force = GBSAForce(**gbsa_params)
        params[:,1:] = gbforce.getStandardParameters(top)
        gbforce.addParticles(params)
        gbforce.finalize()
        system.addForce(gbforce)
```
        1    0.302    0.302    3.795    3.795 openmm_interface.py:2349(initialize_implicit_solvent)

Without Numpy, with the calling function replacing Numpy with list comprehensions:
```
    def initialize_implicit_solvent(self, params):
        '''
        Add a Generalised Born Implicit Solvent (GBIS) formulation.

        Args:
            * params:
                - a :py:class:`SimParams` instance
        '''
        gbsa_params = {
            'solventDielectric':    params.gbsa_solvent_dielectric,
            'soluteDielectric':     params.gbsa_solute_dielectric,
            'SA':                   params.gbsa_sa_method,
            'cutoff':               params.nonbonded_cutoff_distance,
            'kappa':                params.gbsa_kappa,
            'nonbonded_method':     params.gbsa_cutoff_method,
            }

        top = self.topology
        system = self._system
        from simtk.openmm.openmm import NonbondedForce
        for f in system.getForces():
            if isinstance(f, NonbondedForce):
                break
        n = f.getNumParticles()
        params = [[f.getParticleParameters(i)[0].value_in_unit(defaults.OPENMM_CHARGE_UNIT)]
            for i in range(f.getNumParticles())
            ]

        from .custom_forces import GBSAForce
        gbforce = self._gbsa_force = GBSAForce(**gbsa_params)
        # params[:,1:] = gbforce.getStandardParameters(top)
        [p.extend(sp) for p, sp in zip(params, gbforce.getStandardParameters(top))]
        gbforce.addParticles(params)
        gbforce.finalize()
        system.addForce(gbforce)
```
        1    0.004    0.004    6.865    6.865 openmm_interface.py:2349(initialize_implicit_solvent)

... which is about 2-fold faster than the original implementation.

swails:
run jenkins

swails:
run jenkins

swails:
run jenkins

jchodera:
This looks ready to review and merge, @peastman!

wrappers/python/simtk/openmm/app/internal/customgbforces.py
```diff
@@ -34,11 +34,19 @@
 
 from collections import defaultdict
 import copy
+import numpy
```
peastman: I think you forgot to remove this line.

peastman:
Thanks for reminding me!  Just one small change needed (a line that accidentally got left in) and it can be merged.

wrappers/python/simtk/openmm/app/internal/customgbforces.py
tristanic: Argh! Whoops - fixed now.

peastman:
Looks good, thanks!

PR #2274, opened 2019-03-02
Improved ion addition to both addSolvent() and addMembrane()

JoaoRodrigues:
After the discussion in #2100, addressing #2099, I implemented a private `addIons` method that replaces waters (or any molecule really) by ions, up to a given concentration or to a certain number of ions. I replaced the ion addition routines in each `addSolvent` and `addMembrane` by a call to the new `addIons`.

Tested locally and it works for both membrane and non-membrane systems. I cannot figure out what would be a proper test for this though, to add to the unit tests.

Suggestions are welcome!

### DIFF
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -357,7 +357,7 @@ class Modeller(object):
                 n_trials = 10
 
             if n_trials == 0:
-                raise ValueError('Could not add more than {} ions to the system'.format(addedIons))
+                raise ValueError('Could not add more than {} ions to the system'.format(numAddedIons))
 
         # Replace waters/ions in the topology
         modeller.delete(toReplace)
@@ -366,7 +366,7 @@ class Modeller(object):
             element = (positiveElement if i < numPositive else negativeElement)
             newResidue = modeller.topology.addResidue(element.symbol.upper(), ionChain)
             modeller.topology.addAtom(element.symbol, element, newResidue)
-            modeller.positions.append(replaceableMols[water] * nanometer)
+            modeller.positions.append(replaceableMols[water])
 
         # Update topology/positions
         self.topology = modeller.topology
@@ -622,8 +622,18 @@ class Modeller(object):
         self.topology = newTopology
         self.positions = newPositions
 
+        # Convert water list to dictionary (residue:position)
+        waterPos = {}
+        _oxygen = elem.oxygen
+        for chain in newTopology.chains():
+            for residue in chain.residues():
+                if residue.name == 'HOH':
+                    for atom in residue.atoms():
+                        if atom.element == _oxygen:
+                            waterPos[residue] = newPositions[atom.index]
+
         # Add ions to neutralize the system.
-        self._addIons(forcefield, addedWaters, positiveIon=positiveIon, negativeIon=negativeIon, ionicStrength=ionicStrength, neutralize=neutralize)
+        self._addIons(forcefield, waterPos, positiveIon=positiveIon, negativeIon=negativeIon, ionicStrength=ionicStrength, neutralize=neutralize)
 
     class _ResidueData:
         """Inner class used to encapsulate data about the hydrogens for a residue."""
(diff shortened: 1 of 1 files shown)


JoaoRodrigues:
Think this is finally passing tests (except AppVeyor that is complaining for something weird and unrelated I think...).

swails:
run jenkins

peastman:
> except AppVeyor that is complaining for something weird and unrelated I think

Yeah, I really need to get that working again.  I keep fixing it, and it will work for a few weeks and then break again.

wrappers/python/simtk/openmm/app/modeller.py
```diff
@@ -257,6 +258,120 @@ def convertWater(self, model='tip3p'):
...
+
+        numReplaceableMols = len(replaceableMols)
+
+        # Fetch ion elements from user input
+        if positiveIon not in posIonElements:
+            raise ValueError('Illegal value for positive ion: {}'.format(positiveIon))
+        if negativeIon not in negIonElements:
+            raise ValueError('Illegal value for negative ion: {}'.format(negativeIon))
+        positiveElement = posIonElements[positiveIon]
+        negativeElement = negIonElements[negativeIon]
```
peastman: The above code is still present in `addSolvent()` and `addMembrane()` as well.  It could be removed from those.

wrappers/python/simtk/openmm/app/modeller.py
```diff
@@ -257,6 +258,120 @@ def convertWater(self, model='tip3p'):
...
+                nonbonded = system.getForce(i)
+                break
+        else:
+            raise ValueError('The ForceField does not specify a NonbondedForce')
+
+        totalCharge = 0.0
+        for i in range(nonbonded.getNumParticles()):
+            nb_i = nonbonded.getParticleParameters(i)
+            totalCharge += nb_i[0].value_in_unit(elementary_charge)
+        # Round up to integer
```
peastman: It actually rounds to the nearest integer, not up.

peastman:
Looks good!  Just a couple of minor comments on it.

wrappers/python/simtk/openmm/app/modeller.py
JoaoRodrigues: Ah, forgot to remove those. I was going step by step to make sure I didn't forget anything.

wrappers/python/simtk/openmm/app/modeller.py
JoaoRodrigues: Why not `int(round(totalCharge))` then?

swails: Or why not just `int(total_charge + 0.5)`?  The `floor` call here is superfluous.

:smiley:

Personally I always round via truncating the number + 0.5.  That's a very general way of truncating to a set number of sigfigs:

Nearest 10:

```c++
(int)((N + 5) * 0.1) * 10
```

Nearest 100:
```c++
(int)((n + 50) * 0.01) * 100
```

This is equivalent to `round(N, -1)` and `round(N, -2)` in Python, respectively, but not all languages support the "places" argument for round (e.g., C).

swails:
run jenkins

swails:
run jenkins

swails:
why is my 'run jenkins' not working? :(

wrappers/python/simtk/openmm/app/modeller.py
peastman: I don't have much preference between the different ways of doing it.  They're all pretty much equivalent.  Though I just tested them and found what I consider to be a bug in Python's `round()` function.  `round(0.5)` incorrectly returns 0 instead of 1.

peastman:
Because the world is out to get you.  Also because Nothing Ever Works.  Like the computer I just installed routine software updates on, and now it can't see the mouse and keyboard.

swails:
> Also because Nothing Ever Works.

Ha, I proved that was not true last night!  After weeks of trying, I finally got One Thing to work! (exposing nvidia GPUs to docker containers on Gentoo)

wrappers/python/simtk/openmm/app/modeller.py
JoaoRodrigues: See: `https://docs.python.org/3/library/functions.html#round` - they call it a feature :)

OK, I'll just keep whatever is the norm in OpenMM, for consistency.

JoaoRodrigues:
Thanks to both!

PR #2317, opened 2019-06-06
Support CHARMM psf that contains colinear type of lonepair.

jing-huang:
The CHARMM psf parser have been updated to support systems with colinear type of lonepairs. This enhancement is mainly motivated by the recently updated halogen parameters in the additive CGENFF and polarizable Drude force fields.


peastman:
Thanks!

peastman:
Would it be possible to add a test case for this?  It can be quite simple.  I'd suggest

1. Add a minimal PSF file that includes a lonepair of this type under tests/systems.
2. Add a test in TestCharmmFiles.py that loads the file, moves the virtual site atom to an incorrect position, calls `computeVirtualSites()` on the Context, and checks that it's now in the correct position.

jing-huang:
No problem. I'll add a test.

wonpil:
Peter, 

When will this be merged to the main branch? Currently, OpenMM simulations using the halogen-containing systems generated by CHARMM-GUI all failed due to this issue.

Thanks,
Wonpil

peastman:
I'm just waiting for the test case.  Once that's added I can merge it.

wonpil:
Thanks!

peastman:
The test is failing because the rtf file isn't getting copied to the staging area.  You can fix that by adding another line to this block: https://github.com/pandegroup/openmm/blob/master/wrappers/python/CMakeLists.txt#L49-L75

jing-huang:
Thanks Peter. I have corrected the CMakeLists file as well as corrected an assertEqual mistake, however the checks still fail. Could you please help take a look?

peastman:
The test failure is unrelated.  There's a segfault in TestOpenCLBrownianIntegrator, which I need to look into.  That's in the C++ layer, and this PR only changes Python code.  So I'll go ahead and merge this.  Thanks!

wonpil:
Thanks Peter!

PR #2324, opened 2019-06-19
Documentation: Conda+GPU need to pin packages.

Carreau:
Add into the documentation that when installing with conda you likely
need to pin `cudatoolkit` and `cudaXY` or openmm might just fail to find
CUDA.

-- 

At least we struggled with that on our cluster for a couple of hours. Not sure this is the *only* thing that helped; but at least it was necessary without our version of cuda (9.0), on centos 7.

It would be nice if these were pinned in each openmm from each label; but I'm not sure if this is possible as you do not explicitely depend on it

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -80,7 +80,13 @@ where :code:`cuda92` should be replaced with the particular CUDA version
 installed on your computer.  Supported values are :code:`cuda75`, :code:`cuda80`,
 :code:`cuda90`, :code:`cuda91`, :code:`cuda92`, and :code:`cuda100`.  Because
 different CUDA releases are not binary compatible with each other, OpenMM can
-only work with the particular CUDA version it was compiled with.
+only work with the particular CUDA version it was compiled with. You most likely 
+want to also install the corresponding version of the cuda package :code:`cuda92`
+and the corresponding version of :code:`cudatoolkit` from the default channel
+::
+
+   conda install -c omnia/label/cuda92 cuda92
+   conda install 'cudatoolkit==9.2'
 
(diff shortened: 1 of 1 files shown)


peastman:
Currently we assume that CUDA has been installed with the installer from Nvidia, *not* with conda.  Supporting the conda packages for CUDA toolkits is something we've been looking into.

jchodera:
@Carreau : Thanks for the contribution! As @peastman notes, we're working on migrating to `conda-forge` where we would use the `cudatoolkit` package instead of assuming system-installed CUDA.

On your cluster with CUDA 9.0 installed, did it not work to install from the `cuda90` channel?

peastman:
I'm closing this, since it doesn't match our currently supported method of installing the toolkit.  If we start supporting conda packages for the CUDA toolkit in the future, we should update the documentation accordingly.

Carreau:
> On your cluster with CUDA 9.0 installed, did it not work to install from the `cuda90` channel?

No, I had to manually install 'cuda90' and æcondatoolkit==90' to get it to accept cuda as a platform

> I'm closing this, since it doesn't match our currently supported method of installing the toolkit

Well that is strange, we did have cuda install from nvidia on our cluster and followed the instructions; but setting `ld_library_path` and company did not work. Installing cuda90 and cudatoolkit did though and gave a 30 speedup.

But you know your software better than I so i'll let you judge.

peastman:
If it isn't loading the CUDA platform, try running a Python interpreter and entering the following:

```python
from simtk.openmm import *
print(Platform.getPluginLoadFailures())
```

That will display error messages that hopefully provide useful information about why it couldn't be loaded.

PR #2336, opened 2019-07-01
Install apoa1.pdb

peastman:
One of the files used by the benchmark script wasn't getting installed by `make install`.


jchodera:
see also https://github.com/pandegroup/openmm/pull/2261

PR #2338, opened 2019-07-03
Allow ForceField to add custom prefix to templates when loading XML files

tristanic:
The situation here is that I'm using a very large forcefield containing ligand templates from https://github.com/nwmoriarty/amber_library (covering a bit over half of the Chemical Components Dictionary, at last count). These include quite a number of different sets of chiral enantiomers, for which OpenMM's template-matching won't work - so my code tries to assign templates by name-matching first before falling back to OpenMM's automatic matching for those it can't find. That's working well - but now I want to allow users to be able to load in their own ffXML files via an interactive GUI menu. The problem is, chances are high that the residue name they're using is already in the database - e.g. the most commonly-used of all "working title" ligand names, "LIG", officially corresponds to 3-pyridin-4-yl-2,4-dihydroindeno[1,2-c]pyrazole. 

Anyway, this would seem like the easiest way to handle the problem. It allows me to add the prefix "USER_" to all custom residues loaded via the GUI, so my code can look for the ligand definition there first before moving on to the full forcefield.


peastman:
Does this really completely solve the problem?  In addition to residue name conflicts, it's also possible to get conflicts in atom type names.  Do you need to worry about that?

If nothing else, it might be good to rename the argument to `residue_prefix` to avoid confusion in case it's necessary to add an `atom_type_prefix` option in the future.

tristanic:
That’s a good point. I know there are problems with the AMBER MODRNA library overriding the parameters for arginine guanidinium nitrogens with horrible effects (the NH2 groups start spinning like little propellers). Prefixing the atom type names would be one very simple way to address that. 

In all honesty, though, I think that ultimately trying to support “everything” will require a rethink of the forcefield layout. Rather than one big pool of parameters, it’d be better to arrange it as a tree: nodes inherit (and can override) all parameters from their parents, but their overrides only pass down to their children. Would simplify a lot of issues like this. But that’s a discussion for another time, I suppose.

tristanic:
Thought about this a bit more overnight. Atom parameter redefinitions are in general a (much) harder problem that would take a lot of careful thought, and are in my too-hard basket for now. The pipeline I'll be recommending people use is PHENIX's new wrapper around Antechamber, which shouldn't be overriding any time names (looking at the ffXML generated from the entire Moriarty and Case residue library with 19,728 residues, there are just 6 new atom types defined - for the Fe4S4 iron-sulphur cluster, which just isn't captured by existing GAFF types). What I'm more worried about is multiple redefinitions of torsions, angles etc. - one would hope that these are mostly similar enough to not be a huge problem (particularly for my case, where I specifically tell the users this environment is *not* suitable for equilibrium dynamics), but ideally these should be caught and the relevant atoms defined as new types to rule out all chance of conflict. That would take a *lot* more careful coding than I have the resources for right now.

tristanic:
To reduce the parameter-redefinition problem I've rearranged my own code to store the ligands as a big zip file of individual XML files, and only add them to the `ForceField` as needed. That should make the chances of serious parameter collisions much smaller (and make simulation startups somewhat more snappy!).

I had a conversation with David Case on this general subject a few months back. I believe he was going to start a conversation with the AMBER core team about schemes to manage large force fields with potential conflicts, but I don't know if that's gone anywhere yet.

peastman:
> What I'm more worried about is multiple redefinitions of torsions, angles etc. - one would hope that these are mostly similar enough to not be a huge problem

It will be a problem even if they're identical.  If two different files each define a torsion for the same atoms, it will add two torsions to the force.  That's because it creates a separate PeriodicTorsionGenerator from each file, and then they each are invoked independently to add torsions.

tristanic:
Hmm, ok. That’s likely to cause some interesting behaviour! Will clearly have to spend some quality time rethinking my approach here.

peastman:
Generally speaking, a good approach is if each file only defines force field terms that depend on at least one of the types it added itself.  For example, you have a standard file for GAFF that defines all the standard types and parameters involving those types.  But maybe some other molecule just can't be captured using those types.  So a file needs to define a few more types.  But that file shouldn't repeat all the standard GAFF parameter definitions.  It should only include definitions that involve one of the new types it defined.

jchodera:
@tristanic : I hadn't seen the [Amber library](https://github.com/nwmoriarty/amber_library) before. Do you have any information on where this comes from? Does it cover the entire Ligand Expo using GAFF 1.y or 2 parameters? Which Ligand Expo release date is this current for?

It would be extremely useful if we could bundle a compiled ffxml file for all of these PDB ligands along with OpenMM as well. 

I definitely agree that it would be optimal if we could relax the residue name collision problems to allow multiple ligands with distinct topologies to have different names. The names are only useful in warning messages to the user when things go wrong, anyway, and we could in principle also track filenames (if they are available) to give more useful feedback on errors.

tristanic:
The problem is that these files (coming from a semi-automated Antechamber pipeline) use only atom types already defined in GAFF2, but then define new bonds, angles and torsions for combinations of these atoms that *aren’t* parameterised in the GAFF2 core forcefield. That’ll be generally fine when only using one or a few very different ligands in the one simulation, but will clearly cause problems when multiple similar ligands are used. So it’s an issue of finding the neatest way to keep them independent of each other.

jchodera:
@tristanic These `frcmod` files all come from `parmchk2`, right? Do you know if it will always generate the same guess for the same combination of atom types, or if there might be multiple `frcmod` files that contain different guesses for the *same* combination of atom types?

If they're always the same, we can simply slurp both GAFF2 and all the `frcmod` files into a single `ffxml` with all the parameters in it.

peastman:
> it would be optimal if we could relax the residue name collision problems to allow multiple ligands with distinct topologies to have different names.

The issue here is that @tristanic wants to match residues based on name, because he wants to have different templates for chiral molecules that can be distinguished by topology.  OpenMM doesn't care if two templates have the same name, but he does.

tristanic:
@jchodera It's a collaboration between Nigel Moriarty (PHENIX - Lawrence Berkeley National Lab) and David Case (AMBER). It's a work in progress and still far from perfect (in particular, there was a bug that caused various nitrogens to be erroneously treated as primary amines, the effects of which are still being dealt with), but the goal is indeed to eventually cover all non-covalent ligands in Ligand Expo. 

I haven't actually checked if the parameters generated are always the same for the same atoms, but can do so - not today, though. One issue with the monolithic ffXML is that it does start slowing things down (just *loading* the thing takes about 15 seconds - not such a big deal since the `ForceField` can be cached to a pickle file which takes about 1.5 seconds to load, and there is a definite impact on the simulation startup time). Also doesn't completely solve the problem: there's still the issue of ensuring things work properly when the user wants to load in their own custom ligand generated through the same pipeline (PHENIX releases now incorporate AMBER, and - starting with daily builds in the next few days - the command `phenix.elbow ligand.pdb --amber` will produce a .mol2/.frcmod). Discussing with them about going the next step and optionally producing a ffXML as well.

tristanic:
@jchodera no such luck, I'm afraid. They're *mostly* the same (I haven't found any bonds or angles that differ between residues), but I've found a few proper and improper torsion definitions that vary quite substantially (or, to be more precise, choose from  one of two different guesses). Will see if I can hack together a script to check more rigorously.

```
CHU.xml:    <Improper k1="43.932" periodicity1="2" phase1="3.141592653589793" type1="c" type2="cd" type3="o" type4="os"/>
C80.xml:    <Improper k1="4.6024" periodicity1="2" phase1="3.141592653589793" type1="c" type2="cd" type3="o" type4="os"/>
C17.xml:    <Improper k1="4.6024" periodicity1="2" phase1="3.141592653589793" type1="c" type2="cd" type3="o" type4="os"/>
C18.xml:    <Improper k1="43.932" periodicity1="2" phase1="3.141592653589793" type1="c" type2="cd" type3="o" type4="os"/>
```

```
P06.xml:    <Proper k1="2.11292" periodicity1="2" phase1="3.141592653589793" type1="ca" type2="ca" type3="cc" type4="cd"/>
P16.xml:    <Proper k1="2.9288" periodicity1="2" phase1="3.141592653589793" type1="ca" type2="ca" type3="cc" type4="cd"/>
P2X.xml:    <Proper k1="2.11292" periodicity1="2" phase1="3.141592653589793" type1="ca" type2="ca" type3="cc" type4="cd"/>
P30.xml:    <Proper k1="2.9288" periodicity1="2" phase1="3.141592653589793" type1="ca" type2="ca" type3="cc" type4="cd"/>
```

tristanic:
OK, in the set as it stands right now there are 3 bonds, 143 angles, 558 propers and 71 impropers that have conflicting definitions. I don't see any that have more than two definitions. While most of them are close enough so they wouldn't be problematic in practice, I see some angle definitions with equilibrium angles conflicting by more than 10 degrees and the conflicting impropers have ~10-fold different spring constants. So simply merging everything into a single XML will almost certainly leave some residues behaving badly.

tristanic:
@peastman is this actually true?

> It will be a problem even if they're identical. If two different files each define a torsion for the same atoms, it will add two torsions to the force. That's because it creates a separate PeriodicTorsionGenerator from each file, and then they each are invoked independently to add torsions.

Looking at the code:
```
    @staticmethod
    def parseElement(element, ff):
        existing = [f for f in ff._forces if isinstance(f, PeriodicTorsionGenerator)]
        if len(existing) == 0:
            generator = PeriodicTorsionGenerator(ff)
            ff.registerGenerator(generator)
        else:
            generator = existing[0]
```

I have a single `ForceField` object that I'm just calling `loadFile()` on to add the new ffXML file, so `parseElement()` should always return the same generator. By the looks of things, it means that any repeat definition(s) will be added to the `proper` and `improper` lists, but not actually used (`createForce()` will just use the first definition it finds). Still not ideal, but not as bad as stacking up multiple forces on the same torsion.

peastman:
Perhaps you're right.  Test it to be sure it's really working that way, but maybe it isn't a problem.

tristanic:
The version of OpenMM bundled into ChimeraX has had these changes patched into its `forcefield.py` for the last few months, and it's been working with no apparent issues for a fair range of different ligands. Would it be possible to merge this in?

peastman:
Sure, I think this is fine to merge.

PR #2350, opened 2019-07-21
Don't parse resid

### LINKED ISSUE #2348: non-integer residue ID in CHARMM PSF and CRD files
It seems like that current CharmmPsfFile and CharmmCrdFile are not able to handle non-integer residue ID.
It would be great if it can support non-integer residue ID for the PDB file containing insertion code.
The example CHARMM PSF and CRD files are attached.

[non_int_resid.zip](https://github.com/pandegroup/openmm/files/3411363/non_int_resid.zip)

swails:
Since we don't do anything with it anyway and it is not required to be an integer.

Fixes #2348


peastman:
This doesn't completely fix #2348, but it's one part of the fix.  So I'll go ahead and merge it, then do the rest of the fix in a separate PR.

PR #2351, opened 2019-07-22
WIP: Nosé-Hoover Thermostat

andysim:
A while ago I submitted a [Nosé-Hoover chain integrator](https://openmmtools.readthedocs.io/en/latest/api/generated/openmmtools.integrators.NoseHooverChainVelocityVerletIntegrator.html#openmmtools.integrators.NoseHooverChainVelocityVerletIntegrator) to OpenMMTools.  That version was written as a CustomIntegrator, but our tests show that it can be quite slow compared to the Langevin version.  Moreover, Drude cannot be implemented that way.  This PR (written by Andreas Krämer and I) provides a proper first class implementation of NHC that allows an arbitrary number of thermostats, which can be applied to either absolute or relative velocities.  The performance is much better than the CustomIntegrator version, and Drude is supported.

TODO:
- [x] Reference
- [x] CUDA
- [x] OpenCL


peastman:
Thanks, that's fantastic!  There's quite a lot here to digest, so it will take me a while to get through all of it.  For the moment I just want to clarify the overall design.

You've added a new integrator called `VelocityVerletIntegrator`.  But if I understand correctly, the fact that it uses a velocity Verlet (rather than leapfrog) algorithm is kind of incidental.  The real point is that it's a thermostated integrator.  You add one or more `NoseHooverChain` objects to it, which are used for temperature control.  So perhaps a better name would be `NoseHooverIntegrator` or something like that?  The integrator also has options related to MTS functions, but if I understand correctly that's strictly for the interaction with the heat bath, right?  There's only a single step size for integrating the particle positions.

The integrator has two methods called `addNoseHooverChainThermostat()` and `addMaskedNoseHooverChainThermostat()`.  I'm not entirely clear on the difference between them. What exactly is being masked in the latter?

Then you have a subclass called `DrudeVelocityVerletIntegrator`, which is specifically for systems with Drude particles.  It has a method called `addDrudeNoseHooverChainThermostat()` which builds chains based on the `DrudeForce`.  Though is that method actually necessary?  Since the sole purpose of this class is to work with systems that contain DrudeForces, could you just automatically do this in `initialize()`?

andysim:
Tagging @Olllom in case I misspeak.  Thanks very much for the initial feedback, and please take your time with the review.  I just noticed that the CI failed, so I'll have to dig into those logs and figure out what went wrong.

> You've added a new integrator called VelocityVerletIntegrator. But if I understand correctly, the fact that it uses a velocity Verlet (rather than leapfrog) algorithm is kind of incidental. The real point is that it's a thermostated integrator. You add one or more NoseHooverChain objects to it, which are used for temperature control. So perhaps a better name would be NoseHooverIntegrator or something like that? The integrator also has options related to MTS functions, but if I understand correctly that's strictly for the interaction with the heat bath, right? There's only a single step size for integrating the particle positions.

That's true.  The initial idea (you can see from the commit history that our design changed significantly over time) was to have a simple VV integrator that could have extra stuff like thermostats added.  There's no reason anybody would use this instead of the existing `VerletIntegrator`, so I agree with relabeling it NoseHooverIntegrator (rather than the more verbose `NoseHooverChainVelocityVerletIntegrator` that I used for the `CustomIntegrator` implementation).  Your comments about the MTS stuff are correct - the multi-timestepping capabilities only apply to the NH thermostat particles and the user-provided timestep is used for particle propagation as it is in the `VerletIntegrator`.

> The integrator has two methods called addNoseHooverChainThermostat() and addMaskedNoseHooverChainThermostat(). I'm not entirely clear on the difference between them. What exactly is being masked in the latter?

The first adds a thermostat that applies to all particles' absolute velocities, and will most likely represent the vast majority of use cases.  The masked variant allows the user to add a thermostat that applies to either the velocities of a subset of the atoms, or to the relative motion of some atoms relative to "parent" atoms; the former could be used for goofy things like different temperatures for different subsystems, while the latter is needed for Drude.  I will add some comments to clarify, if the consensus is to stick with this design.

> Then you have a subclass called DrudeVelocityVerletIntegrator, which is specifically for systems with Drude particles. It has a method called addDrudeNoseHooverChainThermostat() which builds chains based on the DrudeForce. Though is that method actually necessary? Since the sole purpose of this class is to work with systems that contain DrudeForces, could you just automatically do this in initialize()?

I can't actually remember why we opted not to do this automagically.  It's certainly possible and, now that you mention it, I think it should be the default behavior.  I'll add that in the next round of commits.

peastman:
Ok, that all makes sense.

You've implemented `NoseHooverChain` as a top level public class, with a public constructor and a ton of methods for querying and setting its properties.  But it looks to me like users are never actually supposed to work with it directly?  They get created internally by `addNoseHooverChainThermostat()` and `addMaskedNoseHooverChainThermostat()`, but never directly by the user.  And there's no `getChains()` method, so the user can't access the ones that have been created internally.  I suspect this was a case where you started out with one design, then switched to a different one part way through?

Olllom:
@andysim @peastman 
> So perhaps a better name would be NoseHooverIntegrator or something like that?

I like that, too.

> > Since the sole purpose of this class is to work with systems that contain DrudeForces, could you just automatically do this in initialize()?

> I can't actually remember why we opted not to do this automagically. It's certainly possible and, now that you mention it, I think it should be the default behavior.

The idea was to keep the API relatively simple, while not taking away the more general functionality, such as thermostating protein and water with different temperatures. 

Instead of hiding all this from the public API, we could have a flag `autosetup` in both the normal and the Drude integrator, which defaults to `true` and triggers building the appropriate Nose-Hoover chains with a default configuration (`numMTS=3, numYS=3, chainLength=3`). 
1. This would spare users thinking about these settings, which are not important in most cases. I do not like the idea that all these things appear as parameters of the constructor. Especially for Drude, this would become really messy.
2. If somebody wanted to exploit the flexibility of the implementation, he could set `autosetup` to false and add the Nose-Hoover chains manually.

> You've implemented `NoseHooverChain` as a top level public class, with a public constructor and a ton of methods for querying and setting its properties. But it looks to me like users are never actually supposed to work with it directly?  They get created internally by `addNoseHooverChainThermostat()` and `addMaskedNoseHooverChainThermostat()`, but never directly by the user. And there's no `getChains()` method, so the user can't access the ones that have been created internally. I suspect this was a case where you started out with one design, then switched to a different one part way through?

Partly yes. One goal from the start was to separate the implementation of the thermostat from the integrator, allowing for multiple chains and facilitating adding Nose-Hoover to other integrators. If I  remember correctly, the API of the `NoseHooverChain` class is currently only accessed from the integrator and not meant to be called by the user.

peastman:
> (numMTS=3, numYS=3, chainLength=3).

You could just make those be default values for the arguments in `addNoseHooverChainThermostat()`.

Is the following a correct summary?  The integrator is intended to be used in one of two ways.  You will either

1. call `addNoseHooverChainThermostat()` exactly once to thermostat the whole system (it would never make sense to call it multiple times), or

2. call `addMaskedNoseHooverChainThermostat()` one or more times to apply thermostats to distinct, nonoverlapping parts of the system.

If so, I feel like there ought to be a cleaner API for this, though I'm not certain what it is.

openmmapi/src/VelocityVerletIntegrator.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    // make sure that thermostats do not overlap 
+    for (auto &other_nhc: noseHooverChains) {
+        for (auto &particle: mask){
+            bool isParticleInOtherChain = (std::find(other_nhc.getThermostatedAtoms().begin(), 
+                                                    other_nhc.getThermostatedAtoms().end(),
+                                                    particle) == other_nhc.getThermostatedAtoms().begin());
+            if (isParticleInOtherChain){
+                throw OpenMMException("Found particle " + std::to_string(particle) + "in a different NoseHooverChain, "
```
Olllom: @peastman this will fire, if `addNoseHooverThermostat()` is called twice

Olllom:
> Is the following a correct summary? The integrator is intended to be used in one of two ways. 

Exactly.

> If so, I feel like there ought to be a cleaner API for this, though I'm not certain what it is.

Any suggestions are welcome. Note that there are quite a few sanity checks. For example, if `addNoseHooverChainThermostat()` is called twice, the exception https://github.com/pandegroup/openmm/pull/2351/files#r308443398 will fire. Maybe this is enough to ensure that there is no misuse of either of the methods.

peastman:
What about this idea?  Get rid of `addNoseHooverChainThermostat()`.  Instead have two different constructors.  For the common case where you just want to apply a single thermostat to every particle, there will be one with this signature:

```c++
NoseHooverIntegrator(double temperature, double collisionFrequency, double stepSize,
                     int chainLength=3, int numMTS=3, int numYoshidaSuzuki=3)
```

For the less common case where you want multiple thermostats, you'll use the existing constructor then make multiple calls to `addMaskedNoseHooverChainThermostat()` (but we can probably give it a simpler name, like maybe just `addThermostat()`).

The Drude integrator probably needs only a single constructor, which takes two temperatures and automatically builds the chains based on the DrudeForce.  If you want something more customized than that, you can just use the standard NoseHooverIntegrator.

peastman:
I'm not sure the Drude integrator works correctly.  Or quite possibly I'm just not understanding the code.  But I want to make sure.  Inside `addDrudeNoseHooverChainThermostat()` you build three lists of particles: `drudeParticles` for the Drude particles, `drudeParents` for their parents, and `realParticles` for everything that isn't a Drude.  You then add two thermostats:

```c++
addMaskedNoseHooverChainThermostat(system, realParticles, vector<int>(), temperature, collisionFrequency,
                                   chainLength, numMTS, numYoshidaSuzuki);
addMaskedNoseHooverChainThermostat(system, drudeParticles, drudeParents, drudeTemperature, drudeCollisionFrequency,
                                   chainLength, numMTS, numYoshidaSuzuki);
```

Does that actually produce correct results?  For each Drude/parent pair, the Langevin Drude integrator couples their center of mass to one thermostat and the displacement between them to a second thermostat.  Those are independent degrees of freedom, so the two thermostats don't directly affect each other.  But this is doing something different.  It couples the parent particle to the system thermostat, rather than the center of mass.  And that isn't independent from the displacement.  If you move the parent particle without also moving the Drude, that changes the displacement.  So the two thermostats are both acting on the same degrees of freedom at the same time.

Olllom:
> What about this idea? Get rid of `addNoseHooverChainThermostat()`. Instead have two different constructors. For the common case where you just want to apply a single thermostat to every particle, there will be one with this signature:
> 
> ```c++
> NoseHooverIntegrator(double temperature, double collisionFrequency, double stepSize,
>                      int chainLength=3, int numMTS=3, int numYoshidaSuzuki=3)
> ```
> 
> For the less common case where you want multiple thermostats, you'll use the existing constructor then make multiple calls to `addMaskedNoseHooverChainThermostat()` (but we can probably give it a simpler name, like maybe just `addThermostat()`).
> 
> The Drude integrator probably needs only a single constructor, which takes two temperatures and automatically builds the chains based on the DrudeForce. If you want something more customized than that, you can just use the standard NoseHooverIntegrator.

I like this a lot.

andysim:
> Does that actually produce correct results? For each Drude/parent pair, the Langevin Drude integrator couples their center of mass to one thermostat and the displacement between them to a second thermostat. Those are independent degrees of freedom, so the two thermostats don't directly affect each other. But this is doing something different. It couples the parent particle to the system thermostat, rather than the center of mass. And that isn't independent from the displacement. If you move the parent particle without also moving the Drude, that changes the displacement. So the two thermostats are both acting on the same degrees of freedom at the same time.

Excellent catch!  That's a mistake on our part; our implementation is correct in the limit of a zero mass Drude particle, which is not the reality.  The ensembles we generated in testing so far have looked great, so the error is small, but the nonorthogonality of the two degrees of freedom in the current code does impose a slight order dependence on which thermostat fires first.  I'll fix the code to use the center of mass and address your other helpful feedback.

craabreu:
This is very nice, @andysim and @Olllom! Thanks for sharing it. If I may make a couple of suggestions:
1. It would be interesting to have the thermostat chain optionally integrated in the middle of each step, such as described [here](https://aip.scitation.org/doi/10.1063/1.4991621). This resembles the [BAOAB Langevin integrator](https://aip.scitation.org/doi/10.1063/1.4802990) and the gain in accuracy is worthwhile.
2. With small additional effort (I believe), it would be possible to include other global thermostats such as [Nosé-Hoover-Langevin](https://link.springer.com/article/10.1007%2Fs10955-009-9734-0) and [Canonical Velocity Rescaling](https://aip.scitation.org/doi/10.1063/1.2408420).

openmmapi/include/OpenMM.h
```diff
@@ -61,6 +61,7 @@
 #include "openmm/MonteCarloBarostat.h"
 #include "openmm/MonteCarloMembraneBarostat.h"
 #include "openmm/NonbondedForce.h"
+#include "openmm/NoseHooverChain.h"
```
peastman: It looks to me like NoseHooverChain is an internal class?  It doesn't appear anywhere in NoseHooverIntegrator's public API.  If so, let's move the header into `include/openmm/internal` and not include it here.

openmmapi/include/openmm/NoseHooverIntegrator.h
```diff
@@ -0,0 +1,245 @@
...
+     *        particles should have been added, before calling this function.
+     * @param temperature the target temperature for the system.
+     * @param collisionFrequency the frequency of the interaction with the heat bath (in 1/ps).
+     * @param chainLength the number of beads in the Nose-Hoover chain.
+     * @param numMTS the number of step in the  multiple time step chain propagation algorithm.
+     * @param numYoshidaSuzuki the number of terms in the Yoshida-Suzuki multi time step decomposition
+     *        used in the chain propagation algorithm (must be 1, 3, or 5).
+     */
+    explicit NoseHooverIntegrator(double stepSize, System &system, double temperature, int collisionFrequnency,
+                                  int chainLength = 3, int numMTS = 3, int numYoshidaSuzuki = 3);
```
peastman: Let's make the constructor arguments consistent with LangevinIntegrator:

```c++
NoseHooverIntegrator(double temperature, double collisionFrequnency, double stepSize, int chainLength=3, int numMTS=3, int numYoshidaSuzuki=3)
```

Having similar arguments but in a different order will lead to lots of confusion.

Requiring the System as an argument breaks some design assumptions and will lead to lots of problems.  Users are supposed to be able to create the System and Integrator independently of each other.  Just store whatever information gets passed to the constructor, and then you'll get a reference to the System when `initialize()` is called.

openmmapi/include/openmm/NoseHooverIntegrator.h
```diff
@@ -0,0 +1,245 @@
...
+     * @param system the system to be thermostated.  Note: this must be setup, i.e. all
+     *        particles should have been added, before calling this function.
+     * @param temperature the target temperature for the system.
+     * @param collisionFrequency the frequency of the interaction with the heat bath (in 1/ps).
+     * @param chainLength the number of beads in the Nose-Hoover chain.
+     * @param numMTS the number of step in the  multiple time step chain propagation algorithm.
+     * @param numYoshidaSuzuki the number of terms in the Yoshida-Suzuki multi time step decomposition
+     *        used in the chain propagation algorithm (must be 1, 3, or 5).
+     */
+     int addThermostat(System& system, double temperature, double collisionFrequency,
```
peastman: Here too don't expect a System.  It may not exist yet.

openmmapi/src/NoseHooverIntegrator.cpp
```diff
@@ -0,0 +1,289 @@
...
+    // create and add new chain
+    int chainID = noseHooverChains.size();
+    auto nhc = NoseHooverChain(temperature, relativeTemperature, collisionFrequency, relativeCollisionFrequency,
+                               nDOF, chainLength, numMTS, numYoshidaSuzuki, chainID, thermostatedParticles, thermostatedPairs);
+    noseHooverChains.push_back(nhc);
+    return chainID;
+}
+
+double NoseHooverIntegrator::getTemperature(int chainID) const {
+    if (chainID >= noseHooverChains.size()) {
```
peastman: All these range checks can be simplified to just

```C++
ASSERT_VALID_INDEX(chainId, noseHooverChains);
```

It also checks for negative indices.

openmmapi/src/NoseHooverIntegrator.cpp
```diff
@@ -0,0 +1,289 @@
...
+        throw OpenMMException("Cannot set relative collision frequency for chainID " + std::to_string(chainID)
+                + ". Only " + std::to_string(noseHooverChains.size()) + " have been registered with this integrator."
+        );
+    }
+    noseHooverChains.at(chainID).setDefaultRelativeCollisionFrequency(frequency);
+}
+
+double NoseHooverIntegrator::computeKineticEnergy() {
+    double kE = 0.0;
+    if(noseHooverChains.size()) {
```
peastman: Let's make this `if (noseHooverChains.size() > 0)`.  Maybe just a personal preference, but I find it clearer.

peastman:
The old class name `VelocityVerletIntegrator` still appears in a lot of places, like in comments or the names of files such as TestCudaVelocityVerletIntegrator.cpp.  You can just do a search through the diff to find all of them.

peastman:
This PR seems to have gone quiet.  Are you still working on it?  Anything I can do to help it along?

andysim:
Sorry for the delay and the long silence - I got consumed by other projects.  Your comments have been very helpful and we have incorporated the suggested changes.  We're working on the OpenCL port now, and should have it done by tomorrow.  Thanks for your patience!

peastman:
That's great news, thanks!

Olllom:
@peastman We completed the implementation of the Nose Hoover integrator. The PR is ready for another round of review. The failing test should just be due to a tolerance being overly tight. It will get fixed soon. 

The new version implements your requested changes, as well as the corresponding CUDA, OpenCL, and Python code. Also, we had previously missed that the Drude pairs require special handling by the integrator. Therefore a list of thermostated pairs and atoms are now passed to the integration kernels. The Drude plugin now contains a `DrudeNoseHooverIntegrator` - merely a wrapper around the `NoseHooverIntegrator` that constructs the correct thermostats. The new version also contains a hard wall constraint (with test), some more tests, and serialization for both integrator classes.

peastman:
Fantastic!  I'll take a look shortly.

openmmapi/include/openmm/NoseHooverIntegrator.h
```diff
@@ -0,0 +1,279 @@
...
+    double getMaximumPairDistance() const { return maxPairDistance_; }
+    /**
+     * Sets the maximum distance (in nm) that a connected pair may stray from each other, implemented using a hard
+     * wall. If set to zero, the hard wall constraint is omited and the pairs are free to be separated by any distance.
+     */
+    void setMaximumPairDistance(double distance) { maxPairDistance_ = distance; }
+    /**
+     * Get a list of all individual atoms (i.e. not involved in a connected Drude-like pair) in the system.
+     */
+    const std::vector<int> & getAllAtoms() const { return allAtoms; }
```
peastman: I'm a little confused about this method, both what it's for and the intended behavior.  If it omits any particles that are part of a pair, then it isn't really "all atoms", is it?  Also, if the thermostat is only applied to a subset of particles, should it still return all particles, or only the thermostated ones?  I'm just not clear what this is for.

Public parts of the C++ API should use the word "particles" rather than "atoms".

openmmapi/include/openmm/NoseHooverIntegrator.h
```diff
@@ -0,0 +1,279 @@
...
+    void setRelativeCollisionFrequency(double frequency, int chainID=0);
+    /**
+     * Compute the total (potential + kinetic) heat bath energy for all heat baths
+     * associated with this integrator, at the current time.
+     */
+    double computeHeatBathEnergy();
+    /**
+     * Get the number of Nose-Hoover chains registered with this integrator.
+     */
+    int getNumNoseHooverThermostats() const {
```
peastman: The method names are slightly inconsistent: `addThermostat()` but `getNumNoseHooverThermostats()`.  How about making all of them consistent: `addThermostat()`, `getThermostat()`, `getNumThermostats()`.

peastman:
I'm getting lost trying to trace through all the code to see how massless particles are handled.  Probably it's all correct, but I just want to make sure.

- If you add a thermostat that explicitly lists particles and one of them has mass 0, it throws an exception.
- If you use a default thermostat that acts on the whole system, and there happen to be massless particles in the system, it does not throw an exception.  Instead it makes those particles immobile and omits them when computing the number of degrees of freedom for the thermostat.

Is that what it's currently doing?

openmmapi/src/NoseHooverIntegrator.cpp
```diff
@@ -0,0 +1,354 @@
...
+                                          std::to_string(pair.second) + " in a different NoseHooverChain, "
+                                          "but particles can only be thermostated by one thermostat.");
+                }
+            }
+        }
+
+        // make sure that massless particles are not thermostated
+        for(auto particle: nhc.getThermostatedAtoms()){
+            double mass = system.getParticleMass(particle);
+            if (mass < 1e-8) {
```
peastman: Should this be `if (mass == 0)`?  A particle with tiny mass still has mass.  I'm not sure why someone would create a mass that small, but still...  Setting the mass to exactly 0 means to make that particle immobile.

peastman:
NoseHooverChain has a bunch of methods with "default" in their names, like `getDefaultTemperature()` and `setDefaultCollisionFrequency()`.  In what sense are these defaults?  Usually we use the word "default" to specify the initial value for something that gets stored in the Context.  For example, System has `setDefaultPeriodicBoxVectors()`, and that sets the initial box vectors for new Contexts, but then you can change them by calling `setPeriodicBoxVectors()` on the Context.  But I don't think the values in NoseHooverChain work like that, do they?

openmmapi/src/NoseHooverIntegrator.cpp
```diff
@@ -0,0 +1,354 @@
...
+            kineticEnergy = nhcKernel.getAs<NoseHooverChainKernel>().computeMaskedKineticEnergy(*context, nhc, false);
+            scale = nhcKernel.getAs<NoseHooverChainKernel>().propagateChain(*context, nhc, kineticEnergy, getStepSize());
+            nhcKernel.getAs<NoseHooverChainKernel>().scaleVelocities(*context, nhc, scale);
+        }
+        vvKernel.getAs<IntegrateVelocityVerletStepKernel>().execute(*context, *this, forcesAreValid);
+        for(auto &nhc : noseHooverChains) {
+            kineticEnergy = nhcKernel.getAs<NoseHooverChainKernel>().computeMaskedKineticEnergy(*context, nhc, false);
+            scale = nhcKernel.getAs<NoseHooverChainKernel>().propagateChain(*context, nhc, kineticEnergy, getStepSize());
+            nhcKernel.getAs<NoseHooverChainKernel>().scaleVelocities(*context, nhc, scale);
+        }
```
peastman: I see it's still using the OBABO discretization rather than the BAOAB one.  Is that because you tried it and it didn't work, or have you not tried it yet?  At least with Langevin, BAOAB can allow significantly larger time steps.  I don't know whether the theory carries over to Nose-Hoover through.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -7075,6 +7076,110 @@ double CudaIntegrateVerletStepKernel::computeKineticEnergy(ContextImpl& context,
...
+    int numParticles = numAtoms + 2*numPairs;
+    float maxPairDistance = integrator.getMaximumPairDistance();
+    // Make sure atom and pair metadata is uploaded and has the correct dimensions
+    if (prevMaxPairDistance != maxPairDistance) {
+        std::vector<float> tmp(1, maxPairDistance);
+        maxPairDistanceBuffer.upload(tmp);
+        prevMaxPairDistance = maxPairDistance;
+    }
+    if (numAtoms !=0 && (!atomListBuffer.isInitialized() || atomListBuffer.getSize() != numAtoms)) {
+        atomListBuffer.initialize<int>(cu, atomList.size(), "atomListBuffer");
```
peastman: If `atomListBuffer` is already initialized but `numAtoms` has changed, this line will throw an exception.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -7075,6 +7076,110 @@ double CudaIntegrateVerletStepKernel::computeKineticEnergy(ContextImpl& context,
...
+        std::vector<int2> tmp;
+        std::vector<float> tmp2;
+        for(const auto &pair : pairList) {
+            tmp.push_back(make_int2(std::get<0>(pair), std::get<1>(pair)));
+            tmp2.push_back(std::get<2>(pair));
+        }
+        pairListBuffer.initialize<int2>(cu, pairList.size(), "pairListBuffer");
+        pairListBuffer.upload(tmp);
+        pairTemperatureBuffer.initialize<float>(cu, pairList.size(), "pairTemperatureBuffer");
+        pairTemperatureBuffer.upload(tmp2);
```
peastman: And similarly here.  You can call `resize()` to resize a CudaArray that has already been initialized.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -8287,6 +8392,314 @@ void CudaApplyAndersenThermostatKernel::execute(ContextImpl& context) {
...
+    double relativeTemperature = nhc.getDefaultRelativeTemperature();
+    double relativeFrequency = nhc.getDefaultRelativeCollisionFrequency();
+
+    if (numYS != 1 && numYS != 3 && numYS != 5) {
+        throw OpenMMException("Number of Yoshida Suzuki time steps has to be 1, 3, or 5.");
+    }
+
+    auto & chainState = cu.getIntegrationUtilities().getNoseHooverChainState();
+
+    if (!scaleFactorBuffer.isInitialized() ||scaleFactorBuffer.getSize() == 0) {
```
peastman: A couple more places that need to call `resize()` instead of `initialize()` if it's already initialized.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -8287,6 +8392,314 @@ void CudaApplyAndersenThermostatKernel::execute(ContextImpl& context) {
...
+    bool absChainIsValid = chainState.count(2*chainID) != 0 &&
+                           chainState[2*chainID].isInitialized() &&
+                           chainState[2*chainID].getSize() == chainLength;
+    bool relChainIsValid = chainState.count(2*chainID+1) != 0 &&
+                           chainState[2*chainID+1].isInitialized() &&
+                           chainState[2*chainID+1].getSize() == chainLength;
+
+    if (!absChainIsValid && !relChainIsValid) return 0.0;
+
+    if (!heatBathEnergy.isInitialized() || heatBathEnergy.getSize() == 0) {
```
peastman: Same as above.

openmmapi/include/openmm/NoseHooverIntegrator.h
Olllom: You are right, this name is confusing. We can change it to `getThermostatedIndividualParticles` and `getThermostatedPairs` or something like that.

openmmapi/include/openmm/NoseHooverIntegrator.h
Olllom: OK

peastman:
> Exactly. Do you agree that this is OK?

Perfect.  I figured that was probably what it was doing, but the code was complicated enough I thought I should ask to make sure.

> I think we should also get rid of some of the setters, at least the yoshida-suzuki and chain length, which are not effective after creating a context anyway.

That makes sense.

Time to review a bit more.  Did you know this PR has 6712 added lines of code in 67 files?  I'm about a third of the way through it so far!

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -8287,6 +8392,314 @@ void CudaApplyAndersenThermostatKernel::execute(ContextImpl& context) {
...
+            pairlists[chainID].initialize<int2>(cu, nPairs, "pairlist" + std::to_string(chainID));
+            std::vector<int2> int2vec;
+            for(const auto &p : nhcPairs) int2vec.push_back(make_int2(p.first, p.second));
+            pairlists[chainID].upload(int2vec);
+        }
+        if (pairlists[chainID].getSize() != nPairs) {
+            throw OpenMMException("Number of thermostated pairs changed. Cannot be handled by the same Nose-Hoover thermostat.");
+        }
+    }
+    if (!kineticEnergyBuffer.isInitialized() || kineticEnergyBuffer.getSize() == 0) {
```
peastman: Like above.

platforms/cuda/src/kernels/noseHooverChain.cu
```diff
@@ -0,0 +1,170 @@
...
+    for (int index = blockIdx.x*blockDim.x+threadIdx.x; index < numAtoms; index += blockDim.x*gridDim.x) {
+        int atom = atoms[index];
+        mixed4 v = velm[atom];
+        mixed mass = v.w == 0 ? 0 : 1 / v.w;
+        energy.x += 0.5f * mass * (v.x*v.x + v.y*v.y + v.z*v.z);
+    }
+    energyBuffer[blockIdx.x*blockDim.x+threadIdx.x] = energy;
+}
+
+extern "C" __global__ void computePairsKineticEnergy(mixed2 * __restrict__ energyBuffer, int numPairs,
```
peastman: I'm not sure it's safe to treat the standard energy buffer as mixed2.  It's only guaranteed to be long enough to hold one element per thread, not two.  See https://github.com/openmm/openmm/blob/master/platforms/cuda/src/CudaContext.cpp#L444.  It you need something bigger, you can create your own buffer for accumulating kinetic energy.

platforms/cuda/src/kernels/noseHooverChain.cu
```diff
@@ -0,0 +1,170 @@
...
+
+extern "C" __global__ void scalePairsVelocities(mixed2 * __restrict__ scaleFactor, int numPairs,
+                                                mixed4* __restrict__ velm, const int2 *__restrict__ pairs){
+    const mixed &absScale = scaleFactor[0].x;
+    const mixed &relScale = scaleFactor[0].y;
+    for (int index = blockIdx.x*blockDim.x+threadIdx.x; index < numPairs; index += blockDim.x*gridDim.x) {
+        int atom1 = pairs[index].x;
+        int atom2 = pairs[index].y;
+        mixed4 &v1 = velm[atom1];
+        mixed4 &v2 = velm[atom2];
```
peastman: This might lead to inefficient memory access.  It's better to create a local variable

```
mixed4 v1 = velm[atom1];
```

And then store it back to global memory all at once

```
velm[atom1] = v1;
```

The compiler *might* be smart enough to see what you're doing and consolidate everything into fewer memory transactions, but I wouldn't count on it.

platforms/cuda/src/kernels/velocityVerlet.cu
```diff
@@ -0,0 +1,349 @@
...
+ * Perform the third step of Velocity Verlet integration.
+ *
+ * modify the velocities (velm) after the force update
+ */
+
+extern "C" __global__ void integrateVelocityVerletPart3(int numAtoms, int numPairs, int paddedNumAtoms, mixed2* __restrict__ dt, real4* __restrict__ posq,
+        real4* __restrict__ posqCorrection, mixed4* __restrict__ velm,  const long long* __restrict__ force, const mixed4* __restrict__ posDelta,
+        const int* __restrict__ atomList, const int2* __restrict__ pairList) {
+    mixed2 stepSize = dt[0];
+#if __CUDA_ARCH__ >= 130
```
peastman: Copied from the VerletIntegrator code, undoubtedly. :)  I should probably clean that up at some point.  I think it's OK if we finally drop support for G80 GPUs!  Don't worry about this, what you have is fine.  I'm just amused to see ancient workarounds like this pop up again.

peastman: By the way, the correspond code in the OpenCL platform *is* still relevant, because low end embedded GPUs sometimes still don't have double precision.  But I think it's been about 8 or 9 years since NVIDIA made any that didn't.

peastman:
Note that most of my comments about CUDA also apply to OpenCL, so I won't repeat them.

Olllom:
> Did you know this PR has 6712 added lines of code in 67 files? I'm about a third of the way through it so far!

Painfully aware

> By the way, the corresponding code in the OpenCL platform is still relevant, because low end embedded GPUs sometimes still don't have double precision. 

In that case, I consider it incredibly helpful to have the analog in the CUDA code so that the OpenCL part can be a more or less mindless copy.

peastman:
>In that case, I consider it incredibly helpful to have the analog in the CUDA code so that the OpenCL part can be a more or less mindless copy.

Fine with me!

platforms/reference/src/SimTKReference/ReferenceVelocityVerletDynamics.cpp
```diff
@@ -0,0 +1,274 @@
...
+              inverseMasses[ii] = 1.0/masses[ii];
+       }
+    }
+
+    //// Perform the integration.
+
+    // Regular atoms
+    for (const auto &atom : atomList) {
+        if (masses[atom] != 0.0) {
+            for (int xyz = 0; xyz < 3; ++xyz) {
```
peastman: You don't actually need all these xyz loops.  Vec3 supports standard math operations.  For example, you could just write

```c++

velocities[atom] += 0.5*inverseMasses[atom]*forces[atom]*getDeltaT();
xPrime[atom] = atomCoordinates[atom];
atomCoordinates[atom] += velocities[atom]*getDeltaT();
```

plugins/drude/tests/TestDrudeNoseHoover.h
```diff
@@ -0,0 +1,605 @@
...
+                                    temperatureDrude, frequencyDrude, 0.0005, 
+                                    chainLength, numMTS, numYS);
+    Context context(system, integ, platform);
+    context.setPositions(positions);
+    context.setVelocitiesToTemperature(temperature, randomSeed);
+    std::vector<Vec3> velocities = context.getState(State::Velocities).getVelocities();
+    for (int i = 0; i < numMolecules; i++){
+        Vec3 noize;
+        for (int j = 0; j < 3; j++){
+            noize[j] = float(((i+18311)*(j+18253) * 313419097822414) % 18313) / float(18313);
```
peastman: I take it this is a random number generator? 😀  We do have a proper RNG available for use in tests:

```c++
#include "sfmt/SFMT.h"
...
// Create and initialize an RNG instance
OpenMM_SFMT::SFMT sfmt;
init_gen_rand(0, sfmt);
...
noise[j] = genrand_real2(sfmt); // Uniformly distributed between 0 and 1
```

tests/TestNoseHooverThermostat.h
```diff
@@ -0,0 +1,293 @@
...
+        energies[i] = energy;
+    }
+    double sum = std::accumulate(energies.begin(), energies.end(), 0.0);
+    double mean = sum / energies.size();
+    double sq_sum = std::inner_product(energies.begin(), energies.end(), energies.begin(), 0.0);
+    double std = std::sqrt(sq_sum / energies.size() - mean * mean);
+    double relative_std = std / mean;
+    //std::cout << "Relative STD of energy " << relative_std << std::endl;
+    // Check mean temperature
+    ASSERT_EQUAL_TOL(temperature, temperature, 1e-2);
```
peastman: I think one of those should be `mean_temp`?

tests/TestNoseHooverThermostat.h
```diff
@@ -0,0 +1,293 @@
...
+    context.createCheckpoint(checkpoint);
+    
+    State state = context.getState(State::Positions | State::Velocities);
+    for (size_t i=0; i<100; i++){
+        state = context.getState(State::Positions | State::Velocities);
+#if DEBUG
+        std::cout << "posvel" << state.getPositions()[0] << " " << state.getVelocities()[0] << std::endl;
+#endif
+        integrator.step(1);
+    }
```
peastman: I don't quite understand this loop.  Could it all be simplified to

```c++
integrator.step(100)
State state = context.getState(State::Positions | State::Velocities);
```

tests/TestNoseHooverThermostat.h
```diff
@@ -0,0 +1,293 @@
...
+    State state2 = context.getState(State::Positions | State::Velocities);
+    for (size_t i=0; i<100; i++){
+        state2 = newContext.getState(State::Positions | State::Velocities);
+#if DEBUG
+        std::cout << "posvel" << state2.getPositions()[0] << " " << state2.getVelocities()[0] << std::endl;
+#endif
+        newIntegrator.step(1);
+    }
+
+    for (int i=0; i<3;i++){
```
peastman: Why the loop?  You just repeat the exact same checks three times.

peastman:
THERE!  Reviewing it was certainly much less work than it was for you to write it, but I still count it as an accomplishment!

I just want to say that given the amount of new code, I found remarkably few things to comment on.  This really is very well written.  It's going to be a great feature.  Thanks so much!

plugins/drude/tests/TestDrudeNoseHoover.h
Olllom: :-) We were not sure if the built-in RNG is 100% platform-independent. But if you don't mind we will leave it as is - this would spare us to rewrite the check of positions and velocities below :-)

tests/TestNoseHooverThermostat.h
Olllom: ooops

tests/TestNoseHooverThermostat.h
Olllom: That's right. This is just us debugging. Will be removed.

tests/TestNoseHooverThermostat.h
Olllom: Uhm... yeah. Looks like this was a `xyz` loop before we were aware of `ASSERT_EQUAL_VEC`

Olllom:
@andysim if you find the time, push my changes to see if the looser tolerance fixes the failing CI test (ctest should have run through on m6 by the time you read this)

peastman:
By the way, since you've just gone through the work of implementing a major new feature, I'd be very interested to hear your thoughts on #2471.  How much difference would it have made to you?

peastman:
Do you have an estimate of when this will be finished?  There are going to be merge conflicts between this and #2488.  I don't want to inflict that on you, so I'd rather not merge it until this one is finished.  But whether that's practical depends on how long it will be.

andysim:
Sorry for the slow progress on this end.  The code is basically good to go in its current form (apart from some docs, if you'd like me to add some blurb to the manual).  As a followup, there are a number of things that I would like to try but, in the interests of avoiding merge conflicts and scope creep,  I think a followup PR is the best way to go.  First, I want to convert to the awesome new common interface (I suspect you do also, given the line of inquiry).  From there, I'd like to follow up on the BOAOB approach, [which will need a little experimentation when used with constraints](http://dx.doi.org/10.1021/acs.jpca.9b02771).  I recently found out that there's a [better way to compute temperature when pushing the timestep](doi.org/10.1021/acs.jctc.8b00874), as well as a [better way to handle internal degrees of freedom](http://dx.doi.org/10.1021/acs.jpclett.9b02983) when small molecules are being simulated.  I think these are all worth trying out but in a later PR.  Do you agree, or would you prefer to wait on this one?

andysim:
It looks like the 6th Travis test failed to set up the env correctly - would you mind triggering a restart please?  The Appveyor setup seems to be happy with the latest fix.

peastman:
I just restarted the build that had problems.  Yes, a followup PR for other changes would be fine.  There are two places in the user guide where it should be described: section 3.6.7 has brief descriptions of all the integrators, and then chapter 20 should have a longer description of the algorithm.  All that can wait for the followup PR.

peastman:
Tests are passing, so I'll go ahead and merge.  Thanks for the fantastic new feature!

peastman:
I'm working on merging this with #2488, but I'm running into an error.  It comes from these lines:

https://github.com/openmm/openmm/blob/f6431a42ac7597f33d017a585854bea0159f3b85/platforms/opencl/src/OpenCLKernels.cpp#L9032-L9033

Line 9032 sets `chainState[2*chainID]` to a new, uninitialized OpenCLArray.  Then the very next line tries to query the size of that uninitialized array, which produces an exception.  What was the intended behavior?

andysim:
Sorry about that - my fault.  My intention was to first query the map for a valid entry with the given integer key; if not, call the default constructor to create it just to make sure that the size can possibly be queried.  From there, I was thinking that the default-constructed array would identify itself as having zero size so that the subsequent check would trigger an initialization, which, in hindsight, is kinda silly.  I'm happy to send a patch to your branch if that's the easiest way forward - please let me know how you'd like to proceed.  Kudos for the awesome work on #2488 - it looks like a great improvement and will really ease development in the future

peastman:
It would be easiest if you can just describe the changes here.  I've already done the merge, and I'm just trying to get it working before I commit it.

andysim:
Because you haven't overloaded the `||` operator in the Array class, it will be safe to change 

```
if (chainState.at(2*chainID).getSize() != chainLength) {
```
to
```
if (!chainState.at(2*chainID).isInitialized() || chainState.at(2*chainID).getSize() != chainLength) {
```
because the compiler will shortcut the second condition.  The checks within that block look correct to me - the decision whether to call `resize` or `initialize<>` looks to be correct after your comments during the PR review.  Please let me know if I can clarify any thing else and sorry again for the mess.

peastman:
Thanks, very simple!  Now I just need to figure out why TestDrudeNoseHoover is failing...

andysim:
I wasn't able to locate the offending travis log on the PR or your branch but I'll happily take a look if you can point me in the right direction.  There is a test that checks for regular particles and drude particles being within a certain temperature range after a very short run for SWM4 water; this is stochastic but should only fail infrequently (something like ~1 in 10 from memory).  If that's the offending test, it's possible that we chose to strict a tolerance for checking the temperatures.

peastman:
I haven't checked it in yet, but if I can't figure it out soon I'll go ahead and check it in, and maybe the problem will be more obvious to you.  The failing test is the one that simulates a water box and checks that total energy is conserved, including the energy of the chains.  In the master branch it always passes.  In my merged version it passes on reference, but for CUDA and OpenCL the energy drifts.

peastman:
I merged the changes into #2488.  I'll continue trying to debug that test myself, but if you're inclined to look at it, that would be great.  It's probably just some very minor change.

andysim:
I've been poking around and eventually found that the drift is coming from the `CMMotionRemover`; simply omitting that force from the waterbox test leads to everything passing on all platforms (as you've already noted, the Reference test passes with it present).  I will try to investigate a little more tomorrow and hopefully will be able to find something concrete to address the problem.

PR #2354, opened 2019-07-31
Added CONTRIBUTING and SUPPORT files

peastman:
This is a first draft.  Comments welcome!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+## How to Contribute to OpenMM Development
+
+We welcome anyone who wants to contribute to the project, whether by adding a feature,
+fixing a bug, or improving documentation.  The process is quite simple.
+
+First, it is always best to begin by opening an issue on Github that describes the change you
+want to make.  This gives everyone a chance to discuss it before you put in a lot of work.
+For bug fixes, we will confirm that the behavior is actually a bug and that the proposed fix
+is correct.  For new features, we will decide whether the proposed feature is something we
+want and discuss possible designs for it.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
+## How to Contribute to OpenMM Development
+
+We welcome anyone who wants to contribute to the project, whether by adding a feature,
+fixing a bug, or improving documentation.  The process is quite simple.
+
```
jchodera: Would it make sense to add some structure within this document, like
```
### General process 
...
### Submitting a feature
...
### Fixing a bug
...
### Improving documentation
...
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+First, it is always best to begin by opening an issue on Github that describes the change you
+want to make.  This gives everyone a chance to discuss it before you put in a lot of work.
+For bug fixes, we will confirm that the behavior is actually a bug and that the proposed fix
+is correct.  For new features, we will decide whether the proposed feature is something we
+want and discuss possible designs for it.
+
+Once everyone is in agreement, the next step is to create a pull request with the code changes.
```
jchodera: You can link to https://help.github.com/en/articles/about-pull-requests

peastman: Good idea, I'll add that.

CONTRIBUTING.md
peastman: I'm not sure what we'd say in the different cases.  The process is pretty much the same regardless: open an issue to discuss it, then submit a PR with the changes.  And since the whole document is only a few paragraphs long, it doesn't really need a lot of organization.

PR #2358, opened 2019-08-01
Support for multiple paths in OPENMM_PLUGIN_DIR

tallakahath:
Sometimes you can't have all your plugins in one place, e.g., due to a modular modules system where third party plugins may be built and loaded into immutable locations independent of the main openmm install. This patch supports path styles for OPENMM_PLUGIN_DIR of the type `some_path/plugins:other_path/plugins:etc`, to ease such use cases.

### DIFF
--- a/olla/include/openmm/Platform.h
+++ b/olla/include/openmm/Platform.h
@@ -213,26 +213,24 @@ public:
      */
     static void loadPluginLibrary(const std::string& file);
     /**
-     * Load multiple dynamic libraries (DLLs) which contain OpenMM plugins from a single directory.
-     * This method loops over every file contained in the specified directory and calls loadPluginLibrary()
+     * Load multiple dynamic libraries (DLLs) which contain OpenMM plugins from one or more directories.
+     * Multiple fully-qualified paths can be joined together with ':' on unix-like systems
+     * (or ';' on windows-like systems); each will be searched for plugins, in-order. For example,
+     * '/foo/plugins:/bar/plugins' will search both `/foo/plugins` and `/bar/plugins`. If an
+     * identically-named plugin is encountered twice it will be loaded at both points; be careful!!!
+     *
+     * This method loops over every file contained in the specified directories and calls loadPluginLibrary()
      * for each one.  If an error occurs while trying to load a particular file, that file is simply
      * ignored. You can retrieve a list of all such errors by calling getPluginLoadFailures().
      *
-     * @param directory    the path to the directory containing libraries to load
+     * @param directory    a ':' (unix) or ';' (windows) deliminated list of paths containing libraries to load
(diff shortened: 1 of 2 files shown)


olla/include/openmm/Platform.h
```diff
@@ -35,6 +35,7 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <sstream>
```
peastman: This include isn't actually needed anywhere in the header.  So better to move it to Platform.cpp.  That way it will only get included in that one file, not in everything that includes this header.

peastman:
Good idea!  And the implementation looks fine.  The one thing that's missing is documentation.  Could you update the description of `loadPluginsFromDirectory()` in the header to say you're allowed to give a list of directories instead of just one?

olla/include/openmm/Platform.h
tallakahath: I couldn't get it to compile without adding it there, though I haven't the foggiest idea why (`message` later on is a stringstream, too, and caused no problems). It could be specific to my build system, or maybe I had something misconfigured. I'll remove it and see if, in a completely clean env, it's still required.

peastman: How strange.  You do need to include that header, but you should be able to do it in the .cpp file rather than the .h file.

tallakahath: Ooooh I see the problem: the sstream import was inside ONLY the `#ifdef WIN32` block.

tallakahath: Moved it outside the ifdef block, compiles fine now.

olla/include/openmm/Platform.h
```diff
@@ -227,6 +227,13 @@ class OPENMM_EXPORT Platform {
      * OPENMM_PLUGIN_DIR is set, this returns its value.  Otherwise, it returns a platform
      * specific default location.
      *
+     * On unix-like systems, OPENMM_PLUGIN_DIR is allowed to either be a
```
peastman: This should go in the documentation for `loadPluginsFromDirectory()`, not `getDefaultPluginsDirectory()`.  Any path you tell it to load from can be a set of multiple directories, not just the default one.

jchodera:
>This patch supports path styles for OPENMM_PLUGIN_DIR of the type some_path/plugins:other_path/plugins:etc, to ease such use cases.

This is awesome---thanks for contributing it!

olla/include/openmm/Platform.h
tallakahath: Sorry, that was a typo on my part. I was parsing the comment as applying to whatever it was below, not what it was above.

peastman:
Looks great.  Thanks for implementing this!

Issue #2353, opened 2019-07-30
Missing features in CharmmPsfFile.createSystem()


Olllom:
While working with CHARMM input files in OpenMM, I have come across several shortcomings of the PSF parser `CharmmPsfFile.createSystem()`. I am willing to fix this soon, but before doing so, I wanted to check if anybody else has suggestions.

Here is what I found annoying:
- Although most of the CHARMM force field is parameterized without analytic dispersion correction, there is currently no option in `createSystem()` to manually disable the correction. Instead, the user has to disable it manually by walking through all `NonbondedForce` and `CustomNonbondedForce` objects of the system.
- `createSystem()` sometimes merges residues that have the same RESID from different segments. 
- There is an annoying warning "Detected PSF molecule section that is WRONG. Resetting molecularity," that gets thrown even for some simple PSF files that CHARMM does not have any trouble with. 

CharmmCrdFile:
- It might be useful to (maybe by default??) shift all coordinates by half a box length. Most CHARMM systems are build around the origin and we usually have to rewrap coordinates by hand. Not such a big deal, but anyway.

@jing-huang 
@andysim Reminder to modularize `createSystem` for your plugin.

peastman:
Fixes for these would be great!

Does CHARMM have the ability to apply a dispersion correction?  If so, how (and where) do you specify whether to use it?  Anyway, the default behavior of CharmmPsfFile should definitely be to match what CHARMM would do when processing the same file.

Olllom:
CHARMM has a dispersion correction that gets activated by the `LRC` or `LRC_MS` keywords in NBONDS: https://www.charmm.org/charmm/documentation/by-version/c42b1/params/doc/nbonds/ 
When using the CHARMM force fields, especially for lipid bilayers, the LRC should not be turned on.

One concern regarding backwards compatibility is CHARMM-GUI, which uses the parsers in its OpenMM input files.

peastman:
So `createSystem()` should be able to tell whether the correction is supposed to be enabled or not?  Then it can just automatically do the right thing.

> One concern regarding backwards compatibility is CHARMM-GUI, which uses the parsers in its OpenMM input files.

Do the scripts it generates disable the correction?  If so, this won't affect them.  If not, this will fix a bug.

Olllom:
> So createSystem() should be able to tell whether the correction is supposed to be enabled or not? Then it can just automatically do the right thing.

[...]

PR #2359, opened 2019-08-03
Fixes for #2353

Olllom:
@andysim Do you want to throw in your patch or open a separate pull request?

@peastman Thanks for resetting the resid to `None` for every new chain. That commit resolved the merging of residues.

Changes:
- Three fine-grained tests for issue #2353 
- Disable long-range correction
- MOLNT section is only tested for number of entries

Regarding the LRC keyword: I went through the CHARMM parameter files and did not find a single occurence of LRC. I also talked to some long-time CHARMM users and made sure that LRC would always be specified in UPDATE, ENERGY, or DYNA rather than in the parameter files.


peastman:
Are you sure about the recalculated energies in the tests?  The previous values were computed with CHARMM, so I don't know how they would have included the correction.

Olllom:
In order to make the Drude and NBFIX unit tests work, I recalculated the reference energies in CHARMM.

I am not sure if the CHARMM input scripts for the reference energies belong in the OpenMM repository. For now, I have put them into a [separate repository](https://github.com/Olllom/charmm-vs-openmm-energies).

This repo includes:
- A list of OpenMM energies.
- All input files for the NBFIX and Drude unit tests.
- A CI that runs the calculations in CHARMM and makes sure that the reported energies are correct.

Note that I had to increase the tolerance for the Drude from 0.5 to 1.0 kcal/mol to make the tests work. When using LRC in CHARMM, I missed the reference energy by 0.6 kcal/mol, too. 
@jing-huang It would be great if you could check my [Drude input script](https://github.com/Olllom/charmm-vs-openmm-energies/blob/master/drude/drudetest.inp) against the one you were using when you calculated the reference energies.

peastman:
> When using LRC in CHARMM, I missed the reference energy by 0.6 kcal/mol, too.

If you want the energies to match *with* the correction, I think you would use LRC_MS rather than LRC.

Olllom:
As far as I understand the docs for CHARMM and OpenMM, LRC refers to the Allen-Tildesley correction, i.e. NonbondedForce.setUseDispersionCorrection().
LRC_MS is Shirts’s method, i.e. CustomNonbondedForce.setUseLongRangeCorrection(). Therefore the former is active in the Drude Test, while the latter applies to the NBFIX force. Or am I understanding this wrong?

peastman:
They both use the Michael Shirts method.  The difference is that NonbondedForce can compute the integrals analytically, since we know in advance what the form is.  CustomNonbondedForce needs to compute them numerically.

Olllom:
Thanks for clarifying. 

This makes me wonder if the reference energy in the old unit test for Drude is sound. From what I am seeing the Ewald tolerance is too loose for a tight (0.5 kcal/mol) comparison.

OpenMM with long-range correction on Reference platform; CHARMM with **LRC_MS**.

| Ewald Tolerance | OpenMM Energy (grid size) | CHARMM Energy (grid size) 
| ----------- | --------- | --------- |
| 0.0005 | -1831.54030 (27) | -1835.09303 (30) |
| 0.00005 | -1834.49224 (49) | -1834.97569 (50) | 
| 0.000005 | -1834.72101 (87) | -1834.97259 (90) |
| 0.0000005 | -1834.73929 (150) | -1834.97251 (150) |

So both converge to an energy that is ~0.2 kcal/mol different. The old unit test used an Ewald tolerance of 0.0005 and the reference energy was 1831.54. If this reference energy was really calculated in CHARMM, it might have been luck that the OpenMM energies were so close.


Both without long-range correction

| Ewald Tolerance | OpenMM Energy (grid size) | CHARMM Energy (grid size) 
| ----------- | --------- | --------- |
| 0.0005 | -1784.76586 (27) | -1788.39520 (30) |
| 0.00005 | -1787.71779 (49) | -1788.27787 (50) | 
| 0.000005 | -1787.94657 (87) | -1788.27476 (90) |
| 0.0000005 | -1787.96486 (150) | -1788.27468 (150) |

Again, both converge to an energy that is ~0.2 kcal/mol different.

This makes me believe that the new test is fine and there might be an inaccuracy in the old one instead.

swails:
run jenkins

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -2439,7 +2439,7 @@ def createForce(self, sys, data, nonbondedMethod, nonbondedCutoff, args):
 
         for atom in data.atoms:
             self.force.addParticle((typeToMergedType[data.atomType[atom]],))
-        self.force.setUseLongRangeCorrection(True)
+        self.force.setUseLongRangeCorrection(bool(args.get('useLongRangeCorrection', False)))
```
peastman: This change is incorrect.  On any force field other than CHARMM, the correction should be enabled by default.

Olllom: OK, let me fix that.

peastman: A better solution would be to make this an attribute of the `<LennardJonesForce>` tag.  It could be processed similarly to the existing `lj14scale` attribute: see https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L2359-L2366.

Then we would need to add that attribute to all the CHARMM36 files.  Those files were generated with ParmEd.  For the moment it would probably be reasonable to just add the attribute by hand, but ParmEd would also need to get updated to produce correct files in the future.

Olllom:
Most recent commit:
- added an xml tag `useDispersionCorrection` to the LennardJones and NonbondedForce generators
- applied the keyword argument `useDispersionCorrection` that was already in `ForceField.createSystem()` also to the LennardJones force (not only the `NonbondedForce`)
- included `useDispersionCorrection` into all LennardJones and NonbondedForce generators in CHARMM xml files 

The behavior is as follows:
- If `useDispersionCorrection` is specified in either the xml file or as a keyword argument, use the specified value
- For conflicting settings, raise a warning, but take the keyword argument rather than the one from the xml file
- If it is not specified anywhere, use `True` as the default.

peastman:
Looks good!  Assuming the tests pass, is this ready to merge?

Olllom:
Yes.
Are you opening a PR/Issue for Parmed?

peastman:
I'm not familiar with the ParmEd code, so if you are and you want to update it that would be great.  Otherwise, I can probably figure it out.

peastman:
Merged.  Thanks!

PR #2373, opened 2019-08-15
Amoeba+

leucinw:


peastman:
I'm closing this, because this isn't an appropriate pull request.  To begin with, you just moved every single file in the whole source tree to a different directory.  And then you added a complete copy of Tinker to the source tree.  Neither of those is appropriate.

Please start by opening an issue.  We can discuss exactly what features you want to add, and the most appropriate way of adding them.  Once everyone is in agreement on it, you can create a new pull request containing on the changes relevant to the new feature.

PR #2374, opened 2019-08-20
set currentStep in DCDReporter

Olllom:
This should be the correct behavior.

When the step is initialized as 0 (current master version) postprocessing the trajectories in CHARMM leads to warnings and the first frame is skipped.


peastman:
Looks good, thanks!

And it looks like I *finally* managed to get AppVeyor working again!  Although it timed out running the tests, but hopefully that will fix itself once the cache gets repopulated.

Issue #2369, opened 2019-08-13
AMOEBA vdW softcore support


mjschnie:
This issue is a discussion of adding AMOEBA Buffered-14-7 softcore support to OpenMM. There are potentially two approaches:

1) Add a "CustomNonbonded" force class that supports reduced hydrogen positions. This would allow hydrogen atom vdW centers to be located between their current nuclear coordinates and the bonded heavy atom.

2) Add support directly to the AmoebaVdwForce. This would entail two global softcore parameters (i.e. power on lambda and a softcore offset), a per-particle Lambda parameter, and  rules for how particles should interact if both have Lambda values < 1.0 (i.e. intramolecular softcore).

The former approach is more general, but may not be widely used since reduced hydrogen positions are relatively uncommon. Approach 2 has been implemented in the TINKER-OpenMM branch:
https://github.com/TinkerTools/Tinker-OpenMM

peastman:
I agree the second approach is probably simpler.  What would the functional form be?  That is, how exactly would each interaction be calculated given the global and per-particle parameters?

mjschnie:
It's Equation 2 of the following paper:
https://www.pnas.org/content/105/17/6290

An initial reference implementation is given in the code linked below. This is a bit more complicated than necessary, because it supports both LJ 6-12 and Buffered-14-7. I'd suggest we only merge softcore Buffered 14-7 support (the LJ 6-12 softcore support is already available using a CustomNonbonded Force).

https://github.com/TinkerTools/Tinker-OpenMM/blob/master/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceVdwForce.cpp

peastman:
So in that equation, n and p are global parameters, while lambda is a per-particle parameter?  If so, how does the lambda in the equation get determined from the per-particle parameters?  Is there a combining rule for that?

Do you actually need multiple lambdas (turning on and off multiple groups at once)?  Or is it sufficient to have a single lambda, and the per-particle parameters just specify which particles are affected by it?  I ask because modifying a global parameter is a lot cheaper than modifying per-particle parameters.

mjschnie:
The global parameters are n and alpha (p is the separation distance rij normalized by the minimum energy separation).

We could definitely use a single global lambda value, and then a per-particle "flag" to specify if the atom is being soft-cored. Alternatively, the "lambda combining rules" used in TINKER-OpenMM are given below and use a per-particle lambda float value:

[...]

PR #2377, opened 2019-08-22
Add Softcore support to AmoebaVdwForce.

mjschnie:
This addresses issue "AMOEBA vdW softcore support #2369":
https://github.com/openmm/openmm/issues/2369

- Each atom is configured with an alchemical flag to indicate if it should be affected by the softcore functional form.
- The softcore form is defined by 1) multiplying interactions by Lambda^N and 2) adding Alpha*(Lambda -1)^2 to the reduced separation distance (rho = rij / rmin).  Both N and Alpha are global parameters of the force (defaults are 5 and 0.7, respectively) and can be set.
- Application of the softcore form is governed by a enum AlchemicalMethod that is "None" by default, or can be set to either Decouple or Annihilate. Decouple does not apply the softcore form between two atoms that are both alchemical, while Annihilate does.
- A single Context parameter "vdwLambda" controls the alchemical state, which avoids the need to perform a relatively slow update of per particle parameters in the Context when/if vdwLambda changes.

### DIFF
--- a/plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp
+++ b/plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp
@@ -2343,9 +2343,10 @@ public:
     bool areParticlesIdentical(int particle1, int particle2) {
         int iv1, iv2;
         double sigma1, sigma2, epsilon1, epsilon2, reduction1, reduction2;
-        force.getParticleParameters(particle1, iv1, sigma1, epsilon1, reduction1);
-        force.getParticleParameters(particle2, iv2, sigma2, epsilon2, reduction2);
-        return (sigma1 == sigma2 && epsilon1 == epsilon2 && reduction1 == reduction2);
+        bool isAlchemical1, isAlchemical2;
+        force.getParticleParameters(particle1, iv1, sigma1, epsilon1, reduction1, isAlchemical1);
+        force.getParticleParameters(particle2, iv2, sigma2, epsilon2, reduction2, isAlchemical2);
+        return (sigma1 == sigma2 && epsilon1 == epsilon2 && reduction1 == reduction2 && isAlchemical1 == isAlchemical2);
     }
 private:
     const AmoebaVdwForce& force;
@@ -2378,7 +2379,8 @@ void CudaCalcAmoebaVdwForceKernel::initialize(const System& system, const Amoeba
     for (int i = 0; i < force.getNumParticles(); i++) {
         int ivIndex;
         double sigma, epsilon, reductionFactor;
-        force.getParticleParameters(i, ivIndex, sigma, epsilon, reductionFactor);
+        bool isAlchemical;
+        force.getParticleParameters(i, ivIndex, sigma, epsilon, reductionFactor, isAlchemical);
         sigmaEpsilonVec[i] = make_float2((float) sigma, (float) epsilon);
         bondReductionAtomsVec[i] = ivIndex;
         bondReductionFactorsVec[i] = (float) reductionFactor;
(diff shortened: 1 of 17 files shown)


plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp
```diff
@@ -2449,6 +2468,12 @@ double CudaCalcAmoebaVdwForceKernel::execute(ContextImpl& context, bool includeF
         hasInitializedNonbonded = true;
         nonbonded->initialize(system);
     }
+
+    // Not sure about passing a Context parameter to the Kernal?
+    vector<float> vdwLambdaVec(1, 0);
+    vdwLambdaVec[0] = context.getParameter(AmoebaVdwForce::Lambda());
+    vdwLambda.upload(vdwLambdaVec);
```
peastman: This will work, but it's a little inefficient.  The problem is that `upload()` is a blocking call.  So it doesn't return until the upload is complete, and at that point the GPU's work queue will have been emptied out.  Then it will take some time for it to set up the next kernel, and during that time the GPU won't be doing any work.  It's not that big a deal, but still a slight inefficiency.

The first thing you can do is just put it in an "if" block based on the alchemical method.  In most simulations it will be `None`, so you can just skip this altogether.

The next thing you can do is use the alternate version of `upload()` (https://github.com/openmm/openmm/blob/master/platforms/cuda/include/CudaArray.h#L176) that supports non-blocking copies.  Just pass `false` for the second argument.  But of course there's a complication: non-blocking copies can only be done from pinned memory.  So to do that, you need to allocate the buffer with `cuMemHostAlloc()`, then free it with `cuMemFreeHost()` in the destructor.  For an example of this, take a look at the `pinnedBuffer` field of CudaContext.

peastman: Here's another way of optimizing it: just keep track of what the last value was that you uploaded, and only upload a new value when it changes.  Lambda won't usually change every time step.  If you're only having to upload a new value occasionally, the cost will be totally negligible.

peastman:
This looks really good!  Just a few minor changes needed.

You're getting a test failure in `testAmoebaVdwForce()` in TestAPIUnits.py.  It just needs minor changes to handle the new parameters.

The serialization proxy needs to be backward compatible, so it can still read files written by older versions.  The way to do that is to increment the version number written at the very start of `serialize()`.  Then `deserialize()` can use that to read each file correctly.  If `version <= 2` it will use default values for the new parameters, and for `version >= 3` it will expect them to have been stored in the file.

Could you add a couple of sentences to the documentation at the top of AmoebaVdwForce.h?  Just say that it supports softcore interactions, and give a reference to the paper describing the functional form.

Otherwise, this really looks great.  I'm impressed at how well you managed to understand some very obtuse code.

peastman:
Looks great!  Assuming the tests pass, I think this should be ready to merge.  I'm going to delay actually merging it until we have 7.4 out.  We're right in the middle of building the release candidate now, and if we can avoid having to make a separate branch that will make our lives simpler.

mjschnie:
Thank for the feedback Peter!
I switched the vdwLambda parameter over to using pinned host memory for non-blocking transfers, and also implemented your suggestion to only update the device value of lambda when its changed. I agree that the cost for this should be negligible now, even for alchemical simulations. 
The serialization has been updated to version 3, with backwards compatibility to previous versions. 
I added documentation AmoebaVdwForce.h -- see what you think?

Waiting to merge works for me; in the mean time we'll try this out on some production Monte Carlo Orthogonal Space Tempering simulations.

peastman:
run jenkins

peastman:
@swails looks like Jenkins has stopped working again.

swails:
My computer shut down unexpectedly. I need to turn it back on. I will soon.

peastman:
Thanks!

swails:
run jenkins

swails:
Jenkins stalled for some reason when I brought it back online.  Trying again

swails:
run jenkins

peastman:
7.4 is out, so I'll merge this.  Sorry for the delay.

PR #2378, opened 2019-08-22
Added the W-H combining rule for AMOEBA+ VdW

leucinw:


peastman:
The only thing you changed is the calculation of the dispersion correction coefficient.  Here are the things you need to do to implement this:

- Update the documentation for all relevant methods in AmoebaVdwForce.h to add the new value to the list of allowed combining rules.
- Update the reference implementation in AmoebaReferenceVdwForce.cpp.  That involves another branch in `setEpsilonCombiningRule()`, and possibly a new function to implement the computation.
- Update the CUDA implementation.  That includes a new branch in https://github.com/openmm/openmm/blob/master/plugins/amoeba/platforms/cuda/src/AmoebaCudaKernels.cpp#L2416-L2425, and then a new block in https://github.com/openmm/openmm/blob/master/plugins/amoeba/platforms/cuda/src/kernels/amoebaVdwForce2.cu#L18-L28 to perform the calculation.
- Add a new method in TestReferenceAmoebaVdwForce.cpp and TestCudaAmoebaVdwForce.cpp to test the new combining rule and make sure it produces correct results.
- Add the new combining rule at https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L4434 so ForceField will accept it as a legal value.

peastman:
Much better!  Where did the reference values in the test case come from?  Were those computed with Tinker?

peastman:
Looks like you're getting a compilation error:

```
/AmoebaReferenceVdwForce.cpp: In member function ‘void OpenMM::AmoebaReferenceVdwForce::setEpsilonCombiningRule(const string&)’:
/home/travis/build/openmm/openmm/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceVdwForce.cpp:133:27: error: cannot convert ‘double (OpenMM::AmoebaReferenceVdwForce::*)(double, double, double, double)const’ to ‘OpenMM::CombiningFunction {aka double (OpenMM::AmoebaReferenceVdwForce::*)(double, double)const}’ in assignment
          _combineEpsilons = &AmoebaReferenceVdwForce::whEpsilonCombiningRule;
                           ^
```

leucinw:
Forces and energy are from Tinker, after unit convertion. It seems the
force in openmm has negative sign comparing to Tinker. I will check the
error tonight.

leucinw:
> Looks like you're getting a compilation error:
> 
> ```
> /AmoebaReferenceVdwForce.cpp: In member function ‘void OpenMM::AmoebaReferenceVdwForce::setEpsilonCombiningRule(const string&)’:
> /home/travis/build/openmm/openmm/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceVdwForce.cpp:133:27: error: cannot convert ‘double (OpenMM::AmoebaReferenceVdwForce::*)(double, double, double, double)const’ to ‘OpenMM::CombiningFunction {aka double (OpenMM::AmoebaReferenceVdwForce::*)(double, double)const}’ in assignment
>           _combineEpsilons = &AmoebaReferenceVdwForce::whEpsilonCombiningRule;
>                            ^
> ```

The problem is that W-H combining rule is using both 2 sigmas and 2 epsilons to get the combined epsilon, while other combining rules usually use only sigma or epsilon. Any thoughts on how to deal with this easily?

peastman:
> It seems the force in openmm has negative sign comparing to Tinker.

You're using `testgrad`?  It outputs the gradient of the energy, which is the negative of the force.

> The problem is that W-H combining rule is using both 2 sigmas and 2 epsilons to get the combined epsilon, while other combining rules usually use only sigma or epsilon. Any thoughts on how to deal with this easily?

Yeah, that code is written in a very... curious style.  I didn't write it. :)  Some day I really ought to give all the AMOEBA code a massive cleanup, but I'm not sure it's really worth the effort.  Anyway, if you want to make all the epsilon combining rules take four arguments (epsilons and sigmas) instead of two, that would probably be the simplest solution for now.

leucinw:
> You're using `testgrad`? It outputs the gradient of the energy, which is the negative of the force.
 Oh yes I was using testgrad. It makes sense to me now.

> Anyway, if you want to make all the epsilon combining rules take four arguments (epsilons and sigmas) instead of two, that would probably be the simplest solution for now.
Thanks. I will be doing this way then.

peastman:
7.4 is out, so we can merge this.  Unfortunately, merging #2377 seems to have created conflicts, since it also involved changes to AmoebaVdwForce.  Can you resolve the conflicts?  Hopefully if you just `git pull` it will be able to resolve them automatically.

leucinw:
I checked the conflicts. It should be ok to merge now?

peastman:
Thanks!  All the tests are failing though, with this error:

```
/home/travis/build/openmm/openmm/plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceVdwForce.cpp:399:16: error: redeclaration of ‘double combinedEpsilon’
         double combinedEpsilon = (this->*_combineEpsilons)(epsilons[siteI], epsilons[siteJ]);
                ^
```

peastman:
Now it's getting this error:

```
/TestReferenceAmoebaVdwForce.cpp:416:105: error: a function-definition is not allowed here before ‘{’ token
 void testVdwAlchemical(int power, double alpha, double lambda, AmoebaVdwForce::AlchemicalMethod method) {
                                                                                                         ^
/home/travis/build/openmm/openmm/plugins/amoeba/platforms/reference/tests/TestReferenceAmoebaVdwForce.cpp:2192:1: error: expected ‘}’ at end of input
 }
 ^
```

plugins/amoeba/platforms/reference/tests/TestReferenceAmoebaVdwForce.cpp
```diff
@@ -436,12 +466,8 @@ void testVdwAlchemical(int power, double alpha, double lambda, AmoebaVdwForce::A
     for (int i=0; i<8; i++) {
         expectedForces[i] *= scale;
     }
-
-    double tolerance          = 1.0e-04;
-    compareForcesEnergy(testName, expectedEnergy, energy, expectedForces, forces, tolerance);
```
peastman: I don't think you meant to delete this code, did you?  You just removed the part of this test case that did the actual testing.

plugins/amoeba/platforms/cuda/tests/TestCudaAmoebaVdwForce.cpp
```diff
@@ -396,18 +396,41 @@ void testVdwAmmoniaCubicMeanHhg() {
     compareForcesEnergy(testName, expectedEnergy, energy, expectedForces, forces, tolerance);
 }
 
+
+// test VDW w/ sigmaRule=CubicMean and epsilonRule=W-H
+
+void testVdwAmmoniaCubicMeanWH() {
+
+    std::string testName      = "testVdwAmmoniaCubicMeanWH";
+
```
peastman: Something's clearly wrong in this code.  You begin a function, then go right on to the next one without closing it off, which I guess makes the `testVdwAlchemical` into a local function?

leucinw: You are right. This was caused by my incorrectly deleting the conflict markers.

peastman:
Looks good.  Thanks!

PR #2394, opened 2019-09-17
Enable running OpenCL tests on Mac

peastman:


peastman:
No good.  The OpenCL tests all fail with the error "No compatible OpenCL platform is available".  Presumably that means the VM doesn't have access to any kind of GPU.  It was an experiment.

PR #2397, opened 2019-09-19
Use correct parameters for nucleic acids with GBn2

### LINKED ISSUE #2391: GBn2 uses incorrect parameters for nucleic acids
When creating a system from a prmtop file and using GBn or GBn2 implicit solvent, certain per-particle parameters are assigned based on the element.  It turns out that Amber uses different values for these parameters in nucleic acids and proteins.  OpenMM uses the protein parameters for everything.  Possibly it was based on an older version of Amber that didn't include the nucleic acid support?  We need to fix it to use the correct nucleic acid parameters for these two implicit solvent models.

See https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=10939&p=30328&start=0&view=.

peastman:
Fixes #2391


peastman:
The test case for this PR computes the energy for a DNA molecule and compares it to the value from Amber.

hainm:
hi, I was asking about the complex (with protein). I guess it's a "not (yet)", right?

peastman:
I haven't tested that case.  If you'd like to do it, that would be great.

Issue #204, opened 2013-11-04
Automatic unit compatibility checking


kyleabeauchamp:
```
timestep = 2.0 / u.picoseconds
integrator = mm.VerletIntegrator(timestep)

In [45]: timestep
Out[45]: Quantity(value=2.0, unit=/picosecond)

In [46]: integrator.getStepSize()
Out[46]: Quantity(value=2.0, unit=picosecond)

```

To me, this looks like the unit conversion is trying to be smart and automatically invert per-picoseconds into picoseconds.  However, it seems to be failing.  

IMHO, this sort of automatic inversion should not even be attempted without a warning or error.  

Thoughts?

jchodera:
Let's mark this as a potential "enhancement".  We can try the `accepts_compatible_units` unit checking soon.

jchodera:
Maybe rename the issue "Automatic unit compatibility checking"?

peastman:
This is fixed by #3010.

PR #2411, opened 2019-09-30
Add ability to enforce units on input arguments

### LINKED ISSUE #2406: wrong unit conversion in MonteCarloMembraneBarostat.getSurfaceTension
The surface tension gets multiplied by 1e-25 when setting and getting the surface tension in units.

Part of the problem might be that the comment in MonteCarloMembraneBarostat.h specifies the unit as `bar` rather than `bar*nm` but I don't see how this would give a factor of 1e-25.

```
In [1]: from simtk import unit                                                                                                         

In [2]: from simtk.openmm import MonteCarloMembraneBarostat      

In [5]: barostat = MonteCarloMembraneBarostat(1.0*unit.bar, 1.0*unit.bar*unit.nanometer, 1, 1, 25)                                     

In [6]: barostat.getDefaultPressure()                                                                                                  
Out[6]: Quantity(value=1.0, unit=bar)

In [7]: barostat.getDefaultSurfaceTension()                                                                                            
Out[7]: Quantity(value=1e-25, unit=nanometer*bar)

In [8]: barostat = MonteCarloMembraneBarostat(1.0, 1.0, 1, 1, 25)                                                                      

In [9]: barostat.getDefaultPressure()                                                                                                  
Out[9]: Quantity(value=1.0, unit=bar)

In [10]: barostat.getDefaultSurfaceTension()                                                                                           
Out[10]: Quantity(value=1.0, unit=nanometer*bar)
```

peastman:
Fixes #2406.  At least mostly.  See my comments on that issue.

If units are specified for an argument in swigInputConfig.py, it will now convert the argument to those units.  This makes sure the correct units are used in non-standard cases, and also throws an exception if the user passes inconsistent units.

This could be the basis of a full solution to #204, but that will take a lot of work.  At the moment I've only specified units for a handful of cases.  For all other cases, it still uses the old behavior of assuming the units are right and converting to the standard unit system.


jchodera:
This is an awesome improvement!

Olllom:
Great! Many thanks.

PR #2416, opened 2019-10-03
Fixed error building neighbor list with triclinic box

### LINKED ISSUE #2361: Problem with Amber-prepared truncated octahedral/triclinic system
Hi, I have a bizarre problem and I wonder if you can suggest something how to debug it. I prepared a system (protein+co-factor) with Amber using a truncated octahedral simulation box. I have successfully read the parm7 and rst7 input files in OpenMM (verified that the box vectors were triclinic) and started a vanilla simulation to test the system. I noticed, however, that after proper minimization the simulation invariably crashed with "Exception: Particle coordinate is nan." I started to experiment with more or less stringent simulation parameters that have always worked just fine with orthorhombic systems, but even with the most stringent settings the simulation crashed after many millions of MD steps, which is, I guess, indicating that this is not an equilibration issue. Unfortunately, I cannot see any smoking gun as to why the simulations end up with these nan coordinates. However, I ran a sanity check by preparing the exact same system with Amber except that I built an orthorhombic box and ran the exact same simulations with OpenMM, and found that all of them without exception ran just fine and never crashed. Even simulations with HMR and MTS and a loose Ewald tolerance are running and seem totally healthy. It seems that there might be something about the truncated octahedral/triclinic simulation box that is ultimately causing the crashes, but I have no idea how to debug it since I have not found any signs of bad behavior.

Thanks a million for any suggestions.

Best regards,

   Istvan

peastman:
Fixes #2361.

When using a triclinic box, the neighbor list builder would very occasionally miss an interaction that should have been included.  This only happened with CUDA.

### DIFF
--- a/platforms/cuda/src/kernels/findInteractingBlocks.cu
+++ b/platforms/cuda/src/kernels/findInteractingBlocks.cu
@@ -243,6 +243,7 @@ extern "C" __global__ void findBlocksWithInteractions(real4 periodicBoxSize, rea
         for (int block2Base = block1+1; block2Base < NUM_BLOCKS; block2Base += 32) {
             int block2 = block2Base+indexInWarp;
             bool includeBlock2 = (block2 < NUM_BLOCKS);
+            bool forceInclude = false;
             if (includeBlock2) {
                 real4 blockCenterY = sortedBlockCenter[block2];
                 real4 blockSizeY = sortedBlockBoundingBox[block2];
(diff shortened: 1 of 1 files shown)


peastman:
@swails Jenkins is having problems again.  Either the CUDA or the OpenCL build always fails, and sometimes both.  Here's the complete log:

```
+ docker inspect -f . swails/openmm-all:latest
.
Failed to run image 'swails/openmm-all:latest'. Error: docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"process_linux.go:413: running prestart hook 1 caused \\\"error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: cuda error: no cuda-capable device is detected\\\\n\\\"\"": unknown.
```

swails:
My Ubuntu machine thought a kernel update was important enough to force itself.  That broke the nvidia driver.  I reinstalled the driver and rebooted.  Should be fine now.

swails:
run jenkins

peastman:
Thanks!

Issue #2422, opened 2019-10-09
PDBxFile writer produces invalid mmCIF files


JoaoRodrigues:
Writing PDBx/mmCIF files using `app.PDBxFile` and specifying `keepIds=True` will produce invalid output files. Trying to read them back in fails with the following error:.

```
Traceback (most recent call last):
  File "/home/software/miniconda3/envs/openmm/lib/python3.6/site-packages/simtk/openmm/app/pdbxfile.py", line 150, in __init__
    atom = atomTable[atomKey]
KeyError: ('A', '1', 'ATOM')

During handling of the above exception, another exception occurred:
```

The issue is that residue insertion codes are blank in most PDB files but are recorded in the topology as a single space ` `. This 'feature' then tricks `PDBxFile.writeModel` into writing that single space as the residue ins. code instead of a period (`.`). This, in turn, trips the reader when it parses the different fields and leads to the error above.

Anyway, this is fixed by changing [line 400 of pdbxfile.py](https://github.com/openmm/openmm/blob/e0b3c2bd86fe471f9b75bec10f24365b523152aa/wrappers/python/simtk/openmm/app/pdbxfile.py#L400) to `if res.insertionCode.strip()` instead of the `len()` check.

The following code reproduces the error:

```
from pathlib import Path
import simtk.openmm.app as app

fname = Path(app.__file__).parent / 'data' / 'POPC.pdb' 

mol = app.PDBFile(fname)

# This works
with open('test.cif', 'w') as handle:
    app.PDBxFile.writeFile(mol.topology, mol.positions, handle)

mol2 = app.PDBxFile('test.cif')

# This fails
with open('test.cif', 'w') as handle:
    app.PDBxFile.writeFile(mol.topology, mol.positions, handle, keepIds=True)

mol2 = app.PDBxFile('test.cif')
```

peastman:
Fixed by #2423.

PR #2423, opened 2019-10-09
Fixes writing PDBx/mmCIF files with blank residue insertion codes.

JoaoRodrigues:
Addresses #2422 

Fixes writing invalid PDBx/mmCIF files when residue insertion codes are blank (not empty) in Topology. Replaces the current check `len(res.insertionCode) > 0` with `res.insertionCode.strip()` to account for cases where the insertion code is stored as a single space from reading e.g. PDB files.


peastman:
Thanks!

Issue #2424, opened 2019-10-12
addMembrane fails for membrane patches with >128 lipids


JoaoRodrigues:
There is a problem in `modeller.addMembrane` that causes (at least PDBx/mmCIF) files that are written to have overlapping residue numbers for the lipids. See the code snippet below to reproduce locally (using a OpenMM test file).

While this does not give an error right away, it leads to problems later on if you re-read the file and try to do something with it. For example, only the last 128 POPC lipids have bonds defined in the topology (try reloading and doing res.bonds on the first POPC res). This leads to the following (cryptic) error when trying to build a system from that structure:

```
Traceback (most recent call last):
  File "./md_scripts/openmm//amberff/minimizeSystem.py", line 82, in <module>
    rigidWater=True)
  File "/home/software/miniconda3/envs/openmm/lib/python3.6/site-packages/simtk/openmm/app/forcefield.py", line 1145, in createSystem
    raise ValueError('No template found for residue %d (%s).  %s' % (res.index+1, res.name, _findMatchErrors(self, res)))
ValueError: No template found for residue 389 (POP).  The set of atoms matches POPC, but the bonds are different.  Perhaps the chain is missing a terminal group?
```

A solution is to simply renumber the lipids when adding them to the topology in `addMembrane`. Another is to create several a new chain for every lipid patch we add. I like the first solution better for its simplicity - harder to handle multiple chains later on during analysis.

I'll make a PR to address this.

Code to reproduce this problem:
```
import simtk.openmm.app

mol = app.PDBFile('/home/software/science/openmm-src/wrappers/python/tests/systems/alanine-dipeptide-implicit.pdb') 
ff = app.ForceField('amber14-all.xml', 'amber14/tip3p.xml')

modeller = app.Modeller(mol.topology, mol.positions)
modeller.addMembrane(ff, minimumPadding=5.0)

for r in modeller.topology.residues(): 
    if r.name == 'HOH': continue 
    print(r, r.name, r.id, r.chain, r.chain.id)  # multiple POPC residues with the same numbering
```

peastman:
Fixed by #2425.

PR #2425, opened 2019-10-12
Fixes addMembrane for large(r) systems

JoaoRodrigues:
Addresses #2424 

I also added a new test system (a GPCR taken from OPM and _fixed_ using PDBFixer) to have a realistic test and modified `test_Modeller.py` to check for overlapping numberings in lipids.


peastman:
Thank you!

PR #2428, opened 2019-10-17
Updated Appveyor VM

peastman:
Supposedly this will make it start up faster.


peastman:
It still times out, but it was doing that before.  I'll merge this and then investigate how to speed it up separately.

PR #2429, opened 2019-10-17
Fixed range error when adding lone pairs from PSF file

### LINKED ISSUE #2392: OpenMM 7.4.0 crashes when using the Drude-2013 forcefield
I recently downloaded OpenMM 7.4.0 and tried a simulation using the Drude-2013 forcefield. Right at the beginning of the simulation it crashed: `Exception: Particle coordinate is nan`. When I tried the exact same system with OpenMM 7.3.1 everything went smoothly. 

I also checked whether the used platform makes any difference. Here is what I found:

- Version 7.3.1
All four platforms (Reference, CPU, CUDA, and OpenCL) worked.

- Version 7.4.0
Reference: `Segmentation fault`
CPU: `Exception: Particle coordinate is nan`
CUDA and OpenCL: `ValueError: Energy is NaN`

I hope this helps.

peastman:
Fixes #2392.

@jing-huang could you check that my fix is correct?  The problem was an off-by-one error introduced in #2317.  CHARMM numbers atoms starting at 1, and it appears that it sets the fourth atom to 0 (or negative?) to indicate a collinear lone pair.  OpenMM numbers atoms starting at 0, so it subtracted 1 from the indices.  That means it should be checking that the last atom is >=0, not >0.

This affected any file that satisfied very specific conditions: it contained a non-collinear lone pair, and the third atom it depended on was the very first atom in the file.

### DIFF
--- a/wrappers/python/simtk/openmm/app/charmmpsffile.py
+++ b/wrappers/python/simtk/openmm/app/charmmpsffile.py
@@ -934,7 +934,7 @@ class CharmmPsfFile(object):
                 atom1=lpsite[1]
                 atom2=lpsite[2]
                 atom3=lpsite[3]
-                if atom3 > 0: 
+                if atom3 >= 0: 
                     if lpsite[4] > 0 : # relative lonepair type
                         r = lpsite[4] /10.0 # in nanometer
                         xweights = [-1.0, 0.0, 1.0]


jing-huang:
Yes, everything looks fine. Thanks for getting this fixed!

Issue #2396, opened 2019-09-19
BAOAB integrator


peastman:
Following up on https://github.com/openmm/openmm/issues/2386#issuecomment-531438507, I've been looking into creating a new BAOAB integrator.  As a starting point, I wanted to get a sense for how well LangevinIntegrator (which uses the Langevin Leapfrog algorithm of https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4308582/) compares, and to consider different ways of formulating it.  So I tried to reproduce the tests described in https://royalsocietypublishing.org/doi/full/10.1098/rspa.2016.0138.  I simulated alanine dipeptide in a box of water, using `rigidWater=True` and `constraints=HBonds`.  I tried various timesteps and three different integrators:

- LangevinIntegrator
- BAOABIntegrator from openmmtools
- GeodesicBAOABIntegrator from openmmtools with `K_r=2`.

For each one I computed the average potential energy and temperature over a set of simulations.

The first thing I found was that there was no significant difference in stability between them.  All the integrators were stable at 4 fs.  None was stable at 6 fs.

Here is the average potential energy as a function of step.  (Ignore the "simplified" line.  I'll get to that shortly.)

![image](https://user-images.githubusercontent.com/4379786/65265032-d005d880-dac4-11e9-9453-5ed4157103a1.png)

The BAOAB integrators do much better than LangevinIntegrator as you increase the time step.  An interesting difference is that the average energy increase with LangevinIntegrator (the system is too hot) but decreases with BAOAB (the system is too cold).  The difference between the two BAOAB integrators is very small, but GeodesicBAOABIntegrator seems to be slightly worse than BAOABIntegrator.

Now here is the average temperature.

![image](https://user-images.githubusercontent.com/4379786/65265279-689c5880-dac5-11e9-93ee-5f996916a74a.png)

The situation is reversed.  LangevinIntegrator has no trouble holding the temperature right at 300K.  The BAOAB integrators show significant deviation even at 2 fs, and by 4 fs the temperature is about 9 degrees too low.

How should I interpret this?  The paper on BAOAB said it was more accurate for configurational quantities than kinetic ones, but I wasn't expecting it to be quite that inaccurate.  Should BAOAB be avoided for simulations where you want to compute kinetic quantities?

[...]

PR #2436, opened 2019-10-21
Create BAOABLangevinIntegrator

peastman:
Implements #2396.  Please give it a try!

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -120,7 +120,7 @@ steps.
         forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
         system = forcefield.createSystem(pdb.topology, nonbondedMethod=PME,
                 nonbondedCutoff=1*nanometer, constraints=HBonds)
-        integrator = LangevinIntegrator(300*kelvin, 1/picosecond, 0.002*picoseconds)
+        integrator = BAOABLangevinIntegrator(300*kelvin, 1/picosecond, 0.002*picoseconds)
         simulation = Simulation(pdb.topology, system, integrator)
         simulation.context.setPositions(pdb.positions)
         simulation.minimizeEnergy()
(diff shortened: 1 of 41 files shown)


peastman:
Should I merge this, or does anyone plan to try it first?

jchodera:
Apologies---we hadn't had a chance to comment on this yet due to travels and visitors!
@maxentile @c-matthews @andrrizzi : Can you also take a look?

docs-source/usersguide/application.rst
```diff
@@ -1022,13 +1022,13 @@ Integrators
 OpenMM offers a choice of several different integration methods.  You select
 which one to use by creating an integrator object of the appropriate type.
 
-Langevin Integrator
--------------------
+BAOAB Langevin Integrator
```
jchodera: Would you mind if I contributed a bit more information to this section?

docs-source/usersguide/theory.rst
```diff
@@ -1338,6 +1338,21 @@ The integration is done using a leap-frog method similar to VerletIntegrator.
 :cite:`Izaguirre2010` The same comments about the offset between positions and
 velocities apply to this integrator as to that one.
 
+BAOABLangevinIntegrator
+***********************
+
```
jchodera: Would you mind if I contribute a little more information to this section?

platforms/cuda/src/kernels/baoab.cu
```diff
@@ -0,0 +1,127 @@
+enum {VelScale, NoiseScale};
+
+/**
+ * Perform the first step of BAOAB integration.
```
jchodera: Would it make sense to provide a little bit of documentation about how you split the BAOAB update equations up among kernel calls? It looks like you fused the operations for a few equations here, and it would be good to tie specific equations in the publication to your logic in the implementation for future maintainability or development of variants (e.g. g-BAOAB, MTS versions, etc.).

docs-source/usersguide/application.rst
peastman: Probably simplest if I merge this, then you can create a new PR with additional text?

platforms/cuda/src/kernels/baoab.cu
peastman: It pretty much just goes through the steps in order, nothing unusual at all.  The breaks between kernels just correspond to the places where it applies constraints or computes forces.  All the other integrators have their kernels split in exactly the same way.

docs-source/usersguide/application.rst
jchodera: Sounds good!

platforms/cuda/src/kernels/baoab.cu
jchodera: Can we at least minimally document that in the code? In general, it's great if the code is more readable and maintainable by everyone---if it's not obvious what is going on, a few extra comment lines on how you are breaking up the update equations into kernels is super helpful.

jchodera: Thanks! This is helpful!

Issue #2439, opened 2019-10-23
Bug in addMembrane: wrong number of ions added for membrane systems at a certain concentration


JoaoRodrigues:
When writing #2274 I made two mistakes that cause a miscalculation in the number of ions when `ionicStrength > 0`. This mistake leads to a lower concentration of ions that intended (although the charge of the system remains neutral if `neutralize=True`). The problem comes from:

1. in #2099 and #2274, I remove a layer of waters that is too close to the membrane, to avoid clashes between the lipid heads and the ions. This layer _must_ be counted towards the total number of waters when calculating the molar fraction of ions to be added. Right now, it is not, so we get a lower count here.

2. I left some code from before that generated a list of water molecules from the last two chains in the topology, assuming that these were the ones added. When I refactored the ion adding code into `_addIons`, I created a new chain in the topology _regardless_ of adding ions or now. When the box size in Z is too big and more waters were needed than those in the lipid patches, there is an internal call in `addMembrane` to `addSolvent`, which in turn calls `_addIons` (with neutralize=False and ionicStrength=0). This last step created an empty chain that then threw off the counting of waters in the `waterPos` dictionary.

Apologies for the oversight. If anything, this bug highlights the need for tests whenever a new feature is added (and generally for the code). PR incoming, I'd appreciate a thorough review to make sure I did not miss anything.

Issue #2430, opened 2019-10-18
Create 7.4.1 release


peastman:
There've been enough bug fixes since 7.4 that I think it would make sense to do a 7.4.1 patch.  Here are the changes I'd suggest including.

#2397: Use correct parameters for nucleic acids with GBn2
#2400: Barostats use checkpointed random number generator
#2411: Add ability to enforce units on input arguments
#2413: Make sure correct CUDA context is current
#2416 and #2418: Fixed error building neighbor list with triclinic box
#2423: Fixes writing PDBx/mmCIF files with blank residue insertion codes.
#2425: Fixes addMembrane for large(r) systems

jchodera:
How 'bout we also toss in

#2421: Added missing hartree unit (useful for OpenMM integration into [QCEngine](https://github.com/MolSSI/QCEngine/pull/151))

Otherwise, looks good to me!

peastman:
Sure.  It isn't really a bug fix, but it's a very small change so I'm ok with including it.

jchodera:
Considering this was an accidental omission of an important unit despite the inclusion of its friends, it's not a stretch to consider it a "bugfix".

JoaoRodrigues:
I think there might be a bug in the way ions are added to membrane systems via `addMembrane`. When I refactored the neutralization code into the `_addIons` function I missed that since I am excluding some waters from the pool of replaceable molecules, I am effectively underestimating the number of waters in the box and therefore the final ion concentration is _less_ than expected. I don't expect the effect to be really big since the layer of 'protected' waters is thin, but nevertheless ... This problem was never caught because the solvation/neutralization tests never check for membrane systems.

I'm working on a fix and on improving the tests, but if you don't mind, I'd rather have this shipped in 7.4.1! Sorry for introducing a bug ...

peastman:
Could you clarify what you mean?  What waters are being excluded?

JoaoRodrigues:
See [here](https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L1562). When `_addIons` is called, it gets `waterPos`as the pool of water molecules that will be replaced by ions. Now look [here](https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L329). For the membrane case, because there are some water molecules that are excluded from the replaceable pool (too close to the membrane), the `numIons` calculation is slightly wrong, leading to an effectively lower concentration of ions in the final box.

[...]

PR #2440, opened 2019-10-23
Fixes _addIons behavior for membrane systems.

JoaoRodrigues:
Addresses #2439 and problem raised in #2430. 

I'd rather have someone thoroughly review the changes... It took me quite a few hours to figure this out and I'm not sure I got all the behaviour right (especially the test code). Shoutout to @joaomcteixeira for helping debug this.

Thanks for all the patience!

### DIFF
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -258,13 +258,16 @@ class Modeller(object):
         self.topology = newTopology
         self.positions = newPositions
 
-    def _addIons(self, forcefield, replaceableMols, ionCutoff=0.05*nanometer, positiveIon='Na+', negativeIon='Cl-', ionicStrength=0*molar, neutralize=True):
+    def _addIons(self, forcefield, numWaters, replaceableMols, ionCutoff=0.05*nanometer, positiveIon='Na+', negativeIon='Cl-', ionicStrength=0*molar, neutralize=True):
         """Adds ions to the system by replacing certain molecules.
 
         Parameters
         ----------
         forcefield : ForceField
             the ForceField to use to determine the total charge of the system.
+        numWaters : int
+            the total number of water molecules in the simulation box, used to
+            calculate the number of ions / concentration to add.
         replaceableMols : dict
             the molecules to replace by ions, as a dictionary of residue:positions
         ionCutoff: distance=0.5*nanometer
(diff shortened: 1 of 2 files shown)


JoaoRodrigues:
Appveyor build+tests take a very long time - build failed because it reached the 60min limit.

peastman:
I think this fix is way more complicated than it needs to be!  When calling `_addIons()`, why not just scale the ionic strength by the ratio of total waters to replaceable waters?

JoaoRodrigues:
@peastman so you suggest changing the `ionicStrength` value before passing it to `_addIons`?

Note that there are two changes here: one is the indentation of a whole block of code inside `_addIons` (that was adding an additional chain to the topology even when no ions were added) and another is the `numWaters` fix to correct for the number of waters.

peastman:
> @peastman so you suggest changing the ionicStrength value before passing it to _addIons?

Exactly.  Does that sound reasonable?

JoaoRodrigues:
Very. I was just so down the rabbit hole yesterday that this was the simplest I could think of!

JoaoRodrigues:
@peastman, I implemented what you suggested (`ionicStrength  = (numTotalWaters / numReplaceableWaters) * ionicStrength`) but I run into what I think is a rounding/floating point problem (see below). Any suggestions?

```
numTotalWaters = 15884
numReplaceableWaters = 8758  # after filtering for lipid proximity
ionicStrength = (numTotalWaters / numReplaceableWaters) * ionicStrength
# ionicStrength changes from 1M to 1.8136560858643527M

#numIons = 286.38742669930247
# vs. using the total number of waters directly in the numIons formula
# numIons = 286.5342960288809
#
# Test Result 
>       self.assertEqual(resCount['NA'], expected_sodium)
E       AssertionError: 286 != 287
```

peastman:
Is the system charged?  Here's the actual code that calculates the number of ions to add:

```python
numIons = (numReplaceableMols - numPositive - numNegative) * ionicStrength / (55.4 * molar)
```

So really the value we want to use for `ionicStrength` is `(totalWaters-abs(charge)) / (replaceableWaters-abs(charge))`.  So maybe we should add an argument to `_addIons()` with the total number of waters and let it scale the ionic strength itself.

JoaoRodrigues:
> So maybe we should add an argument to _addIons() with the total number of waters and let it scale the ionic strength itself.

That's what I did at first :)

peastman:
This is the only thing holding up 7.4.1.  Can we get it finished up fairly soon?

JoaoRodrigues:
I was about to write you. What is your suggestion then about the code? I can clean up the commits so it's more clear what I changed.

peastman:
How about what I described in https://github.com/openmm/openmm/pull/2440#issuecomment-545982792.  Would that work?

JoaoRodrigues:
That's what I did :) See [here](https://github.com/openmm/openmm/pull/2440/files#diff-79aa2f2e44418fd8c23efbab342f57b5R261)

peastman:
Sorry, I must have misunderstood your comment above!  I thought you were saying there was an error in the current version, and since you hadn't made any more commits since then, I was just waiting.

JoaoRodrigues:
No problem. My editor chomped a ton of white space from the files, so diffing is a mess and makes it much harder to see what I changed. In short, this is what I did:

- Added an argument to `_addIons` that specifies to _total_ number of water molecules in the system. I then use this value in the `numIons` formula to calculate the number of ions to add.
- Put all the ion adding code inside an `if numIons > 0` (or so) statement to avoid adding new (empty) chains to the topology when there aren't ions to be added. This avoid further problems down the road in `addMembrane`.
- adapted `TestModeller.py` to check for the number of ions properly.

JoaoRodrigues:
Thanks!

joaomcteixeira:
:clap:

Issue #2432, opened 2019-10-19
addHydrogens should only optimize positions of new atoms.


JoaoRodrigues:
My understanding of `addHydrogens` is that it adds only "missing" atoms, so it should not change anything else in the structure. The code however, when performing the post-addition optimization, restrains only heavy atoms. I think the restraints should include all atoms that were previously in the structure. Since `addHydrogens` already keeps track of the indices of the atoms it adds, it's trivial to check who to restraint later on. 

This change would help people (like me) that might want to make small changes to a structure but want to keep the rest of the structure intact. Would it be a welcome change, or is this a feature and not a "bug"?

jchodera:
[`Modeller.addHydrogens()`](https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/modeller.py#L938-L961) already has the behavior that heavy atom masses are set to zero prior to minimization so that they should not move during minimization. If they _are_ moving, I suppose this is a bug, and we should track down what is going on.

JoaoRodrigues:
Indeed. But I'm talking about hydrogens that _already exist_ in the structure, not only heavy atoms.

jchodera:
Yep, sorry, I originally misread your comment and just updated this above!

JoaoRodrigues:
Thanks for the reply! I saw your updated comment as I was typing :)

I'm in favour of such an option. I think there are some cases where it's helpful and it's sort of an unexpected behavior from the docstring/function name. I will make the PR later this weekend and if it's fine with you and @peastman , you can merge.

peastman:
Sounds reasonable to me.  I'm not sure an option is even needed.  If the model contains existing hydrogens, can we reasonably assume they're at locations that aren't totally absurd?

JoaoRodrigues:
@peastman if there are locations that are absurd, they can also be cleared anyways with a more explicit call to a minimizer, so yes, maybe an option doesn't make that much sense other than add some sort of backwards compatibility.

peastman:
Having said that, I do need to add: I think both behaviors are reasonable, and it's not obvious that either one is necessarily better than the other.  I suspect this may be a case where about 98% of the time it doesn't matter which way it works, and then each of the behaviors is more convenient than the other about 1% of the time.

jchodera:
> Sounds reasonable to me. I'm not sure an option is even needed.

[...]

PR #2443, opened 2019-10-24
Only optimize newly added hydrogen atoms during addHydrogens()

JoaoRodrigues:
As discussed in #2432, I added a check that only optimizes the positions of newly added hydrogen atoms. Everything else in the system is kept fixed. I began by adding an option but ended up removing it following your (@jchodera and @peastman) opinions.

Also created an additional test that checks that the positions are the same.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -120,7 +120,7 @@ steps.
         forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
         system = forcefield.createSystem(pdb.topology, nonbondedMethod=PME,
                 nonbondedCutoff=1*nanometer, constraints=HBonds)
-        integrator = LangevinIntegrator(300*kelvin, 1/picosecond, 0.002*picoseconds)
+        integrator = BAOABLangevinIntegrator(300*kelvin, 1/picosecond, 0.002*picoseconds)
         simulation = Simulation(pdb.topology, system, integrator)
         simulation.context.setPositions(pdb.positions)
         simulation.minimizeEnergy()
@@ -210,10 +210,10 @@ convenient and less error-prone.  We could have equivalently specified
 The units system will be described in more detail later, in Section :ref:`units-and-dimensional-analysis`.
 ::
 
-    integrator = LangevinIntegrator(300*kelvin, 1/picosecond, 0.002*picoseconds)
+    integrator = BAOABLangevinIntegrator(300*kelvin, 1/picosecond, 0.002*picoseconds)
 
 This line creates the integrator to use for advancing the equations of motion.
-It specifies a :class:`LangevinIntegrator`, which performs Langevin dynamics,
+It specifies a :class:`BAOABLangevinIntegrator`, which performs Langevin dynamics,
 and assigns it to a variable called :code:`integrator`\ .  It also specifies
 the values of three parameters that are specific to Langevin dynamics: the
 simulation temperature (300 K), the friction coefficient (1 ps\ :sup:`-1`\ ), and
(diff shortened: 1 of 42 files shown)


wrappers/python/simtk/openmm/app/modeller.py
```diff
@@ -715,6 +715,9 @@ def addHydrogens(self, forcefield=None, pH=7.0, variants=None, platform=None):
         Definitions for standard amino acids and nucleotides are built in.  You can call loadHydrogenDefinitions() to load
         additional definitions for other residue types.
 
+        By default, except when variants is not set, existing hydrogen atoms in the structure are kept and their positions
+        are not optimized.
```
peastman: This sentence is kind of confusing.  `variants` not being set *is* the default.  Also, I don't see where it treats them differently if `variants` is set?

wrappers/python/tests/TestModeller.py
```diff
@@ -593,6 +593,40 @@ def test_addHydrogensPdb3(self):
...
+        # Fetch 'new' positions
+        new_positions = modeller.positions.value_in_unit(nanometers)
+        newH = [atom.index for atom in topology_after.atoms() if atom.element == element.hydrogen]
+        newH_pos = [new_positions[i] for i in newH][:-1*n_deleted]
+
+        # added all Hs back in?
+        self.assertEqual(len(newH), len(oriH) + n_deleted)
+
+        # Are the old ones at the same position?
+        xyz_diff = any([sum(o-n) for o, n in zip(oriH_pos, newH_pos)])
```
peastman: I'd change this to

```python
xyz_diff = any([norm(o-n) > 1e-6 for o, n in zip(oriH_pos, newH_pos)])
```

The positions will have been passed to a Context and back again, so depending on the platform it's possible they'll have been converted to single precision along the way.  Also, I prefer to avoid implicit conversions of other types to booleans, since they're confusing to read and sometimes lead to unexpected results.

wrappers/python/simtk/openmm/app/modeller.py
JoaoRodrigues: Too much passive voice alright... My understanding of the `variants` option is that, e.g., if you have a HIE and you specify HID, that epsilon proton is going to be removed, thus affecting an existing atom.

How about: `Only the positions of newly added hydrogen atoms will be optimized.` ?

wrappers/python/tests/TestModeller.py
JoaoRodrigues: Good point. My first thought was to use the `numpy.isclose` function but I didn't want to import numpy, convert all positions, etc, just for this. I'll use the norm function then and that tolerance factor.

wrappers/python/simtk/openmm/app/modeller.py
peastman: `variants` can affect which hydrogens get added or removed, but it doesn't affect which ones get optimized.  Also, users won't know what "optimized" refers to here.  That's an implementation detail.

How about adding a sentence just before the last paragraph: "In all cases, the positions of existing hydrogens are not modified."

JoaoRodrigues: Sounds good!

JoaoRodrigues:
Made the edit to the docstring. Also refactored the tests to check if both the existing atoms are kept fixed and if the newly added ones are optimized. Easier to catch if something accidentally restraints everything.

peastman:
Github seems to be confused.  It's listing all the other changes that have been added to the main branch recently as new changes in this PR.  Any idea what's going on?

JoaoRodrigues:
You pushed a few commits while I was working on this. I then pulled from master to keep it on track, but I guess it did not make a clean history. I'll make a new PR with a clean history.

JoaoRodrigues:
Interesting... I reset history and GitHub thinks the PR was merged ...

PR #2444, opened 2019-10-24
make non-primitave types pre increment instead of post increment

FrazerClews:
doing this will make the code more efficient as using post increment usually involves keeping a copy of the previous value and extra code


peastman:
Thanks!  This won't actually affect performance.  All these changes are in places that are totally irrelevant as far as speed is concerned.  But as a matter of principle, I agree it's better to use pre-increment rather than post-increment.

PR #2446, opened 2019-10-24
Attempting to debug hang on appveyor

peastman:


JoaoRodrigues:
Thanks for looking into this. I was going to profile the `addMembrane` function to see where it's taking most of the time. Let me know if I can help you out.

peastman:
It's been happening for a while.  I've been meaning to debug it, but just hadn't gotten around to it until now.

JoaoRodrigues:
OK. By the way, you can do `py.test TestModeller.py -k "test_addMembrane"` to run only a specific test from the file. Learned that with all this debugging/testing.

peastman:
Why is it hiding all output from the test?  When I run exactly the same command locally it doesn't.  And I'm not specifying `-b`.  Sigh.

JoaoRodrigues:
The `print()` statements? Besides using `-s`, maybe the last snippet here will help: https://docs.pytest.org/en/latest/capture.html

peastman:
It isn't using pytest right now.  I was having similar problems with that, and adding -s didn't help.  So I switched it to just invoking the module directly (which invokes `unittest.main()`).

peastman:
It looks like it isn't really hanging.  Appveyor is just really really slow.  To embed a protein in a membrane, it starts with a scaled down version of the protein and then scales it up to full size in 50 steps, running a bit of dynamics after each step.  Running the test on my laptop, each of those steps takes about one second.  On Appveyor, each step takes about one minute!  I first thought it might be using the reference platform, but it isn't.  It's using CPU.  On my laptop the platform creates 8 threads and on Appveyor it only creates 2, but that only explains a small part of the difference.

JoaoRodrigues:
Yeah, Appveyor is _really_ slow... Would it help to have a dedicated machine to run these?

peastman:
Setting up another jenkins box would be an option, but that's a lot of trouble in its own way.

jchodera:
Could we use [circle-ci](https://circleci.com/blog/windows-general-availability-announcement/) as a replacement?

peastman:
Setting up another jenkins box would be an option, but that's a lot of trouble in its own way.

PR #2447, opened 2019-10-25
Avoid unnecessary force evaluation when computing kinetic energy

peastman:
This is an optimization to `context.getState(State::Energy)`.  For a leapfrog integrator, we have to first shift the velocities by half a step, which requires computing the forces.  But for non-leapfrog integrators, that isn't necessary.  This change allows it to skip the force evaluation for those integrators.


jchodera:
Nice!

PR #2448, opened 2019-10-25
Restraint existing atoms during addHydrogens

JoaoRodrigues:
Second attempt at #2443 with a cleaner git history.


peastman:
Git does very strange things...

peastman:
Thanks!

JoaoRodrigues:
Forgot to link the original issue: #2432

zhang-ivy:
I can see that this has been merged into master, however it looks like it’s not in the OpenMM 7.4.1  release, as the change isn’t in the 7.4_branch, even though the merge predates the release. @peastman: Do you know if this is intentional or not? If not, could we include this in the next release?

jchodera:
I've queued this for the OpenMM 7.5 project board!

PR #2452, opened 2019-10-28
Allow multiple matching templates if they assign identical parameters

### LINKED ISSUE #2426: There doesn't appear to be a way to solvate CHARMM systems
Hi there,

I'm fairly new to using OpenMM, so this may be an oversight on my part, but I can't seem to solvate a CHARMM system using Modeller.getSolvate in 7.3.1.dev-4a269c0. I'm using the Python interface (installed via conda), but I think this is probably a general issue. 

I've had a good look throught the last full version of the Python API (for version 6.3.0, so slightly behind the version I'm using, but I couldn't find a more up to date version: [http://docs.openmm.org/6.3.0/api-python/namespacesimtk_1_1openmm.html](http://docs.openmm.org/6.3.0/api-python/namespacesimtk_1_1openmm.html)) and I can't see anything obvious.

Modeller.getSolvate requires a ForceField instance to obtain the Van der Waals radii for the atoms in the system. I think this should be able to accept a CharmmParameterSet, but I hit an error as this doesn't have a toSystem method.

This is a toy example I'm working with:

```python
from simtk.openmm import app
from simtk import openmm as mm

# Parameters - the forcefield
psf_path = "/users/travis/psf_file.psf" 
charmm_folder = "/users/travis/.charmm/"
rtf_file = charmm_folder + "top_all36_prot.rtf"
param_file = charmm_folder + "par_all36_prot.prm"
water_stream = charmm_folder + "toppar_water_ions.str"
param_set = app.CharmmParameterSet(rtf_file, param_file, water_stream)

# The system and coordinates.
psf = app.CharmmPsfFile(psf_path)
topology = psf.topology
# I'm building coordinates using openbabel, but they could be loaded from pdb
coordinates = np.array()  # A numpy array containing molecular coordinates
coordinates = unit.quantity.Quantity(coordinate_array.tolist(), unit=unit.angstrom)

modeller = app.Modeller(topology, coordinates)
padding = 7.5 * unit.angstrom
# I think this should work.
modeller.addSolvent(param_set, padding=padding)
```

But I receive an AttributeError:

```AttributeError: 'CharmmParameterSet' object has no attribute 'createSystem'```

This is perhaps as expected, as the typical method of creating CHARMM systems is to initialise a CharmmPsfFile and use CharmmPsfFile.createSystem. 

Are there any workarounds to this problem? For instance, is it possible to obtain a ForceField from a Simulation or System?

Thank you for your time.

### PR DESCRIPTION
**Pull request by peastman**
Fixes #2426.  Patches can lead to two identical templates that both match the same residue.  Allow this as long as the templates really are identical.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -888,7 +888,23 @@ class ForceField(object):
                 template = allMatches[0][0]
                 matches = allMatches[0][1]
             elif len(allMatches) > 1:
-                raise Exception('Multiple matching templates found for residue %d (%s): %s.' % (res.index+1, res.name, ', '.join(match[0].name for match in allMatches)))
+                # We found multiple matches.  This is OK if and only if they assign identical types and parameters to all atoms.
+                t1, m1 = allMatches[0]
+                atoms1 = [t1.atoms[m] for m in m1]
+                allIdentical = True
+                for t2, m2 in allMatches[1:]:
+                    atoms2 = [t2.atoms[m] for m in m2]
+                    if any(a1.type != a2.type or a1.parameters != a2.parameters for a1,a2 in zip(atoms1, atoms2)):
+                        allIdentical = False
+                    # Properly comparing virtual sites really needs a much more complicated analysis.  This simple check
+                    # could easily fail for templates containing vsites, even if they're actually identical.  Since we
+                    # currently have no force fields that include both patches and vsites, I'm not going to worry about it now.
+                    if t1.virtualSites != t2.virtualSites:
+                        allIdentical = False
+                if not allIdentical:
+                    raise Exception('Multiple matching templates found for residue %d (%s): %s.' % (res.index+1, res.name, ', '.join(match[0].name for match in allMatches)))
+                template = allMatches[0][0]
+                matches = allMatches[0][1]
         return [template, matches]
 
     def _buildBondedToAtomList(self, topology):
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -632,6 +632,29 @@ class ForceField(object):
             atom = self.getAtomIndexByName(atom_name)
             self.addExternalBond(atom)
 
+        def areParametersIdentical(self, template2, matchingAtoms, matchingAtoms2):
+            """Get whether this template and another one both assign identical atom types and parameters to all atoms.
+            
+            Parameters
+            ----------
+            template2: _TemplateData
+                the template to compare this one to
+            matchingAtoms: list
+                the indices of atoms in this template that atoms of the residue are matched to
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: wrappers/python/simtk/openmm/app/forcefield.py [hunk 1]
```diff
@@ -888,7 +888,23 @@ def _getResidueTemplateMatches(self, res, bondedToAtom, templateSignatures=None,
...
+                    atoms2 = [t2.atoms[m] for m in m2]
+                    if any(a1.type != a2.type or a1.parameters != a2.parameters for a1,a2 in zip(atoms1, atoms2)):
+                        allIdentical = False
+                    # Properly comparing virtual sites really needs a much more complicated analysis.  This simple check
+                    # could easily fail for templates containing vsites, even if they're actually identical.  Since we
+                    # currently have no force fields that include both patches and vsites, I'm not going to worry about it now.
+                    if t1.virtualSites != t2.virtualSites:
+                        allIdentical = False
+                if not allIdentical:
+                    raise Exception('Multiple matching templates found for residue %d (%s): %s.' % (res.index+1, res.name, ', '.join(match[0].name for match in allMatches)))
```
jchodera: Want to change this to `Multiple distinct matching templates` or `Multiple non-identical matching templates`?

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -888,7 +888,23 @@ def _getResidueTemplateMatches(self, res, bondedToAtom, templateSignatures=None,
                 template = allMatches[0][0]
```
jchodera: Line 869 lists the incorrect type in the docstring:
```
        template : _ForceFieldTemplate
            The matching forcefield residue template, or None if no matches are found.
```
`_ForceFieldTemplate` should be `_TemplateData`.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -888,7 +888,23 @@ def _getResidueTemplateMatches(self, res, bondedToAtom, templateSignatures=None,
...
                 matches = allMatches[0][1]
             elif len(allMatches) > 1:
-                raise Exception('Multiple matching templates found for residue %d (%s): %s.' % (res.index+1, res.name, ', '.join(match[0].name for match in allMatches)))
+                # We found multiple matches.  This is OK if and only if they assign identical types and parameters to all atoms.
+                t1, m1 = allMatches[0]
```
jchodera: I worry that putting the logic for comparing templates here could cause issues if we neglect to update this when updating the information carried in `_TemplateData`? What about moving the method to check if the templates are identical to a method within the [`_TemplateData` class](https://github.com/openmm/openmm/blob/d88ddb3b7ac4bda2d636daaee4b0572bae9f6b25/wrappers/python/simtk/openmm/app/forcefield.py#L587)?

peastman:
Thanks!  I made changes based on the comments.

peastman:
Ok if I merge this?

PR #2277, opened 2019-03-05
Enable docker for GPU builds

swails:
I finally got nvidia-docker2 installed and actually working on my Gentoo
machine (I had to disable their restriction of using GPU devices to
users in the 'video' group, since the nvidia container runtime hook
didn't run as a user with those privileges, and I can't figure out what
user it was running as).

The other GPU-enabled node I added to the Jenkins array runs Ubuntu, and
that one already has nvidia-docker2 installed and running correctly on
it.

PR #2453, opened 2019-10-28
Cherrypick changes for 7.4.1

peastman:
See #2430.


peastman:
@swails the CPU build passes.  But for the CUDA build, all the tests fail with

```
In file included from /opt/cuda/10.1/bin/../targets/x86_64-linux/include/cuda_runtime.h:83,
                 from <command-line>:
/opt/cuda/10.1/bin/../targets/x86_64-linux/include/crt/host_config.h:129:2: error: #error -- unsupported GNU version! gcc versions later than 8 are not supported!
  129 | #error -- unsupported GNU version! gcc versions later than 8 are not supported!
      |  ^~~~~
```

and for the OpenCL build, they all fail with

```
exception: Error initializing context: clGetPlatformIDs (-1001)
```

Has something changed?

swails:
> Has something changed?

Yes -- the master branch now runs the GPU tests inside docker containers.  The target branch here still runs them on the host OS (which is subject to whichever GCC version I have activated at the time -- in this case GCC 9.1).

It would be best to checkout the `devtools` directory from the `master` branch.

peastman:
When did you make the changes?  I can't find the PR.

peastman:
Those are commits, but what PR did they come from?

jchodera:
For the record, it's very, very dangerous that we haven't yet turned on branch protection to avoid this:
image

peastman:
Thanks!  I didn't realize those settings existed.  I *think* I've successfully added a rule that will prevent direct pushes to the master branch.  Can you check and make sure it looks ok?

jchodera:
I'd also check this option
image
otherwise there is still a list of people who can do this.

peastman:
Done, thanks.

peastman:
CUDA passed this time, but the OpenCL build failed to start up.

```
Failed to run image 'swails/openmm-all:latest'. Error: docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"process_linux.go:432: running prestart hook 1 caused \\\"error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: cuda error: no cuda-capable device is detected\\\\n\\\"\"": unknown.
```

swails:
It was #2277

peastman:
No wonder I couldn't find it!  The PR was opened in March but not merged until September, so it was way down the list.  Mystery solved. :)

swails:
I'm nothing if not deliberate :rofl:

peastman:
Any idea why the jenkins OpenCL build can't start?

swails:
run jenkins

swails:
It's because Ubuntu sucks. It sneaks kernel updates into the upgrade process which breaks the currently installed Nvidia drivers. To fix it I have to reboot, kill the desktop environment, reinstall the kernel drivers, then reboot again. 

I try to mark the node offline when I notice a kernel update, but I think it updates itself sometimes, and I don't always notice a kernel update in the list of available updates.

peastman:
Here are a couple of pages that look like they might be helpful.

https://askubuntu.com/questions/678630/how-can-i-avoid-kernel-updates
https://hollysydney.com/article/prevent-ubuntu-upgrading-kernel

peastman:
Ok, changes are merged.  We're ready to try building the release candidate from the head of `7.4_branch`.

PR #2455, opened 2019-10-29
make std::string pass as references, not values

FrazerClews:
there were a number of constructors that didn't have the std::string being passed as a reference which is very expensive, this makes them pass by reference if they don't get chaged


peastman:
Thanks!

PR #2456, opened 2019-10-30
Convert from AppVeyor to Azure Pipelines

Lnaden:
AppVeyor has been painful slow, (e.g. #2446). Azure Pipeline's have been making steady headway at getting better and marketshare for CI (Conda-Forge is now pure Azure). However, the barrier to entry is a bit steep due to how complex you can make the builds.

So in my off time, I decided to see if I can port the existing AppVeyor script to an equivalent Azure script. Thus far mostly successful, but I could use some help finishing it since I don't fully understand compilation.

You can see the current implementation's outputs here: https://dev.azure.com/lnaden/OpenMM/_build/results?buildId=25

Outstanding issues and questions to resolve (I could use help or suggestions here):
- [ ] cclash is dropped relative to the AppVeyor build, I'm not sure if its needed
- [ ] The OpenMM build step is > 10 min, not sure if thats normal or could be sped up.
- [ ] There are several very slow tests:
    * TestReferenceCustomManyParticleForce ~ 100s
    * TestReferenceVariableLangevinIntegrator > 120s
    * TestReferenceRpmd > 120s
- [ ] There are several failing tests
    * TestReferenceCustomNonbondedForce > 180s (timeout)
    * TestCpuPme fails with error code `0xc0000135`
    * Python tests can't find `simtk` module, although it looks like it should

Even with the issues, the builds start up instantly and take ~30 min to complete.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,6 +25,15 @@ steps:
     echo ###vso[task.setvariable variable=VSPATH]%VSPATH%
   displayName: "Find the VS2015 path to call vcvarsall.bat properly in the CMake blocks later"
 
+# Use Python version
+# Use the specified version of Python from the tool cache, optionally adding it to the PATH
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '3.x'
+    addToPath: true
+    architecture: 'x64' # Options: x86, x64 (this argument applies only on Windows agents)
+  displayName: 'Set the Python version to correct the paths for later'
+
 - script: python -m pip install --upgrade pytest numpy cython
   displayName: 'Install Python tools'
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -130,7 +130,7 @@ steps:
     pip list
     python -c 'help("modules")'
     python -m site
-    ls $(python -m site --user-site)
+    ls "$(python -m site --user-site)"
   workingDirectory: '$(Build.SourcesDirectory)/build'
   displayName: "Debug stuff"
 
@@ -139,7 +139,7 @@ steps:
     pip list
     python -c 'help("modules")'
     python -m site
-    ls $(python -m site --user-site)
+    ls "$(python -m site --user-site)"
   workingDirectory: '$(Build.SourcesDirectory)/build'
   displayName: "Debug stuff with vsvarsall to see if its different"
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -115,13 +115,13 @@ steps:
   displayName: "Generate OpenMM NMake files with CMake"
 
 - script: |
-    # call "$(VSPATH)\VC\Auxiliary\Build\vcvarsall.bat" x64
+    call "$(VSPATH)\VC\Auxiliary\Build\vcvarsall.bat" x64
     cmake --build . --target install
   workingDirectory: '$(Build.SourcesDirectory)/build'
   displayName: "Build and Install OpenMM"
 
 - script: |
-    # call "$(VSPATH)\VC\Auxiliary\Build\vcvarsall.bat" x64
+    call "$(VSPATH)\VC\Auxiliary\Build\vcvarsall.bat" x64
     cmake --build . --target PythonInstall
   workingDirectory: '$(Build.SourcesDirectory)/build'
   displayName: "Install OpenMM's Python Components"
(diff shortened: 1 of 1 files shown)


Lnaden:
Since this is meant to be as close to a 1:1 port, there are probably many optimizations or other ways to do things which are better than having to fight the AppVeyor builds.

peastman:
Thanks!  It would be really interesting to get this working and see how it compares to AppVeyor.

> cclash is dropped relative to the AppVeyor build, I'm not sure if its needed

That's an optimization because Visual Studio is so slow.  It caches compiled object files and lets them be reused from one build to the next if the source hasn't changed.  When everything's working it helps a lot.  Since we haven't had a successful AppVeyor build in a while, I don't think it's currently using any cached files.

> The OpenMM build step is > 10 min, not sure if thats normal or could be sped up.

That's probably about normal.  Visual Studio is really slow.

> There are several very slow tests:

Looking at a recent AppVeyor build, the times for those tests are 88 seconds for TestReferenceCustomManyParticleForce, 107 seconds for TestReferenceVariableLangevinIntegrator, and 106 seconds for TestReferenceRpmd.  So in all cases, Azure takes slightly longer than AppVeyor.  Not an encouraging sign.

> TestReferenceCustomNonbondedForce > 180s (timeout)

This also times out on AppVeyor.

> TestCpuPme fails with error code 0xc0000135

According to https://stackoverflow.com/questions/11432940/what-does-error-code-0xc0000135-mean-when-starting-a-net-application that's STATUS_DLL_NOT_FOUND.  Possibly it's failing to link to FFTW?

> Python tests can't find simtk module, although it looks like it should

No idea.

> Even with the issues, the builds start up instantly and take ~30 min to complete.

AppVeyor takes 28 minutes to get to the point where it starts running Python tests, so pretty similar (assuming Azure really did go through the work of building and installing the Python modules).

Lnaden:
One thing I had to do was call `vcvarsall.bat` every time a `script` block had to invoke the anything needing the MSVC compilers since each block is designed to be isolated from one another, so I am a bit concerned about the linking in that context. However, file system changes persist.

>  Possibly it's failing to link to FFTW?

Possibly, I do set `-DFFTW_LIBRARY=$(Build.SourcesDirectory)/fftw/libfftw3f-3.lib -DFFTW_INCLUDES=$(Build.SourcesDirectory)/fftw` which should exist, but I never run adjust the PATH variable globally (`set PATH=C:\fftw;%PATH%`), is that needed to find the FFTW linking?

>  Azure takes slightly longer than AppVeyor. Not an encouraging sign.

I'm not sure what hardware the difference between AppVeyor and Azure build's are, so that might not be a catastrophic sign, but something to look at.

> No idea.

I'll keep debugging that as I can.

> assuming Azure really did go through the work of building and installing the Python modules

It did build them, you can expand each of the build stages and in the `Install OpenMM's Python Components` section, I see the build outputting blocks like:

```
...
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\baseunit.py to baseunit.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\constants.py to constants.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\doctests.py to doctests.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\mymatrix.py to mymatrix.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\prefix.py to prefix.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\quantity.py to quantity.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\standard_dimensions.py to standard_dimensions.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\unit.py to unit.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\unit_definitions.py to unit_definitions.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\unit_math.py to unit_math.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\unit_operators.py to unit_operators.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\unit\__init__.py to __init__.cpython-36.pyc
byte-compiling /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\simtk\__init__.py to __init__.cpython-36.pyc
running install_egg_info
Writing /hostedtoolcache\windows\Python\3.6.8\x64\Lib\site-packages\OpenMM-7.5.0-py3.6.egg-info
[100%] Built target PythonInstall
```

So I am pretty sure that is all being run correctly, I just can't find them in the next section despite the Python binary coming from the same environment/location of `c:\hostedtoolcache\windows\python\3.6.8\x64\python.exe` Something to explore.

Lnaden:
There is one discrepancy I forgot to mention, the comments for the AppVeyor file say: 

> Setup shell for VS2015, x64

But then use the `os: Visual Studio 2019` so I don't think it was compiling against VS2015 like the comment said. Is that intended or should we try to force the VS2015 variables through the `-vsvars_ver=14.0` flag of the `vsvarsall.bat` calls?

peastman:
> but I never run adjust the PATH variable globally (set PATH=C:\fftw;%PATH%), is that needed to find the FFTW linking?

Probably.  The CMake variables only affect build time.  I don't think they do anything for linking at run time.

> so I don't think it was compiling against VS2015 like the comment said

Sorry, just a mistake on my part.  I just updated it to 2019 a couple of weeks ago (in #2428), and I didn't notice the comment.

Lnaden:
Superseded by #2534

PR #2457, opened 2019-10-30
Apply timeout to Python tests in AppVeyor

peastman:
This is an attempt at improving the behavior on AppVeyor.  This should make the slow test time out after three minutes, instead of hanging the whole build.


peastman:
I'm not sure that's an improvement.  It does stop after three minutes, but then the whole test process exists and it doesn't run any more tests.  That isn't what happens on my computer.  A difference with Windows perhaps?  At https://pypi.org/project/pytest-timeout/ it says, "Note that while by default on POSIX systems py.test will continue to execute the tests after a test has timed, out this is not always possible."

peastman:
I'm closing this since it didn't help.

PR #2466, opened 2019-11-11
Get and scale velocities using numpy during simulated tempering

ljmartin:
This improves the speed of _attemptTemperatureChange during simulated tempering, for those using numpy. List comprehension is required for those not using numpy, hence the if/else.


peastman:
Looks great, thanks!

PR #2475, opened 2019-11-18
ThreeParticleAverageSite Doxygen fix #2473

### LINKED ISSUE #2473: Minor: ThreeParticleAverageSite weight3 Doxygen fix
I think [this](https://github.com/openmm/openmm/blob/f9106ddb6a0d72c960c73b23ed0e07cd7190a80f/openmmapi/include/openmm/VirtualSite.h#L115) parameter should be weight3.

mjw99:


peastman:
Thanks!

Issue #2474, opened 2019-11-18
Optimisations to context creation


tristanic:
With all the recent improvements to simulation preparation in OpenMM 7.4.0, the time for starting a new simulation in ISOLDE is down to 1-3 seconds in typical use cases (~200 residues, implicit solvent, MDFF, CMAP, and a half-dozen custom forces). Which is fantastic, and greatly improves usability for an interactive environment that involves constantly stopping and starting lots of small local simulations.

I'm still curious about what more can be done here, though.  Profiling startup of a fairly typical simulation gives me the following:

```
         334589 function calls (330304 primitive calls) in 2.173 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    1.569    1.569    1.569    1.569 {built-in method _openmm.new_Context}
        1    0.039    0.039    0.087    0.087 forcefield.py:2030(createForce)
        1    0.030    0.030    0.237    0.237 forcefield.py:1026(createSystem)
        1    0.029    0.029    0.033    0.033 forcefield.py:1925(createForce)
```
... showing that preparing the kernels and finalising the `Context` is now the rate-limiting step. It's also surprisingly variable - starting the same simulation multiple times over shows the time required for `new_Context()` varying by about +/- 0.5 seconds. I'm wondering if there are any possibilities for easy gains here? Some thoughts:

- could kernel preparation be parallelised over the available CPU cores?
- Is there scope to cache the Lepton `CompiledExpression`(s) for a given custom force so they only need to be re-computed if the energy function changes? For example, could the Python session maintain a dict of {energy function hash : Lepton object} so each function only needs to be parsed at first use? 

A case in point for the second one: for MDFF I have two different `CustomCompoundBondForce` implementations. The first just uses `Continuous3DFunction` for the map interpolation. The second is designed for large maps where the `Continuous3DFunction` array size gets too big for the GPU, and instead performs the cubic interpolation on-the-fly - so all that's stored on the GPU is a `Discrete3DFunction` containing the map grid points. Obviously makes for a much more complex energy function (although surprisingly the performance difference is barely noticeable once the simulation is running). But the difference in complexity shows in simulation startup time: if I force the use of this lower-memory-usage implementation for the same test case as timed above, then the `new_Context()` call time increases to about 4.5 seconds.

No real pressure - all these timings are already great for almost all tasks.

peastman:
[...]

PR #2476, opened 2019-11-19
Optimized processing of expressions with tabulated functions

peastman:
See #2474.  On the reference and CPU platforms, this can make context creation much faster.


peastman:
run jenkins

Issue #2471, opened 2019-11-15
Unify code between OpenCL and CUDA


peastman:
This is an idea I've been thinking about recently.  It isn't necessarily something we want to do right away.  I just want to record my thoughts and see what other people have to say about it.

Large parts of the CUDA and OpenCL platforms are nearly identical.  Converting a feature from one to the other is largely a matter of duplicating the code, then doing a series of find/replace translations.  This certainly isn't ideal.  Having duplicated code means changes have to be made in two places, and it creates a risk that a bug fixed in one won't get fixed in the other.  It also increases the work needed to implement new features, since you need to create two similar but not identical versions of it.

In addition, there are lots of other compute APIs on the horizon.  AMD is pushing ROCm.  Apple wants everyone to use Metal.  Supposedly Intel is working on yet something else for their upcoming discrete GPUs, though I haven't seen any details.  And who knows what else might come along in the future?  Currently I have no plans to add support for any of these, but it's possible we might want to some day, and we'd be much more likely to do it if it didn't require so much work.

I think there's a lot of room to reduce the amount of duplication.  For example, OpenCLArray and CudaArray could share an abstract parent class that defines a common API.  Likewise for OpenCLContext and CudaContext.  The code used to define kernel arguments is different in OpenCL and CUDA, but it would be easy to create an abstract ComputeKernel class that provides a unified API for both of them.  This would allow a large fraction of the C++ code to be shared between the two platforms, written in terms of abstract superclasses.

It might even be possible to share kernel code between the platforms.  CUDA and OpenCL code are really very similar.  A lot of the differences are minor matters of syntax.  We could define a common set of functions and macros that would cover most of them.  For example, `KERNEL` would be defined as `__kernel` on OpenCL and as `extern "C" __global__` on CUDA.  `GLOBAL_SIZE` would be defined as `get_global_size(0)` on OpenCL and as `(blockDim.x*gridDim.x)` on CUDA.  This would make it possible to write a single kernel that would compile on both.

Clearly it would be a huge task to convert everything to work this way.  But in the long run, it might still be a good idea.

jchodera:
This is an excellent and forward-looking idea, but this problem can't be unique to OpenMM. How are other codes that need to support a variety of GPU (or other hardware accelerated) languages?

Are libraries like XLA advanced enough to allow the expressivity you need while also supporting the variety of backends you envision? These libraries have a great deal of additional developer time supporting them, which could be advantageous in getting access to new backends for free.

[...]

PR #2488, opened 2019-11-27
Common compute framework to unify CUDA and OpenCL code

peastman:
This is the beginning of an implementation of #2471.  So far I've only implemented one class with the common API, CMMotionRemover, but that should give a sense of how this will work.  Instead of having separate OpenCLRemoveCMMotionKernel and CudaRemoveCMMotionKernel classes, there's a single CommonRemoveCMMotionKernel class used by both platforms.  Just based on whether a CudaContext or OpenCLContext gets passed to its constructor, it gets implemented with the correct one.

I'm trying to follow these principles:

1. Use of this framework should always be optional.  If you want to write separate OpenCL and CUDA implementations of a class, that's fine.
2. It should preserve source compatibility with existing plugins.
3. The common API doesn't need to expose every last feature of the OpenCL and CUDA platforms.  I want to keep it relatively simple while still allowing about 90% of forces and integrators to be implemented with it.

By the way, I've learned a lot about C++ while implementing this.  I knew in an abstract sense that templates could get really complicated, but I never before dreamed of writing a method with the signature

```c++
template <class T>
typename std::enable_if<std::is_fundamental<T>::value, void>::type addArg(T& value)
```


peastman:
My latest commit finally passed a milestone: this PR now deletes more lines of code than it adds.  Lots more to still to do!

peastman:
Some of the travis builds are failing, because it's using an archaic version of gcc (4.8.4) that doesn't have full C++11 support.  I could fix that in a few ways: by telling it to install a newer version, by switching to a newer build environment (the next oldest, Xenial, has gcc 5.4, which includes full C++11 support), or by switching to clang (Trusty has clang 3.5, which also has full C++11 support).  See https://docs.travis-ci.com/user/languages/cpp/.

Any preference between these?

andysim:
I'm painfully aware of the incompleteness of GCC 4.8, given that we're running CentOS 7 on many of our systems and that's the default compiler.  When compiling with Intel's icpc, the underlying GCC is still used on the backend so brace yourself for a number of error reports including the message `error: namespace "std" has no member "is_trivially_copyable"`.  Thankfully the fix is simple, the user just needs to add `-gxx-name=/path/to/a/non/ancient/gcc/compiler/bin/g++` to the `CXXFLAGS`, as well as prepending `/path/to/a/non/ancient/gcc/compiler/lib64` to `LD_LIBRARY_PATH` before running.

In terms of Travis, I don't have any strong opinion but I think that GCC compatibility is important to guarantee, even though clang is faster and produces faster code.  When we used Travis for Psi4, we often found timeouts when installing dependencies via conda / apt-get, so if it's possible to spin up an environment with a compatible GCC baked in, that's probably the way to go.

andysim:
I simplified the test [patch is here](https://github.com/openmm/openmm/files/4010246/patch.txt) and it looks like the center of mass motion is indeed nonzero on Cuda, but zero on Reference.

peastman:
You're completely right—it was a problem in CMMotionRemover.  Thanks!  It would have taken me a very long time to track that down.

For travis, I think it will be simplest to just update the build environment to xenial.  That has a gcc version that's still pretty old, but new enough to support the full C++11 spec.

jchodera:
I know OpenMM isn't a Python-only project, but we've been trying to standardize the computational molecular sciences community about the [MolSSI Cookiecutter](https://github.com/MolSSI/cookiecutter-cms) originally developed by @lnaden and now extensively improved by [MolSSI](http://molssi.org). Hundreds of open source CMS projects use this structure now, so it is very familiar to many developers and makes it easier for others to help contribute time to help maintain OpenMM's CI infrastructure.

peastman:
This is getting close to being complete.  I still have a few things to fix, and there's more code I might merge in future PRs, like the RPMD plugin.  According to Github, there are 18,908 lines added and 32,017 lines deleted in 210 files, so looking over everything might take a little while. :)  But if you want an overview of the changes, take a look at the new section in the developer guide.  Search for developer.rst in the list of changes.

peastman:
run jenkins

peastman:
@swails any idea why jenkins isn't running?

peastman:
All tests are passing on Travis and AppVeyor (though AppVeyor times out, as usual).  Jenkins isn't running, but I ran all the CUDA and OpenCL tests locally.  So unless someone objects, I'm going to go ahead and merge this.

andysim:
-13,000 LoC!  I've always had issues with OpenCL tests failing on my MacOS laptop, so I'm curious to see if these changes help.  I think this one is ready to go, and a huge improvement, regardless of the outcome of those tests.  Excellent idea and execution 😃

swails:
run jenkins

swails:
I triggered it by hand. I moved Jenkins to a Linode to improve reliability and availability. We may have to fix some settings. 

It will still use my machines as slaves, but won’t rely on one of them to serve Jenkins.

peastman:
run jenkins

peastman:
Thanks!  And it caught a bug that was missed in my local testing.  Could you trigger it again?

peastman:
And it is merged!  Hopefully I didn't break too many things in the process.  Keep an eye out for anything that isn't working.

tristanic:
Holy cow. @peastman, you're a machine!

PR #2490, opened 2019-12-02
Signature modification to Modeller.addExtraParticles()

mcs95:
Within Modeller.addExtraParticles(), add the ability to pass down
an option to modify compiled.matchResidueToTemplate() 's
ignoreExternalBonds parameter.

https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=11264&p=0&start=0&view=&sid=e384beeb08f3bfd767dcfb546006ad63

### DIFF
--- a/wrappers/python/simtk/openmm/app/modeller.py
+++ b/wrappers/python/simtk/openmm/app/modeller.py
@@ -1004,7 +1004,7 @@ class Modeller(object):
         del context
         return actualVariants
 
-    def addExtraParticles(self, forcefield):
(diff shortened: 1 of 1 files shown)


peastman:
Could you add a description of the argument to the docstring?

```
ignoreExternalBonds : boolean=False
    If true, ignore external bonds when matching residues to templates.  This is
    useful when the Topology represents one piece of a larger molecule, so chains are
    not terminated properly.
```

Otherwise this looks good.

peastman:
Looks great!

PR #2497, opened 2019-12-06
Fixed a potential bug in charmmpsffile parser when building 1-3 and 1-4 exclusions

z-gong:
Currently, the 1-3 and 1-4 exclusions are built from angle_list and dihedral_list. This is correct for most force fields for bio simulations. However, in some force fields (especially coarse-grained FFs and FFs for solid materials), some angles and dihedrals are intentionally ignored. Therefore, the 1-3 and 1-4 exclusion list should be built from bonds.


peastman:
Could you add a test case to verify this is working correctly?  The simplest way is to create some relatively small system (a few hundred atoms or less) that requires this change to work correctly, then compute the energy and verify it matches the value computed by CHARMM.  You can add the files in wrappers/python/tests/systems, and put the test in TestCharmmFiles.py.  You can mostly just copy this test, making minor changes as appropriate: https://github.com/openmm/openmm/blob/master/wrappers/python/tests/TestCharmmFiles.py#L123-L141.

wrappers/python/simtk/openmm/app/charmmpsffile.py
```diff
@@ -466,6 +467,37 @@ def __init__(self, psf_name, periodicBoxVectors=None, unitCellDimensions=None):
...
+        for bond in self.bond_list:
+            a2, a3 = bond.atom1, bond.atom2
+            for a1 in a2.bond_partners:
+                pair = (min(a1.idx, a3.idx), max(a1.idx, a3.idx),)
+                if a1 != a3:
+                    pair_13_set.add(pair)
+            for a4 in a3.bond_partners:
+                pair = (min(a2.idx, a4.idx), max(a2.idx, a4.idx),)
+                if a2 != a4:
+                    pair_13_set.add(pair)
```
peastman: Is this loop actually needed?  Shouldn't the loop above already have found all triples?

z-gong: Yes, it is needed. For a five-member ring, there are five bonds: (1-2, 2-3, 3-4, 4-5, 1-5). If it only considers a1-a2-a3 but not a2-a3-a4, then the pair_13_set will be (5-1-2, 1-2-3, 2-3-4, 3-4-5, 2-1-5), so 4-5-1 is missed.

peastman: Ok, just wanted to make sure!

z-gong:
I've added a test for MoS2 monolayer, which contains a lot of bridged and fused rings. Some angles and all dihedrals are intentionally ignored in psf file. Since I don't have access to CHARMM, I calculate the reference energy with NAMD. The NAMD input script is also attached.

peastman:
Looks good.  Thank you!

btywoniuk:
@z-gong  Could you please explain to me how you created .pdb, .psf and .prm for MoS2?? 
I would like to recreate a bigger structure (and hopefully convert it to Gromacs formats).

z-gong:
@btywoniuk I'm using fftool (https://github.com/agiliopadua/fftool) to build the input files for MoS2. It can export topology and forcefield in GROMACS and LAMMPS format. I made some modifications so that it can export NAMD files also. The fftool is quite easy to use (see README and examples). Essentially, you need to prepare an XYZ file for MoS2, which can be generated from CIF file with VESTA or whatever tool you like. And you also need to prepare a text file that lists the ff parameters.

PR #2503, opened 2019-12-12
Fix Lepton optimization with -ffast-math

jhenin:
Lepton::ParsedExpression::getConstantValue() used to return NaN to indicate non-constant
expressions. This can break with -ffast-math optimization where NaN comparisons are
not defined.

This patch defines isConstant() to test whether an expression is constant before
calling getConstantValue(). getConstantValue() now throws an exception if called on a
non-constant expression.


peastman:
Thanks!

PR #2507, opened 2019-12-19
Lipid

el446:
Add new lipid systems for Modeller.addMembrane().


peastman:
Looks good.  Thank you!

PR #2511, opened 2020-01-03
Add support to ForceField for handling SMIRNOFF trefoil impropers

### LINKED ISSUE #2509: OpenMM ForceField handling for SMIRNOFF impropers
I'm finishing up [this `openmm-forcefields` PR](https://github.com/choderalab/openmm-forcefields/pull/79) to add both GAFF and SMIRNOFF residue template generators for the OpenMM `ForceField` facility, and having a bit of trouble getting `ForceField` to replicate the [SMIRNOFF improper behavior](https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#impropertorsions) in which all three impropers around a trefoil are added. That is, if we have `i-j(k)-l`, where `j` is the central atom, all three torsions (`j-i-k-l`, `j-k-l-i`, `j-l-i-k`) are added. 

The SMIRNOFF residue template generator creates a unique atom type for each atom of the small molecule and then generates improper terms for each of these impropers, but only one improper seems to be added to the system. I'm trying to make sense of the [`_matchImproper`](https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L1744) code to see if one of the other modes besides `default` may be more appropriate, but it might be the case that I need to add a `smirnoff` mode here. 

Any ideas if there is a trick I could use to exploit one of the existing modes, or any pointers for adding a new mode to ensure all three are added?

jchodera:
This PR adds a new `ordering='smirnoff'` mode to improper torsion handling in `PeriodicTorsionGenerator` to make it possible to get exact agreement with [SMIRNOFF small molecule force fields](https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html), which use trefoil-style torsions to avoid improper torsion ambiguity issues that plague GAFF, AMBER, and CHARMM. 

This feature will be used by the [`SMIRNOFFTemplateGenerator`](https://github.com/choderalab/openmm-forcefields/tree/update-amber#examples-using-smirnofftemplategenerator-to-generate-small-molecule-smirnoff-parameters) that provides support for parameterizing small molecules on the fly with the [`openforcefield` toolkit](http://openforcefield.org).

Fixes #2509

### DIFF
--- a/wrappers/python/tests/TestForceField.py
+++ b/wrappers/python/tests/TestForceField.py
@@ -921,23 +921,6 @@ class TestForceField(unittest.TestCase):
     <Type name="[H]C(=O)[H]$H1#2" element="H" mass="1.007947" class="[H]C(=O)[H]$H1#2"/>
     <Type name="[H]C(=O)[H]$H2#3" element="H" mass="1.007947" class="[H]C(=O)[H]$H2#3"/>
   </AtomTypes>
-  <NonbondedForce coulomb14scale="0.833333" lj14scale="0.5">
-    <UseAttributeFromResidue name="charge"/>
-    <Atom sigma="0.3399669508423535" epsilon="0.359824" class="[H]C(=O)[H]$C1#0"/>
-    <Atom sigma="0.2959921901149463" epsilon="0.87864" class="[H]C(=O)[H]$O1#1"/>
-    <Atom sigma="0.2510552587719476" epsilon="0.06276" class="[H]C(=O)[H]$H1#2"/>
-    <Atom sigma="0.2510552587719476" epsilon="0.06276" class="[H]C(=O)[H]$H2#3"/>
-  </NonbondedForce>
-  <HarmonicBondForce>
-    <Bond class1="[H]C(=O)[H]$C1#0" class2="[H]C(=O)[H]$O1#1" length="0.122768286085" k="475133.08130977117"/>
-    <Bond class1="[H]C(=O)[H]$C1#0" class2="[H]C(=O)[H]$H1#2" length="0.10850661293899999" k="338125.5447433327"/>
-    <Bond class1="[H]C(=O)[H]$C1#0" class2="[H]C(=O)[H]$H2#3" length="0.10850661293899999" k="338125.5447433327"/>
-  </HarmonicBondForce>
-  <HarmonicAngleForce>
-    <Angle class1="[H]C(=O)[H]$O1#1" class2="[H]C(=O)[H]$C1#0" class3="[H]C(=O)[H]$H1#2" angle="2.4141195890059883" k="279.7597371776727"/>
-    <Angle class1="[H]C(=O)[H]$O1#1" class2="[H]C(=O)[H]$C1#0" class3="[H]C(=O)[H]$H2#3" angle="2.4141195890059883" k="279.7597371776727"/>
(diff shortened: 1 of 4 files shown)


peastman:
Looks good.  Do you want to add a test for it?  Perhaps just an alternate version of this one:

https://github.com/openmm/openmm/blob/cb4f20167c2c003d38c098f264e8c9e4852afa19/wrappers/python/tests/TestForceField.py#L880-L908.

jchodera:
Great idea! I've just added a test.

More generally, it would be great to use a [pull request template](https://help.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository) to remind contributors to add tests, update docs, and update the changelog (if we have one). We use one for the [openforcefield toolkit](https://github.com/openforcefield/openforcefield/blob/master/.github/pull_request_template.md) that is minimal but helpful!

jchodera:
@peastman : Am I just unable to find this, or did we neglect to discuss the [`ordering` attribute in `<PeriodicTorsionForce>` in the user docs](http://docs.openmm.org/latest/userguide/application.html#periodictorsionforce)?

jchodera:
@peastman: I've added a draft of docs for the `ordering` section, but you'll likely want to update this further yourself after merging.

jchodera:
Looks like all tests pass except for some stochastic failures. 

I still think we should relax the criteria for those stochastic tests (or increase the amount of sampling done) to reduce the false positive rate, since the high false positive rate makes the travis badge pretty useless and will dissuade users from adopting what appears to be a broken source tree.

jchodera:
@peastman: I kept retriggering all the Travis tests until they passed. This is ready to merge---thanks!

wrappers/python/tests/TestForceField.py
```diff
@@ -907,6 +907,72 @@ def test_ImpropersOrdering(self):
...
+  <HarmonicBondForce>
+    <Bond class1="[H]C(=O)[H]$C1#0" class2="[H]C(=O)[H]$O1#1" length="0.122768286085" k="475133.08130977117"/>
+    <Bond class1="[H]C(=O)[H]$C1#0" class2="[H]C(=O)[H]$H1#2" length="0.10850661293899999" k="338125.5447433327"/>
+    <Bond class1="[H]C(=O)[H]$C1#0" class2="[H]C(=O)[H]$H2#3" length="0.10850661293899999" k="338125.5447433327"/>
+  </HarmonicBondForce>
+  <HarmonicAngleForce>
+    <Angle class1="[H]C(=O)[H]$O1#1" class2="[H]C(=O)[H]$C1#0" class3="[H]C(=O)[H]$H1#2" angle="2.4141195890059883" k="279.7597371776727"/>
+    <Angle class1="[H]C(=O)[H]$O1#1" class2="[H]C(=O)[H]$C1#0" class3="[H]C(=O)[H]$H2#3" angle="2.4141195890059883" k="279.7597371776727"/>
+    <Angle class1="[H]C(=O)[H]$H1#2" class2="[H]C(=O)[H]$C1#0" class3="[H]C(=O)[H]$H2#3" angle="2.2670423843424943" k="330.7820750751862"/>
+  </HarmonicAngleForce>
```
peastman: The three blocks above can all be deleted.  Since the test case only looks at the torsions, we don't need to include other forces.

peastman:
The text in the manual looks fine.  I suggested a change to the test.  Once that's done I think this is ready to merge.

wrappers/python/tests/TestForceField.py
jchodera: Done!

jchodera:
Done! Thanks!

Issue #2513, opened 2020-01-06
Drude Stability


Olllom:
After trying the new `DrudeNoseHooverIntegrator` and also the `DrudeLangevinIntegrator` on a few practical systems, I want to just summarize some stability issues that we encountered in the process for future reference. I currently don't have time to implement this but I think that these things need to be fixed to make Drude easy to use in OpenMM.

Tagging @andysim @jing-huang  in case I misspeak.

### Temperature Reporting
The temperature from the `StateDataReporter` does not distinguish between real atoms and Drude particles so that the reported temperatures are usually ~30% low. Of course, this does not directly affect stability but it makes the monitoring and detection of problems unnecessarily hard.

### Velocity Generation
`Context.setVelocitiesToTemperature` does not distinguish between Drude and real particles. If the initial Drude temperature is around 300 K, equilibrating the system becomes a challenge. This might have been the problem in  https://github.com/openmm/openmm/issues/1968 

### Drude Masses
As mentioned in https://github.com/openmm/openmm/issues/1766, Drude particle masses are initialized to zero if the parameter files don't explicitly specify a nonzero mass. The CHARMM PSF parser should at least have a keyword `drudeMass`. It should also print a warning if a massless Drude particle is encountered and `drudeMass is None`. (Most people use the extended Lagrangian integrators.)

### Sensible Defaults
It would make sense to provide defaults for the Drude collision frequency and temperature in the Drude integrators so people don't have to guess. Also, the hard wall constraint is switched off by default. It could be useful to set the default to 0.025. 

### How to Implement
The implementation of some of these things would formally belong into the `Context` rather than the integrators, but this is a bit complicated due to Drude being a plugin.

Instead, why not implement an empty `AbstractDrudeForce` in the main API, from which only `DrudeForce` derives. The `Context` could then implement something called `hasDrudeForce` and at least raise warnings, when something like `Context.setVelocitiesToTemperature` is called.

The Drude specific code could then go into a central place in the plugin, either a Drude integrator base class or a utility class, like 
```
class DrudeUtils {
    ...
    void setVelocitiesToTemperature(Context& context, double temperature, double drudeTemperature=1);
    void optimizeDrudeParticlePositions(Context& context, tolerance);
    double computeTemperature(Context& context);
    double computeDrudeTemperature(Context& context);
}
```
see also https://github.com/openmm/openmm/issues/905

peastman:
Thanks for the summary.  Here are some thoughts on them.

[...]

PR #2518, opened 2020-01-09
Modify temperature initialization mechanism

andysim:
Per the discussion in #2513, this PR will let the integrators determine how temperatures should be assigned, so that Drude particles have the appropriate relative and center of mass velocities assigned.

TODO:
- [x] Default integrator implementation
- [x] Drude integrator override


jchodera:
That would also be useful for leapfrog integrators where the velocity is lagged by half a timestep! Currently, we are initializing velocities incorrectly for those integrators, since we should be subtracting a half-kick from the velocities after assignment from the Maxwell-Boltzmann distribution.

peastman:
> Currently, we are initializing velocities incorrectly for those integrators

I don't think there's anything incorrect about it.  We're just initializing the distribution at a different point in time (on the half step).  In equilibrium the kinetic energy distribution is independent of time, so it doesn't matter which point you initialize it at.  In practice, of course, your system is almost never in equilibrium when you first initialize it, so this is just picking a reasonable starting point for equilibration.

andysim:
I was going to suggest the exact same thing!  Will make that change now

andysim:
I also want to consolidate the tests into a single header file, like the main code.  Unless there's a reason why they were written out explicitly for each platform?

peastman:
> Unless there's a reason why they were written out explicitly for each platform?

No, I just never got around to consolidating them!  Originally the main library tests were also separate.  Then I consolidated them, but I was tired enough after doing it that I put off the plugins until later.  And then never did it.

andysim:
Great!  I'll make that change as part of this PR

jchodera:
> I don't think there's anything incorrect about it. We're just initializing the distribution at a different point in time (on the half step). In equilibrium the kinetic energy distribution is independent of time, so it doesn't matter which point you initialize it at. In practice, of course, your system is almost never in equilibrium when you first initialize it, so this is just picking a reasonable starting point for equilibration.

If your claim were true, then the distribution would also be correct if we randomized the velocity _every_ timestep. This is simple to check for a harmonic oscillator. Check out this [Jupyter notebook on google colab](https://colab.research.google.com/drive/1vwsHkNugH8Mp8uNFkeR9GtsYNS7FbYpH).

Here's the result of redrawing the velocity every timestep for leapfrog (velocities are lagged by half a timestep) and velocity Verlet. I'm looking at the position distribution to remove any concern about which estimator is correct for kinetic energies or instantaneous temperatures:
image
While the effect is likely mininized by only doing this infrequently, that doesn't change the fact that it's still wrong.

The velocity in these integrators is just an additional intermediate variable, and the intermediate variables at the beginning of the timestep for velocity Verlet and leapfrog Verlet are different and related by a force-dependent quantity, `a*dt/2`. It's still possible to initialize the velocities for leapfrog Verlet correctly, but it involves correcting the velocities by `-a*dt/2`.

andysim:
Ok, here's a quick summary of the status of this PR.  The temperature initialization is now delegated to the integrators and is tested for every integrator.  I added a `DrudeIntegrator` class that's sandwiched between `Integrator` and the Drude SCF and Langevin integrators to reduce code duplication (the `DrudeNoseHoover` is derived from `NoseHoover` instead, but the temperature initialization code lives in a different location that's common to all Drude integrators).

As part of the process I moved all the Drude tests to headers common to all platforms, which makes it much easier to tweak tests and makes this a negative LoC change, even with all the comments added.  Yay!  I don't think the Drude SCF and Langevin integrators or the DrudeForce were being tested in single/mixed precision before.  I had to loosen the finite difference tests in `DrudeForce` for the single/mixed tests, which seems reasonable to me.  There's a [test in the SCF integrator](https://github.com/openmm/openmm/pull/2518/files#diff-36cfe84c93ae2b0bd4aae3848e2fc39aR121) that looks at forces on Drude particles and that varies quite significantly between single and double precision.  Without knowing exactly why this discrepancy is so large, I'm not comfortable with just lowering the tolerance.  Apart from that, the tests seem to be in good shape, so I think this one's ready for review:

```
      Start 400: TestReferenceDrudeForce
 1/31 Test #400: TestReferenceDrudeForce ...................   Passed    0.03 sec
      Start 401: TestReferenceDrudeLangevinIntegrator
 2/31 Test #401: TestReferenceDrudeLangevinIntegrator ......   Passed   83.68 sec
      Start 402: TestReferenceDrudeNoseHoover
 3/31 Test #402: TestReferenceDrudeNoseHoover ..............   Passed   50.74 sec
      Start 403: TestReferenceDrudeSCFIntegrator
 4/31 Test #403: TestReferenceDrudeSCFIntegrator ...........   Passed  137.15 sec
      Start 404: TestOpenCLDrudeForceSingle
 5/31 Test #404: TestOpenCLDrudeForceSingle ................   Passed    2.13 sec
      Start 405: TestOpenCLDrudeForceMixed
 6/31 Test #405: TestOpenCLDrudeForceMixed .................   Passed    2.84 sec
      Start 406: TestOpenCLDrudeForceDouble
 7/31 Test #406: TestOpenCLDrudeForceDouble ................   Passed    0.55 sec
      Start 407: TestOpenCLDrudeLangevinIntegratorSingle
 8/31 Test #407: TestOpenCLDrudeLangevinIntegratorSingle ...   Passed   29.16 sec
      Start 408: TestOpenCLDrudeLangevinIntegratorMixed
 9/31 Test #408: TestOpenCLDrudeLangevinIntegratorMixed ....   Passed   29.99 sec
      Start 409: TestOpenCLDrudeLangevinIntegratorDouble
10/31 Test #409: TestOpenCLDrudeLangevinIntegratorDouble ...   Passed   36.39 sec
      Start 410: TestOpenCLDrudeNoseHooverSingle
11/31 Test #410: TestOpenCLDrudeNoseHooverSingle ...........   Passed   23.53 sec
      Start 411: TestOpenCLDrudeNoseHooverMixed
12/31 Test #411: TestOpenCLDrudeNoseHooverMixed ............   Passed   24.65 sec
      Start 412: TestOpenCLDrudeNoseHooverDouble
13/31 Test #412: TestOpenCLDrudeNoseHooverDouble ...........   Passed   24.99 sec
      Start 413: TestOpenCLDrudeSCFIntegratorSingle
14/31 Test #413: TestOpenCLDrudeSCFIntegratorSingle ........***Failed    0.36 sec
      Start 414: TestOpenCLDrudeSCFIntegratorMixed
15/31 Test #414: TestOpenCLDrudeSCFIntegratorMixed .........***Failed    0.44 sec
      Start 415: TestOpenCLDrudeSCFIntegratorDouble
16/31 Test #415: TestOpenCLDrudeSCFIntegratorDouble ........   Passed   19.69 sec
      Start 416: TestCudaDrudeForceSingle
17/31 Test #416: TestCudaDrudeForceSingle ..................   Passed    0.43 sec
      Start 417: TestCudaDrudeForceMixed
18/31 Test #417: TestCudaDrudeForceMixed ...................   Passed    0.42 sec
      Start 418: TestCudaDrudeForceDouble
19/31 Test #418: TestCudaDrudeForceDouble ..................   Passed    0.42 sec
      Start 419: TestCudaDrudeLangevinIntegratorSingle
20/31 Test #419: TestCudaDrudeLangevinIntegratorSingle .....   Passed   24.67 sec
      Start 420: TestCudaDrudeLangevinIntegratorMixed
21/31 Test #420: TestCudaDrudeLangevinIntegratorMixed ......   Passed   25.75 sec
      Start 421: TestCudaDrudeLangevinIntegratorDouble
22/31 Test #421: TestCudaDrudeLangevinIntegratorDouble .....   Passed   25.51 sec
      Start 422: TestCudaDrudeNoseHooverSingle
23/31 Test #422: TestCudaDrudeNoseHooverSingle .............   Passed   23.40 sec
      Start 423: TestCudaDrudeNoseHooverMixed
24/31 Test #423: TestCudaDrudeNoseHooverMixed ..............   Passed   24.08 sec
      Start 424: TestCudaDrudeNoseHooverDouble
25/31 Test #424: TestCudaDrudeNoseHooverDouble .............   Passed   23.97 sec
      Start 425: TestCudaDrudeSCFIntegratorSingle
26/31 Test #425: TestCudaDrudeSCFIntegratorSingle ..........***Failed    0.50 sec
      Start 426: TestCudaDrudeSCFIntegratorMixed
27/31 Test #426: TestCudaDrudeSCFIntegratorMixed ...........***Failed    0.41 sec
      Start 427: TestCudaDrudeSCFIntegratorDouble
28/31 Test #427: TestCudaDrudeSCFIntegratorDouble ..........   Passed   19.83 sec
      Start 428: TestSerializeDrudeForce
29/31 Test #428: TestSerializeDrudeForce ...................   Passed    0.01 sec
      Start 429: TestSerializeDrudeLangevinIntegrator
30/31 Test #429: TestSerializeDrudeLangevinIntegrator ......   Passed    0.01 sec
      Start 430: TestSerializeDrudeNoseHooverIntegrator
31/31 Test #430: TestSerializeDrudeNoseHooverIntegrator ....   Passed    0.01 sec

87% tests passed, 4 tests failed out of 31

Total Test time (real) = 635.82 sec

The following tests FAILED:
        413 - TestOpenCLDrudeSCFIntegratorSingle (Failed)
        414 - TestOpenCLDrudeSCFIntegratorMixed (Failed)
        425 - TestCudaDrudeSCFIntegratorSingle (Failed)
        426 - TestCudaDrudeSCFIntegratorMixed (Failed)
```

andysim:
Regarding @jchodera's comments, I think the way to handle the velocity offsets is to add a public member to `Integrator` specifying the offset to be applied.  That way it could be correctly set for the many Python-based custom integrators in OpenMMTools.  Based on the example posted above I think it's a good idea to do this, but I think a separate PR makes more sense - this one already got a little out of hand with all the changes to the tests!

peastman:
Also regarding velocity offsets, if you haven't been following the discussion at #2532, you might want to take a look.  We might be switching from BAOAB to the leapfrog version of the same splitting, since it turns out that gives much more accurate velocities.  Also see http://dx.doi.org/10.1021/acs.jpca.9b02771, which gives lots of information about using Nose-Hoover with both versions of the splitting.

peastman:
I think the Travis error was a sphinx related one that I've since fixed through an update to the autodoc-doxygen package.   AppVeyor timed out, as it always did.  We've since dropped it and replaced it with Azure.  The Jenkins failures are the only ones I'd be concerned about, especially since they're all in tests related to this change: TestOpenCLDrudeNoseHoover, TestCudaNoseHooverThermostat, and TestCudaDrudeSCFIntegrator.

We still need the change to account for the half step offset in leapfrog integrators, but I can do that in a separate PR after this is merged.

andysim:
Ah, I didn't see the Jenkins log before.  Sorry about that. I'm looking into it now - likely just a tolerance that I need to loosen

andysim:
The rebase fixed the merge conflict and build problem.  Could somebody run jenkins please, so I can see what the error is there?

andysim:
The holdup with this one is that I chose some tolerances for a couple of stochastic tests, based on the K20 and V100 cards we have.  The success rate was high enough on those platforms, but it seems the tolerance I chose is too tight for whatever card is being used on the Jenkins server.  If there's a way I can activate the Jenkins tests on demand, I can probably make the necessary changes in a couple of iterations.  I'll try to find some consumer grade cards to test locally in the meantime.

peastman:
Let me see if I can reproduce the errors locally.  That would be the easiest way.

peastman:
For `TestNoseHooverThermostat.testDimerBox()`, I can make the errors go away by just increasing `nSteps` to 5000.  However, it then sometimes fails two lines later where it checks the energy standard deviation.  Maybe that tolerance just needs to be a bit bigger?

On my laptop I haven't been able to reproduce the hardwall constraint violation.  I'll try a different computer next.

andysim:
Yes, longer runs sound great for that one - I was trying to find a balance between good testing and quick runs.  Loosening the subsequent threshold is fine too; I chose the values based on tests on a K20 card with a limited amount of trials.  The hardwall constraint is baffling to me; I've never seen that fail before.  I can understand that it could possibly cross the boundary before the violation is detected, but it's strange that I've never seen that before.

peastman:
No success reproducing the hardwall violation.  I'm really confused by that!  I don't see anything in the code that looks like it might behave differently on different GPUs.  There's no shared memory or synchronization or anything like that.

andysim:
Adding to the mystery, the previous Jenkins run didn't show a hardwall constraint violation, IIRC

peastman:
Maybe we should just add the fixes to TestNoseHooverThermostat, then go ahead and merge this PR?  Then if the error continues to appear we can debug it in a separate PR.

peastman:
Is it ok if I merge this?  I'll deal with the tests separately.

andysim:
Yes, I think this one is ready to go.  It definitely improves the maintainability of the Drude tests quite a bit, but a little trial-and-error is needed to refine some of the testing tolerances.

Issue #2499, opened 2019-12-08
PowerPC & ARM fixes


jaimergp:
(Comes from https://github.com/conda-forge/openmm-feedstock/pull/16#issuecomment-562988195).

Trying to build OpenMM for PPC & ARM on CF has uncovered some platform code that must be fixed. For details, check the link above. I am just leaving this here as the reference issue for a potential PR.

jchodera:
From @peastman:
> For the PPC intrinsics error, possibly we just need to upgrade to the latest version of cl.hpp. I see it no longer imports the Intel intrinsics. https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl.hpp

Sounds like it would be an easy thing to tackle this now and see if that fixes the `ppc64le` issue?

jaimergp:
I will submit a PR here, but also will try to patch it at the recipe level first to see if that works, because I don't think the CI in this repo tackles `ppc64le`, right?

peastman:
Correct.

peastman:
I believe this has all been fixed, and it works correctly on both PPC and ARM.  If there are still any outstanding issues, feel free to reopen this.

PR #2522, opened 2020-01-14
Update cl.hpp

jaimergp:
Update `cl.hpp` with new version from https://github.com/KhronosGroup/OpenCL-CLHPP/commit/c5063c34675155fffc20185e74ba82a9dfd5dc84

This tries to address https://github.com/openmm/openmm/issues/2499

### DIFF
--- a/platforms/opencl/src/cl.hpp
+++ b/platforms/opencl/src/cl.hpp
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008-2013 The Khronos Group Inc.
+ * Copyright (c) 2008-2015 The Khronos Group Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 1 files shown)


jaimergp:
This change doesn't fix the problems we were seeing with the large differences in forces. Refer to https://github.com/conda-forge/openmm-feedstock/pull/16#issuecomment-574121825.

peastman:
We weren't expecting it to fix that.  This only affects the OpenCL platform, but that's a difference between the Reference and CPU platforms.  This does fix the errors about Intel intrinsics?

jaimergp:
Apparently. The build log doesn't contain any more references to `DNO_WARN_X86_INTRINSICS` after I removed that flag in the `conda-forge` PR.

peastman:
Good enough then.  Thanks!

PR #2523, opened 2020-01-14
testInstallation: raise exception if large differences are observed

jaimergp:
We are seeing large differences in forces in the `ppc64le` build attempts. However, these are not caught by the CI because the exit code for those errors is still `0`. Raising an exception if errors are observed will change that behavior.


jaimergp:
Can we get this in OpenMM 7.5? It would simplify the conda recipes a bit.

peastman:
It was merged back in January, so it will be in 7.5.

PR #2527, opened 2020-01-16
[WIP] Convert Nose Hoover integrator to Common plaform

andysim:
This PR aims to clean up the NHC code by porting it to the new common architecture.  It also fixes a subtle bug in the previous implementation; after reordering the atoms, the forces are not reordered and are also not marked as invalid.  Therefore every 250 steps on Cuda and OpenCL we were using incorrect forces; in a lipid bilayer/water system we observed hot water / frozen bilayer as a result.  For homogeneous systems the bug doesn't seem to have much of an impact.  The fix is easy - simply check to see if reordering has happened and recompute forces if so.  Sorting the forces is another possible approach, but the reordering is infrequent enough that recomputing seems reasonable to me.

TODO:
- [x] Fix bug in forces after reordering
- [x] Debug velocity Verlet integrator on CUDA and OpenCL
- [ ] Debug Nose Hoover machinery on CUDA and OpenCL
- [ ] Remove old platform-specific code

### DIFF
--- a/platforms/common/src/kernels/velocityVerlet.cc
+++ b/platforms/common/src/kernels/velocityVerlet.cc
@@ -1,0 +1,373 @@
+/**
+ * Perform the first step of Velocity Verlet integration.
+ */
+
+KERNEL void integrateVelocityVerletPart1(int numAtoms, int numPairs, int paddedNumAtoms, GLOBAL const mixed2* RESTRICT dt, GLOBAL const real4* RESTRICT posq,
+                                         GLOBAL mixed4* RESTRICT velm, GLOBAL const mm_long* RESTRICT force, GLOBAL mixed4* RESTRICT posDelta,
+                                         GLOBAL const int* RESTRICT atomList, GLOBAL const int2* RESTRICT pairList
+#ifdef USE_MIXED_PRECISION
+                                        ,GLOBAL const real4* RESTRICT posqCorrection
+#endif
+    ){
+    const mixed2 stepSize = dt[0];
+    const mixed dtPos = stepSize.y;
+    const mixed dtVel = 0.5f*(stepSize.x+stepSize.y);
+    const mixed scale = 0.5f * dtVel/(mixed) 0x100000000;
+    int index = GLOBAL_ID;
+    while (index < numAtoms) {
+        int atom = atomList[index];
+        mixed4 velocity = velm[atom];
+        if (velocity.w != 0.0) {
+#ifdef USE_MIXED_PRECISION
+            real4 pos1 = posq[atom];
+            real4 pos2 = posqCorrection[atom];
+            mixed4 pos = make_mixed4(pos1.x+(mixed)pos2.x, pos1.y+(mixed)pos2.y, pos1.z+(mixed)pos2.z, pos1.w);
+#else
+            real4 pos = posq[atom];
+#endif
+            velocity.x += scale*force[atom]*velocity.w;
+            velocity.y += scale*force[atom+paddedNumAtoms]*velocity.w;
+            velocity.z += scale*force[atom+paddedNumAtoms*2]*velocity.w;
+            pos.x = velocity.x*dtPos;
+            pos.y = velocity.y*dtPos;
+            pos.z = velocity.z*dtPos;
+            posDelta[atom] = pos;
+            velm[atom] = velocity;
+        }
+        index += GLOBAL_SIZE;
+    }
+    index = GLOBAL_ID;
+    while (index < numPairs){
+        int atom1 = pairList[index].x;
+        int atom2 = pairList[index].y;
+        mixed4 v1 = velm[atom1];
+        mixed4 v2 = velm[atom2];
+        mixed m1 = v1.w == 0.0f ? 0.0f : 1.0f / v1.w;
+        mixed m2 = v2.w == 0.0f ? 0.0f : 1.0f / v2.w;
+        mixed mass1fract = m1 / (m1 + m2);
+        mixed mass2fract = m2 / (m1 + m2);
+        mixed invRedMass = (m1 * m2 != 0.0f) ? (m1 + m2)/(m1 * m2) : 0.0f;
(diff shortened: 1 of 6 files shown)


peastman:
I think NoseHooverIntegrator also needs the fixes from #2524.  I should have fixed it at the same time, but didn't think about it.  There were two issues.  One is that if `updateContextState()` returned true to indicate forces were invalid, it wasn't updating the flag.  The other is that if you call `getState()` and ask for energy but not forces, it leaves the force buffer containing incorrect values.  My fix was just to set `forcesAreValid = false` whenever the user asks for energies.  That's not ideal, since if the user also asked for forces then they really are valid.  Also, if the forces were previously valid it would be more efficient to just copy them out and restore them.  But all of that would be a lot more complicated, so I went for the simple fix to start with.  I'll try to do a better one in the future.

andysim:
I *think* this one's good to go now.  I'd appreciate a Jenkins pass as a sanity check, but all looks well on a K20 card at least.  I don't fully understand the problem with the force computations at this point, so I'd also appreciate some double checking of 90f0d82.

peastman:
I started a jenkins build.  The force changes look fine.  Though another advantage of switching to the LFMiddle discretization is that all that business of tracking whether the forces are valid becomes unnecessary.

andysim:
I agree with the switch to the "middle scheme" - it's something I mentioned in the original NHC PR.  There are a few things I'll have to play with, like exactly where to impose velocity and position constraints within that scheme, so I think a followup PR is the best way to go.  Now that the GPU codes are consolidated, this should be much easier.  The performance should be a lot better too - the NHC propagation is done on only one thread currently, and this would be done once instead of twice.

peastman:
Agreed, that's a separate PR.  You can look at LangevinMiddleIntegrator to see how we're applying constraints.  The paper I mentioned tests all possible ways of applying constraints, and we're using the one they recommend.

platforms/common/src/kernels/velocityVerlet.cc
```diff
@@ -133,49 +147,52 @@ extern "C" __global__ void integrateVelocityVerletPart2(int numAtoms, mixed2* __
...
+KERNEL void integrateVelocityVerletPart3(int numAtoms, int numPairs, int paddedNumAtoms, GLOBAL mixed2* RESTRICT dt, GLOBAL real4* RESTRICT posq,
+                                         GLOBAL mixed4* RESTRICT velm, GLOBAL const mm_long* RESTRICT force, GLOBAL const mixed4* RESTRICT posDelta,
+                                         GLOBAL const int* RESTRICT atomList, GLOBAL const int2* RESTRICT pairList
+#ifdef USE_MIXED_PRECISION
+                                        ,GLOBAL const real4* RESTRICT posqCorrection
+#endif
+    ){
     mixed2 stepSize = dt[0];
-#if __CUDA_ARCH__ >= 130
+#ifndef SUPPORTS_DOUBLE_PRECISION
```
peastman: I think this should be `#ifdef`, not `#ifndef`?

platforms/common/src/kernels/velocityVerlet.cc
```diff
@@ -133,49 +147,52 @@ extern "C" __global__ void integrateVelocityVerletPart2(int numAtoms, mixed2* __
...
+    while(index < numAtoms) {
         int atom = atomList[index];
         mixed4 velocity = velm[atom];
         if (velocity.w != 0.0) {
             mixed4 deltaXconstrained = posDelta[atom];
             velocity.x += scale*force[atom]*velocity.w + (deltaXconstrained.x - velocity.x*stepSize.y)*oneOverDt;
             velocity.y += scale*force[atom+paddedNumAtoms]*velocity.w + (deltaXconstrained.y - velocity.y*stepSize.y)*oneOverDt;
             velocity.z += scale*force[atom+paddedNumAtoms*2]*velocity.w + (deltaXconstrained.z - velocity.z*stepSize.y)*oneOverDt;
-#if __CUDA_ARCH__ < 130
+#ifdef SUPPORTS_DOUBLE_PRECISION
```
peastman: And this one should be `#ifndef`.

platforms/common/src/kernels/velocityVerlet.cc
```diff
@@ -194,27 +211,33 @@ extern "C" __global__ void integrateVelocityVerletPart3(int numAtoms, int numPai
...
+        relVel.x += relFrc.x * invRedMass;
+        relVel.y += relFrc.y * invRedMass;
+        relVel.z += relFrc.z * invRedMass;
         if (v1.w != 0.0f) {
             mixed4 deltaXconstrained = posDelta[atom1];
             v1.x = comVel.x - relVel.x*mass2fract + (deltaXconstrained.x - v1.x*stepSize.y)*oneOverDt;
             v1.y = comVel.y - relVel.y*mass2fract + (deltaXconstrained.y - v1.y*stepSize.y)*oneOverDt;
             v1.z = comVel.z - relVel.z*mass2fract + (deltaXconstrained.z - v1.z*stepSize.y)*oneOverDt;
-#if __CUDA_ARCH__ < 130
+#ifdef SUPPORTS_DOUBLE_PRECISION
```
peastman: Same issue here.

platforms/common/src/kernels/velocityVerlet.cc
```diff
@@ -226,58 +249,57 @@ extern "C" __global__ void integrateVelocityVerletPart3(int numAtoms, int numPai
             v2.x = comVel.x + relVel.x*mass1fract + (deltaXconstrained.x - v2.x*stepSize.y)*oneOverDt;
             v2.y = comVel.y + relVel.y*mass1fract + (deltaXconstrained.y - v2.y*stepSize.y)*oneOverDt;
             v2.z = comVel.z + relVel.z*mass1fract + (deltaXconstrained.z - v2.z*stepSize.y)*oneOverDt;
-#if __CUDA_ARCH__ < 130
+#ifdef SUPPORTS_DOUBLE_PRECISION
```
peastman: And here.

andysim: Oh wow, I got it wrong in all places 🤦‍♂

PR #2554, opened 2020-02-04
Fixed inconsistency between plaforms in constraints with CustomIntegrator

peastman:
Constraints are applied relative to a base configuration that determines which directions the constraint forces are applied along.  Different platforms were using different base configurations when you executed the following sequence of steps:

1. Update positions
2. Update positions again
3. Apply constraints

The Reference and CPU platforms used the positions at the start of step 1 as the base.  The CUDA and OpenCL platforms used the positions at the start of step 2.  I changed them to behave the same as Reference.


jchodera:
@peastman: Is this PR attached to any issue?

jchodera:
@peastman: Let's try to track which PRs and issues are going to be in 7.4.2 (if we need a bugfix release) or 7.5.0 by using the Projects or Milestones tagging!

peastman:
> @peastman: Is this PR attached to any issue?

No, it was just something I came across.  Neither behavior was really "incorrect", but all platforms should behave consistently.

PR #2561, opened 2020-02-15
Converted BAOABLangevinIntegrator to LangevinMiddleIntegrator

### LINKED ISSUE #2532: Discretizations of Langevin integrator
I've been running more tests of BAOAB.  In #2396 and #2520 I tested it on alanine dipeptide in a box of water, using hbond contraints and rigid water.  I found that it produced excellent potential energies, but the temperature was too low at large step sizes.  I also wanted to see how it worked without constraints, since we might want to use it for AMOEBA.  So I repeated those tests with flexible water and no constraints.  Here are the results.  The columns are step size (in fs), mean potential energy, and mean temperature.

```
0.5 -35605.70052568342 kJ/mol 297.62917866031023 K
1.0 -35609.767674540715 kJ/mol 290.01451980187807 K
1.5 -35640.367635378985 kJ/mol 277.7200127610364 K
2.0 -35623.32820068326 kJ/mol 260.3315873536176 K
2.5 -35591.90934715343 kJ/mol 238.60981627728958 K
```

The potential energies are excellent.  Even at 2.5 fs it still has very little error.  But the kinetic energies are terrible!  Before we handwaved this as, "It has less error in configurational quantities than kinetic ones."  But this isn't just a matter of the error being a bit larger.  The temperatures aren't even in the right range.

But clearly there must be accurate velocities in there somewhere.  You can't get an accurate trajectory (and hence accurate potential energy distribution) otherwise.  That implies the problem must just be in how we're calculating the temperature.

Since openmmtools makes it easy to use any discretization of the integrator you want, I decided to repeat the test with all possible cyclic permutations of `V R O R V`.  Keep in mind that all of these are equivalent.  When running a simulation with any one of them, it cycles through exactly the same sequence of operations.  The only difference is what point in the cycle you interpret as the "break between steps".  Here are the results.

```
V R O R V

0.5 -35638.50715852432 kJ/mol 297.44406396220063 K
1.0 -35611.054894770445 kJ/mol 290.0160817851261 K
1.5 -35625.51474560968 kJ/mol 277.8675158094801 K
2.0 -35664.53782076183 kJ/mol 260.17400740815174 K
2.5 -35575.531359334906 kJ/mol 238.43197405628177 K

V V R O R

0.5 -35616.709575415734 kJ/mol 300.16092415193197 K
1.0 -35632.38941835361 kJ/mol 299.88658398452867 K
1.5 -35642.16039110886 kJ/mol 300.0406027048148 K
2.0 -35636.90849051754 kJ/mol 300.0756093508294 K
2.5 -35573.218970007474 kJ/mol 302.09220953946925 K

R V V R O

0.5 -35703.033396773084 kJ/mol 300.0059357696807 K
1.0 -35962.15707769471 kJ/mol 299.78669020251635 K
1.5 -36249.157901376406 kJ/mol 300.02509772648847 K
2.0 -36771.86778721861 kJ/mol 300.0954166688417 K
2.5 -37408.47504859062 kJ/mol 302.4584345887388 K

O R V V R

0.5 -35708.3051064537 kJ/mol 300.05461671506856 K
1.0 -35937.06525132866 kJ/mol 299.7084628213554 K
1.5 -36285.485792754305 kJ/mol 300.0714472598585 K
2.0 -36750.93079633593 kJ/mol 300.2960435429621 K
2.5 -37395.53765795346 kJ/mol 302.3787520536727 K

R O R V V

0.5 -35658.89274174247 kJ/mol 299.7312231314666 K
1.0 -35602.84066143191 kJ/mol 300.19659045010724 K
1.5 -35631.42077158848 kJ/mol 299.9624030452655 K
2.0 -35629.41289981576 kJ/mol 299.87705344743586 K
2.5 -35620.44282742511 kJ/mol 301.8949595144138 K
```

It's clear that two of these are much better than the others.  VVROR and RORVV both give excellent measures of both kinetic and potential energy.  RVVRO and ORVVR both give accurate temperatures but the potential energy distribution is off.  And VRORV gives good potential energies, but the velocities are way too low.  That doesn't mean the simulation is inaccurate.  Again, all of these lead to identical simulations.  It's just that in between the two V substeps, the "velocity" parameter doesn't actually correspond to the velocities at any point in the physical trajectory.

VVROR happens to be the "LFMiddle" method that Amber uses.  It looks to me like they knew what they were doing when they chose it over VRORV (which they refer to as "VVMiddle").  In fact, I found the following in the Amber manual explaining their choice:

> For any general systems, “LFMiddle” shares the same accuracy as “VVMiddle” for sampling the marginal distribution of coordinates. In addition, “LFMiddle” leads to the exact marginal distribution of momenta in the harmonic limit.

And they cite this paper for it: http://engine.scichina.com/doi/10.1360/N972018-00908 

Given that, I think we should consider switching to that discretization.  The "velocities" you get from VRORV are simply wrong.  They don't correspond to physical velocities at any point in the trajectory.

peastman:
Fixes #2532.


jchodera:
Nooooooooo! Can't we have _both_ a velocity Verlet and leapfrog version of this integrator natively?

PR #2572, opened 2020-02-24
PDBxFile marks heterogens with HETATM

peastman:
Fixes part of https://github.com/pandegroup/pdbfixer/issues/195.


peastman:
Something seems to be wrong with travis.  It's failing to install SWIG.  Since this is a Python only change, and it passed on Azure, I'm going to go ahead and merge it.

PR #2596, opened 2020-03-12
Accept more variants of labels in CHARMM files

### LINKED ISSUE #2502: CharmmParameterSet Parsing
Just hit this bug:
The CharmmParameterSet parser for .prm files expects section titles (like "ANGLES", "BONDS") as full words and plural, while CHARMM parses only the first four letters. 
This means, e.g., that "ANGL" is a valid section title in CHARMM but not in OpenMM.

peastman:
Fixes #2502.  @Olllom could you review this and make sure it looks correct?

### DIFF
--- a/wrappers/python/simtk/openmm/app/charmmparameterset.py
+++ b/wrappers/python/simtk/openmm/app/charmmparameterset.py
@@ -9,10 +9,10 @@ Structures at Stanford, funded under the NIH Roadmap for Medical Research,
 grant U54 GM072970. See https://simtk.org.  This code was originally part of
 the ParmEd program and was ported for use with OpenMM.
 
-Copyright (c) 2014 the Authors
+Copyright (c) 2014-2020 the Authors
 
 Author: Jason M. Swails
-Contributors: Jing Huang
+Contributors: Jing Huang, Peter Eastman
 Date: Sep. 17, 2014
 
 Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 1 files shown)


Olllom:
The change is correct.
Instead of testing for `"IMPHI"` and `"THETA"` you can test for `"IMPH"` and `"THET"`, respectively.

See
 https://www.charmm.org/charmm/documentation/by-version/c40b1/params/doc/parmfile/

I just tested it here:
https://github.com/Olllom/charmm-vs-openmm-energies/commit/079e9151cd58738c4876ef912bc01679791d2bb2
https://github.com/Olllom/charmm-vs-openmm-energies/runs/518615946

peastman:
>  Instead of testing for "IMPHI" and "THETA" you can test for "IMPH" and "THET", respectively.

Are you sure about that?  That isn't what the documentation says.  For example, it lists the option

```
IMPRoper or IMPHI
 atom atom atom atom force_constant periodicity phase
```

with lower case indicating that IMPROPER can be shortened to IMPR, but no optional letters in IMPHI.

Olllom:
Yes, I found some examples online and I also tested it. The link refers to a repo that runs CHARMM in the CI pipeline. Energies are unchanged when replacing the keywords by their four-letter counterparts. By the way, you could also replace the search for NBXMOD by a search for fields that start with NBXM if you want.

peastman:
Ok, thanks.  I'll fix that.

Issue #2593, opened 2020-03-11
Around 2 % overestimation for density under NPT simulation by OpenMM compared with other softwares


z-gong:
Hello, I performed NPT simulation at 300 K and 1 bar with Langevin integrator and MonteCarloBarostat on a simple LJ-126 liquid (Martini coarse-grained water model, vdW interactions are truncated at 1.2 nm without dispersion correction) with the script attached. The density I get is 1.039 g/mL, which is 1.7 % larger than the value produced by GROMACS and LAMMPS (1.022 g/mL). For GROMACS, I'm using Langevin integrator and Paranello-Rahman barostat. For LAMMPS, I'm using Nose-Hoover thermostat and barostat.

I also tried the new Nose-Hoover integrator in the master branch, and the overestimation is the same. I compared the energy calculated on exactly the same coordinates, OpenMM, GROMACS and LAMMPS give identical potential energy. It seems energy calculation and thermostat are correct, so the problem can only come from barostat.

I also tested SPC/E water at 300 K and 1 bar, the density I get is 1.017 g/mL, 2 % larger than literature value (0.998 g/mL), whereas Lammps and GROMACS give the correct result.

Is there any mistake in my script or is this difference expected between MonteCarloBarostat and virial based barostats? 2 % overestimation of density may sound small, but it means there's a huge difference in pressure.

```python
import sys, random
from simtk import openmm as mm, unit
from simtk.openmm import app

system = mm.System()
system.setDefaultPeriodicBoxVectors([4.8,0,0],[0,4.8,0],[0,0,4.8])
for i in range(1000):
    system.addParticle(72.06)

nbforce = mm.NonbondedForce()
nbforce.setNonbondedMethod(mm.NonbondedForce.CutoffPeriodic)
nbforce.setCutoffDistance(1.2)
nbforce.setUseDispersionCorrection(False)
for i in range(1000):
    nbforce.addParticle(0, 0.47, 5.0)
system.addForce(nbforce)

barostat = mm.MonteCarloBarostat(1 * unit.bar, 300)
system.addForce(barostat)

top = app.Topology()
chain = top.addChain()
for i in range(1000):
    residue = top.addResidue('WAT', chain)
    top.addAtom('W'+str(i), None, residue)

integrator = mm.LangevinIntegrator(300, 1, 0.01)  # the suggested step size is 0.03 ps
sim = app.Simulation(top, system, integrator)
sim.reporters.append(app.StateDataReporter('martini.log', 1000, step=True,
[...]

PR #2598, opened 2020-03-16
bug fix for handling constraints and rigidWater in CharmmPsfFile

z-gong:
As mentioned in issue #2593, the angle constraints are not added correctly


wrappers/python/tests/TestCharmmFiles.py
```diff
@@ -361,6 +361,40 @@ def test_Nonbonded_Exclusion(self):
...
+        # the system is made of one water molecule and one methanol molecule
+        hBonds_water = [[0, 1], [1, 2]]
+        hAngles_water = [[0, 2]]
+        hBonds_methanol = [[3, 4], [3, 5], [3, 6], [7, 8]]
+        allBonds_methanol = hBonds_methanol + [[3, 7]]
+        hAngles_methanol = [[4, 5], [4, 6], [5, 6], [3, 8]]
+        system = psf.createSystem(params, constraints=None, rigidWater=False)
+        self.assertEqual(system.getNumConstraints(), 0)
+        system = psf.createSystem(params, constraints=None, rigidWater=True)
+        self.assertCountEqual([system.getConstraintParameters(i)[:2] for i in range(system.getNumConstraints())],
```
peastman: Instead of just checking the number of constraints, how about verifying that the two lists of constraints are the same?  Alternatively you could call `validateConstraints()`: https://github.com/openmm/openmm/blob/master/wrappers/python/tests/validateConstraints.py

z-gong: assertCountEqual means they have the same elements. Indeed the name is very misleading, and it doesn't work under python2. I changed it to assertEqual on the sorted lists

peastman: That is indeed a very misleading name!  Python 2 isn't important.  We've dropped support for it.

peastman:
Looks good.  Is this ready to merge?

z-gong:
Yes it is.

peastman:
Thanks!

PR #2600, opened 2020-03-18
Nosé Hoover Middle scheme

andysim:
My plan is to make a new class modeled on the current NHC integrator, similar to the current Langevin / LangevinMiddle setup.  Here's the plan:

- [x] Clone NH integrator to make new NHM integrator

- [x] Add tests by copying existing NH tests

- [x] Convert reference platform to use the middle scheme

- [x] Convert common platform to use the middle scheme


peastman:
The only reason we have separate LangevinIntegrator and LangevinMiddleIntegrator classes is for backward compatibility.  Since Nose-Hoover will be a new feature in the next release, I don't think there's any benefit to having a non-middle version of it.  It's much less accurate and forces you to take smaller time steps.

andysim:
Fair point.  I currently have both versions around, which will help with debugging, but I'll clean everything up at the end.

andysim:
The original code had a separate kernel for the Nosé-Hoover chain and for the VV integrator, which was to allow mixing and matching of integrators and thermostats.  Given that we're just targeting the "middle" leapfrog scheme, which is what the LangevinMiddleIntegrator, appears to do, I decided I'm going to nuke the standalone Nosé-Hoover kernel.  I'll rename all the 'VelocityVerletIntegrator' code as just 'NoseHooverIntegrator' and make it clear in the docs that the method implemented is "LFMiddle".  Sound reasonable?  I'm also going to consolidate the NoseHooverIntegrator and NoseHooverThermostat tests, to reflect this change in design.

peastman:
That sounds fine.

andysim:
Sorry, I forgot to change the title before to reflect that I think this one's ready for review/merge. 
 The tolerances of the tests *seem* to be appropriate, but may need adjustments.  Once this is in, I'm happy to aid in reconciling the changes with #2651 if you like.  The failing Jenkins tests are all tests that pass on my machine, and are related to the CMAP potential.  I may be able to dig into that later, but I don't think this PR is the culprit.

peastman:
Thanks!  I'll go over it shortly.  I already have an open issue about the jenkins tests (#2648).  They pass for me too, so I don't know what's going on with them.

peastman:
Looks very good!

The one minor issue I see is that some comments and method names claim it's using BAOAB, which it isn't.  BAOAB is the same as VV-Middle.  Or in the other terminology, LF-Middle could be described as BBAOA.  It starts with a double length B step (update the velocities for a full step rather than a half step), and then there's no velocity update (B step) at the end.  So for example, the `ABstep()` method is incorrectly named, since it really only performs a position update.  (It does, however, also apply constraint forces to the velocities.)

peastman:
I'm going to go ahead and merge this, then make the changes myself.  I hope that's ok!

andysim:
Oops, sorry about that - it fell off my radar.  I originally implemented VV middle but changed it to LF middle to be consistent with the Langevin thermostat.  Excellent catch to notice that :)

peastman:
After merging this, I get several test failures on my local computer:

```
$ ./TestReferenceNoseHooverIntegrator 
exception: Assertion failure at TestNoseHooverIntegrator.h:78.   Expected [0.0224696, 0, 0], found [0, 0, 0]
$ ./TestReferenceDrudeNoseHoover 
exception: Assertion failure at TestDrudeNoseHoover.h:305
$ ./TestOpenCLNoseHooverIntegrator mixed
OpenCL internal error: [CL_DEVICE_NOT_AVAILABLE] : OpenCL Error : Error: Build Program driver returned (518)
OpenCL internal error: OpenCL Warning : clBuildProgram failed: could not build program for 0x1021c00 (AMD Radeon R9 M370X Compute Engine) (err:-2)
OpenCL internal error: [CL_BUILD_ERROR] : OpenCL Build Error : Compiler build log:
Error returned by cvms_element_build_from_source

exception: Error compiling kernel: Error returned by cvms_element_build_from_source
$ ./TestOpenCLDrudeNoseHoover mixed
OpenCL internal error: [CL_DEVICE_NOT_AVAILABLE] : OpenCL Error : Error: Build Program driver returned (518)
OpenCL internal error: OpenCL Warning : clBuildProgram failed: could not build program for 0x1021c00 (AMD Radeon R9 M370X Compute Engine) (err:-2)
OpenCL internal error: [CL_BUILD_ERROR] : OpenCL Build Error : Compiler build log:
Error returned by cvms_element_build_from_source

exception: Error compiling kernel: Error returned by cvms_element_build_from_source
```

I'll start by seeing if I can track down the OpenCL compiler errors.

andysim:
How strange.  It's working on the K20 and V100 cards I've tried.  I'll see if I can get access to something with an AMD video card.  The `CL_DEVICE_NOT_AVAILABLE` error is very strange to me

peastman:
I suspect there's a compiler bug.  I'm trying to track down exactly what's triggering it.  I've narrowed it down to `propagateNoseHooverChain()`, but so far haven't been able to pin it to a particular line.

andysim:
Ah, good to know.  Which exact compiler are you using?  I've not been able to reproduce it yet.

peastman:
Found it!  I had to change all instances of `EXP()` to `exp()`.  In single or mixed precision `EXP` is defined as `native_exp`, which only supports single precision.  When using that in integrator code in mixed precision, that led to it being called with a double precision argument.  It was legitimate for the compiler to view that as an error, but it was failing to produce a proper error message.

There also was a second error in that kernel: at line 25 it defines a variable called `aa`, and then defines another one with the same name at line 35.

The reference platform errors were my fault.  I made a mistake in renaming methods.

This is on a Mac, by the way.

andysim:
Excellent news :) My daily driver is a ~2013 macbook pro, but has always had issues with the OpenCL tests so I only run them on our servers.  Now Apple has finally released a 13" version with a physical escape key I might finally update, about 3 years later than I should have, so I should be able to test OpenCL on OSX again.

Sorry for the EXP mixup - I should have realized that the `exp` vs `expf` behavior is associated with the `real` type, not `mixed`.  Your solution is a good one, but doesn't that result in the slower `exp` being called for the single and mixed cases?  If so, would it be worth defining a `MIXED_EXP` or something like that, analogous to the definition of `EXP`, but tied to the `mixed` type instead of `real`?  I haven't profiled since the switch to the middle scheme, but that kernel runs on only 1 thread, so I think it could be important.

The shadowed variable is was a bad mistake on my part :(

peastman:
> doesn't that result in the slower exp being called for the single and mixed cases?

What was the intended behavior in mixed precision?  Did you want that function to be evaluated in single or in double?  If single, then instead of switching to `exp` we could use `EXP` and instead cast the argument to `real`.

I'm not sure about the effect in single precision mode.  Do you want to benchmark it and see?

>  Now Apple has finally released a 13" version with a physical escape key

And not just that.  It also comes with a non-broken keyboard, for the first time in several years!

jchodera:
> Now Apple has finally released a 13" version with a physical escape key I might finally update

Oh my goodness, I am ordering one of these today just for the escape key!

andysim:
Ok, I tried some tests and found a negligible performance hit from using `exp` instead of `EXP` so I think that's a good way to go.  I only did a simple test on a K20 for a 55k atom system and 2fs timestep.  The code gets about half of the throughput of the LangevinIntegrator, pretty much independent of the NHC chain parameters, so I need to do some profiling.  I suspect that having a custom kernel for the common case of all atoms being controlled by the same thermostat will help greatly.  Accessing an atom list lookup for this case could be causing an inefficiency due to the fact that it is wasting time accessing an array that's just an identity mapping.

peastman:
Great.  I'll merge #2670 once the tests pass.  I'll also go ahead and do the cleanup to take advantage of the new checkpointing system.

andysim:
Thanks - that cleanup will be very useful going forwards.  The workaround we came up with took a while to figure out and isn't too aesthetically pleasing.  I think the new system looks a lot cleaner and will greatly aid development going forwards.  Just out of interest have you ever seen memory issues when trying to run `nvprof`?  I see stuff like this
```
(ommdev) :~/programming/openmm/build (master) > nvprof -s ./TestCudaHarmonicAngleForce 
==26171== NVPROF is profiling process 26171, command: ./TestCudaHarmonicAngleForce
  what():  Error deleting array angleParams: CUDA_ERROR_ILLEGAL_ADDRESS (700)
```
while the normal test runs through `cuda-memcheck` just fine, as well as in standalone mode.  It happens for each of the few different tests that I've tried so far.

peastman:
I haven't seen that.  When you see a CUDA error deleting an array, it usually means the error really occurred at some earlier point and just wasn't detected until then.  If the test runs successfully in cuda-memcheck but fails in nvprof, I'm not sure what to make of that.

PR #2605, opened 2020-03-23
Remove unused variables

alexhenrie:
This dead code was identified with the help of [scan-build](https://clang-analyzer.llvm.org/scan-build.html). Every variable that I am proposing to remove is initialized to the result of some function call, making it very unlikely that the compiler's optimizer will automatically eliminate the unnecessary code.


peastman:
Looks good.  Thanks!

Issue #2604, opened 2020-03-23
Question about metadynamics implementation


ajsilveira:
Hi @peastman . I have a question regarding the implementation of well-tempered metadynamics.
When adding a gaussian, the collective variable is normalized. However, 
biasWidth enters the equation https://github.com/openmm/openmm/blob/2e86184b4b720f267d78d6c2902fff908695c135/wrappers/python/simtk/openmm/app/metadynamics.py#L199
without being divided by (maxValue - minValue). Should the user provide a normalized 
standard deviation? Also, I don't understand the role of gridWidth in that equation. 
Could you please provide more details about this? Thanks!

craabreu:
Hi all,
There is another possible issue with `metadynamics.py`. While deposited Gaussians are correctly handled in the case of a periodic `BiasVariable`, nothing is passed to `Continuous1DFunction` objects (or its 2D/3D counterparts) with respect to periodic boundaries.
This could be easily solved for 1D functions because periodic splines are already available here:
https://github.com/openmm/openmm/blob/f9106ddb6a0d72c960c73b23ed0e07cd7190a80f/openmmapi/src/SplineFitter.cpp#L77
However, periodic 2D/3D splines would require further implementation.
@peastman, am I missing something here?

craabreu:
I tested `Metadynamics` by making it deposit a single Gaussian and comparing the result with the exact expression. Please see the script and the graphical comparison below. The employed PDB file is [here](https://github.com/openmm/openmm/blob/master/wrappers/python/tests/systems/alanine-dipeptide-implicit.pdb). Clearly, the deposited Gaussian does not have the specified standard deviation. 

```
from simtk import openmm
from simtk.openmm import app
import matplotlib.pyplot as plt
import numpy as np

npoints = 181
sigma = 0.35
nsteps = 500

pdb = app.PDBFile('alanine-dipeptide-implicit.pdb')
forcefield = app.ForceField('amber99sb.xml', 'tip3p.xml')
system = forcefield.createSystem(pdb.topology, nonbondedMethod=app.NoCutoff)

psi = openmm.CustomTorsionForce('theta')
psi.addTorsion(6, 8, 14, 16)
psi_bias = app.BiasVariable(psi, -np.pi, np.pi, sigma, True, npoints)
meta = app.Metadynamics(system, [psi_bias], 300, 5, 1, nsteps)

integrator = openmm.VerletIntegrator(0.001)
platform = openmm.Platform.getPlatformByName('Reference')
[...]

PR #2611, opened 2020-03-26
[WIP] Changes in Metadynamics

craabreu:
This PR addresses some topics discussed in #2604. The intended tasks are:

- [x] Use the already available `SplineFilter::createPeriodicSpline` method to enable periodicity in `Continuous1DFunction` by means of an optional boolean flag `periodic`, whose default value is `False`.

- [x] Use `Continuous1DFunction` with `periodic=True` in `Metadynamics` whenever it deals with a periodic bias variable in the one-dimensional case. This is required to ensure continuity of the biasing potential derivative at the borders of the variable domain.

- [x] Fix the widths of the Gaussian biasing potentials deposited by `Metadynamics` so that they match the specified values.

- [x] Fix a bug in `Metadynamics` regarding the dimensions of the bias matrices.

- [x] Alter the way `Continuous2DFunction` and `Continuous3DFunction` are used when periodic variables are present. In this case, it would be difficult to implement periodic splines, especially considering that periodicity could occur in some directions only. As a workaround, the domain along the periodic directions is extended, such as proposed [in item 2 here](https://stackoverflow.com/a/25106574/4284505), so that the derivatives at the borders are approximately continuous. The number of expansion points in each periodic direction is the minimum between the original number of grid intervals in such direction and a value specified through a new optional argument `gridExpansion` (default=20).

- [x] Make sure that the new `CustomCVForce` which `Metadynamics` adds to the passed system is the only force in its force group.

- [x] For consistency with the general OpenMM API, allow the specified `minValue`,  `maxValue` and `BiasWidth` to contain units of measurement.


peastman:
Thanks, this is looking really good!

As an alternative to an entirely new ContinuousPeriodic1DFunction, what about just adding an optional boolean flag to Continuous1DFunction to tell it whether to be periodic?  It will be a simpler API, and it will add a lot less code overall.

> In this case, it would be difficult to implement periodic splines, especially considering that periodicity could occur in some directions only.

How important do you think that case is?  Would it be reasonable to say functions are either periodic in all dimensions, or else non-periodic in all dimensions?

craabreu:
Thanks, @peastman 

> As an alternative to an entirely new ContinuousPeriodic1DFunction, what about just adding an optional boolean flag to Continuous1DFunction to tell it whether to be periodic? It will be a simpler API, and it will add a lot less code overall.

I agree. This is a cleaner alternative.

> How important do you think that case is? Would it be reasonable to say functions are either periodic in all dimensions, or else non-periodic in all dimensions?

I think it would be important to keep this option available. For instance, in this recent [paper](https://doi.org/10.1021/acs.jctc.9b01181), they use a dihedral angle (periodic) and a radius of gyration (non-periodic) as collective variables.

craabreu:
Hi @peastman 

I have a question regarding the implementation of an optional boolean flag to `Continuous1DFunction`. It is clear to me that backward compatibility would be preserved for the constructor and for the parameter setter, but how could I preserve it for the getter method below?

https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/openmmapi/include/openmm/TabulatedFunction.h#L92

craabreu:
I managed to preserve backward compatibility for `Continuous1DFunction` by keeping `getFunctionParameters` unchanged and including a new method `getPeriodicityStatus`. I hope this is okay.

There is a bug in the allocation of matrices in `Metadynamics`, which I solved in the temporary `WellTemperedMetadynamics` class. I intend to eventually substitute one class for the other. The bug comes from the fact that a `reversed` function is used in this line:

https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/wrappers/python/simtk/openmm/app/metadynamics.py#L206

So, the order should have also been reverted here:

https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/wrappers/python/simtk/openmm/app/metadynamics.py#L123

and here:

https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/wrappers/python/simtk/openmm/app/metadynamics.py#L124

peastman:
That's about what I was thinking.  Let's call the method `getPeriodic()`, and also add a corresponding `setPeriodic()`.

craabreu:
Hi @peastman 
I devised the script below in order to test the implementation of `Continuous1DFunction` with periodic spline. Differently from what I expected, having `periodic=True` did not make the derivative have the same value at both extremes. Please see the graphical comparison with SciPy, which exhibits the expected behavior. If the test is correct, this problem might affect other classes such as `CMAPTorsionForce` and `AmoebaTorsionTorsionForce`, which also make use of periodic splines. Would you please help me finding out what is going on here?

```py
from scipy import interpolate
from simtk import openmm as mm
from simtk import unit
from simtk.openmm import app
import matplotlib.pyplot as plt
import numpy as np

gridWidth = 21
nPoints = 1001
nGaussians = 100
sigma = 0.1

xgrid = np.linspace(0, 1, num=gridWidth)
ygrid = np.zeros(gridWidth)
np.random.seed(0)
for center in np.random.uniform(size=nGaussians):
    dist = np.abs(xgrid - center)
    dist = np.min(np.array([dist, np.abs(dist-1)]), axis=0)
    ygrid += np.exp(-0.5*(dist/sigma)**2)

fig, ax = plt.subplots(2, 2, sharex=True, sharey=True)

def make_plot(ax0, x, y, dydx, title):
    ax0.set_title(title)
    ax1 = ax0.twinx()
    ax0.plot(xgrid, ygrid, 'bo')
    ax0.plot(x, y, 'b-')
    ax1.plot(x, dydx, 'r-')
    ax1.hlines((dydx[0], dydx[-1]), x[0], x[-1], color='red', ls='--')

for k, periodic in enumerate([False, True]):
    table = mm.Continuous1DFunction(ygrid, 0, 1, periodic=periodic)

    xcoord = mm.CustomExternalForce('x')
    xcoord.addParticle(0)
    cvforce = mm.CustomCVForce('table(xcoord)')
    cvforce.addCollectiveVariable('xcoord', xcoord)
    cvforce.addTabulatedFunction('table', table)

    system = mm.System()
    system.addParticle(1)
    system.addForce(cvforce)

    integrator = mm.CustomIntegrator(0)
    platform = mm.Platform.getPlatformByName('Reference')
    context = mm.Context(system, integrator, platform)

    x = np.linspace(0, 1, num=nPoints)
    y = np.empty(nPoints)
    dydx = np.empty(nPoints)
    for i in range(nPoints):
        context.setPositions([mm.Vec3(x[i], 0, 0)])
        state = context.getState(getEnergy=True, getForces=True)
        y[i] = state.getPotentialEnergy().value_in_unit(unit.kilojoules_per_mole)
        dydx[i] = -state.getForces()[0].x
    make_plot(ax[k, 0], x, y, dydx, f'OpenMM (periodic={periodic})')

    cs = interpolate.CubicSpline(xgrid, ygrid, bc_type='periodic' if periodic else 'natural')
    y = cs(x)
    dydx = cs.derivative()(x)
    make_plot(ax[k, 1], x, y, dydx, f'SciPy (periodic={periodic})')

plt.show()
```
Figure_1

craabreu:
The script below aims at checking the effectiveness of the grid expansion strategy at making the bias-potential derivative approximately continuous at the domain borders. A 1D case was considered for simplicity. It shows that the discontinuity decays exponentially with the number of expansion points.

```py
from scipy import interpolate
from simtk import openmm as mm
from simtk import unit
from simtk.openmm import app
import matplotlib.pyplot as plt
import numpy as np

xcoord = mm.CustomExternalForce('x')
xcoord.addParticle(0)
cvforce = mm.CustomCVForce('table(xcoord)')
cvforce.addCollectiveVariable('xcoord', xcoord)
table = mm.Continuous1DFunction(np.empty(2), 0, 1)
cvforce.addTabulatedFunction('table', table)

system = mm.System()
system.addParticle(1)
system.addForce(cvforce)

integrator = mm.CustomIntegrator(0)
platform = mm.Platform.getPlatformByName('Reference')
context = mm.Context(system, integrator, platform)

np.random.seed(123)
centers = np.random.uniform(size=100)
extraWidths = np.arange(0, 21, 2)
differences = np.empty(len(extraWidths))
for gridWidth in [21, 51, 101, 181]:
    xgrid = np.linspace(0, 1, num=gridWidth)
    ygrid = np.zeros(gridWidth)
    for center in centers:
        dist = np.abs(xgrid - center)
        dist = np.min(np.array([dist, np.abs(dist-1)]), axis=0)
        ygrid += np.exp(-0.5*(dist/0.1)**2)

    for j, extraWidth in enumerate(extraWidths):
        n = extraWidth + 1
        ygrid_expanded = np.hstack((ygrid[-n:-1], ygrid, ygrid[1:n]))
        extraRange = extraWidth/(gridWidth - 1)
        table.setFunctionParameters(ygrid_expanded, -extraRange, 1+extraRange)
        cvforce.updateParametersInContext(context)
        force_value = np.empty(2)
        for i in range(2):
            context.setPositions([mm.Vec3(i, 0, 0)])
            force_value[i] = context.getState(getForces=True).getForces()[0].x
        differences[j] = np.abs(force_value[1] - force_value[0])

    plt.plot(extraWidths, differences, 'o', label=f'gridWidth = {gridWidth}')

plt.xticks(extraWidths)
plt.xlabel('Number of expansion points')
plt.yscale('log')
plt.ylabel('Deviation in the end-point derivatives')
plt.legend()
plt.show()
```

Figure_1

peastman:
I believe the derivative *ought* to be identical at the end points, regardless of the grid size.  Perhaps there's an off by one error somewhere in the indexing or something like that?

It will be easier to debug with a lower level test that only involves the minimal amount of code.  `testPeriodicSpline()` in TestSplineFitter.cpp would be a good place to start.  It goes directly to SplineFitter, so we don't need to worry about the CustomExternalForce or CustomCVForce introducing problems.  Can you reproduce this in a test that calls SplineFitter directly?

openmmapi/include/openmm/TabulatedFunction.h
```diff
@@ -90,6 +91,12 @@ class OPENMM_EXPORT Continuous1DFunction : public TabulatedFunction {
      * @param[out] max            the value of x corresponding to the last element of values
      */
     void getFunctionParameters(std::vector<double>& values, double& min, double& max) const;
+    /**
+     * Get the periodicity status of the tabulated function.
+     *
+     * @param periodic       whether the function is periodic
+     */
+    void getPeriodic(bool& periodic) const;
```
peastman: It can just return the value directly: `bool getPeriodic()`.  Returning values through references is only needed when a function has to return multiple values.

craabreu: Done!

tests/TestSplineFitter.cpp
```diff
@@ -82,6 +82,12 @@ void testPeriodicSpline() {
         ASSERT_EQUAL_TOL(sin((double)i), SplineFitter::evaluateSpline(x, y, deriv, i), 0.05);
         ASSERT_EQUAL_TOL(cos((double)i), SplineFitter::evaluateSplineDerivative(x, y, deriv, i), 0.05);
     }
+    double ya[] = {15.579, 16.235, 17.325, 18.741, 20.454, 22.517, 24.944, 27.554, 29.942, 31.657,
+                   32.486, 32.612, 32.494, 32.532, 32.785, 32.917, 32.402, 30.842, 28.229, 24.989,
+                   21.762, 19.074, 17.147, 15.970, 15.467, 15.579};
+    y.assign(ya, ya+y.size());
+    ASSERT_EQUAL_TOL(SplineFitter::evaluateSplineDerivative(x, y, deriv, x[0]),
+                     SplineFitter::evaluateSplineDerivative(x, y, deriv, x[x.size()-1]), 1e-6);
```
craabreu, line 90: I added this test, which fails. This is the resulting error message:

```
exception: Assertion failure at TestSplineFitter.cpp:90.  Expected 1.15465, found 0.265326
```

tests/TestSplineFitter.cpp
```diff
@@ -82,10 +82,14 @@ void testPeriodicSpline() {
...
+    for (unsigned int i = 0; i < x.size()-1; i++)
+        x[i] = i/(x.size()-1.0);
     double ya[] = {15.579, 16.235, 17.325, 18.741, 20.454, 22.517, 24.944, 27.554, 29.942, 31.657,
                    32.486, 32.612, 32.494, 32.532, 32.785, 32.917, 32.402, 30.842, 28.229, 24.989,
                    21.762, 19.074, 17.147, 15.970, 15.467, 15.579};
     y.assign(ya, ya+y.size());
+    SplineFitter::createPeriodicSpline(x, y, deriv);
     ASSERT_EQUAL_TOL(SplineFitter::evaluateSplineDerivative(x, y, deriv, x[0]),
                      SplineFitter::evaluateSplineDerivative(x, y, deriv, x[x.size()-1]), 1e-6);
 }
```
craabreu: I added this test, which fails. This is the resulting error message:

```
exception: Assertion failure at TestSplineFitter.cpp:94.  Expected 13.9418, found 11.5266
```

craabreu: Sorry. Line 86 is wrong. I will fix it and post a new comment.

craabreu: Done! The proper error message is this:

```
exception: Assertion failure at TestSplineFitter.cpp:94.  Expected 9.88119, found 10.0784
```

craabreu: This issue has been addressed in a different PR (#2629).

craabreu:
I think that this PR is ready for review. I'm sorry I didn't deactivate the automatic removal of trailing spaces.

peastman:
You switched how it generates the spline coefficients if `periodic=true`, but it also needs to take that into account when evaluating it.  Currently, it will still return 0 outside the range of tabulated values.  `evaluate()` and `evaluateDerivative()` in ReferenceContinuous1DFunction need to be updated to be periodic.  Likewise, ExpressionUtilities needs to generate code that computes periodic values.

Continuous1DFunctionProxy needs to maintain compatibility with old files.  Increment the version number it writes to the file to 2.  Then in `deserialize()`, if `version < 2` it should assume nonperiodic rather than expecting the attribute to be specified in the file.

peastman:
I think this PR fell between the cracks.  I see you pushed some changes since my last comments.  Is this ready for another review?

peastman:
Just checking in again.  Is this ready for another review?

craabreu:
Sorry for the delay. I believe this is now ready for a new review.

peastman:
Most of this looks good.  The one part I'm not entirely comfortable with is the `gridExpansion` feature.  The problem is that updating the tabulated functions for multidimensional metadynamics is already a significant bottleneck, and adding that much padding will massively slow it down.  Consider that 40\*40\*40=64,000.  And that's a lower limit on the amount of padding you'd be adding to a 3D function!  It's pretty much guaranteed to be unusable.

The "right" way to fix this is to support periodic multidimensional splines.  If we don't want to add that right now, it's ok.  But instead of adding a workaround that isn't actually useful in practice, it's probably better to just not support multiple periodic CVs, and throw an exception if the user request them.  Then in the future we can remove the exception once proper support is added.

craabreu:
> But instead of adding a workaround that isn't actually useful in practice, it's probably better to just not support multiple periodic CVs, and throw an exception if the user request them. 

I've been using the grid expansion feature in my own research, but only in the two-dimensional case (Ramachandran free-energy surfaces, which is a rather common application). It has been pretty satisfactory, both in terms of continuity and efficiency. I didn't try any 3D calculation in practice, though.

> The "right" way to fix this is to support periodic multidimensional splines. 

The `SplineFitter.create2DNaturalSpline` relies on first-order derivatives with respect to x and y obtained via 1D splines. In these cases, it would be straightforward to impose periodic boundary conditions or not, according to the user specification. However, it also relies on cross derivatives, which is estimated by taking the derivatives w.r.t. x of the previously computed derivatives w.r.t. y. This order is chosen by convention. I'm not sure if this would be correct when direction x is periodic and direction y is not, or vice-versa. If we can guarantee it is correct, then it would be straightforward to support periodic multidimensional splines (the 3D case follows the same principle). Anyway, I believe it is safe to support, right now, the special case when all directions are periodic.

peastman:
> I'm not sure if this would be correct when direction x is periodic and direction y is not, or vice-versa.

I can't think of any reason it would make a difference.  Periodicity just matters when computing the spline coefficients.  Once you get to evaluation, the calculation is identical regardless of how the spline was determined.

> Anyway, I believe it is safe to support, right now, the special case when all directions are periodic.

That would be great!

peastman:
We're getting ready to make the 7.5 release, so if we want to include this it needs to be finished up soon.  If you can implement periodicity for multidimensional splines, that would be great.  If you don't have time for that right now, let's just take out the grid expansion option and have it throw an exception instead.  We can then add support in the next release.

craabreu:
I have implemented the periodic versions of 2D/3D splines and tabulated functions. Now I need to:

- [ ] Update serialization of 2D/3D continuous function objects and corresponding tests
- [ ] Update `ForceField.py` in order to include possible explicit specification of the `periodic` argument
- [ ] Include test cases for periodic 2D/3D continuous functions
- [ ] Replace grid expansion by periodic functions in `Metadynamics.py`

Please note that I have removed the `setPeriodic()` method, meaning that the periodicity feature of a tabulated function will be immutable once defined at construction time. The reason is that `setFunctionParameters()` performs sanity checks which depend on whether the function is periodic or not. Therefore, if the user called `setFunctionParameters()` first, then `setPeriodic()`, these checks would be meaningless.

craabreu:
I think this is ready for a new review.

peastman:
Travis is reporting that it doesn't compile.

```
/Users/travis/build/openmm/openmm/openmmapi/src/SplineFitter.cpp:92:9: error: call to 'abs' is ambiguous
    if (not_equal(y[0], y[n-1]))
        ^~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/openmm/openmm/openmmapi/src/SplineFitter.cpp:40:26: note: expanded from macro 'not_equal'
#define not_equal(a, b) (abs((a)-(b)) > 1e-15 + 1e-15*abs(b))  // same as scipy.interpolate()
                         ^~~
/Applications/Xcode-9.3.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdlib.h:137:6: note: candidate function
int      abs(int) __pure2;
         ^
/Applications/Xcode-9.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/stdlib.h:111:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long      abs(     long __x) _NOEXCEPT {return  labs(__x);}
                                           ^
/Applications/Xcode-9.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/stdlib.h:113:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long long abs(long long __x) _NOEXCEPT {return llabs(__x);}
```

You probably meant that to be `fabs()` rather than `abs()`?

openmmapi/src/SplineFitter.cpp
```diff
@@ -37,6 +37,15 @@
 using namespace OpenMM;
 using namespace std;
 
+#define not_equal(a, b) (abs((a)-(b)) > 1e-15 + 1e-15*abs(b))  // same as scipy.interpolate()
```
peastman, line 40: Preprocessor macros should be all upper case: `NOT_EQUAL()`.

peastman: Or alternatively you could just make this a function rather than a macro:

```c++
static bool notEqual(double a, double b) {
  return (fabs(a-b) > 1e-15 + 1e-15*fabs(b));
}

Comment (craabreu) [reply openmmapi/src/SplineFitter.cpp]: Done!

File: serialization/tests/TestSerializeTabulatedFunctions.cpp [hunk 5]
```diff
@@ -65,6 +68,35 @@ void testContinuous1DFunction() {
...
+
+    double min2, max2;
+    vector<double> values2;
+    copy->getFunctionParameters(values2, min2, max2);
+    ASSERT_EQUAL(min, min2);
+    ASSERT_EQUAL(max, max2);
+    ASSERT_EQUAL(values.size(), values2.size());
+    for (int j = 0; j < (int) values.size(); j++)
+        ASSERT_EQUAL(values[j], values2[j]);
+
```
peastman, line 97: This should also check that `getPeriodic()` returns the correct value.  Same for the other tests below.

wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -86,18 +86,42 @@ def _parseFunctions(element):
                 params[key] = int(function.attrib[key])
             elif key.endswith('min') or key.endswith('max'):
                 params[key] = float(function.attrib[key])
+        if functionType.startswith('Continuous'):
+            params['periodic'] = eval(function.attrib.get('periodic', 'False').title())
```
peastman, line 90: This is dangerous.  It will execute any Python code put into the attribute.  A better way of doing it is

```python
params['periodic'] = (function.attrib.get('periodic', 'false').lower() == 'true')
```

peastman:
TestMetadynamics.testHarmonicOscillator() is failing on all the builds:

```
>           assert abs(fe[i]-e) < 1.0*kilojoules_per_mole
E           AssertionError: assert Quantity(value=3.039012257389733, unit=kilojoule/mole) < (1.0 * Unit({BaseUnit(base_dim=BaseDimension("amount"), name="mole", symbol="mol"): -1.0, ScaledUnit(factor=1.0, master=gram*nanometer**2/(picosecond**2), name='kilojoule', symbol='kJ'): 1.0}))
E            +  where Quantity(value=3.039012257389733, unit=kilojoule/mole) = abs((Quantity(value=10.239012257389746, unit=kilojoule/mole) - Quantity(value=7.200000000000013, unit=kilojoule/mole)))
```

wrappers/python/simtk/openmm/app/forcefield.py
craabreu: If the user makes a mistake when trying to enter `true`, this will evaluate to `false` without notice. How about the alternative below?

```py
        if functionType.startswith('Continuous'):
            periodicStr = function.attrib.get('periodic', 'false').lower()
            if periodicStr in ['true', 'false', 'yes', 'no', '1', '0']:
                params['periodic'] = periodicStr in ['true', 'yes', '1']
            else:
                raise ValueError('ForceField: non-boolean value for periodic attribute in tabulated function definition')
```

serialization/tests/TestSerializeTabulatedFunctions.cpp
craabreu: Ok.

craabreu:
> TestMetadynamics.testHarmonicOscillator() is failing on all the builds:

Two issues here:

1. There is randomness in this test. In my local system, it passes more often than not. I was able to eliminate randomness by adding the following line:

```py
integrator.setRandomNumberSeed(4321)
```

2. The test behaves differently in this new version of `Metadynamics` due to the correction of the issue mentioned in https://github.com/openmm/openmm/issues/2604#issuecomment-604640319. Even though `biasWidth` is specified as `0.02` (see below), in the original `Metadynamics` the standard deviation of the deposited Gaussians was actually `sqrt(2*biasWidth/gridWidth)*(maxValue-minValue) = 0.00431`. Note that `gridWidth=31` due to a floating-point rounding error.

https://github.com/openmm/openmm/blob/eec9cd69a6f2c1e92d365691bbc62eae621b481c/wrappers/python/tests/TestMetadynamics.py#L19

https://github.com/openmm/openmm/blob/eec9cd69a6f2c1e92d365691bbc62eae621b481c/wrappers/python/simtk/openmm/app/metadynamics.py#L298

https://github.com/openmm/openmm/blob/eec9cd69a6f2c1e92d365691bbc62eae621b481c/wrappers/python/simtk/openmm/app/metadynamics.py#L199

If we simply substitute `0.00431` for `0.02` in a new version of the test, then `gridWidth` will evaluate to 140. The only way to have the same behavior as before is by doing:

```py
bias = BiasVariable(cv, 0.94, 1.06, 0.00431, gridWidth=31)
```

peastman:
I think this is ready to merge!  The failing test case is because the Mac build is still using Python 2.7, which we're not supporting anymore.  I updated it to use 3.8 in #2764 

Thanks so much.  This is fantastic!

PR #2612, opened 2020-03-27
Update Azure VM to newer version

andysim:
When I first set up Azure CI I used the oldest VS version, in keeping with the OpenMM ethos of supporting legacy hardware and software stacks where possible.  Recently Microsoft issued the warning

```
On March 23, 2020, Azure Pipelines will remove Windows Server 2012 R2 with Visual Studio 2015(`vs2015-win2012r2`), macOS High Sierra 10.13 (`macOS-10.13`), and Windows Server Core 1803 (`win1803`) from our hosted pools. This includes builds using the `Hosted` agent pool. A series of brownouts to highlight builds and releases using these soon-to-be-removed images will be conducted over the next couple of weeks. During these brownouts, any builds and releases using the vs2015-win2012r2, macOS-10.13, or win1803 images will fail.
```

To address the deprecation, this PR updates the VM to a newer, but not latest, version.

### DIFF
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -2,9 +2,9 @@ jobs:
 
   # Configure, build, install, and test job
   - job: 'windows_build'
-    displayName: 'Windows VS2015'
+    displayName: 'Windows VS2017'
     pool:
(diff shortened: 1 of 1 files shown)


peastman:
Thanks!

Issue #2614, opened 2020-03-28
CharmmPsfFile parser constrains are broken with CHARMM TIP3 water topology


YevChern:
Hi. After changes introduced in #2598 ```CharmmPsfFile``` parser isn't able to handle CHARMM TIP3 water topology properly when constrains are on. The reason is that TIP3 topology from [CHARMM36](http://mackerell.umaryland.edu/charmm_ff.shtml) force field contain additional bond between hydrogen atoms that is needed, as I understand, for constrains implementation with CHARMM software. With the new logic for constrains addition introduced in #2598 , constrains between hydrogen atoms in water are added twice when HBonds or AllBonds is defined (first [as any other bond](https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/wrappers/python/simtk/openmm/app/charmmpsffile.py#L951-L959), then [when angles are constrained](https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/wrappers/python/simtk/openmm/app/charmmpsffile.py#L1025-L1048)).

Solution would be to omit adding constrain between water hydrogens when we cycle through bonds, so it is then added when angles are constrained with rigidWater.

YevChern:
```CharmmPsfFile``` is now ignoring H-H water bond when reading a file which resolves the issue.

PR #2615, opened 2020-03-28
Fix for CharmmPsfFile parser constrains with CHARMM TIP3 water topology

YevChern:
As described in #2614 , ```CharmmPsfFile``` parser is not handling constrains with CHARMM TIP3 water topology properly. Proposed changes should be enough to fixed it.

### DIFF
--- a/wrappers/python/simtk/openmm/app/charmmpsffile.py
+++ b/wrappers/python/simtk/openmm/app/charmmpsffile.py
@@ -272,6 +272,9 @@ class CharmmPsfFile(object):
                 drudepair_list.append([min(id1,id2), max(id1,id2)])
             elif (atom_list[id1].name[0:2]=='LP' or atom_list[id2].name[0:2]=='LP' or atom_list[id1].name=='OM' or atom_list[id2].name=='OM'):
                 pass
+            # Ignore H-H bond in water if present
+            elif atom_list[id1].name[0]=='H' and atom_list[id2].name[0]=='H' and (atom_list[id1].residue.resname in WATNAMES):
+                pass
             else:
                 bond_list.append(Bond(atom_list[id1], atom_list[id2]))
         bond_list.changed = False
--- a/wrappers/python/tests/TestCharmmFiles.py
+++ b/wrappers/python/tests/TestCharmmFiles.py
@@ -432,56 +432,77 @@ class TestCharmmFiles(unittest.TestCase):
         system_openmm = forcefield.createSystem(topology, constraints=None, rigidWater=True, nonbondedMethod=PME,
                                                 nonbondedCutoff=8 * angstroms)
         self.assertEqual(
-            sorted(system_charmm.getConstraintParameters(i) for i in range(system_charmm.getNumConstraints())),
-            sorted(system_charmm.getConstraintParameters(j) for j in range(system_charmm.getNumConstraints())))
+            sorted(system_charmm.getConstraintParameters(i)[:2] for i in range(system_charmm.getNumConstraints())),
+            sorted(system_openmm.getConstraintParameters(j)[:2] for j in range(system_openmm.getNumConstraints())))
+        for i in range(system_charmm.getNumConstraints()):
+            self.assertAlmostEqual(system_charmm.getConstraintParameters(i)[2],
+                                   system_openmm.getConstraintParameters(i)[2], delta=1e-7 * nanometers)
 
         system_charmm = psf.createSystem(params, constraints=HBonds, rigidWater=False, nonbondedMethod=PME,
                                          nonbondedCutoff=8 * angstroms)
         system_openmm = forcefield.createSystem(topology, constraints=HBonds, rigidWater=False, nonbondedMethod=PME,
                                                 nonbondedCutoff=8 * angstroms)
         self.assertEqual(
-            sorted(system_charmm.getConstraintParameters(i) for i in range(system_charmm.getNumConstraints())),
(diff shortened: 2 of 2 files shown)


z-gong:
If the PSF already contains H-H bonds explicitly, then it will fail if someone uses constraints=HBonds and rigidWater=False.

Probably it's better to check the logic of adding angle constraints instead of bond
https://github.com/openmm/openmm/blob/691ee61997aa13f8a0bf0c7fffd72a818a0d0c8d/wrappers/python/simtk/openmm/app/charmmpsffile.py#L1023-L1027
``` python
if bond constraint already exists:
    continue
```

YevChern:
The behavior then might be a bit counterintuitive as one would effectively get rigid water with rigidWater=False because of that explicit H-H water bond. But, strictly speaking, this is what user asking for by having H-H bond in water in their topology. Though, I wasn't aware that this bond was there in official distribution of the CHARMM force field before #2598.

I agree, it is better avoid adding second constrain during angle, not bond, constraining and probably issue a warning that the water hydrogens are constrained and the water is still rigid due to explicit bond in topology in the case of constraints=HBonds/AllBonds and rigidWater=False.

peastman:
> But, strictly speaking, this is what user asking for by having H-H bond in water in their topology.

The user probably didn't ask for it, CHARMM just added it automatically.  I'd say this is more of an implementation difference in how the two programs handle constraints: CHARMM has to create a fake "bond" to allow the distance to be constrained, but it sets the spring constant for that bond to 0 because you don't really want a bond there.  OpenMM doesn't need that bond because it has other ways of making water rigid.

If the user specifies `constraints=HBonds, rigidWater=False`, I think the correct behavior would be for the H-O-H angle to be flexible?  (Not that I can think of any reason someone would choose that combination of parameters, but *if* they did.)

YevChern:
I can see the following scenarios how we can handle explicit H-H water bond:

- Assume that H-H water bond is present in topology only for constrains implementation with CHRAMM and ignore it. Then we would add bond and angle constrains as we normally do. That will give us expected behavior with ```constraints=HBonds``` and ```rigidWater=False```, but user will not be able to override constrain distance for H-H bond with explicit bond in topology if, for whatever reason, he decides to set it to some non-standard value. With this solution we make assumptions about what a user actually wants with a given topology instead of just following the instructions given, which is probably not good.

- Take H-H water bond if present in topology and constrain it. At the same time we would avoid adding H-O-H angle constrain. That would give rigid water with ```constraints=HBonds``` and ```rigidWater=False```, but will allow non-standard values for H-H constrain distance. We would also issue a warning if ```constraints=HBonds``` and ```rigidWater=False``` that the water is still made rigid because of the explicit H-H bond in topology. If this is not what the user wants, one would have to edit topology and delete that H-H bond. In this case we make less assumptions on what user actually wants from the software and just following the instructions, but put some more responsibility on a user to make sure that the topology provided corresponds to what user expects from it.

- Not do anything special with explicit H-H water bond. In this case user will not be able to run a simulation before doing something about that H-H bond - either deleting it or defining ```constraints=HBonds```, ```rigidWater=False``` and getting rigid water that way. In this case we do zero assumptions on what user actually wants and just following the instructions given. This solution is probably not a good one as for many users spending time figuring difference between internal implementations of constrains in CHARMM and OpenMM would be a waste of time and we should make things just work for them.

I personally prefer second solution as it still preserves the ability to override angle constraining with some custom value of H-H bond. I don't know why would someone do that, but it's always better to have more options then less. What do you think?

peastman:
I think this is largely just a question of implementation.  Regardless of how it's implemented, I believe this is what the result should be:

- If `rigidWater=True`, the H-H distance should be constrained.
- If `rigidWater=False` and `constraints` is either `HBonds` or `AllBonds`, the H-H distance should *not* be constrained.
- Regardless of these settings, the Topology object returned by `topology` should *not* include a bond between the hydrogens.  Topology objects represent the chemical structure of the molecule, independent of modelling choices like force field or constraints.

As long as that's the result, we can  implement it internally however is most straightforward.

Here's another way of looking at it.  A user should be able to follow these steps:

1. Use CHARMM to create a psf file using the CHARMM36 force field.
2. Load that file into OpenMM.
3. Create a System by calling `createSystem()` on it.
4. Create a ForceField using OpenMM's implementation of CHARMM36.
5. Call `createSystem()` on the ForceField, passing in the Topology object created from the psf file.

The Systems created in steps 3 and 5 should produce identical energies and have identical constraints.

YevChern:
Sorry for a prolonged silence. The easiest way would be to ignore the H-H water bond when we read PSF file.
Now everything should work as expected.

swails:
@peastman -- any idea what this is about:

```

/tmp/openmmTempKernel0x5574c59a3dc0_5995.cu(771): error: calling a constexpr __host__ function("fmin") from a __global__ function("computeBondedForces") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.



/tmp/openmmTempKernel0x5574c59a3dc0_5995.cu(772): error: calling a constexpr __host__ function("fmin") from a __global__ function("computeBondedForces") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
```?

peastman:
I don't think that's related to this PR?  Could you open a separate issue for it?

peastman:
Looks good!  Is this ready to merge?

peastman:
I opened an issue about the jenkins error: #2648.  It clearly isn't caused by this change, since the error is in a C++ test and this PR only affects Python code.

YevChern:
@peastman yes.

peastman:
Thanks!

Issue #2618, opened 2020-03-29
ARM support and improvements


mark-mb:
I saw that the current support for Arm was a bit outdated it, so I started brushing it up.

 - What set inputs and benchmarks would you recommend to use to tweak and test improvements?
 - What would you want to see updated in the Arm part of OpenMM?

peastman:
> What would you want to see updated in the Arm part of OpenMM?

#2499 also relates to ARM fixes.  @jaimergp may have thoughts about this.

> What set inputs and benchmarks would you recommend to use to tweak and test improvements?

You can use the benchmark.py script in the examples folder to run lots of different benchmarks.  A good default one is

```
python benchmark.py --platform=CPU --seconds=10 --test=pme
```

I'm assuming you're interested in the CPU platform.  The speed of ARM specific code is mostly irrelevant when running on a GPU.

PR #2619, opened 2020-03-29
ARM64 support

mark-mb:
This changes allow OpenMM to compile and run the tests in an A72 platform. Discussion for further work in #2618


peastman:
@jaimergp could you take a look at this and make sure it looks ok?

peastman:
Could you comment on the changes to JAMA?  I get very nervous about changes to complex numerical libraries that I don't understand!  Are these changes that should be contributed to the upstream package?

jaimergp:
Let's see if the CF build works with these changes!

jaimergp:
> Could you comment on the changes to JAMA? I get very nervous about changes to complex numerical libraries that I don't understand! Are these changes that should be contributed to the upstream package?

According to the commit message: "This was already present in JAMA 1.0.3, but not in OpenMM"

peastman:
In that case, shall we just update to the latest version of JAMA?

jaimergp:
I guess there are no downsides to that!

mark-mb:
> Could you comment on the changes to JAMA? I get very nervous about changes to complex numerical libraries that I don't understand! Are these changes that should be contributed to the upstream package?

One of the tests was stuck in an infinite loop because JAMA::SVC wasn't converging. It was comparing something like 1e-324 to 0, which, of course, are different. The solution in these cases is usually to add some wiggle room, but, like you, I wouldn't be comfortable just changing it.

I checked latest JAMA version source (latest, meaning November 2012), and there was that same "tiny" variable set to pow(2, -966). In the C++ version, I set it to the smallest representable double possible. That fixed the issue, and the test passed.

peastman:
In that case, how about if we remove those changes from this PR, and then in a separate PR update to the latest JAMA version?  I feel better about just keeping in sync with the standard package.

mark-mb:
I'll do that. Is JAMA code somewhere directly in C++, or do I have to translate it from Java to C++?

peastman:
It's at https://math.nist.gov/tnt/download.html.

jaimergp:
That indeed worked! https://cloud.drone.io/conda-forge/openmm-feedstock/13/1/2

mark-mb:
I moved the JAMA changes to openmm#2626.

The Jenkins failures in this pull request seem unrelated to the patch. Not sure what procedure you usually follow in these cases.

peastman:
Thank you!

PR #2616, opened 2020-03-28
Minor Python tweaks

RasmusWL:
I read the [post](https://foldingathome.org/2020/03/10/covid19-update/) from Folding@Home about their effort on simulating SARS-CoV-2, and that they were using OpenMM.

Here is a small effort in trying to support your project :blush: it's mostly just minor tweaks making the Python code more idiomatic.

I used the alerts from https://lgtm.com/projects/g/openmm/openmm/?mode=tree to find opportunities for improvements, and might do a few more PRs based on this if you're happy about this PR :blush: (for full disclosure, I work on this project in my day-to-day work).

PR #2620, opened 2020-03-30
Fix bug in __ne__ implementation from #2616

RasmusWL:
Wow, that was slightly embarrassing :flushed:

I also stumbled upon the fact that you it's slightly more correct to implement `__ne__` by using `not self == other` instead of `not self.__eq__(other)`. (explained [here](https://stackoverflow.com/a/30676267/538973))


peastman:
Oops!

Issue #2578, opened 2020-02-26
Remaining changes for Nose-Hoover


peastman:
There are some changes that still need to be made to NoseHooverIntegrator before we can release 7.5:

- Finish up #2527 to convert it to the common framework.
- Finish up #2518 to change how temperature initialization is done.
- Convert it to the LFMiddle discretization, which allows much larger timesteps.  This is what we now use in LangevinMiddleIntegrator.  Also see http://dx.doi.org/10.1021/acs.jpca.9b02771, which has details about using this discretization with Nose-Hoover.
- Write the documentation for it in the user guide.  In particlular,
  - Add a brief description of it in section 3.6.7.
  - Add a sentence or two in chapter 16 to add it to the list of integrators for Drude particles.
  - Add a detailed description in chapter 20.

cc @andysim @Olllom

z-gong:
I've noticed that some logic in the velocityVerlet integrator can be simplified. Herein, when integrating a drude pair, the velocity and force are transformed to center of mass and relative motions, on which the velocities are integrated, and then transformed back to real atoms and drude particles. This is not really necessary. The forces can operate directly on the real and drude particles to integrate their velocity.
https://github.com/openmm/openmm/blob/a9223eea076bf54cc5028064d664b2951be94579/platforms/reference/src/SimTKReference/ReferenceVelocityVerletDynamics.cpp#L103-L139

andysim:
I think the temperature initialization stuff is good to go already - I'll take a look at what Travis is unhappy about in a bit, but I don't think it's related to my changes.  I think the only difficult task remaining is the conversion to the common platform.  The integrator itself is working, but I'm getting strange memory errors in the thermostat.  I'll dig around more today and try to resolve them.

andysim:
I think the temperature initialization stuff is good to go already - I'll take a look at what Travis is unhappy about in a bit, but I don't think it's related to my changes.  I think the only difficult task remaining is the conversion to the common platform.  The integrator itself is working, but I'm getting strange memory errors in the thermostat.  I'll dig around more today and try to resolve them.

peastman:
Are you planning to do the conversion to LFMiddle or should I start on that?

andysim:
Sorry for the delay - crazy times over here!  I'm working on it right now while social distancing responsibly.  Feel free to chime in on #2600 if you think my approach is bad.

peastman:
[...]

PR #2621, opened 2020-03-30
Nosé Hoover Integrator docs

andysim:
The text is probably too verbose, but here is an attempt at documenting the Nosé-Hoover integrator, which is on the to-do list for #2578.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -1059,6 +1059,23 @@ sampling, and therefore is preferred for most applications.  Also note that
 :code:`LangevinIntegrator`\ , like :code:`LangevinMiddleIntegrator`\ , is a leapfrog
 integrator, so the velocities are offset by half a time step from the positions.
 
+Nosé-Hoover Integrator
+----------------------
+
+The :code:`NoseHooverIntegrator` uses the same "middle" leapfrog propagation
+algorithm as :code:`LangevinMiddleIntegrator`, but replaces the stochastic
+temperature control with a velocity scaling algorithm that produces more
+accurate transport properties :cite:`Basconi2013`.  This velocity scaling
+results from propagating a chain of extra variables, which slightly reduces the
+computational efficiency with respect to :code:`LangevinMiddleIntegrator`.  The
+integrator is minimally created by specifying only the timestep::
+
+    integrator = NoseHooverIntegrator(0.004*picoseconds)
+
+However, many optional settings may be tuned as described XXXXXXXX.  Note that
+for this integrator, the velocities are offset by half a time step from the
+positions.
+
 Leapfrog Verlet Integrator
 --------------------------
 
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -1071,7 +1071,7 @@ computational efficiency with respect to :code:`LangevinMiddleIntegrator`.  The
 thermostated integrator is minimally created with syntax analogous to the
 :code:`LangevinMiddleIntegrator` example above::
 
-    NoseHooverIntegrator integrator(300*kelvin, 25/picosecond,
+    NoseHooverIntegrator integrator(300*kelvin, 1/picosecond,
                                     0.004*picoseconds);
 
 The first argument specifies the target temperature.  The second specifies the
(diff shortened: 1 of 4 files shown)


docs-source/usersguide/application.rst
```diff
@@ -1059,6 +1059,64 @@ sampling, and therefore is preferred for most applications.  Also note that
...
+The :code:`NoseHooverIntegrator` uses the same "middle" leapfrog propagation
+algorithm as :code:`LangevinMiddleIntegrator`, but replaces the stochastic
+temperature control with a velocity scaling algorithm that produces more
+accurate transport properties :cite:`Basconi2013`.  This velocity scaling
+results from propagating a chain of extra variables, which slightly reduces the
+computational efficiency with respect to :code:`LangevinMiddleIntegrator`.  The
+thermostated integrator is minimally created with syntax analogous to the
+:code:`LangevinMiddleIntegrator` example above::
+
+    NoseHooverIntegrator integrator(300*kelvin, 25/picosecond,
```
peastman: Is 25/ps a good choice?  How does this value relate to the friction coefficient?  For Langevin, we recommend a friction coefficient of 1/ps.

docs-source/usersguide/application.rst
```diff
@@ -1059,6 +1059,64 @@ sampling, and therefore is preferred for most applications.  Also note that
...
+                                    0.004*picoseconds);
+
+The first argument specifies the target temperature.  The second specifies the
+frequency of interaction with the heat bath: a lower value interacts minimally,
+yielding the microcanonical ensemble in the limit of a zero frequency, while a
+larger frequency will perturb the system greater, keeping it closer to the
+target temperature.  The third argument is the integration timestep that, like
+the other arguments, must be specified with units.
+
+More verbose invocation is possible::
```
peastman: This is probably more information than is needed here.  The rest of the section could be replaced with something like, "This integrator supports lots of other options, including the ability to couple different parts of the system to thermostats at different temperatures.  See the API documentation for details."

docs-source/usersguide/application.rst
Olllom: Yes, that is a reasonable choice. Nosé-Hoover preserves the ensemble for position and momentum. In contrast, the Langevin thermostat introduces artificial damping through the (partial) randomization of velocities and therefore preserves the ensemble only for positions. As a consequence, the fricition coefficient must be small for Langevin in order to retain the kinetics. For Nose-Hoover, the collision frequency can be chosen according to (mainly) numerical considerations such as speed of convergence and stability.

peastman: I would say it differently.  Langevin has an implicit heat bath (represented by the random and damping forces) while Nose-Hoover models the heat bath explicitly.  But either way the heat bath creates friction on the physical degrees of freedom, and the stronger that friction, the more it will slow them down.

I ran some tests to measure the effect of this.  I simulated alanine dipeptide in explicit solvent for 1 ns, and computed the autocorrelation functions for the phi and psi angles.  I repeated this with friction coefficients of 1/ps and 25/ps.  For phi, the higher friction only slowed it down a little:

image

But for psi it slowed things down quite a lot:

image

peastman:
I'd like to get this PR finished up.  Based on the data above, it looks to me like the friction coefficient really does affect the dynamics, and 25/ps is much too large a value.  Do you see any reason not to use 1/ps, the same value we use with other thermostats?

Olllom:
No, I don't see anything wrong with using 1/ps as a default.
That said, I find you ala2 results quite surprising; so surprising that I repeated your test and also found some discrepancies between 1/ps and 25/ps for the psi and phi autocorrelation. So 1/ps seems to be a better default.

Originally, we used quite large collision frequencies to get stable Drude simulations but the instability was partly caused by a bug in the thermostat as we found out later. The original Drude publications used fairly large collision frequencies, (like 10/ps for the regular and 200/ps for the Drude thermostat in https://aip.scitation.org/doi/10.1063/1.1598191) 

Subsequently, we ran a few tests for water diffusion in bulk and in bilayers, which did not seem to be sensitive to the collision frequency at all, even when we went as large as 50/ps.

To wrap up: I agree that 1/ps is good for the NoseHooverIntegrator. It is not a sensible default for the Drude thermostat.

peastman:
We've been having more discussions about this offline.  Getting good statistics is harder than I thought.  Even after 100 ns the results still aren't fully converged.  Based on the data we've collected so far, it looks like 1/ps and 25/ps are both reasonable values.  I'm fine with whichever one you want to use.  If you can make the other changes, we can get this PR merged.

andysim:
Ok, I updated the default shown in the docs, but also recommended a larger value for equilibration runs.  I'm happy to change it further if anybody objects, or if our tests reveal anything new.

peastman:
Looks great!

Issue #1968, opened 2018-01-10
Drude: Particle coordinate is nan


chemlove:
I am new to openMM and I want to use the charmm_polar_2013.xml to perform MD on the protein-water system. The openMM version is 7.1.

When I placed just 4 (water-4.pdb) or 100 (water-100.pdb) water molecules in the box, the simulation could go on with the DrudeSCFIntegrator and print the potential energy. 
However, when I placed ~1000 water molecules in this box, or only placed my protein in this box, the simulation crashed with DrudeSCFIntegrator and the error information is:
```
In [328]: for i in range(1,10):
   .....:         simulation.step(1000)
   .....:         
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-328-14cc9c512e83> in <module>()
      1 for i in range(1,10):
----> 2         simulation.step(1000)
      3 

/opt/gpu/anaconda-2.3.0/lib/python2.7/site-packages/simtk/openmm/app/simulation.pyc in step(self, steps)
    130     def step(self, steps):
    131         """Advance the simulation by integrating a specified number of time steps."""
--> 132         self._simulate(endStep=self.currentStep+steps)
    133 
    134     def runForClockTime(self, time, checkpointFile=None, stateFile=None, checkpointInterval=None):

/opt/gpu/anaconda-2.3.0/lib/python2.7/site-packages/simtk/openmm/app/simulation.pyc in _simulate(self, endStep, endTime)
    217                 for reporter, next in zip(self.reporters, nextReport):
    218                     if next[0] == nextSteps:
--> 219                         reporter.report(self, state)
    220 
    221     def saveCheckpoint(self, file):

/opt/gpu/anaconda-2.3.0/lib/python2.7/site-packages/simtk/openmm/app/pdbreporter.pyc in report(self, simulation, state)
     90             self._topology = simulation.topology
     91             self._nextModel += 1
---> 92         PDBFile.writeModel(simulation.topology, state.getPositions(), self._out, self._nextModel)
     93         self._nextModel += 1
     94         if hasattr(self._out, 'flush') and callable(self._out.flush):

/opt/gpu/anaconda-2.3.0/lib/python2.7/site-packages/simtk/openmm/app/pdbfile.pyc in writeModel(topology, positions, file, modelIndex, keepIds, extraParticleIdentif
ier)    328             positions = positions.value_in_unit(angstroms)
    329         if any(math.isnan(norm(pos)) for pos in positions):
[...]

PR #2625, opened 2020-03-31
Added drudeMass argument for setting mass of Drude particles

peastman:
See #1968 and #2513.


peastman:
For some reason travis builds have stopped showing up.  They're still getting run.  You can find the results if you go directly to the travis website.  I think this just started happening a few days ago.

PR #2642, opened 2020-04-13
Lepton: define M_PI if needed with MingW (in addition to MSVC)

giacomofiorin:
This small proposed change enables using MingW to build Lepton when it is linked with LAMMPS.

A PR affecting the LAMMPS copy is already open (see below) and this PR is needed to avoid deviations from the original.  Any improvements to this PR will be applied verbatim to the copies as well.

https://github.com/lammps/lammps/pull/1992


peastman:
Looks good.  Thanks!

PR #2645, opened 2020-04-15
[WIP] Refactored and optimised AVX CPU code.

dwtowner:
Pulled out some vector utility functions (gather pair, reduce) and refactored the AVX CPU code to make use of them.

@peastman  First commit in a series. This commit gives about 5% boost, but the complete set will ultimately give about 20%.


peastman:
You have some really impressive looking code in here!  The optimizations look great.  And I like the way you unified the different versions of the code.

There's just one problem: this change breaks all other vector architectures!

fvec4 and ivec4 act as an abstraction layer.  We have implementations for multiple architectures.  That lets us write vectorized code once and compile it on multiple platforms.  But it only works if all implementations have identical APIs.  So any feature you add to vectorize_sse.h, you also have to add the same feature to vectorize_ppc.h, vectorize_neon.h, and vectorize_pnacl.h.

dwtowner:
Yes, I realised that yesterday, but never got around to commenting on it here. I think it should be straightforward to add those API features to the other targets, but I have no machine to test on just at the moment. I can probably put in the code I as I think it should be and then let the CI system try it out.

peastman:
That sounds great then.  I'm working on the CI right now: #2653

Once I get that working, I'll make a more detailed pass through the code and offer more specific comments.

mark-mb:
I personally think it's still better to have a vector mask or integer or similar type. Reducing from a vector mask to an int8, and then expanding that int8 back to a vector may be simple and efficient in SSE or AVX, but it may not be that easy in other vector implementations. Maybe having a mask or integer type in the vectorization API could hide those details.

dwtowner:
A mask type in the simd api is probably the right thing to do. In my patch
I can change the getmaskfromcompare and blend functions to work on that
type and the implementation is then free to choose whether the mask is
bitwise or element- wise as appropriate. I'll make that change and you can
all see what you think.

peastman:
By the way, CI is now working for ARM and PPC.  That should help with the testing.

dwtowner:
I decided to back out some of the changes I'd made to make it easier to port across different SIMD platforms, so rather than getting in a mess I've created a new PR in #2661. This omits some of the optimisations regarding conversion to and from integer masks, and concentrates just on getting one unified API across all SIMD types and one templated algorithm which works with them.

ivec4 is still present, but in its new form it can probably be removed pretty easily. That's a change too far for the moment though.

The changes have slowed down SSE/AVX a little, but the code is still faster than the current master.

I have done what I think is right for ARM/Neon/NaCl but I don't have a way of building on those machines at the moment so will have to wait for CI to pick things up. All new API functions are tested though, so it shouldn't be too painful I hope.

If this commit is accepted, I can then restore some of the other optimisations back onto the IA-specific vector types to bring the speed back up there, and then commit my AVX-512 changes too.

dwtowner:
Abandoned in favour of #2661

PR #2651, opened 2020-04-21
CustomIntegrator variables are saved to checkpoints

### LINKED ISSUE #2650: Context.{create|load}Checkpoint() should save/restore CustomIntegrator parameters
We recently realized that `Context.{create|load}Checkpoint()` does not save/restore `CustomIntegrator` parameters, resulting in simulations that make use of `CustomIntegrator`s may not properly restore their state if resuming from a checkpoint.

peastman:
Fixes #2650.

### DIFF
--- a/openmmapi/include/openmm/CustomIntegrator.h
+++ b/openmmapi/include/openmm/CustomIntegrator.h
@@ -668,6 +668,16 @@ protected:
      * Get whether computeKineticEnergy() expects forces to have been computed.
      */
     bool kineticEnergyRequiresForce() const;
+    /**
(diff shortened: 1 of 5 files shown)


jchodera:
@peastman: Will this work for `CustomCompoundIntegrator` as well?

peastman:
Do you mean a CompoundIntegrator that includes a CustomIntegrator as one of its components?  No, thanks for pointing that out.  I need to add support to CompoundIntegrator as well.

jchodera:
Yes! There could even be multiple `CustomIntegrator` components.

PR #2653, opened 2020-04-23
Added CI builds for PPC and ARM

peastman:
#2645 made me realize we really ought to have CI testing for PPC and ARM.  Travis has alpha level support for this (https://docs.travis-ci.com/user/multi-cpu-architectures/), so I'm trying it out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,6 +141,7 @@ before_install:
       cd cmake-3.17.1;
       ./bootstrap;
       make -j 4;
+      sudo make install;
       export PATH=${HOME}/OpenMM/bin:${PATH};
     fi
   #- if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,11 +136,11 @@ before_install:
       brew install -y https://raw.githubusercontent.com/Homebrew/homebrew-core/5b680fb58fedfb00cd07a7f69f5a621bb9240f3b/Formula/doxygen.rb;
       sudo pip install -U pytest --ignore-installed six;
     fi
-  - if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
-      CMAKE_URL="http://cmake.org/files/v3.7/cmake-3.7.2-Linux-x86_64.tar.gz";
-      mkdir $HOME/cmake && travis_retry wget --no-check-certificate --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C $HOME/cmake;
-      export PATH=${HOME}/cmake/bin:${PATH};
-    fi
+  #- if [[ "${TRAVIS_OS_NAME}" == "linux" ]]; then
+  #    CMAKE_URL="http://cmake.org/files/v3.7/cmake-3.7.2-Linux-x86_64.tar.gz";
+  #    mkdir $HOME/cmake && travis_retry wget --no-check-certificate --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C $HOME/cmake;
+  #    export PATH=${HOME}/cmake/bin:${PATH};
+  #  fi
   - if [[ "$OPENCL" == "true" ]]; then
       wget http://s3.amazonaws.com/omnia-ci/AMD-APP-SDKInstaller-v3.0.130.135-GA-linux64.tar.bz2;
       tar -xjf AMD-APP-SDK*.tar.bz2;
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 addons:
   apt:
     packages:
-      - gromacs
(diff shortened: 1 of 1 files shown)


peastman:
And now, of course, travis has just stopped building at all, for no reason I've been able to figure out.

jchodera:
@jaimergp may have experience with this!

peastman:
The issue I'm dealing with right now is that the preinstalled version of cmake for PPC and ARM (which is also the version provided by apt) is very old.  I'm trying to get it to compile from source, which probably isn't a great long term solution (it's slow), but can serve as a temporary one.

jchodera:
@peastman : If we went to a conda-based environment, you can find `ppc64le` and `arm` versions of recent `cmake`:
https://anaconda.org/conda-forge/cmake/files

jaimergp:
Unpacking the conda tarballs might work in the meantime, but at least we could try to recreate the build environments using conda envs (with strict pinning to known working versions), instead of system-provided packages, if you want.

Alternatively, I don't know if this [apt repo ](https://apt.kitware.com/) includes extra architectures.

peastman:
Thanks.  If I can't get the current approach to work, I"ll trying using conda.  The Kitware repo doesn't have packages for ARM or PPC.

peastman:
I don't think conda is available for ARM.  At https://repo.anaconda.com/miniconda/ they have some very old (from 2015) builds for ARMv7, but nothing for ARM64.

peastman:
I'm kind of stuck on this.  If I install cmake with the command `make install`, it fails with permission errors (of course):

```
Install the project...
-- Install configuration: "Release"
CMake Error at Source/kwsys/cmake_install.cmake:41 (file):
  file cannot create directory: /usr/local/doc/cmake-3.17/cmsys.  Maybe need
  administrative privileges.
  cmake_install.cmake:42 (include)
Makefile:93: recipe for target 'install' failed
make: *** [install] Error 1
```

If instead I do `sudo make install` it gets further and lists lots of files it's installing, but then it still fails with a permission error:

```
CMake Error at cmake_install.cmake:150 (file):
  file failed to open for writing (Permission denied):
    /home/travis/build/openmm/openmm/cmake-3.17.1/install_manifest.txt
Makefile:93: recipe for target 'install' failed
make: *** [install] Error 1
```

peastman:
Finally, success!  Or rather, the PPC build succeeded and the ARM build was making steady progress until it reached the 50 minute limit and got stopped.  Now to see if I can make them faster.

jaimergp:
Drone.io has native ARM support: 

- Example build: https://cloud.drone.io/conda-forge/openmm-feedstock/13
- YML steps: https://github.com/conda-forge/openmm-feedstock/blob/0287adc0f23915e32218d78c7f85c9d21a8c194f/.drone.yml

peastman:
I can get it working by having travis build cmake from source.  It just takes a long time (about 8 minutes on PPC, 20 minutes on ARM).  Is there somewhere I could have travis upload those builds to, from which we could then just download them in the future?

jaimergp:
You could:

* Use [GitHub releases](https://docs.travis-ci.com/user/deployment/releases/) (only for tagged releases)
* Upload to [WeTransfer ](https://developers.wetransfer.com/) (7 days only, but free)
* [Upload to S3](https://docs.travis-ci.com/user/uploading-artifacts/) (no PRs, paid)
* [Use Google Drive](https://developers.google.com/drive/api/v3/manage-uploads)
* Publish to an `artifacts` label in Anaconda's omnia?

peastman:
I need somewhere I can upload to from travis using just a command line (so nothing requiring a web API) and without embedding any sensitive passwords or keys in the PR.  Once I've got the files off travis, there are lots of places I can host them.

jchodera:
We should just switch to GitHub Actions, which is lightning-fast and can store [build artifacts](https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts).

We can also easily add some Azure GPU builders that will run tests and benchmarks.

jchodera:
Travis appears to support [the upload of build artifacts to S3](https://docs.travis-ci.com/user/uploading-artifacts/). Could we just set up an S3 bucket?

jchodera:
Oh! Travis can automagically upload artifacts to [GitHub Releases too](https://docs.travis-ci.com/user/deployment/releases/).

peastman:
I was hoping for something rather simpler.  Just a public FTP server or some such so I could let it run a single time, post the compiled binary, and then put it somewhere where it will forever after be available.  Build artifacts require that we merge this change (which it is certainly *not* ready for!), create a release, and have the build successfully run for the release (and it isn't currently able to run successfully).  All of that just to transfer one file out of the VM.

peastman:
Finally, success!  I managed to get a newer cmake to work by upgrading to a newer OS version.

PR #2654, opened 2020-04-24
CompoundIntegrator handles checkpoints correctly

### LINKED ISSUE #2650: Context.{create|load}Checkpoint() should save/restore CustomIntegrator parameters

peastman:
Fixes #2650.


jchodera:
Does this also store the `currentIntegrator` value, and test that it can resume from the correct integrator when checkpointed in `currentIntegrator != 0`?

peastman:
It does now!

PR #2658, opened 2020-04-25
Improvements for ARM64 vector implementation

mark-mb:
Little optimizations and bug fixing

### DIFF
--- a/openmmapi/include/openmm/internal/vectorize8.h
+++ b/openmmapi/include/openmm/internal/vectorize8.h
@@ -75,55 +75,55 @@ public:
     void store(float* v) const {
         _mm256_storeu_ps(v, val);
     }
-    fvec8 operator+(const fvec8& other) const {
+    fvec8 operator+(fvec8 other) const {
         return _mm256_add_ps(val, other);
     }
-    fvec8 operator-(const fvec8& other) const {
+    fvec8 operator-(fvec8 other) const {
         return _mm256_sub_ps(val, other);
     }
-    fvec8 operator*(const fvec8& other) const {
+    fvec8 operator*(fvec8 other) const {
         return _mm256_mul_ps(val, other);
     }
-    fvec8 operator/(const fvec8& other) const {
+    fvec8 operator/(fvec8 other) const {
         return _mm256_div_ps(val, other);
     }
-    void operator+=(const fvec8& other) {
+    void operator+=(fvec8 other) {
         val = _mm256_add_ps(val, other);
     }
-    void operator-=(const fvec8& other) {
+    void operator-=(fvec8 other) {
         val = _mm256_sub_ps(val, other);
     }
-    void operator*=(const fvec8& other) {
+    void operator*=(fvec8 other) {
         val = _mm256_mul_ps(val, other);
     }
-    void operator/=(const fvec8& other) {
+    void operator/=(fvec8 other) {
         val = _mm256_div_ps(val, other);
     }
     fvec8 operator-() const {
         return _mm256_sub_ps(_mm256_set1_ps(0.0f), val);
     }
-    fvec8 operator&(const fvec8& other) const {
+    fvec8 operator&(fvec8 other) const {
         return _mm256_and_ps(val, other);
     }
-    fvec8 operator|(const fvec8& other) const {
+    fvec8 operator|(fvec8& other) const {
         return _mm256_or_ps(val, other);
     }
-    fvec8 operator==(const fvec8& other) const {
+    fvec8 operator==(fvec8 other) const {
         return _mm256_cmp_ps(val, other, _CMP_EQ_OQ);
     }
-    fvec8 operator!=(const fvec8& other) const {
+    fvec8 operator!=(fvec8 other) const {
         return _mm256_cmp_ps(val, other, _CMP_NEQ_OQ);
     }
-    fvec8 operator>(const fvec8& other) const {
+    fvec8 operator>(fvec8 other) const {
         return _mm256_cmp_ps(val, other, _CMP_GT_OQ);
     }
-    fvec8 operator<(const fvec8& other) const {
+    fvec8 operator<(fvec8 other) const {
         return _mm256_cmp_ps(val, other, _CMP_LT_OQ);
     }
-    fvec8 operator>=(const fvec8& other) const {
+    fvec8 operator>=(fvec8 other) const {
         return _mm256_cmp_ps(val, other, _CMP_GE_OQ);
     }
-    fvec8 operator<=(const fvec8& other) const {
+    fvec8 operator<=(fvec8 other) const {
         return _mm256_cmp_ps(val, other, _CMP_LE_OQ);
     }
     operator ivec8() const;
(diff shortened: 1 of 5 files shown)


peastman:
Thanks!  Could you describe the changes and the reasons for them?

Note that any changes to the API will also require the corresponding changes in vectorize_ppc.h and vectorize_pnacl.h.

cc @dwtowner who is currently working on other changes to the vectorization code in #2645.  The two of you might want to coordinate.

mark-mb:
The results of comparisons for vector instructions is a mask of ones or a mask of zeros. But if we convert those masks back to fvec4 or ivec4, we may end up using the wrong value after conversion. For example, when comparing two fvec4s:
 - In SSE we were returning a mask of ones interpreted as float (so -NaN, as it is not representable)
 - In Neon, we were returning a mask of ones converted from uint32 to float (so -1 represented in float).

Trying to use these values as masks later usually works, but there are some corner cases where it breaks:
 - In SSE, everything seems to work. Even if float -NaN converted to integer results in INT_MIN (0x80000000), this works because apparently blend() works when any bits of the element are set (no need to have a mask of all ones)
 - In Neon, if we do any operations on the fvec4 mask without converting back to integer, (for example, a binary & between a fvec4 value and a fvec4 mask), the resulting value will be wrong.

To cut on misunderstandings and things that (seem to) work by chance. I added a mvec4 (or "mask vector") type, which is basically a uint32_t vector, that can only be casted, not reinterpreted.

I have deliberately intended to make it work without needing to touch anything else (by means of implicit type conversions), because I don't have available PPC machines to test it. In any case, PPC and PNACL seem to get around these issues by returning an ivec4 for fvec4 comparison operations.

The other changes are very small optimizations in Arm vectorization:
 - Reduce any() from 4 to 1 instructions
 - Allow the compiler to optimize initialization of vector elements.

All these changes together give between 2 and 8% improvement (depending on the test, the machine, etc.) in my tests using benchmark.py.

peastman:
> In any case, PPC and PNACL seem to get around these issues by returning an ivec4 for fvec4 comparison operations.

Would it make sense to just have SSE and NEON do the same thing, rather than introducing a new type?

mark-mb:
> > In any case, PPC and PNACL seem to get around these issues by returning an ivec4 for fvec4 comparison operations.
> 
> Would it make sense to just have SSE and NEON do the same thing, rather than introducing a new type?

Maybe. But current fvec4 <-> ivec4 type conversions in Neon and SSE use conversion operations. So, if the resulting ivec4 is converted again into fvec4 (for example, to & an fvec4 with a mask), there will be issues. To fix this, I could:
 - Check if fvec4 <-> ivec4 conversions are used genuinely anywhere (value to value). If there are none, change the conversions into direct casting
 - Check all cases when the result of an fvec4 comparison is converted back to fvec4 to do further operations and fix those.

I opted to add a new type, and keep the changes contained in the vectorization part.

If you think that going through the codebase and fixing / clarifying / making more explicit the use of masks is better than adding a new type, let me know.

And maybe I mentioned PPC prematurely. It seems that if if there are conversions between types (from a mask represented in an ivec4 to fvec4), the same cases seems to apply.

peastman:
I think it's worth looking through the code and seeing.  More importantly, we want all implementations to behave consistently as much as possible.  So if there are functions where some implementations return a fvec4 and others return an ivec4, it would definitely be good to make them consistent.

dwtowner:
In my avx512 implementation I don't have ivec at all and doing the same in
other simd implementations should be easy. Everything can be done using an
int bit mask instead. It's faster that way too. The comparisons return a
mask and the blend accepts a mask.  The only other place i used ivec was in
the gather operation for table approximation and it is easy to just use
explicit intrinsic in just that place to convert when needed.

mark-mb:
Yes, the implementation in #2645 also solves these issues, and probably more elegantly than I'm doing. Unless you prefer to get those corner cases fixed in the meantime, I'll just drop my patch.

peastman:
Since you're both working on the same code at the same time, and solving similar problems, I think it makes sense to collaborate.  I'll leave it to the two of you to figure out the best way of splitting (or not splitting) the changes between PRs.

mark-mb:
I've rebased these changes and added a few more on top of the latest API changes

@dwtowner I made a few very minor modifications in the vectorization API. Hopefully these don't cause many issues with your ongoing refactoring:
 - All input vectors are passed as const reference
 - ARM returns ivecs as masks, like PPC and PNACL implementations. Current implementation was correct, but there were lots of unnecessary conversion instructions.

mark-mb:
> I'm not convinced of the need to use references parameters. For register-like types (which would include SIMD types) pass-by-value can be more efficient and easier for the compiler to optimise. If ARM64 benefits from pass-by-reference then I think changing it for that target is fine, but I don't think it should be applied across all targets.

In this particular case, I have no preference, I was just making the API consistent. The "old" API calls passed by reference, and the "new" API calls passed by value. If you think there are benefits in passing by value, I can change it to pass by value everywhere.

I haven't benchmarked this change in isolation, so I don't have numbers for it one way or the other.

dwtowner:
Here's a godbolt example showing the difference: https://godbolt.org/z/AFFrsK In that case the reference has forced the value into memory, when it could have stayed in a register. Obviously it depends upon the exact context (e.g., if the compiler inlines, then it can elide the memory access), but I think pass-by-value will have better potential for speed.

peastman:
Maybe benchmark both versions and see if one is measurably faster?

mark-mb:
I ran both versions (one passing by const ref, and one passing by value), and I found no significant difference in both ARM and Intel with GCC. If there is any noticeable difference in ICC, I'd have no issue to pick one way or the other.

I haven't looked at the disassembly of OpenMM, but I'd argue that any compiler with optimizations enabled (as this is the case), will inline these very small functions, and decide what is the best way to pass their inputs and get their outputs. In your particular example, the compiler doesn't have any context information to decide what to do, so it just treats them as value and reference respectively. But if you do an actual use of those functions (https://godbolt.org/z/PrJJzf), the compiler just inlines them. Even more, because I'm passing the same input and the functions are basically the same, it seems that it's just doing the operation once and copying the result.

Also @peastman, it seems that the ARM64 CI job timed out. Not sure if it can be retriggered, or what should be done about it.

dwtowner:
Absolutely. I agree that if the compiler can see the function definition at the call site it will likely do the right thing anyway. If it can't see the definition for any reason though, then it will do like my example, in which case passing-by-value will give slightly better code (although to be fair, the difference may be negligible).

peastman:
What's the status of this PR?  It sounds like we want to stick with passing by value?

mark-mb:
I took from this discussion that there is little to no difference on passing by value vs passing by const reference. From me, the PR is ready to merge.

If you prefer to change it to pass by value. I'll update the PR

dwtowner:
I think that passing by value will give the best code (on IA's ABI at
least) irrespective of whether the compiler inlines the function or not.
Although pass-by-reference will give identical code when inlined it will be
(marginally) less than optimal when that doesn't happen. So, I think
pass-by-value should be favoured where possible.

mark-mb:
Is there anything else you'd like to see to get this merged?

peastman:
Sorry, I'd been waiting for you to post that you were done with changes.  I'll take a look!

peastman:
Looks good.  Thanks!

PR #2661, opened 2020-04-27
[WIP] CPU: Refactored code to be generic across vector CPU platforms.

dwtowner:
Ewald and non-Ewald interactions now share a common code base, templated on
their interaction type.

The vec4 and vec8 implementations have been replaced by a single generic implementation
class which is templated on SIMD type. Currently works for SIMD4 and SIMD8 types, but
can be extended in future to support other types (e.g., AVX-512).

Modified runtime CPU support to lay groundwork for future SIMD types.

Pulled out some vector utility functions (gather pair, reduce),
and refactored the AVX CPU code to make use of them.


peastman:
Thanks!  I'll try to make a detailed pass through the code soon.  One general comment is that it should follow the code formatting used everywhere else in OpenMM.  Opening braces should always go on the same line, not the next line.  Comments at the top of classes or methods should use the format

```c++
/**
 * This function does something.
 */
```

rather than

```c++
/// This function does something.
```

Parameters should be documented with `@param` rather than `\param`

mark-mb:
Given my comments in #2658, I'm still concerned about type conversions that give the wrong value:

1) `const auto include = FVEC::expandBitsToMask(~exclusions[i]) & (r2 < cutoffDistance*cutoffDistance);`
2) `const auto t = (r>switchingDistance) & ((r-switchingDistance)*invSwitchingInterval);`

1) works (I think, from the top of my head), because the right part of the & is converted from fvec4 to ivec4 and used as a mask
2) works in SSE / AVX, but it doesn't work in the current implementation of Neon. The left part of the & isn't a mask, but the representation of -1 as float.

(The issue with 2) wasn't present in CpuNonBondedForceVec4, because that was using blend() instead of &)

Some possible solutions:
 - Get comparison operators to return ivec4's and have an ivec4 as input parameter of fvec4 operator &
 - Add a mask type that's only used for these operations and can be casted to fvec and ivec (as in my original patch)
 - ...

Also note that you are including "vectorize_sse.h" directly. That's probably why CI is not very happy.

openmmapi/include/openmm/internal/vectorize_neon.h
```diff
@@ -101,6 +109,14 @@ class fvec4 {
     void store(float* v) const {
         vst1q_f32(v, val);
     }
+
+    /// Store only the lower three elements of the vector.
+    void storeVec3(float* v) const {
```
peastman, line 114: This name is potentially confusing, since people will assume it refers to the `Vec3` class.  Could we call it something different?

openmmapi/include/openmm/internal/vectorize_neon.h
```diff
@@ -374,4 +417,37 @@ static inline fvec4 ceil(const fvec4& v) {
...
+///   X0 X1 X2 X3
+///   Y0 Y1 Y2 Y3
+///   Z0 Z1 Z2 Z3
+/// Each vector of values needs to be summed into a single value, and then stored into
+/// the output vector:
+///   output[0] = (X0 + X1 + X2 + X3)
+///   output[1] = (Y0 + Y1 + Y2 + Y3)
+///   output[2] = (Z0 + Z1 + Z2 + Z3)
+///   output[3] = undefined
+static inline fvec4 reduceToVec3(const fvec4 x, const fvec4 y, const fvec4 z)
```
peastman, line 445: Similarly here.

peastman:
vectorize_pnacl.h doesn't get tested by CI, since we aren't currently using it anywhere.  I do want to keep it working though.  The name is misleading: it really is just an implementation that uses the LLVM portable vector syntax.  You can test it locally by just replacing vectorize_sse.h with vectorize_pnacl.h.  It should work on all architectures.  This makes it useful for debugging.

openmmapi/include/openmm/internal/vectorize_sse.h
```diff
@@ -32,7 +32,7 @@
  * USE OR OTHER DEALINGS IN THE SOFTWARE.                                     *
  * -------------------------------------------------------------------------- */
 
-#include <smmintrin.h>
+#include <immintrin.h>
```
peastman, line 35: Are you sure this will work when AVX is disabled?

openmmapi/include/openmm/internal/vectorize_neon.h
dwtowner: Happy for a better name to be used, but not sure what it should be. Suggestions welcome. It's storing/reducing to 3 elements in memory, so Vec3 seemed the best fit.

openmmapi/include/openmm/internal/vectorize_sse.h
dwtowner: Should work as I never use other than this, although I don't use old compilers so maybe they break.

vectorize_sse is used by vectorize8, so when used in that context I wanted to be able to call AVX intrinsics. Using just the original smmintrin file it would have to change to:

#ifdef __AVX__
#include <immintrin.h>
#else 
#include <smmintrin.h>
#endif

Happy to do that if you'd prefer.

platforms/cpu/include/CpuNonbondedForce.h
```diff
@@ -203,8 +203,8 @@ class CpuNonbondedForce {
         float inverseRcut6Expterm;
         std::atomic<int> atomicCounter;
 
-        static const float TWO_OVER_SQRT_PI;
-        static const int NUM_TABLE_POINTS;
+        static constexpr float TWO_OVER_SQRT_PI = 1.128379167095513;
```
peastman, line 206: I think the original version of this was better: `TWO_OVER_SQRT_PI = (float) (2/sqrt(PI_M))`.  It's clearer, and constants embedded as long strings of digits are a common source of errors.

openmmapi/include/openmm/internal/vectorize_sse.h
peastman: That's probably safest, thanks.

openmmapi/include/openmm/internal/vectorize8.h
```diff
@@ -51,6 +51,19 @@ class fvec8 {
...
     fvec8(__m256 v) : val(v) {}
     fvec8(const float* v) : val(_mm256_loadu_ps(v)) {}
+
+    /** Create a vector by gathering individual indexes of data from a table. Element i of the vector will
+     * be loaded from table[idx[i]].
+     * @param table The table from which to do a lookup.
+     * @param indexes The indexes to gather.
+     */
+    fvec8(const float* table, const int idx[8])
+    {
```
peastman, line 61: Still some incorrect formatting here.  Opening brace should be at the end of the line.  Also in various other places below.

jchodera:
> One general comment is that it should follow the code formatting used everywhere else in OpenMM. 

@peastman : Do you want to add a few sentences about formatting to the [how to contribute guide](https://github.com/openmm/openmm/blob/master/CONTRIBUTING.md)? Seems like we overlooked this!

platforms/cpu/include/CpuNonbondedForce.h
dwtowner: sqrt isn't a constexpr function though. The code generated from the original was doing a runtime conversion from float to vector where it was used.

dwtowner:
I don't much like the side-effecting way of using & with different types either.

We could just swap both those cases to using blend instead? That's more explicit about intent.

The return type of the comparison operators won't necessarily be an ivec (e.g., AVX-512 returns int for those operators and uses int for blends resulting from those comparisons). Each platform should be free to choose the most appropriate type.

Something like this might work:

struct Simd {
  // Needed to explicit declare mask variables in those places where auto fails.
  using ElementMask = /* Your type here */

  // Needed for converting the exclusion mask into something that can be manipulated in vector elements.
  static ElementMask getMaskFromInt(int);

  // Comparisons return some suitable type
  ElementMask operator<= (Simd, Simd);

  blend(Simd, Simd, ElementMask);

  // Combine masks - not needed if the mask is an int, but might be required for other platform masks.
  // ElementMask operator& (ElementMask, ElementMask)
};

mark-mb:
> We could just swap both those cases to using blend instead? That's more explicit about intent.

That would be more explicit and clear and less error-prone.

I'm not sure about the performance difference (if any) between calling the SIMD intrinsic blend vs the intrinsic and. Maybe something like "blendWithZero" would allow to optimize this case?

> Something like this might work:
> 
> struct Simd {
> // Needed to explicit declare mask variables in those places where auto fails.
> using ElementMask = /* Your type here */
> 
> // Needed for converting the exclusion mask into something that can be manipulated in vector elements.
> static ElementMask getMaskFromInt(int);
> 
> // Comparisons return some suitable type
> ElementMask operator<= (Simd, Simd);
> 
> blend(Simd, Simd, ElementMask);
> 
> // Combine masks - not needed if the mask is an int, but might be required for other platform masks.
> // ElementMask operator& (ElementMask, ElementMask)
> };

That would also work. Not sure what is best, so I'll let you decide

peastman:
According to the Intel intrinsics guide, `_mm_and_ps()` has a latency of 1 and througput of 1.  `_mm_blendv_ps()` has a latency of 2 and throughput of 0.7.  So changing that to blend would hurt performance.  I haven't checked other architectures, but it will likely be slower on them too.

platforms/cpu/include/CpuNonbondedForce.h
peastman: No, it was doing a conversion at startup time when the field was initialized.

dwtowner:
> 7

Okay, the intrinsics guide is wrong. I've reported that error to get that fixed. For blend typical performance is latency of 1 and throughput of 0.5 (i.e., the same as an AND).

peastman:
On Windows compilation is failing with this error:

```
D:\a\1\s\openmmapi\include\openmm/internal/vectorize8.h(184): error C2678: binary '&': no operator found which takes a left-hand operand of type '__m256i' (or there is no acceptable conversion)
```

On Mac there's a linker error:

```
[ 68%] Linking CXX shared library ../../../libOpenMMCPU.dylib
Undefined symbols for architecture x86_64:
  "OpenMM::CpuNonbondedForce::NUM_TABLE_POINTS", referenced from:
      OpenMM::CpuNonbondedForce::threadComputeDirect(OpenMM::ThreadPool&, int) in CpuNonbondedForce.cpp.o
      OpenMM::CpuNonbondedForce::exptermsApprox(float) in CpuNonbondedForce.cpp.o
      OpenMM::CpuNonbondedForce::dExptermsApprox(float) in CpuNonbondedForce.cpp.o
      OpenMM::CpuNonbondedForce::erfcApprox(float) in CpuNonbondedForce.cpp.o
ld: symbol(s) not found for architecture x86_64
```

And on ARM it's failing because it's trying to compile the AVX code:

```
In file included from /home/travis/build/openmm/openmm/platforms/cpu/src/CpuNonbondedForceVec8.cpp:26:0:
/home/travis/build/openmm/openmm/openmmapi/include/openmm/internal/vectorize8.h:36:10: fatal error: immintrin.h: No such file or directory
 #include <immintrin.h>
          ^~~~~~~~~~~~~
compilation terminated.
```

platforms/cpu/include/CpuNonbondedForce.h
dwtowner: I realise that the initialisation did the first conversion, but on SSE and AVX also did a float-load, followed by a broadcast at each point of use of that value too. On IA at least, it is better to use a broadcast-load instruction directly. Switching to constexpr fixed that.

Doesn't really matter though, as for the moment I have backed out this change as it was breaking the MAC build.

openmmapi/include/openmm/internal/vectorize8.h
dwtowner: Done

dwtowner:
I've removed the use of & and replaced by an explicit blendZero function as suggested by @mark-mb . On SSE and AVX* the belendZero doesn't make any difference, but it provides a hook for a slightly more optimal code sequence on other platforms if that is possible.

I've addressed all 3 CI build issues, although getting past them may uncover more.

I'm doing local builds with LLVM/NaCl so they shouldn't cause any problems.

peastman:
Windows is still failing with the same error message:

```
D:\a\1\s\openmmapi\include\openmm/internal/vectorize8.h(183): error C2678: binary '&': no operator found which takes a left-hand operand of type '__m256i' (or there is no acceptable conversion)
```

ARM now has a different error:

```
/home/travis/build/openmm/openmm/platforms/cpu/src/CpuNonbondedForceVec8.cpp: In function ‘OpenMM::CpuNonbondedForce* createCpuNonbondedForceVec8()’:
/home/travis/build/openmm/openmm/platforms/cpu/src/CpuNonbondedForceVec8.cpp:52:10: error: ‘OpenMMException’ was not declared in this scope
    throw OpenMMException("Internal error: OpenMM was compiled without AVX support");
          ^~~~~~~~~~~~~~~
/home/travis/build/openmm/openmm/platforms/cpu/src/CpuNonbondedForceVec8.cpp:52:10: note: suggested alternative: ‘_ZTISt9exception’
    throw OpenMMException("Internal error: OpenMM was compiled without AVX support");
          ^~~~~~~~~~~~~~~
          _ZTISt9exception
```

PPC is failing with this error:

```
/home/travis/build/openmm/openmm/platforms/cpu/./include/CpuNonbondedForceFvec.h:234:17: error: invalid initialization of reference of type ‘const __m128i& {aka const __vector(4) int&}’ from expression of type ‘const fvec4’
         if (!any(include))
              ~~~^~~~~~~~~
In file included from /home/travis/build/openmm/openmm/openmmapi/include/openmm/internal/vectorize.h:38:0,
                 from /home/travis/build/openmm/openmm/platforms/cpu/./include/CpuNonbondedForce.h:32,
                 from /home/travis/build/openmm/openmm/platforms/cpu/./include/CpuNonbondedForceFvec.h:28,
                 from /home/travis/build/openmm/openmm/platforms/cpu/src/CpuNonbondedForceVec4.cpp:25:
/home/travis/build/openmm/openmm/openmmapi/include/openmm/internal/vectorize_ppc.h:346:20: note: in passing argument 1 of ‘bool any(const __m128i&)’
 static inline bool any(const __m128i& v) {
                    ^~~
```

The Mac error does seem to be fixed!

platforms/cpu/include/CpuNonbondedForce.h
peastman: Maybe create the fvec version of it inside the method, before the start of the loop that uses it?

dwtowner:
I've now fixed all the build errors themselves, but the ARM CI build seems to hang. I don't have an ARM machine so can't reproduce locally to isolate the problem. Also, the test in question seems to be a reference test; does that even call the CPU code?

peastman:
Let me try restarting that build and see if it's reproducible.

peastman:
It passed when I reran it, so all tests are now looking good!  Is this ready to merge?

I tested this on my desktop machine, which is a fairly typical 4 core, 8 thread i7.  For DHFR, the performance went from 16.16 to 16.72 ns/day, or about a 3.4% improvement.  For ApoA1 it went from 3.73 to 3.81 ns/day, or about a 2.2% improvement.

dwtowner:
Yes, ready to merge.

I removed quite a few of the optimisations to make it easier to make the
code generic across all the different platforms. Now that the groundwork is
laid, it should be easier to add both generic and platform specific
optimisations on top of those changes. I'll start merging those other
optimisations in as new PRs.

peastman:
Sounds good, thank you!

jaimergp:
I think this is breaking all the nightlies for [Windows in `omnia-dev`](https://dev.azure.com/OmniaMD/conda-dev-recipes/_build/results?buildId=235&view=logs&j=7a85d380-999d-5da8-5957-f1313d17ab9c&t=8184c4ff-f7d2-5957-d3c7-ed8493067802&l=3573):

```
jom: C:\Miniconda\conda-bld\openmm_1588298931782\work\build\platforms\cpu\sharedTarget\CMakeFiles\OpenMMCPU.dir\build.make [platforms\cpu\sharedTarget\CMakeFiles\OpenMMCPU.dir\__\src\CpuNonbondedForceVec8.cpp.obj] Error 2
jom: C:\Miniconda\conda-bld\openmm_1588298931782\work\build\CMakeFiles\Makefile2 [platforms\cpu\sharedTarget\CMakeFiles\OpenMMCPU.dir\all] Error 2
jom: C:\Miniconda\conda-bld\openmm_1588298931782\work\build\Makefile [all] Error 2
[conda-build-all] Error: one or more packages failed to build
  openmm-7.5.0-py36_cuda80_1: Command '['cmd.exe', '/d', '/c', 'conda_build.bat']' returned non-zero exit status 2.
```

peastman:
It looks like these are the error messages:

```
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(340): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(341): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(342): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(343): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(344): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(345): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(346): error C3861: '_mm256_extract_epi32': identifier not found
%SRC_DIR%\openmmapi\include\openmm/internal/vectorize8.h(347): error C3861: '_mm256_extract_epi32': identifier not found
```

And I don't see that intrinsic listed at https://docs.microsoft.com/en-us/cpp/intrinsics/x64-amd64-intrinsics-list?view=vs-2019?

jaimergp:
Looks like it's an [Intel thing](https://software.intel.com/en-us/cpp-compiler-developer-guide-and-reference-mm256-extract-epi8-16-32-64).

dwtowner:
Sorry, I rarely use non-intel compilers so didn't realise it wasn't a universally supported intrinsic.  I've switched to a different method (which produces identical code) in #2668.

PR #2662, opened 2020-04-28
Use coarser granularity to split work in PME kernels

mark-mb:
Reduce necessary synchronization between threads, and improve memory locality.

### DIFF
--- a/plugins/cpupme/src/CpuPmeKernels.cpp
+++ b/plugins/cpupme/src/CpuPmeKernels.cpp
@@ -66,96 +66,102 @@ static void spreadCharge(float* posq, float* grid, int gridx, int gridy, int gri
     float posInBox[4] = {0,0,0,0};
     memset(grid, 0, sizeof(float)*gridx*gridy*gridz);
 
-    int i = threadIndex;
+    const int groupSize = max(1, numParticles / (10 * numThreads));
+    int start = groupSize * threadIndex;
     while (true) {
         if (!deterministic)
-            i = atomicCounter++;
-        if (i >= numParticles)
+            start = atomicCounter.fetch_add(groupSize);
+
+        if (start >= numParticles)
             break;
 
-        // Find the position relative to the nearest grid point.
+        int end = min(start + groupSize, numParticles);
+        for (int i = start; i < end; ++i) {
+            // Find the position relative to the nearest grid point.
 
-        fvec4 pos(&posq[4*i]);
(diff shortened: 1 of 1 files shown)


peastman:
What sort of benchmarks have you tested this on?  My experience is that statically splitting the work leads to worse scaling with number of cores, since you often get just one thread that keeps running long after all the others have finished.

If the atomic counter is actually adding significant overhead, we can reduce that pretty easily without losing the benefit.  Instead of having threads grab one particle at a time, break the particles into `numThreads*10` blocks and have threads grab a block of particles at a time.  That's fine grained enough to give pretty good balancing between threads.  And since on average each thread only accesses the counter 10 times in total, the overhead is completely negligible.  You can see an example of doing that in CpuNonbondedForce:

https://github.com/openmm/openmm/blob/d51ef9e650836517690b9481f12cf3f0e44e24e6/platforms/cpu/src/CpuNonbondedForce.cpp#L438-L444

mark-mb:
> What sort of benchmarks have you tested this on?

I tested it on the benchmarks available in examples/benchmarks.py. I didn't see any slowdown in platforms with very few cores, and I saw a significant improvement in both Intel and Arm for platforms with more cores.

For Intel with 96 cores, I got 15% improvement in pme, 2% (could be noise) in apoa1pme, and 30% in apoa1ljpme

For Arm with 16 cores, I got 35% improvement in pme, 35% in apoa1pme and 50% in apoa1ljpme

> My experience is that statically splitting the work leads to worse scaling with number of cores, since you often get just one thread that keeps running long after all the others have finished.

That depends on the workload. In this particular case, every iteration is very homogeneous in the amount of work, because the number of iterations in the internal loops are fixed.

If the number of iterations to be done for each particle was variable depending, for example, on the input data, then I would agree with you.

> If the atomic counter is actually adding significant overhead, we can reduce that pretty easily without losing the benefit. Instead of having threads grab one particle at a time, break the particles into `numThreads*10` blocks and have threads grab a block of particles at a time. That's fine grained enough to give pretty good balancing between threads. And since on average each thread only accesses the counter 10 times in total, the overhead is completely negligible.

If you are still not convinced that this implementation gives improvements, I'll do that.

peastman:
Give it a try and see.  I suspect it will do at least as well, and perhaps better.

mark-mb:
The results are very similar. There seems to be a slight improvement in some cases. I pushed the new version

peastman:
Looks great.  Thanks!

PR #2666, opened 2020-04-30
change CMakeLists.txt to use syntax supported by cmake versions < 3.12

csquaredphd:
The `add_compile_definitions()` was added in cmake version 3.12.  

However, CMakeLists.txt has the minimum version set to 3.1: `CMAKE_MINIMUM_REQUIRED(VERSION 3.1)`.  

So, either the syntax or the minimum should be changed to avoid errors.  This PR does the former.

It has been built and tested on an ARM64 platform running Ubuntu 18.04.4 LTS, which provides cmake version 3.10.2.


peastman:
Looks good, thanks!

PR #2667, opened 2020-05-01
Replace inline assembly with compiler intrinsic for _IRR_DEBUG_BREAK_IF

thtrummer:
Inline assembly isn't supported by the x64 Microsoft compiler (msvc).

### DIFF
--- a/libraries/irrxml/include/irrTypes.h
+++ b/libraries/irrxml/include/irrTypes.h
@@ -2,6 +2,8 @@
 // This file is part of the "Irrlicht Engine".
 // For conditions of distribution and use, see copyright notice in irrlicht.h
 
+// NOTE: This file has been changed from the orginal source distribution.
+
 #ifndef __IRR_TYPES_H_INCLUDED__
 #define __IRR_TYPES_H_INCLUDED__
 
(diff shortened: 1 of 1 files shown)


peastman:
What compiler are you having problems with?  We've done builds with just about every version of Visual Studio since 2008, and this works with all of them.

thtrummer:
`Microsoft (R) C/C++ Optimizing Compiler Version 19.25.28614 for x64`
`Copyright (C) Microsoft Corporation.  All rights reserved.`

`openmm\libraries\irrxml\include\irrArray.h(227,3): error C4235: nonstandard extension used: '__asm' keyword not supported on this architecture`

However, the problem isn't so much the compiler version as trying a 64-bit build (which the library may not support?). Microsoft has never supported inline assembly for 64-bit targets (AFAIR).

peastman:
I see what's happening: that blocked is ifdef'ed with

```c++
 #if defined(WIN32) && defined(_MSC_VER) && defined(_DEBUG)
```

We never compile in debug mode on Windows, because debug libraries are incompatible with anything built in release mode.  That's why we've never seen this.

Anyway, this change looks reasonable.  The irrXML license includes the clause

>   2. Altered source versions must be plainly marked as such, and must not be
>    misrepresented as being the original software.

Could you add a comment to that line documenting that you've changed it, and also one at the top of the file stating that it's modified?

thtrummer:
I've added the comments as requested. Feel free to change them as you see fit.

peastman:
Looks good, thanks!

PR #2674, opened 2020-05-08
AMOEBA Update, 8 May 2020

jayponder:
Implements stretch-torsion, angle-torsion, vdw special pairs as part of the AMOEBA plugin.


peastman:
Could you describe the new features that have been added to AmoebaVdwForce and how they're used?

Are AmoebaStretchTorsionForce and AmoebaAngleTorsionForce really needed?  They look to me like they could be implemented in just a few lines with [CustomCompoundBondForce](http://docs.openmm.org/latest/api-c++/generated/OpenMM.CustomCompoundBondForce.html#_CPPv2N6OpenMM23CustomCompoundBondForceE).  What are the missing features that require creating new classes for them?  (The AMOEBA implementation currently has a bunch of specialized force classes that exist only because they predate the custom forces.  My goal is to eventually get rid of them.)

jayponder:
The changes to the AmoebaVdwForce allow use of Vdw "special pairs". This means that specific radius and well-depth values can be specified for a pair of atom types, and those values will override what would results from use of the general Vdw "combining rules". It is primarily of use where the Vdw interactions are strong and the Vdw functional form and combining rules begin to break down. As one example, use of special pair Vdw parameters for interactions between simple monoatomic cations and anions is the "magic bullet" that fixes the behavior of AMOEBA for "concentrated" salt solutions (beyond the lower mM regime). Use of fixed special pair values provides agreement with activity coefficient data,, for example, up the saturation limit of many simple inorganic salt solutions.

As far as we can tell, OpenMM is not set up to use Vdw parameters other than from a combining rule. The method used in the pull request is to construct and then use a reduced matrix of Vdw parameters between the pairs of atom types/classes in the system being computed.

jayponder:
I agree the bond-torsion and angle-torsion terms COULD perhaps be implemented as custom force terms. But the 360 degree periodicity of the torsions underneath the coupling to a second energy term (bond or angle) makes it fairly messy to implement this with the "1-line" format required by the CustomForce mechanism. Currently all the other AMOEBA terms are implemented explicitly as reference code and CUDA code, including "cross terms" such as stretch-bend. Just for consistency it would seem appropriate to treat the previously unimplemented stretch-torsion and angle-torsion terms in the same fashion as all the other AMOEBA terms.

We would also note that exactly these same energy terms will be incorporated into the AMOEBA+ and HIPPO models as they are extended to general classes of molecules. (Both stretch-torsion and angle-torsion are used mostly to treat "anomeric effects" that are not handled by any other terms in AMOEBA.)

peastman:
I'm having a bit of trouble understanding the intended use of the API.  For example, what is `addCondensedType()` for?  As far as I can tell, there's no information associated with a type.  It's just an arbitrary integer.  Then there's `getUsePairwiseVdw()`.  The name is very confusing, since the interaction is always pairwise.  And shouldn't `setUseLennardJones()` take a `bool` argument to say whether to use it?

> But the 360 degree periodicity of the torsions underneath the coupling to a second energy term (bond or angle) makes it fairly messy to implement this with the "1-line" format required by the CustomForce mechanism.

Could you describe the exact form of the potential function for these terms?

peastman:
Could you post the exact form of the potential functions?  I'm pretty sure these can be implemented with custom forces, and that will be a far simpler and more maintainable way of doing it.

peastman:
Based on the documentation you sent me, I believe the following should be roughly correct for the stretch-torsion term.  I haven't tested this, but I think it should be about right:

```python
expression = """
k11*(distance(p1, p2)-length1)*(1+cos(1*phi+phi1)) +
k12*(distance(p1, p2)-length1)*(1+cos(2*phi+phi2)) +
k13*(distance(p1, p2)-length1)*(1+cos(3*phi+phi3)) +
k21*(distance(p2, p3)-length2)*(1+cos(1*phi+phi1)) +
k22*(distance(p2, p3)-length2)*(1+cos(2*phi+phi2)) +
k23*(distance(p2, p3)-length2)*(1+cos(3*phi+phi3)) +
k31*(distance(p3, p4)-length3)*(1+cos(1*phi+phi1)) +
k32*(distance(p3, p4)-length3)*(1+cos(2*phi+phi2)) +
k33*(distance(p3, p4)-length3)*(1+cos(3*phi+phi3));
phi=dihedral(p1, p2, p3, p4)
"""
force = CustomCompoundBondForce(4, expression)
force.addPerBondParameter("length1")
force.addPerBondParameter("length2")
force.addPerBondParameter("length3")
force.addPerBondParameter("phi1")
force.addPerBondParameter("phi2")
force.addPerBondParameter("phi3")
for i in range(3):
  for j in range(3):
    force.addPerBondParameter("k%d%d" % (i+1, j+1))
```

The angle-torsion term is very similar but a little bit simpler:

```python
expression = """
k11*(angle(p1, p2, p3)-angle1)*(1+cos(1*phi+phi1)) +
k12*(angle(p1, p2, p3)-angle1)*(1+cos(2*phi+phi2)) +
k13*(angle(p1, p2, p3)-angle1)*(1+cos(3*phi+phi3)) +
k21*(angle(p2, p3, p4)-angle2)*(1+cos(1*phi+phi1)) +
k22*(angle(p2, p3, p4)-angle2)*(1+cos(2*phi+phi2)) +
k23*(angle(p2, p3, p4)-angle2)*(1+cos(3*phi+phi3)) +
phi=dihedral(p1, p2, p3, p4)
"""
force = CustomCompoundBondForce(4, expression)
force.addPerBondParameter("angle1")
force.addPerBondParameter("angle2")
force.addPerBondParameter("phi1")
force.addPerBondParameter("phi2")
force.addPerBondParameter("phi3")
for i in range(2):
  for j in range(3):
    force.addPerBondParameter("k%d%d" % (i+1, j+1))
```

peastman:
Any thoughts on the comments above?  I'd like to keep this moving forward so we can hopefully get the vdw changes into 7.5.

jayponder:
In answer to the questions above regarding the vdw special pair mechanism... The names of the routines are perhaps not optimal, and can of course be changed as people see fit. I agree that some of the names are not very correctly descriptive.

As mentioned above, the point of the vdw special pair code is to allow use of vdw parameters (ie, radius and well depth for L-J or for the buffered 14-7 used by AMOEBA) that are not derived from "combining rules". The way this works in canonical Tinker is to store a square matrix where the row and column labels are the unique vdw atom classes found in the system being computed. First we compute and store at each matrix element the two combining rule-based vdw parameters (vdw atom size and well depth) for the interaction of that element's row and column vdw classes. In a second pass, we loop over all the vdw parameters for "special pairs" where the parameters for interaction of a particular pair of vdw classes differ from what is given by the combining rules. The parameters at elements of the parameter square matrix are replaced with these vdw "special pair" values as appropriate. Then when computing the vdw energy/gradient in Tinker by looping over a neighbor list, we look up in this stored square matrix the vdw parameters for each pair interaction.

This is more-or-less what the code in the pull request is doing. The pull request code builds the parameter matrix over the "condensed types" for the system being computed.

peastman:
I think I'm starting to understand, and I see why I was confused before.  It's adding two distinct features: 1) the option to use LJ instead of 14-7, and 2) the ability to use nonstandard parameters for particular pairs (like NBFIX terms in CHARMM).  "Condensed types" really just means atom types.  `addCondensedType()` doesn't mean to define a new type.  It's called once per particle, and specifies what type to use for each one.  Is that all correct?

How about this as an API?  The class can be used in two distinct modes.  In one mode, which corresponds to how it works right now, interaction parameters are defined per-particle and always use a combining rule.  In the alternate mode, they're defined per-type.  We'll add an alternate version of `addParticle()` that takes a type in place of sigma and epsilon.  Which mode it operates in depends on which version you use.  You must use the same version for all particles.

We'll add the following methods which are used when operating in the type-based mode.

```c++
int getNumParticleTypes() const;
int addParticleType(double sigma, double epsilon);
void getParticleTypeParameters(int typeIndex, double& sigma, double& epsilon) const;
void setParticleTypeParameters(int typeIndex, double sigma, double epsilon);

int getNumTypePairs() const;
int addTypePair(int type1, int type2, double sigma, double epsilon);
void getTypePairParameters(int pairIndex, int& type1, int& type2, double& sigma, double& epsilon) const;
void setTypePairParameters(int pairIndex, int type1, int type2, double sigma, double epsilon);
```

Each "type pair" defines the parameters to use between a particular pair of types.  All other interactions would use the combining rule with the parameters for the two types.

How does that sound?

jayponder:
Yes, I think that sounds like it should work. Sorry, I should have made the analogy to CHARMM NBFIX earlier, as that's more-or-less what is being done here.

peastman:
Ok, I can try implementing that.  I don't think it should be too difficult.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceVdwForce.cpp
```diff
@@ -244,6 +244,16 @@ double AmoebaReferenceVdwForce::calculatePairIxn(double combinedSigma, double co
...
 
+    if (usesLJ) {
+        double pp1 = combinedSigma / r_ij;
+        double pp2 = pp1 * pp1;
+        double pp3 = pp2 * pp1;
+        double pp6 = pp3 * pp3;
+        double pp12 = pp6 * pp6;
+        energy = combinedEpsilon * (pp12 - 2.0 * pp6);
+        dEdR = combinedEpsilon * (pp12 - pp6) * (-12.0) / r_ij;
+    }
```
peastman, line 255: This isn't the correct form for Lennard-Jones.  It should be `4*epsilon*((sigma/r)^12 - (sigma/r)^6)`.  The function used here is what you would use if you were specifying the distance by rmin rather than sigma.

jayponder: Well... Tinker uses Rmin instead of sigma as it's internal size descriptor for L-J, and for all of the two-parameter "Vdw" functions. In fact, the "sigma trick" only works for L-J, so using the Rmin values is in some sense more general. (In other words, only L-J has a "sigma", but essentially all two-parameter Vdw forms have an "Rmin".) The code in the pull request is correct for use in Tinker-OpenMM, where the parameters are passed in from Tinker. But it all depends on where the parameters will be coming from.

peastman: Sigma and rmin are both fine ways of parametrizing LJ.  But the variable is called `combinedSigma`, and all the public functions in AmoebaVdwForce likewise refer to it as sigma.  If we ask the user to specify sigma but then interpret it as rmin instead, the result will be wrong.

jayponder: Yes, agreed. The variable should be called combinedRmin or something similar, and any other such references to Sigma should be Rmin, as that is really very confusing. Again Tinker does always use Rmin internally as, for example, there is no clean definition of a "sigma" for the buffered 14-7 function in AMOEBA. Sorry for the confusion.

peastman: I don't think we can rename it.  The public API has always referred to it as sigma, and also has methods like `setSigmaCombiningRule()`.  We can't change those without breaking compatibility.  So I'll keep those as they are and treat it as being sigma for LJ.  Just remember to scale your rmin values by 2^(-1/6) to convert them to sigmas when calling `addParticle()`.

jayponder: Sure, that's fine. I keep forgetting that the AMOEBA plugin is not standalone code :) Just do whatever makes sense to you in the broader context of OpenMM, and we'll make the changes in our interface code, etc. to match. Thanks for looking at this carefully, and for all your help.

peastman:
This is superseded by #2765.

Issue #2676, opened 2020-05-11
Neighbour exclusions masks are too small for wider SIMD


dwtowner:
I've added AVX-512 support to my local branch as a new fvec16 type. It took me a while to figure out why it didn't work properly, but I've now realised that the neighbour exclusion list returns a char for each block, which is too few bits for AVX-512. Somehow the neighbour exclusion list needs to be bumped up to be able to cope with 16 exclusion bits. Is there any reason why the type of the neighbor exclusion list couldn't be changed to something bigger than char? What I mean is, this:

std::vector<std::vector<WIDER_TYPE> > blockExclusions;

Any gotchas to look out for if I did this sort of change?

peastman:
I think I used char just because that was wide enough for all types we support.  Most likely we could widen it to short or even int with no effect on performance.  Just test it to make sure it doesn't slow anything down .

dwtowner:
Tested with int16_t. No slow-down on x86. I've submitted a PR.

PR #2686, opened 2020-05-18
CPU: Widen mask for exclusion bits (#2676)

dwtowner:
The exclusion mask was 8-bits wide, but future CPU support will require
more bits than this (e.g., AVX-512 will be 16-bit).


peastman:
Looks good.  The test failure is a stochastic error in an OpenCL test, not related to this.

PR #2687, opened 2020-05-19
Update DrudeNoseHooverIntegratorProxy.cpp

thtrummer:
Add header for alternative operator spellings. This is required by msvc when standard conformance mode is not enabled (DrudeNoseHooverIntegratorProxy.cpp(53): error C2065: 'not': undeclared identifier).


peastman:
I think it would be better to just change the use of `not` to `!`.  This header doesn't exist anymore in the latest version of the C++ standard.

PR #2690, opened 2020-05-21
Replaced alternative spelling of not-operator with '!'

thtrummer:
This is required by msvc when standard conformance mode is not enabled (DrudeNoseHooverIntegratorProxy.cpp(53): error C2065: 'not': undeclared identifier).


peastman:
Thanks!

PR #2692, opened 2020-05-21
Cpu generic vector test

dwtowner:
Pulled generic vector tests out into a separate header file which can then be easily reused for all available SIMD types. Currently only vec4/vec8 are tested, but other widths can be easily created by
adding extra test files (e.g., AVX-512). Note that the tests mostly check the API required by the CpuNonbondedFvec class, and ignore other APIs (e.g., ivecX). Some of the original tests have been left in place to verify the original APIs, but older tests have been removed where replaced by the new API.

### DIFF
--- a/openmmapi/include/openmm/internal/vectorize8.h
+++ b/openmmapi/include/openmm/internal/vectorize8.h
@@ -46,7 +46,7 @@ class fvec8 {
 public:
     __m256 val;
 
-    fvec8() {}
+    fvec8() = default;
     fvec8(float v) : val(_mm256_set1_ps(v)) {}
     fvec8(float v1, float v2, float v3, float v4, float v5, float v6, float v7, float v8) : val(_mm256_set_ps(v8, v7, v6, v5, v4, v3, v2, v1)) {}
     fvec8(__m256 v) : val(v) {}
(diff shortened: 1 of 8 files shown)


peastman:
Thanks, and sorry for taking a while to review this!  Just to clarify your comment:

> Note that the tests mostly check the API required by the CpuNonbondedFvec class, and ignore other APIs (e.g., ivecX). Some of the original tests have been left in place to verify the original APIs, but older tests have been removed where replaced by the new API.

Is this still testing all features that were tested before?  What tests have been removed?

dwtowner:
TL;DR; Everything that was tested, still is tested.

I have replaced many of the original tests with a new SIMD-width-agnostic way of testing. For example, the new test structure goes through all the maths functions (min, max, abs, etc), arithmetic, and utility functions (transpose, masking, gather-load, etc.), and does so for all supported SIMD widths. I have removed the older tests which do the same thing as they have been superceded. I have left in place the tests for ivec4 and ivec8 because my new API doesn't use those or test them, and any future wider SIMD wouldn't need an ivec16 either, so I haven't bothered updating those tests, just left them as they were.

peastman:
Ok, looks good then.

Issue #2409, opened 2019-09-30
 CUDA - Error computing forces with CUDA platform


ALANLI0258:
I have just installed openmm, and cuda on win 10. After I run "python -m simtk.testInstallation", there is an error. Who can help me deal with it? 
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
OpenMM Version: 7.4
Git Revision: b71e92d9ccef7c98cfd6b862eb34095d94fa1b05

There are 4 Platforms available:

1 Reference - Successfully computed forces
2 CPU - Successfully computed forces
nvcc fatal   : Cannot find compiler 'cl.exe' in PATH
3 CUDA - Error computing forces with CUDA platform
4 OpenCL - Successfully computed forces

CUDA platform error: Error loading CUDA module: CUDA_ERROR_FILE_NOT_FOUND (301)

Median difference in forces between platforms:

Reference vs. CPU: 6.30486e-06
Reference vs. OpenCL: 6.75312e-06
CPU vs. OpenCL: 8.10386e-07

All differences are within tolerance.
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

peastman:
See https://stackoverflow.com/questions/8125826/error-compiling-cuda-from-command-prompt for solutions to this problem.

nvcc is the CUDA kernel compiler.  But it can't work on it's own.  It also requires a "host compiler" to help it out.  On Windows that means cl, which comes with Visual Studio.  So you need to be sure you have Visual Studio installed, and the PATH set correctly so that nvcc can find it.

informatorius:
Can openmm be adapted to not use CUDA SDK compiler but Nvidia CUDA runtime driver?
https://stackoverflow.com/questions/35343729/can-i-use-cuda-without-using-nvcc-on-my-host-code

peastman:
Yes.  Just set `OPENMM_CUDA_COMPILER` to be empty (or anything that isn't actually the path to nvcc).  It will see that nvcc isn't available and, as long as nvrtc is in the library path, it will use that.

The next release of OpenMM will prefer the runtime compiler over nvcc by default.

PR #2693, opened 2020-05-22
Removed name of variadic argument list

thtrummer:
Naming the argument list of a variadic macro is a GNU extension which is not supported by msvc. Since CUDA uses the system preprocessor this will fail to build kernels on Windows (fixes error C2010: '.': unexpected in macro parameter list).


thtrummer:
#2409 might be related.

peastman:
Thanks!

PR #2694, opened 2020-05-22
Fix invalid call to __host__ function in computeBondedForces

thtrummer:
Explicitly cast the second parameter to the type of the first one so the compiler can pick an overload that is supported in device code (fixes error: calling a \_\_host__ function("fmin<float, int, (int)0> ") from a \_\_global__ function("computeBondedForces") is not allowed).


thtrummer:
Depending on the magnitude of `size` it may make sense to cast both parameters to a `double`(even for the single precision case).

peastman:
Excellent!  Let's see if this makes Jenkins happy.

peastman:
@swails Jenkins seems to have stopped working again.  The webhook is getting posted to it, but it returns this response:

```html
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 400 Provided signature [0a201c714dc107c80b1d6587ab844f22174415fa] did not match to calculated</title>
</head>
<body><h2>HTTP ERROR 400 Provided signature [0a201c714dc107c80b1d6587ab844f22174415fa] did not match to calculated</h2>
<table>
<tr><th>URI:</th><td>/github-webhook/</td></tr>
<tr><th>STATUS:</th><td>400</td></tr>
<tr><th>MESSAGE:</th><td>Provided signature [0a201c714dc107c80b1d6587ab844f22174415fa] did not match to calculated</td></tr>
<tr><th>SERVLET:</th><td>Stapler</td></tr>
</table>
<hr><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.4.27.v20200227</a><hr/>

</body>
</html>
````

swails:
I re-registered all hooks and expanded the set of authors that can trigger a build with a PR.  Unfortunately, the plugin I'm forced to use now doesn't have as convenient of a mechanism for whitelisting a PR build.

thtrummer:
`<kernel>:115:24: error: variadic macros not supported in OpenCL
#define make_TempType3(...) make_double3(__VA_ARGS__)`

Did this work before with `#define make_TempType3(a...)` or was it broken as well?

peastman:
Thanks!  Will I need to do anything in the future to make it work?

> Has this worked before with #define make_TempType3(a...) or was it broken as well?

It worked before.  It looks like we may need to use one syntax for CUDA and the other for OpenCL.  You can do that by checking `#ifdef __CUDA_ARCH__`.

thtrummer:
> It worked before. It looks like we may need to use one syntax for CUDA and the other for OpenCL.

Is there a reason this shouldn't work for both cases?
`#define make_TempType3 make_double3`

peastman:
Maybe?  Try it out and see.

peastman:
@swails The jenkins webhook failed again with the same error.

jchodera:
> I re-registered all hooks and expanded the set of authors that can trigger a build with a PR. Unfortunately, the plugin I'm forced to use now doesn't have as convenient of a mechanism for whitelisting a PR build.

Hopefully the GitHub Actions + Azure runner that @jaimergp is working on will help supplant the need for Jenkins.

swails:
> Hopefully the GitHub Actions + Azure runner that @jaimergp is working on will help supplant the need for Jenkins.

They'll let you test on GPUs?

swails:
> @swails The jenkins webhook failed again with the same error.

I'll email you tonight with the changes/settings that need to be set for the Jenkins webhook.

peastman:
> They'll let you test on GPUs?

That's what we're hoping!

swails:
@peastman, I think I have something.  Jenkins is working fine for other projects I'm helping out with on GitHub, so I think there must be something different about how access to OpenMM is granted that's causing the problems.  In particular, it doesn't seem that Jenkins was able to recreate the necessary hooks in this repo.

Can you check the permissions you granted to your user token you sent me?  See [this webpage](https://support.cloudbees.com/hc/en-us/articles/234710368-GitHub-Permissions-and-API-token-Scopes-for-Jenkins) for the scope needed for the job.  While I used to use the ghprb plugin (GitHub Pull Request Builder), it was previously announced deprecated (and since revived, I think), so I switched to a multibranch pipeline.

I've highlighted the necessary scopes in an image below:

scopes

In particular, I need `repo` and `admin:repo_hook` permissions turned on for the access key I'm using.

Let me know when those changes are made and I'll try re-registering the hook from Jenkins.

peastman:
Ok, I turned on those two permission.

swails:
I created a new job for this that matched what I did for the other projects (before it was set to an organization folder, I narrowed the scope to just this repository) and then re-registered hooks (according to the Jenkins syslogs, the hook registration on this repo worked).

The new job is here: https://jenkins.jasonswails.com/job/openmm-github/job/openmm/

It did a full branch scan and is currently building every PR that's open right now, so it'll take awhile to work through that backlog.  Once it does, it should get back to working again.

peastman:
Thanks!  And it looks like this successfully fixed the errors on Jenkins. :)

Issue #2679, opened 2020-05-14
"Ghost" CustomNonBondedForce


Exarctus:
Hey,

I've been looking for a way to attach a custom non-bonded force to a set of atoms, without this force being added to the total force acting on those atoms. This force needs to be computed at every step during MD, and written to an output file.

Does anyone know if there's a clean way to do this? I was thinking this could be achieved with adding 3 per-particle parameters that get updated from the energy expression, but that doesn't seem like a clean way to go about it...

Many thanks for any help!

craabreu:
Hi @Exarctus 
If you need to compute the energy resulting from this `CustomNonbondedForce`, a possible workaround could go like this:

```py
cvforce = openmm.CustomCVForce('0')
cvforce.addCollectiveVariable('my_energy', my_custom_nonbonded_force)
system.addForce(cvforce)
```

Then, you can extend the `StateDataReporter` class so that it becomes able to report the value of this collective variable:

```py
class MyStateDataReporter(openmm.app.StateDataReporter):
   def __init__(self, file, report_interval, cvforce, **kwargs):
        super().__init__(file, report_interval, **kwargs)
        self._cvforce = cvforce

    def _constructHeaders(self):
        return super()._constructHeaders() + ['my_energy']

    def _constructReportValues(self, simulation, state):
        my_energy = self._cvforce.getCollectiveVariableValues(simulation.context)[0]
        return super()._constructReportValues(simulation, state) + [my_energy]
```

A caveat of this approach is that `my_custom_nonbonded_force` will be internally evaluated at every time step, even if `report_interval > 1`. This can be very time-consuming. Alternatively, you could define a `CustomIntegrator` which ignores forces of a given force group, and then use `context.getState` from time to time to evaluate the energy of this force group.

craabreu:
> I've been looking for a way to attach a custom non-bonded force to a set of atoms, without this force being added to the total force acting on those atoms.

If this is a small set of atoms, maybe a simpler solution would be to store their positions over time and just do post-processing.

peastman:
This question has come up a few times before.  People want to have a Force they can compute, but don't want it to affect integration.  Using a CustomIntegrator has generally been the best suggestion.  Perhaps we should add this as a feature to all the standard integrators as well?  There could be a method `setIntegrationForceGroups()` to tell it which groups should be evaluated when performing integration.

craabreu:
[...]

PR #2701, opened 2020-05-27
Added Integrator.setIntegrationForceGroups()

peastman:
See #2679.

### DIFF
--- a/openmmapi/include/openmm/Integrator.h
+++ b/openmmapi/include/openmm/Integrator.h
@@ -87,13 +87,13 @@ public:
     virtual void step(int steps) = 0;
     /**
      * Get which force groups to use for integration.  By default, all force groups
-     * are included.  This is interpreted as a set if bit flags: the forces from group i
+     * are included.  This is interpreted as a set of bit flags: the forces from group i
      * will be included if (groups&(1<<i)) != 0.
      */
     virtual int getIntegrationForceGroups() const;
     /**
      * Set which force groups to use for integration.  By default, all force groups
-     * are included.  This is interpreted as a set if bit flags: the forces from group i
+     * are included.  This is interpreted as a set of bit flags: the forces from group i
      * will be included if (groups&(1<<i)) != 0.
      */
     virtual void setIntegrationForceGroups(int groups);
--- a/openmmapi/include/openmm/Integrator.h
+++ b/openmmapi/include/openmm/Integrator.h
@@ -85,6 +85,18 @@ public:
      * @param steps   the number of time steps to take
      */
     virtual void step(int steps) = 0;
+    /**
+     * Get which force groups to use for integration.  By default, all force groups
+     * are included.  This is interpreted as a set if bit flags: the forces from group i
+     * will be included if (groups&(1<<i)) != 0.
+     */
+    virtual int getIntegrationForceGroups() const;
+    /**
+     * Set which force groups to use for integration.  By default, all force groups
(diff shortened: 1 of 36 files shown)


openmmapi/include/openmm/Integrator.h
```diff
@@ -85,6 +85,18 @@ class OPENMM_EXPORT Integrator {
      * @param steps   the number of time steps to take
      */
     virtual void step(int steps) = 0;
+    /**
+     * Get which force groups to use for integration.  By default, all force groups
+     * are included.  This is interpreted as a set if bit flags: the forces from group i
```
craabreu, line 90: Small typo in "set of bit flags", both here and in line 96.

craabreu:
@peastman Don't you believe that this new option should automatically change how potential energy is computed by calling `Context.getState(getPotentialEnergy=True)` without explicit group specification (and, consequently, the value reported by `StateDataReporter` with option `PotentialEnergy=True`)?

peastman:
I'm not sure there's any good way of doing that.  The argument to `getState()` is an int containing bit flags.  The default value is 0xFFFFFFFF, which indicates a specific set of groups.  Likewise any other possible default value would indicate a particular set of groups.  How would you make the API work?

craabreu:
How about simply moving the `setIntegrationForceGroups()` method from `Integrator` to `Context`? Maybe with another name such as `setActiveForceGroups()`, so that one of its purposes could be the definition of `getState()`'s default bit flag.

peastman:
That wouldn't make any difference.  Currently `getState()` is declared like this:

```c++
State getState(int types, bool enforcePeriodicBox=false, int groups=0xFFFFFFFF) const;
```

You're suggesting to change it to something like this:

```c++
State getState(int types, bool enforcePeriodicBox=false, int groups=getIntegrator().getIntegrationForceGroups()) const;
```

But that isn't legal.  Default values for arguments aren't allowed to refer to the object the method is being called on.

craabreu:
I was thinking of something like moving the `forceGroups` property from `Integrator` to `Context` and then doing this:

```cpp
State getState(int types, bool enforcePeriodicBox=false, int groups=-1) const;
```

Followed by replacing `groups` by `groups > 0 ? groups : forceGroups` in this line:

https://github.com/openmm/openmm/blob/b0d1358247cceee29e431d92224a9dc81d09b6b0/openmmapi/src/Context.cpp#L99

Something similar would have to be done for `ContextImpl.calcForcesAndEnergy()`. I believe that this would not be a real change in API, since backward compatibility seems to be fully preserved.

peastman:
That would break backward compatibility by changing the meaning of `groups=0`.  Currently it means don't include any groups.

craabreu:
> That would break backward compatibility by changing the meaning of groups=0. Currently it means don't include any groups.

Sorry. My mistake. The correct form would be `groups >= 0 ? groups : forceGroups`.

peastman:
That still doesn't work.  It would mean that any time group 31 was included (hence the high order bit was set and the value was negative), it would ignore the groups they requested and instead return the default ones.

z-gong:
Since the purpose is to calculate some user-defined 'force' or 'energy' without actually exerted on the system, could it be possible to add an attribute to the `Force` object instead of changing the logic of the integrator, barostat, etc.?

There could be an API like `force.setUsedInIntegration(False)` so that the force and energy evaluation will just return zero during normal integration.

An argument could be added to `getState` so that these forces are evaluated only when required, like `getState(includingUnusedForces=True)`

PR #2713, opened 2020-06-01
Adding SOVERSION for OpenMM main shared library

### LINKED ISSUE #2281: Shared libraries need versions
Building OpenMM produces the following public shared libraries without versions in their sonames:

* libOpenMMAmoeba.so
* libOpenMMRPMD.so
* libOpenMM.so
* libOpenMMDrude.so

Without sonames the shared libraries cannot be represented in the shlibs system, and if linked by binaries its interface cannot safely change. Therefore no backward-compatible way exists to migrate programs linked against them to a new binary interface.

I have found this issue while packaging OpenMM for Debian.

merkys:
This partially solves #2281 by adding a SOVERSION for main OpenMM shared library (`OpenMM`). Other libraries should get their SOVERSIONs too to fully solve #2281.


peastman:
Is 0 the right value to use?  Should it match the OpenMM version number?

merkys:
> Is 0 the right value to use? Should it match the OpenMM version number?

Good point. In that case, `0` could be substituted with `${OPENMM_MAJOR_VERSION}` or `"${OPENMM_MAJOR_VERSION}.${OPENMM_MINOR_VERSION}"`. The only issue is that the SONAME has to be increased each time symbols are modified or removed from the shared library. If such changes happen only in major releases, `${OPENMM_MAJOR_VERSION}` should be enough, and `"${OPENMM_MAJOR_VERSION}.${OPENMM_MINOR_VERSION}"` would provide even more granularity.

peastman:
I think `${OPENMM_MAJOR_VERSION}.${OPENMM_MINOR_VERSION}` would be best.  Feature releases often only update the second digit.

merkys:
> I think `${OPENMM_MAJOR_VERSION}.${OPENMM_MINOR_VERSION}` would be best. Feature releases often only update the second digit.

Adjusted the PR accordingly.

PR #2715, opened 2020-06-01
Fixed check for changed exceptions

### LINKED ISSUE #2709: Calling NonbondedForce.updateParametersInContext() without changing parameters causes exception
The attached code snippet (and serialized XML files) does _not_ make any changes to the `NonbondedForce` before calling `updateParametersInContext()`, and yet still fails with 
```
simtk.openmm.OpenMMException: updateParametersInContext: The number of non-excluded exceptions has changed
```
This shouldn't be able to happen, as no changes at all are made to the exceptions. 

Simply run:
```bash
python test-update-parameters.py
```
[RUN49.zip](https://github.com/openmm/openmm/files/4707269/RUN49.zip)

Here's the code, for convenience:
```python

def read(filename):
    """Read serialized object from XML file."""
    with open(filename, 'rt') as infile:
        return openmm.XmlSerializer.deserialize(infile.read())

# Read system, state, and integrator
system = read('system.xml')
state = read('state.xml')

for platform_name in ['CPU', 'OpenCL', 'CUDA']:
    platform = openmm.Platform.getPlatformByName(platform_name)
    print(platform_name)

    integrator = read('integrator.xml')
    context.setState(state)
    state2 = context.getState(getEnergy=True, getForces=True)
    
    for force_index, force in enumerate(system.getForces()):
        force_name = force.__class__.__name__
        if force_name == 'NonbondedForce':
            print(' Updating unchanged NonbondedForce...')
            force.updateParametersInContext(context)
            state2 = context.getState(getEnergy=True, getForces=True)

    del context, integrator
```

peastman:
Fixes #2709.  This happened on the Reference and CPU platforms if you were using a parameter offset to modify an exception.

### DIFF
--- a/platforms/cpu/src/CpuKernels.cpp
+++ b/platforms/cpu/src/CpuKernels.cpp
@@ -712,12 +712,23 @@ double CpuCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeFo
 void CpuCalcNonbondedForceKernel::copyParametersToContext(ContextImpl& context, const NonbondedForce& force) {
     if (force.getNumParticles() != numParticles)
         throw OpenMMException("updateParametersInContext: The number of particles has changed");
+
+    // Identify which exceptions are 1-4 interactions.
+
+    set<int> exceptionsWithOffsets;
+    for (int i = 0; i < force.getNumExceptionParameterOffsets(); i++) {
+        string param;
+        int exception;
+        double charge, sigma, epsilon;
+        force.getExceptionParameterOffset(i, param, exception, charge, sigma, epsilon);
+        exceptionsWithOffsets.insert(exception);
+    }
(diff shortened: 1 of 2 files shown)


jchodera:
Thanks! Too bad we just put out 7.4.2...

swails:
Jenkins ran this time, so it seems that the integration is working as expected again.

Jenkins at this point has stabilized:
* I'm using arguably the most popular (read maintained) plugin Jenkins has to offer
* Jenkins is hosted in a cloud-based VM with HA (even if the build agents are at my house)
* The builds are all containerized to isolate it from most details of the executors

It should tide us over until GPUs are GA on Azure pipelines for open source projects like this one.

peastman:
Thanks!

PR #2736, opened 2020-06-09
Initial HIP Platform implementation for AMD GPUs on ROCm

arghdos:


arghdos:
Hello Dr. Eastman,

I have been working on a port of OpenMM's CUDA platform for AMD's HIP language to target AMD GPUs on the ROCm stack for use in Folding @ Home (among other things).

The current code is up to date with the latest release of the OpenMM master branch, and supports all the same functionality as the current CUDA platform, though I haven't ported any of the platform specific plugins yet (cudacompiler, drude, rpmd, amoeba).
The code itself is currently very similar to the CUDA platform, the only substantial changes were needed to account for differences in the wavefront/warp size on AMD/NVIDIA GPUs.
Essentially this entailed making the non-bonded forces able to use either unsigned longs or regular unsigned ints, and being very careful shifting, bitwise functions, etc., and removing hard-coded wave32 assumptions.  Note that this has some knock on effects in the common platform, as some of the kernels there had hardcoded assumed sizes (e.g., "launched w/ a single block of 64 threads"). 

`make test` shows 100% passing on my system, and I have run some basic benchmarking using `examples/benchmark.py` to ensure that the python bindings work as expected.

Although HIP supports running on NVidia GPUs I haven't pursued this code path yet, however given the work done to make the HIP platform wavefront/warp-size agnostic, the only real challenge I foresee there is finding a way to have the CMake build system be compatible.

In particular, I noticed that you have a fairly extensive CI / test / build server set-up, and I have not attempted to get that up and running.  Nor have I looked at updating the documentation much yet.  Given the scope of this PR, I wanted to submit what we have already for review while we discuss next steps.

peastman:
Thanks!  That's a very impressive amount of work!

Can you comment on the motivation for this?  What are the advantages of using HIP instead of OpenCL?  Adding a new platform is obviously not something we do lightly.  It's a ton of additional code that then needs to be maintained, tested on multiple platforms, and added to whenever we add a new feature.  So the bar is very high, and there need to be really compelling advantages to justify it.

jchodera:
Wow! Any idea of how the HIP performance compares to OpenCL on AMD GPUs?

arghdos:
Couple of note: first, the failing tests are specific to my branch, likely due to the things in the common platform I changed (probably need to propagate the expected wavefront size to OpenCL for AMD).   Before I spend more time to go and fix that, let's discuss how (or if) we think this work will be useful.

>I'd be very surprised if they don't support it. OpenCL has been included in all AMD drivers for many years, and I've heard no suggestion from them that they plan to drop it. It also is a standard component of ROCm, so any platform that supports it will necessarily support OpenCL

Sorry, I didn't mean to imply that the OpenCL runtime wouldn't be there... but I can also tell you as someone working on HPC applications for said systems, a vanishingly small number of the codes we are responsible for are using OpenCL.  That means, we're not doing much at all in the way of optimization, bug fixes, new features, et al.  Specifically, I am unlikely to work on optimizing the OpenCL OpenMM backend because a customer had asked us for a HIP port to fit into their workflow (which uses HIP only libraries).

I'm having troubles getting the OpenCL platform to show up in my install at the moment.  Will hopefully get you some performance numbers later today.

>We've already got that. :)
>http://docs.openmm.org/development/developerguide/developer.html#common-compute
>But we still end up needing a lot of code specific to each platform.

One difference between this HIP platform and the OpenCL platform is that by and large the HIP files are carbon copies of the CUDA platform (largely the modifications are due to equivalent of `sed s/CUDA/HIP/g`).  The HIPContext file is where the bulk of the actual changes occurred to remove the Context API, which isn't really implemented for HIP, and setting AMD specific values for warp sizes / thread block sizes, etc.  The kernels themselves are again mostly identical though for some of the sorting kernels etc. I've added explicit definitions of the block size / tile size for sanity.

Thinking forward, would this new platform be better received if I did something along the lines of what @giadefa suggested?  I.e., if there was a "GPU" platform, that can either run CUDA or HIP, where a call to something like `gpuEventCreate` can be mapped to either `cuEventCreate` or `hipEventCreate` depending on the backend selected?  This way we could minimize the introduction of new code to only the specific things needed for HIP / AMD GPUs (e.g., determining the tileflags type).

peastman:
Let me explain where I'm coming from.  We have a *very* small development team.  In fact, I'm the only person who is specifically paid to work full time on OpenMM!  So we have to be strategic about development.  Anything that creates a large ongoing maintenance burden needs to meet a high bar to justify that burden.

NVIDIA is the dominant GPU vendor for HPC.  That's what the overwhelming majority of our users run their production simulations on.  It's essential that we have the best possible performance on NVIDIA GPUs, and to get that we need a CUDA specific implementation.

In an ideal world we'd put in the same amount of work for every brand, with a platform specific to each vendor using whatever technology gets the very best performance on their hardware.  HIP for AMD on Linux, Metal for MacOS, etc.  But we just don't have the resources to do that, so instead we rely on OpenCL as a single platform that works everywhere.  It's supported on NVIDIA, AMD, and Intel.  It's available on Linux, Windows, and MacOS.  It even works on Qualcomm, so if we decide to revive the Folding@Home client for Android we'll be covered there too.

That's not to say we don't care about those platforms.  We absolutely do!  I'm willing to put in quite a bit of work to get the best performance on each platform—but only if it doesn't involve taking on a substantial long term burden to maintain it.

Ideally AMD's OpenCL would be just as fast as HIP.  I certainly don't know of any reason it can't be.  But AMD has to make their own decisions about how to prioritize resources.

giadefa:
HI,
in our hand openCL has always been a lot slower even on the same gpu, say
an Nvidia one. There is no reason for it to be slower apart that no vendor
cares about it. I am not sure that it will still be around in the future
either.

openCL as well as HIP are very similar to low level CUDA, so it should be
easy to wrap it around at the low level that Nick is proposing. You could
have a single 99% identical code for openCL, HIP and CUDA. Now this makes
sense if AMD start to do some nice GPU, so Nick, when you have performances
it would be nice to know.

g

peastman:
> in our hand openCL has always been a lot slower even on the same gpu, say
an Nvidia one.

That's been my experience too.  In most of our benchmarks OpenCL is about 30% slower than CUDA on the same GPU.  Partly that's because CUDA can use features that aren't available in OpenCL, like shuffles and warp vote operations, but that doesn't explain all of it.

OpenCL is the platform that everyone supports and no one embraces.  NVIDIA really wants people to use CUDA, Microsoft wants them to use DirectX, Apple wants them to use Metal, and apparently AMD now wants them to use HIP.

jchodera:
@peastman: Could we find some folks at AMD who would be willing to help us support HIP? Also, the new Playstation 5 uses an AMD GPU, so while it may be far-fetched we could build OpenMM Folding@home cores for it anytime soon, this _would_ seem to be appealing...

vvsteg:
@arghdos: After the problem with large system sizes has been recently resolved, there remains a problem with the OpenCL backend for systems larger 2 million atoms (as reported by @Dann239 in #2805). Strangely enough, this problem reveals itself on AMD GPUs only (see #2805). We have tested the OpenCL backend on AMD GPUs with ROCm and amdgpu-pro drivers. Could you point out the possible reason of this problem?

ThWuensche:
@arghdos I would be interested to test F@H work units on the HIP port of openmm for performance comparision. However I fail in compiling with path settings for HIP, resulting in hip/hip_runtime.h not found. I have ROCM 3.5 or 3.7 in /opt/rocm. How should I set variables for bulding HIP enabled opemMM? Given the big performance advantage of CUDA over OpenCL for NVidia GPUs I would be interested whether the same advantage is achieved with HIP for AMD GPUs (Radeon VII).

arghdos:
@ThWuensche -- something like:

```
cmake .. -DHIP_ROOT_PATH=/opt/rocm-3.7.0/hip
```
should do the trick.

ThWuensche:
I don't see whether my responses to the notification e-mail have been delivered. So will repeat them here:

Dear Nick Curtis,

Thank you for your reply, I had tried that before, but it does not work.
I get the following build error:

[ 31%] Building CXX object
platforms/hip/sharedTarget/CMakeFiles/OpenMMHIP.dir/__/src/HipArray.cpp.o
In file included from
/home/tw/src/openmm-hip/platforms/hip/src/HipArray.cpp:29:
/home/tw/src/openmm-hip/platforms/hip/./include/HipArray.h:35:10: fatal
error: hip/hip_runtime.h: Datei oder Verzeichnis nicht gefunden
 #include <hip/hip_runtime.h>
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: ***
[platforms/hip/sharedTarget/CMakeFiles/OpenMMHIP.dir/build.make:90:
platforms/hip/sharedTarget/CMakeFiles/OpenMMHIP.dir/__/src/HipArray.cpp.o]
Fehler 1
make[1]: *** [CMakeFiles/Makefile2:4525:
platforms/hip/sharedTarget/CMakeFiles/OpenMMHIP.dir/all] Fehler 2
make: *** [Makefile:141: all] Fehler 2

But /opt/rocm/hip/include/hip/hip_runtime.h exists. Do I need additional
settings for the include path? Might something in the cmake scripts work
on your side, but be broken in other environments? I did not dig through
all the cmake-magic to see how include paths are set in this step.

I'm working on Debian Buster.

Your hint how to correct that would be welcome.

Best regards,

Thomas Wuensche

**and**

I'm further playing with the build. Seems that this command does not put
the include path into the makefiles, a grep does not deliver that path.

After setting shell variables C_INCLUDE_PATH and CPLUS_INCLUDE_PATH to
that directory, the compiler finds the include file. However further on
the build breaks with

/opt/rocm/hip/include/hip/hip_runtime.h:60:2: error: #error ("Must
define exactly one of __HIP_PLATFORM_HCC__ or __HIP_PLATFORM_NVCC__");

Obviously I miss some setup. Any hints?

arghdos:
@ThWuensche 

>/opt/rocm/hip/include/hip/hip_runtime.h:60:2: error: #error ("Must
define exactly one of HIP_PLATFORM_HCC or HIP_PLATFORM_NVCC");

do: `export HIP_PLATFORM=hcc`  

re: the missing include, try the above export and then run `make VERBOSE=1` and show me the failing command line.  I think something in our cmake changed between ROCm 3.6 and now... not sure exactly what.

ThWuensche:
Actually I have ROCM 3.5.1 on that box.

Here is the output, please note that C_INCLUDE_PATH and CPLUS_INCLUDE_PATH are still set. Seems that setting HIP_PLATFORM didn't change the game:

[ 29%] Building CXX object platforms/hip/sharedTarget/CMakeFiles/OpenMMHIP.dir/__/src/HipArray.cpp.o
cd /home/tw/src/openmm-hip/build/platforms/hip/sharedTarget && /usr/bin/c++  -DOPENMM_BUILD_VERSION=0 -DOPENMM_LIBRARY_NAME=OpenMM -DOPENMM_MAJOR_VERSION=7 -DOPENMM_MINOR_VERSION=5 -DOpenMMHIP_EXPORTS -I/home/tw/src/openmm-hip/build/platforms/hip/../common/src -I/home/tw/src/openmm-hip/build/platforms/hip/src -I/home/tw/src/openmm-hip/platforms/hip/src -I/home/tw/src/openmm-hip/platforms/hip/../common/include -I/home/tw/src/openmm-hip/platforms/hip/./include -I/home/tw/src/openmm-hip/src -I/home/tw/src/openmm-hip/libraries/asmjit -I/home/tw/src/openmm-hip/libraries/vecmath/include -I/home/tw/src/openmm-hip/libraries/irrxml/include -I/home/tw/src/openmm-hip/serialization/include -I/home/tw/src/openmm-hip/platforms/reference/include -I/home/tw/src/openmm-hip/libraries/csha1/include -I/home/tw/src/openmm-hip/libraries/hilbert/include -I/home/tw/src/openmm-hip/libraries/lbfgs/include -I/home/tw/src/openmm-hip/libraries/sfmt/include -I/home/tw/src/openmm-hip/libraries/lepton/include -I/home/tw/src/openmm-hip/libraries/quern/include -I/home/tw/src/openmm-hip/libraries/jama/include -I/home/tw/src/openmm-hip/olla/include -I/home/tw/src/openmm-hip/openmmapi/include -I/home/tw/src/openmm-hip/./include -I/home/tw/src/openmm-hip/tests  -O3 -DNDEBUG -fPIC   -msse2 -DLEPTON_USE_JIT -DOPENMM_COMMON_BUILDING_SHARED_LIBRARY -std=gnu++11 -o CMakeFiles/OpenMMHIP.dir/__/src/HipArray.cpp.o -c /home/tw/src/openmm-hip/platforms/hip/src/HipArray.cpp
In file included from /home/tw/src/openmm-hip/platforms/hip/./include/HipArray.h:35,
                 from /home/tw/src/openmm-hip/platforms/hip/src/HipArray.cpp:29:
/opt/rocm/hip/include/hip/hip_runtime.h:60:2: error: #error ("Must define exactly one of __HIP_PLATFORM_HCC__ or __HIP_PLATFORM_NVCC__");
 #error("Must define exactly one of __HIP_PLATFORM_HCC__ or __HIP_PLATFORM_NVCC__");
  ^~~~~
In file included from /opt/rocm/hip/include/hip/hip_runtime.h:64,
                 from /home/tw/src/openmm-hip/platforms/hip/./include/HipArray.h:35,
                 from /home/tw/src/openmm-hip/platforms/hip/src/HipArray.cpp:29:
/opt/rocm/hip/include/hip/hip_runtime_api.h:367:2: error: #error ("Must define exactly one of __HIP_PLATFORM_HCC__ or __HIP_PLATFORM_NVCC__");
 #error("Must define exactly one of __HIP_PLATFORM_HCC__ or __HIP_PLATFORM_NVCC__");
  ^~~~~

arghdos:
Oh I see what's happening -- `/opt/rocm/bin/hipconfig` is supposed to be called on line 45 of `FindMMHIP.cmake` to store the compilation flags you need to compile HIP host code:

```
    EXECUTE_PROCESS(COMMAND ${HIP_HIPCONFIG_EXECUTABLE} -C OUTPUT_VARIABLE HIPCXXFLAGS)
```

Not sure why, but it doesn't look like that's happening on your system.

arghdos:
As a temporary workaround, you should be able to do something like:

```
-DCMAKE_CXX_FLAGS=`/opt/rocm-3.7.0/bin/hipconfig -C
```

ThWuensche:
That takes me a little further. I have installed ROCM 3.7, installed rocm-libs. 

Had to add -I/opt/rocm/include, otherwise libfft.h was not found.

Had to change the include in roctracer_ext.h from roctracer.h to roctracer/roctracer.h

Then I was stuck at:

[ 29%] Building CXX object platforms/hip/sharedTarget/CMakeFiles/OpenMMHIP.dir/__/src/HipContext.cpp.o
In file included from /opt/rocm/include/roctracer/roctracer_ext.h:35,
                 from /home/tw/src/openmm-hip/platforms/hip/src/HipContext.cpp:61:
/opt/rocm/include/roctracer/roctracer.h:45:10: fatal error: ext/prof_protocol.h: Datei oder Verzeichnis nicht gefunden
 #include <ext/prof_protocol.h>
          ^~~~~~~~~~~~~~~~~~~~~

After adding  -I/opt/rocm/include/roctracer to the CMAKE_CXX_FLAGS the build went on. That probably makes the above mentioned change to the path for roctracer.h unneccessary.

Now I'm at:

[ 31%] Linking CXX shared library ../../../libOpenMMHIP.so
[ 33%] Built target OpenMMHIP
Scanning dependencies of target TestHipVerletIntegrator
[ 33%] Building CXX object platforms/hip/tests/CMakeFiles/TestHipVerletIntegrator.dir/TestHipVerletIntegrator.cpp.o
[ 33%] Linking CXX executable ../../../TestHipVerletIntegrator
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `hipDriverGetVersion'
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `hipModuleLaunchKernel'
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `hipfftExecD2Z'

Probably doesn't find the shared library

ThWuensche:
Added -L "path_of_build_dir", but that did not fix it. Here is the output with VERBOSE=1:

make -f platforms/hip/tests/CMakeFiles/TestHipVerletIntegrator.dir/build.make platforms/hip/tests/CMakeFiles/TestHipVerletIntegrator.dir/build
make[2]: Verzeichnis „/home/tw/src/openmm-hip/build“ wird betreten
[ 33%] Linking CXX executable ../../../TestHipVerletIntegrator
cd /home/tw/src/openmm-hip/build/platforms/hip/tests && /usr/bin/cmake -E cmake_link_script CMakeFiles/TestHipVerletIntegrator.dir/link.txt --verbose=1
/usr/bin/c++  -D__HIP_PLATFORM_HCC__=  -I/opt/rocm/include -I/opt/rocm/hip/include -I/opt/rocm/include/roctracer -I/opt/rocm/llvm/bin/../lib/clang/11.0.0 -I/opt/rocm/hsa/include -D__HIP_ROCclr__ -L/home/tw/src/openmm-hip/build -O3 -DNDEBUG  -rdynamic -msse2 -DLEPTON_USE_JIT -Wl,--no-as-needed -lrt CMakeFiles/TestHipVerletIntegrator.dir/TestHipVerletIntegrator.cpp.o  -o ../../../TestHipVerletIntegrator -Wl,-rpath,/home/tw/src/openmm-hip/build ../../../libOpenMMHIP.so ../../../libOpenMM.so.7.5 -ldl -lpthread 
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `hipDriverGetVersion'
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `hipModuleLaunchKernel'
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `hipfftExecD2Z'
/usr/bin/ld: ../../../libOpenMMHIP.so: undefined reference to `roctracer_stop'
....

ThWuensche:
Probably that uninitialized memory bugfix needs to be integrated (and extended to hip) to run successfully on ROCM. It had made my Radeon VIIs and other AMD GPUs fail on the COVID moonshot type free energy calculations:

https://github.com/openmm/openmm/commit/cea03eea70ed71f248dbeb6ea447e83053e5c6af

I could patch my source, but probably would be better to integrate it into your fork.

ThWuensche:
Ok, finally could build it with the following cmake command line:

cmake .. -DHIP_ROOT_PATH=/opt/rocm/hip -DCMAKE_CXX_FLAGS="-D__HIP_PLATFORM_HCC__=  -I/opt/rocm/include -I/opt/rocm/hip/include -I/opt/rocm/include/roctracer -I/opt/rocm/llvm/bin/../lib/clang/11.0.0 -I/opt/rocm/hsa/include -D__HIP_ROCclr__ -L/home/tw/src/openmm-hip/build -L/opt/rocm/lib -lamdhip64 -lrocfft -lroctracer64"

Probably most of the values added manually should be the result of autoconfiguration, which obviously didn't work on my system

ThWuensche:
Running a F@H 13424 job now on openMM HIP, runs for a varying number of steps, then breaks with particle coordinate NaN. Could not find the place for the uninitialized memory bug patch, code seems quite different.

Trying to run that job on openCL in that same build I get:

Creating openCL Context...
Creating HIP Context...
  File "../../../../../fahcore_openmm_run.py", line 30, in <module>
    context_ocl = openmm.Context(system, integrator_ocl, platform_ocl)
  File "/usr/local/lib/python3.7/dist-packages/simtk/openmm/openmm.py", line 20600, in __init__
    this = _openmm.new_Context(*args)
simtk.openmm.OpenMMException: Error compiling kernel: /tmp/comgr-f4290c/input/CompileSource:247:50: error: use of undeclared identifier 'static_cast'
                bool isExcluded = (((exclusions>>static_cast<tileflags_t>(tj))&1) == 0);
                                                 ^
/tmp/comgr-f4290c/input/CompileSource:247:62: error: expected expression
                bool isExcluded = (((exclusions>>static_cast<tileflags_t>(tj))&1) == 0);
                                                             ^
/tmp/comgr-f4290c/input/CompileSource:120:21: note: expanded from macro 'tileflags_t'
#define tileflags_t long
                    ^
/tmp/comgr-f4290c/input/CompileSource:459:50: error: use of undeclared identifier 'static_cast'
                bool isExcluded = (((exclusions>>static_cast<tileflags_t>(tj))&1) == 0);
                                                 ^
/tmp/comgr-f4290c/input/CompileSource:459:62: error: expected expression
                bool isExcluded = (((exclusions>>static_cast<tileflags_t>(tj))&1) == 0);
                                                             ^
/tmp/comgr-f4290c/input/CompileSource:120:21: note: expanded from macro 'tileflags_t'
#define tileflags_t long
                    ^
4 errors generated.
Error: Failed to compile opencl source (from CL or HIP source to LLVM IR).

That in the moment seems beyond my reach ...

ThWuensche:
@peastman, @jchodera: Don't know whether my experiments are of interest to you

ThWuensche:
After finding the file and adding the unititialized memory bugfix, a 13424 WU seems to run without NaN on platform HIP.

Made a first benchmark, 100000 steps of project 13424 WU.

With openMM platform HIP:

```
completed 100000 steps

real    16m0,904s
user    6m16,274s
sys     0m16,519s

```
With openMM platform openCL:

```
completed 100000 steps

real    16m26,770s
user    3m57,749s
sys     0m8,744s

```

So the HIP port is working, but so far no significant performance benefit over openCL.

jchodera:
> @peastman, @jchodera: Don't know whether my experiments are of interest to you

I've been following along! Thanks for investigating this!

Would be curious to think what @peastman thinks about the viability of integrating this into the new unified framework.

ThWuensche:
@arghdos 

Dear Nick Curtis,

I had sent you an e-mail as reply to a message from github, but don't know whether you receive these messages. Therefor I repeat it here. Not to spam this thread on github, my e-mail is wuensche@ems-wuensche.com

Dear Nick Curtis,

as you probably have seen, I got your openMM HIP port working for
Folding@Home workloads. However there are no noteworthy performance
improvements so far, quite unlike the nVidia CUDA port in F@H, which
gives something like 25%-30% performance improvement. But you mentio-
ned yourself that this is a functionality PR, not performance.

I have benchmark results, if you are interested, I probably can
forward them to you.

Are you / is AMD further interested in getting that into F@H. If so
and if relevant performance improvements can be achieved, I would
be interested to support that. I personally like to see AMD and open
source as an alternative to nVidia closed solutions in that field.
But with the new product announcements and the additional performance
boost for nVidia CUDA, in that field chances look bad, if AMD is not
making big steps forward. Looking at results in F@H, the current
implementation can only in a few workloads get close to nVidia GPUs
like GTX1080ti, and that only if nVidia is running openCL, no chance
with nVidia running CUDA.

What benchmarks also show is that AMD GPUs (I have Radeon VIIs) are
performing extremely bad at small project sizes / atom counts.
nVidia GPUs, also these with wide architecture, perform much better
in that tasks. Do you have an idea what could be the reason for that
behaviour? Could it be kernel schedule latency?

As mentioned, I very much would like to see and help that AMD can
catch up and provide good results based on open source solutions.

Best regards,

Thomas Wuensche

ThWuensche:
@arghdos As far as I understand your port consists of two parts, adaption of openMM infrastructure to support larger warp sizes and decoupling hard coded assumptions on one hand and then the real HIP part on the other hand. Could these two parts of the port be decoupled, first a patch that allows the 64 thread CUs of GCN as alternative to the NVidia 32 thread warp sizes, so makes code infrastructure more flexible in a way that could also be used for openCL on GCN devices. Then, if that is in place and does not generate performance loss for other platforms, it would be easier to maintain the HIP port in parallel to the other platforms, without creating large differences for changes it the openMM general infrastructure.

peastman:
The OpenCL platform already works fine with GPUs that have a width of 64 (or 16 for that matter).  There are some places in the code where I use the word "warp" to refer to a group of 32 threads, but that's just a shorthand.  It doesn't need to match any particular aspect of the hardware.

ThWuensche:
@arghdos

Dear Nick Curtis, as you helped me to build openMM HIP port in the beginning I assume you're still involved. However I did not see a response to below question, maybe the question slipped by. Are you, is AMD planning to go ahead with performance optimization with this port?

> Dear Nick Curtis,
> 
> I had sent you an e-mail as reply to a message from github, but don't know whether you receive these messages. Therefor I repeat it here. Not to spam this thread on github, my e-mail is [wuensche@ems-wuensche.com](mailto:wuensche@ems-wuensche.com)
> 
> Dear Nick Curtis,
> 
> as you probably have seen, I got your openMM HIP port working for
> Folding@Home workloads. However there are no noteworthy performance
> improvements so far, quite unlike the nVidia CUDA port in F@H, which
> gives something like 25%-30% performance improvement. But you mentio-
> ned yourself that this is a functionality PR, not performance.
> 
> I have benchmark results, if you are interested, I probably can
> forward them to you.
> 
> Are you / is AMD further interested in getting that into F@H. If so
> and if relevant performance improvements can be achieved, I would
> be interested to support that. I personally like to see AMD and open
> source as an alternative to nVidia closed solutions in that field.
> But with the new product announcements and the additional performance
> boost for nVidia CUDA, in that field chances look bad, if AMD is not
> making big steps forward. Looking at results in F@H, the current
> implementation can only in a few workloads get close to nVidia GPUs
> like GTX1080ti, and that only if nVidia is running openCL, no chance
> with nVidia running CUDA.
> 
> What benchmarks also show is that AMD GPUs (I have Radeon VIIs) are
> performing extremely bad at small project sizes / atom counts.
> nVidia GPUs, also these with wide architecture, perform much better
> in that tasks. Do you have an idea what could be the reason for that
> behaviour? Could it be kernel schedule latency?
> 
> As mentioned, I very much would like to see and help that AMD can
> catch up and provide good results based on open source solutions.
> 
> Best regards,
> 
> Thomas Wuensche

peastman:
This is superseded by #3338, so I'm closing it.

Issue #2520, opened 2020-01-13
Recommended simulation settings


peastman:
Now that we have BAOABLangevinIntegrator, we should think about updating our recommended simulation settings.  These settings come up in lots of places:

- They're used in examples.
- They're recommended in the manual.
- We run benchmarks with them.
- OpenMM-Setup uses them as defaults.

Currently we have two sets of recommended settings:

- LangevinIntegrator, 2 fs time step, 1/ps friction, Hbonds constraints
- LangevinIntegrator, 5 fs time step, 1/ps friction, AllBonds constraints, HMR

If we want to start recommending BAOABLangevinIntegrator instead, we need to select settings to recommend with it.  We can presumably increase the time step (to make the simulation run faster), decrease the friction (to make sampling more efficient), or both.  What values can we use while still only constraining hbonds?  Is there still value in also recommending settings with AllBonds?  Is HMR useful, or does that only help when using AllBonds?

We should look at what settings are recommended with other programs, but also run our own tests.

jchodera:
I think our default recommendation should be:
* mixed precision, which is essential for good constraint behavior
* BAOAB, which produces minimal error in configuration space: https://doi.org/10.3390/e20050318
* HMR to make larger timesteps more stable: https://doi.org/10.1021/ct5010406
* 4 fs timesteps, which seem generally stable for most systems when using HMR
* collision rate of 1/ps, which seems minimally perturbative: https://doi.org/10.1063/1.4802990

I don't think most other packages have a good sense of what to recommend and why. There seems to be a fundamental misunderstanding about what is relevant to molecular simulation, even up to the recent [gromacs paper that still quantifies error in terms of drift](https://chemrxiv.org/articles/On_The_Importance_of_Accurate_Algorithms_for_Reliable_Molecular_Dynamics_Simulations/11474583). For sampling from configuration space, we can safely recommend BAOAB, which makes [minimal error in the configuration space distribution up to the point where the integrator breaks](https://doi.org/10.3390/e20050318). HMR does not perturb the configuration distribution.

peastman:
> mixed precision, which is essential for good constraint behavior

What is this referring to?  I know you had issues when you were specifying a super tight tolerance like 1e-10.  But a tolerance of 1e-6 is accurate enough even for most constant energy simulations, and it works fine in single precision.

> HMR to make larger timesteps more stable

[...]

PR #2754, opened 2020-06-22
Updated benchmark script

peastman:
This updates the benchmarking script to use LangevinMiddleIntegrator and a larger time step where possible.  Here are the speeds I get for the PME benchmark with different possible combinations of settings.

LangevinIntegrator, 2 fs, HBonds constraints: 417
LangevinMiddleIntegrator, 4 fs, HBonds constraints: 740
LangevinMiddleIntegrator, 5 fs, AllBonds constraints, HMR: 568
Langevintegrator, 5 fs, AllBonds constraints, HMR: 775

The ordering is the same for most benchmarks.  Using LangevinMiddleIntegrator with the larger step size produces a big speed improvement compared to LangevinIntegrator.  But the very fastest combination is to use LangevinIntegrator with HMR and more constraints.  Since LangevinMiddleIntegrator needs to apply constraints more often, it takes a big hit from using AllBonds constraints.

Therefore if you specify `--heavy-hydrogens`, I made the script continue to use LangevinIntegrator in place of LangevinMiddleIntegrator.

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -88,12 +88,13 @@ def runOneTest(testName, options):
             dt = 0.005*unit.picoseconds
             constraints = app.AllBonds
             hydrogenMass = 4*unit.amu
+            integ = mm.LangevinIntegrator(300*unit.kelvin, friction, dt)
         else:
-            dt = 0.002*unit.picoseconds
+            dt = 0.004*unit.picoseconds
             constraints = app.HBonds
             hydrogenMass = None
+            integ = mm.LangevinMiddleIntegrator(300*unit.kelvin, friction, dt)
         system = ff.createSystem(pdb.topology, nonbondedMethod=method, nonbondedCutoff=cutoff, constraints=constraints, hydrogenMass=hydrogenMass)
-        integ = mm.LangevinIntegrator(300*unit.kelvin, friction, dt)
     print('Step Size: %g fs' % dt.value_in_unit(unit.femtoseconds))
     properties = {}
     initialSteps = 5


jchodera:
Can we deprecate the AllBonds constraints? This can significantly modify the effective torsions, and it doesn't seem to give any significant performance advantage over LangevinMiddleIntegrator + 4 fs + HBonds.

jchodera:
cc our previous discussion here: https://github.com/openmm/openmm/issues/2520

jchodera:
Presumably, there should also be an MTS Langevin scheme that should give performance just as good as any AllBonds combination but without the pathologies associated with it.

peastman:
I wouldn't deprecate that option.  It works fine and does exactly what it says.  Whether it's appropriate for a particular force field is a different matter.

Experimenting with different MTS protocols could be interesting.  We could probably find something that allows a 5 fs outer time step.

jchodera:
> I wouldn't deprecate that option. 

We shouldn't remove the AllBonds option. I'm just saying we should remove it from the benchmark options for Amber force fields, since its inclusion comes with our tacit approval that this is a sensible combination. It is not.

> It works fine and does exactly what it says. Whether it's appropriate for a particular force field is a different matter.

Are there force fields for which this is an appropriate combination? If so, can we use that force field instead?

Again, I think we want OpenMM to "do the right thing" by default. If one of our standard benchmarks does not do the right thing, why do we want this?

jchodera:
Another alternative to all-bonds would be to use the now-popular virtual sites approach to representing hydrogens that gromacs has popularized. There are at least publications about this:
https://onlinelibrary.wiley.com/doi/full/10.1002/jcc.26198?af=R

peastman:
I'm fine with not including those on our benchmarks page.

> Are there force fields for which this is an appropriate combination?

I think it's pretty common in coarse grained models.

> Another alternative to all-bonds would be to use the now-popular virtual sites approach to representing hydrogens that gromacs has popularized.

Using vsites for hydrogen is in addition to constraining all bonds, not instead of it.  Doing that removes even more degrees of freedom than just constraining bonds.

PR #2756, opened 2020-06-24
XTC reporter

jadeshi:
Hey all,
I made an XTCReporter for OpenMM that I'd like to share since I think it could potentially be useful to people, especially for saving long MD simulations with limited disk space. I modeled this after the existing PDBReporter as closely as possible. You use it basically exactly the same way as PDBReporter, supplying a file name and the interval you want to save frames. One additional thing you have to do is explicitly specify whether you want to wrap coordinates using periodic boundary conditions (False by default). An example ala2 simulation utilizing it is included in examples/xtcreporter_example.

Enjoy!


jadeshi:
Also realized I'm not using the most updated version of OpenMM, so my __init__.py file update deletes some newer import statements. What's really needed is just to add the "from .xtcreporter import XTCReporter" line. Nothing else needs to change.

peastman:
Thanks!  If you're going to depend on MDTraj though, why not just use its XTCReporter class directly?  https://github.com/mdtraj/mdtraj/blob/master/mdtraj/reporters/xtcreporter.py

jadeshi:
Oh wow that exists now? I must be way behind the times...

peastman:
Closing abandoned PRs.

PR #2764, opened 2020-07-01
Use Python 3.6 for Mac build

peastman:
The travis file didn't specify what Python to use for the Mac build, and it was defaulting to 2.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,11 +128,8 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew cask uninstall oclint;
       brew install fftw;
-      brew install python3;
+      brew upgrade python;
       brew install -y https://raw.githubusercontent.com/Homebrew/homebrew-core/5b680fb58fedfb00cd07a7f69f5a621bb9240f3b/Formula/doxygen.rb;
-      pip3 install virtualenv;
-      virtualenv -p python3 ~/venv;
-      source ~/venv/bin/activate;
       sudo pip install -U pytest --ignore-installed six;
     fi
   # The cmake version installed by apt on ARM and PPC is very old,
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,11 +128,11 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew cask uninstall oclint;
       brew install fftw;
-      brew install python3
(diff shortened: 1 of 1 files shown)


jchodera:
Good catch, though hopefully we can get the GitHub Actions version working soon!

Lnaden:
If you're trying to use Python + OSX on Travis, I have run into some problems with this before and have notes!

From my notes I left in the [MolSSI Cookiecutter](https://github.com/MolSSI/cookiecutter-cms/blob/master/cihelpers/osx_travis_py_helper.sh) (and the code i'm posting here you is from that, feel free to use it):

> Travis decided to remove pip from their OSX images post xcode 7.3
    Travis official stance: Python package development is not supported in OSX.
    There is technically a pip2 located in the path, but installing packages through it actually installs to a Python
    Iin the Brew cellar, which is NOT the same as the python on the path. So mapping pip to $(which pip2) won't work.
    Easy install does not work, there is no ensurepip, and short of re-mapping the python to the version in the Cellar
    (which I don't want to do since I fully expect even that to disappear), I've burned out of ideas.
    ...
    SO! This file exists as a helper to set up the pyenv
    ...
    SIGH - LNN
> IMPORTANT: This should ONLY be used for OSX in Travis CI, not for local production builds of any kind

And here's the code (you'll need to set a $PYTHON_VER variable either in the script or as a global for Travis; i have it set up to use a `language: generic` in the `.travis.yml` matrix:

```bash
# Use PyEnv, don't let brew update EVERYTHING ELSE that we don't need it to
HOMEBREW_NO_AUTO_UPDATE=1 brew upgrade pyenv
# Pyenv requires minor revision, get the latest
PYENV_VERSION=$(pyenv install --list |grep $PYTHON_VER | sed -n "s/^[ \t]*\(${PYTHON_VER}\.*[0-9]*\).*/\1/p" | tail -n 1)
# Install version
pyenv install -f $PYENV_VERSION
# Use version for this
pyenv global $PYENV_VERSION
# Setup up path shims
eval "$(pyenv init -)"
```

Lnaden:
That script that block is in is then `source`'d in the `.travis.yml` file in the `before_install` section

peastman:
Thanks!  I've been trying to do something similar following the instructions at https://github.com/travis-ci/travis-ci/issues/9929#issuecomment-500073011.  Not working yet, but we'll see...

PR #2765, opened 2020-07-01
Added new features to AmoebaVdwForce

peastman:
cc @jayponder

This supersedes #2674.  It adds the ability to specify parameters by atom type, and to override the combining rules for interactions between specific pairs of types.  It also adds the option to use Lennard-Jones instead of buffered 14-7.

Regardless of which way you specify the parameters, I made it always build a matrix of atom types and use that when doing the calculations.  That let me simplify the code a lot.  I also took the opportunity to do a major cleanup of the reference implementation while I was at it.

I haven't updated the ForceField class to support specifying parameters by type.  Is this planned to be used in a future version of the AMOEBA force field?  If so we'll need to add that too, but first we'll need to discuss the design a bit.  For example, should the vdw atom types be the same as the atom types defined by the force field, or does the `<AmoebaVdwForce>` tag need to be able to define its own atom types separately from those?

Please look this over and let me know what you think!  If you can test it out, that would be even better.


jayponder:
Peter, that all sounds fine. Building the full matrix of parameter is more general and seems like the way to go. The matrix is usually small and should not really affect storage or compute efficiency. It's what Tinker has done since the beginning.

As to specifying vdw by atom "type" (instead of the coarser "class" ), this is mostly in Tinker, and hence the original pull request, because it's how OPLS does things (at least the Jorgensen OPLS, Schrodinger OPLS is more ad hoc). It is not planned to do vdw by type in AMOEBA, as when we find we need different vdw parameters, we generally just add a new class.

peastman:
> As to specifying vdw by atom "type" (instead of the coarser "class" )

What I meant was, will a future version of the force field use the feature added here, where you can override the combining rules?

jayponder:
I can't say whether a future version of our force fields will need to override combining rules. We will probably only use that feature if we "need" to. That said, in the longer view, combining rules (and indeed atoms types themselves) are an anachronism, and will go into the dustbin of history- maybe sooner rather than later.

Do you think the use of the full matrix of vdw parameters, as opposed to doing combining rules on the fly, would have any affect on the speed of OpenMM for simple partial charge force fields like Amber?

peastman:
In that case we can probably omit it for right now.  If we need that feature for a force field in the future, we can add it then.

> Do you think the use of the full matrix of vdw parameters, as opposed to doing combining rules on the fly, would have any affect on the speed of OpenMM for simple partial charge force fields like Amber?

It tends to slow them down a little bit, since computation is faster than memory access on the GPU.  That depends on the size of the matrix, since if the whole thing fits easily in cache then it doesn't make much difference.  In AMOEBA it makes no difference since the vdw kernel is negligible compared to the multipoles.

Let me know if everything else looks ok, or if you see anything that should be changed.

jayponder:
Just to be clear, we do need the matrix of vdw parameters (ie, the capability to override combining rules) in current AMOEBA. That's why we added it in the pull request! It is mostly used for ion interactions in current AMOEBA. In answer to what I thought you were asking, I'm not sure how much we will use it in the future in new force fields like HIPPO. But it does need to be in the code to run correct AMOEBA.

peastman:
I finally tracked down the cause of the test failure: a bug in gcc. :(

```c++
printf("before %d\n", typeForParams.size());
typeForParams[params] = typeForParams.size();
printf("set to %d\n", typeForParams[params]);
```

With clang that produces exactly the output you'd expect:

```
before 0
set to 0
before 1
set to 1
```

But with gcc we get

```
before 0
set to 1
before 1
set to 2
```

It's messing up the execution order, effectively performing the insertion into the map before computing the value it's going to insert.  This is the second gcc bug I've had to work around in the past week!

peastman:
@jayponder is this ok to merge?

jayponder:
I haven't had a chance to test it yet, but looking over the code it seems fine/correct. I'd suggest you go ahead with merging, and either I or someone else will test it on a system with ions to make sure it fully agrees with our Tinker reference code. Thanks.

peastman:
Ok, thanks!

PR #2778, opened 2020-07-13
Added check for unsupported OpenCL implementation

peastman:
When selecting a platform, it will always prefer a supported one over an unsupported one.  If it's forced to use an unsupported platform, it prints a warning to the console.

### DIFF
--- a/platforms/opencl/src/OpenCLContext.cpp
+++ b/platforms/opencl/src/OpenCLContext.cpp
@@ -69,6 +69,14 @@ static void CL_CALLBACK errorCallback(const char* errinfo, const void* private_i
     std::cerr << "OpenCL internal error: " << errinfo << std::endl;
 }
 
+static bool isSupported(cl::Platform platform) {
+    string vendor = platform.getInfo<CL_PLATFORM_VENDOR>();
+    return (vendor.find("NVIDIA") == 0 ||
+            vendor.find("Advanced Micro Devices") == 0 ||
+            vendor.find("Apple") == 0 ||
+            vendor.find("Intel") == 0);
+}
+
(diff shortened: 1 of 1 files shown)


jaimergp:
Wow, I didn't know this was in! My tests with PPC/ARM show that this indeed working! See here:

* [ARM](https://cloud.drone.io/conda-forge/openmm-feedstock/23/1/2) (line 7116)
* [Linux 64](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=188196&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=6d4b912b-175d-51da-0fd9-4d30fe1eb4e7&l=4862)
* [MacOS](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=188196&view=logs&j=9c5ef928-2cd6-52e5-dbe6-9d173a7d951b&t=20c71c51-4b27-578b-485d-06ade2de1d00&l=6461)
* [PPC](https://travis-ci.com/github/conda-forge/openmm-feedstock/jobs/364327404#L8586)

I notice, however, that this [writes directy to stdout](https://github.com/openmm/openmm/pull/2778/files#diff-5709cad511dfdd57f7e206b3b5ea37e7R192). Would it be more explicit to write to stderr? Is it possible to throw warnings in the Python layer too?

peastman:
> Is it possible to throw warnings in the Python layer too?

The same message should get printed whether you're using Python or C++.  Or do you mean throw an exception and refuse to run?

jaimergp:
Not necessarily an exception, but a [warning](https://docs.python.org/3/library/warnings.html) object. Or maybe do throw an exception, but catch it at Python level and depending on a boolean either emit a warning or raise the exception so we have more programmatic control over the Platform. Really thinking out loud now, we can talk about this in the call later.

PR #2790, opened 2020-07-29
States can save integrator parameters

### LINKED ISSUE #2649: Allow Context.{get|set}State() to save/restore CustomIntegrator variables
It would be useful if we could allow `Context.{get|set}State()` to save and restore `CustomIntegrator` variables.

We've recently discovered a use case where suspending/resuming a simulation in the Folding@home core could make use of this feature without requiring we store `CustomIntegrator` parameters outside of the serialized `State`.

The only API extension required would be to add a `getIntegratorParameters=True` optional argument to `Context.getState()` which would also retrieve any integrator parameters.

If present in the `State` object, they would automatically be reset when `Context.setState(state)` is called.

peastman:
Fixes #2649.

### DIFF
--- a/olla/include/openmm/kernels.h
+++ b/olla/include/openmm/kernels.h
@@ -1136,6 +1136,22 @@ public:
      * Load the chain states from a checkpoint.
      */
     virtual void loadCheckpoint(ContextImpl& context, std::istream& stream) = 0;
+    /**
+     * Get the internal states of all chains.
+     * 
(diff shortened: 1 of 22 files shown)


jchodera:
Woohoo! This looks great!

Do we need to update the user guide at all?

peastman:
I don't think so.  The only user visible change is one extra flag you can pass to `getState()`, and `Simulation.saveState()` includes that flag automatically.

PR #2794, opened 2020-08-04
error in _mbondi3_radii

ajasja:
_mbondi3_radii did not set the GLU and ASP oxygen radii

As described in this [forum post](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=12300&p=0&start=0&view=&sid=2ca8b267a6c47694ea1822f218f80375), I made a system with tleap, that included `set default PBRadii mbondi3` in the in file.

However OpenMM complained about incorrect radii for type O2. The value in the topology file is the correct 1.4.
I have tracked the error to `_mbondi3_radii` in costomgbforces.py.  I have updated that part of the code to read like it's [parmed counterpart](https://parmed.github.io/ParmEd/html/_modules/parmed/tools/changeradii.html#mbondi3).

### DIFF
--- a/wrappers/python/simtk/openmm/app/internal/customgbforces.py
+++ b/wrappers/python/simtk/openmm/app/internal/customgbforces.py
@@ -342,12 +342,19 @@ def _mbondi3_radii(topology, all_bonds = None):
         all_bonds = _get_bonded_atom_list(topology)
     radii = _mbondi2_radii(topology, all_bonds=all_bonds)
     for i, atom in enumerate(topology.atoms()):
-        # carboxylate and HH/HE (ARG)
-        if _is_carboxylateO(atom, all_bonds):
-            radii[i] = 1.4
+        if atom.residue.name in ('GLU', 'ASP', 'GL4', 'AS4'):
+            if atom.name.startswith('OE') or atom.name.startswith('OD'):
+                radii[i] = 1.4
         elif atom.residue.name == 'ARG':
             if atom.name.startswith('HH') or atom.name.startswith('HE'):
                 radii[i] = 1.17
+        
+        # Adjust carboxylate O radii on C-Termini. Don't just do the end
+        # residue, since we can have C-termini in the middle as well
+        # (i.e., 2-chain dimers)
+        if atom.name == 'OXT':
+            radii[i] = 1.4
+            radii[i-1] = 1.4
     return radii  # Converted to nanometers above
 
 


peastman:
@swails does this change look correct to you?

There's a test failure in TestCharmmFiles.py, because it checks to make sure the forces haven't changed since the last version of OpenMM.  This does change them of course.  Possibly we just need to update the values that test compares to, but first we need to validate this is now producing correct results.  The best way of doing that would be to compare to forces from Amber using GBn2.

wrappers/python/simtk/openmm/app/internal/customgbforces.py
```diff
@@ -342,12 +342,19 @@ def _mbondi3_radii(topology, all_bonds = None):
...
+            if atom.name.startswith('OE') or atom.name.startswith('OD'):
+                radii[i] = 1.4
         elif atom.residue.name == 'ARG':
             if atom.name.startswith('HH') or atom.name.startswith('HE'):
                 radii[i] = 1.17
+        
+        # Adjust carboxylate O radii on C-Termini. Don't just do the end
+        # residue, since we can have C-termini in the middle as well
+        # (i.e., 2-chain dimers)
+        if atom.name == 'OXT':
```
peastman, line 355: This line is dangerous because it just looks at the atom name without considering what residue it's in.  Atom names have no defined meaning except in the context of a particular residue.  I assume this is only meant to apply to proteins?  If so we should also check the residue name to make sure it's a protein residue.

swails:
The proposed change here looks Amber-specific.  The fix should be added to `_is_carboxylateO` to catch whatever case was missed here.

peastman:
Note that this code was basically copied straight from ParmEd.  Does it need fixing there too?

wrappers/python/simtk/openmm/app/internal/customgbforces.py
ajasja: Yes, probably. I must confess I don't completely understand why the [i-1] radius gets changes as well.

ajasja:
> There's a test failure in TestCharmmFiles.py, because it checks to make sure the forces haven't changed since the last version of OpenMM. **This does change them of course.**

So OpenMM is not using the radii in the topology? I thought these values were just for reference. 
OpenMM does seem to [use the topology provided radii](https://github.com/openmm/openmm/blob/49a00025f7302dc1d1e1745996d27e9e62598d5f/wrappers/python/simtk/openmm/app/internal/amber_file_parser.py#L1022), so I don't know why the energies would be different. Especially not in a Charmm test that is not connected to Amber code-path at all.

peastman:
The code you modified is used by multiple file readers, not just Amber.  The Amber reader gets radii from the file and just uses these ones as a safety check.  Other readers use them directly.

ajasja:
Aaa, that makes sense (I see a call in the `charmmpsffile.py` to getStandardParameters).
I also saw an (abandoned?) implementation of `_mbondi3_radii` which is very similar to the one in parmed

https://github.com/openmm/openmm/blob/fde2ca6d99ff55481ff8bc909f72053a8182fa1f/wrappers/python/simtk/openmm/app/charmmpsffile.py#L1807

wrappers/python/simtk/openmm/app/internal/customgbforces.py
swails: The `[i - 1]` radius gets changed because in the Amber files, the "other" terminal carboxylate oxygen comes immediately before the OXT atom name.

This solution is Amber-specific, and the original code *should* properly handle everything.  If it doesn't, then `_is_carboxylateO` should be fixed so that it identifies all carboxylate oxygens.

PR #2800, opened 2020-08-08
Fixed typo in forcefield.py: 2DFunction -> 3DFunction

Dann239:


peastman:
Thanks!

jchodera:
Doesn't this mean we should add to our test coverage?

PR #2802, opened 2020-08-11
Updated to latest values for physical constants

### LINKED ISSUE #2427: Update values for physical constants
Earlier this year there was a major redefinition of the SI unit system: https://en.wikipedia.org/wiki/2019_redefinition_of_the_SI_base_units.  As part of that, several values that used to be experimental quantities (and hence varied depending on which experiment you took them from) were redefined to have precise values.  That includes Planck's constant, Boltzmann's constant, Avogadro's number, and the electron charge.

Should we update the values in OpenMM to use the new definitions?  That will of course change people's results, though only very slighly.  For example, OpenMM uses 6.0221367e23 for Avogadro's number, but it's now defined to be 6.02214076e23.

peastman:
Fixes #2427.  I also tried to reduce the number of places where constants are scattered through the code.

### DIFF
--- a/platforms/reference/include/SimTKOpenMMRealType.h
+++ b/platforms/reference/include/SimTKOpenMMRealType.h
@@ -1,5 +1,5 @@
 
-/* Portions copyright (c) 2006-2017 Stanford University and Simbios.
+/* Portions copyright (c) 2006-2020 Stanford University and Simbios.
  * Contributors: Pande Group
  *
  * Permission is hereby granted, free of charge, to any person obtaining
@@ -34,10 +34,6 @@
 #define SIN            sin
 #define COS            cos
 #define TAN            tan
-
-// LOG is used in Vishal's gpu code; modifying LOG -> LN 
-#define LN             log
-
 #define EXP            exp
 #define FABS           fabs
 #define ACOS           acos
@@ -49,24 +45,6 @@
 
 #define ATOF           atof
 
-#define PI_M               3.141592653589
-#define TWO_SIX            1.122462048309372981
-#define RADIAN            57.29577951308
-#define RADIAN_TO_DEGREE  57.29577951308
-#define LOG_TEN            2.302585092994045684
-#define SQRT_TWO           1.41421356237309504
-#define DEGREE_TO_RADIAN   0.01745329252
-#define RADIAN_INVERSE     0.01745329252
-
-#define DOT3(u,v) ((u[0])*(v[0]) + (u[1])*(v[1]) + (u[2])*(v[2]))
-
-#define MATRIXDOT3(u,v) u[0]*v[0] + u[1]*v[1] + u[2]*v[2] + \
-                        u[3]*v[3] + u[4]*v[4] + u[5]*v[5] + \
-                        u[6]*v[6] + u[7]*v[7] + u[8]*v[8]
-
-
-// physics constants -- from Gromacs physics.h
-
 #ifndef M_PI
 #ifdef _PI
 #define M_PI _PI
@@ -74,6 +52,15 @@
 #define M_PI        3.14159265358979323846
 #endif
 #endif
+
+#define PI_M               M_PI
+#define RADIAN             (180/M_PI)
+#define SQRT_TWO           1.41421356237309504
+
+#define DOT3(u,v) ((u[0])*(v[0]) + (u[1])*(v[1]) + (u[2])*(v[2]))
+
+
+// Physical constants are from https://pml.nist.gov/cuu/Constants
 
 #define ANGSTROM     (1e-10) 
 #define KILO         (1e3)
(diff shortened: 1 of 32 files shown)


platforms/reference/include/SimTKOpenMMRealType.h
```diff
@@ -75,6 +53,15 @@
...
 #endif
 
+#define PI_M               M_PI
+#define RADIAN             (180/M_PI)
+#define SQRT_TWO           1.41421356237309504
+
+#define DOT3(u,v) ((u[0])*(v[0]) + (u[1])*(v[1]) + (u[2])*(v[2]))
+
+
+// Physical constants are from https://pml.nist.gov/cuu/Constants
```
jchodera, line 63: @peastman : This is insufficient provenance information, since it doesn't provide a versioned reference for the physical constants you are using. 

You'll recall from https://github.com/openmm/openmm/issues/2427 that the appropriate standards body is CODATA, and the page that you point to recommends the [CODATA 2018](https://pml.nist.gov/cuu/Constants/) constants.

Let's 
* explicitly state in the header that this is the CODATA 2018 set
* add to the changelog that we have updated to CODATA 2018
* note in the user and dev guides that we use CODATA 2018 physical constants

peastman: > You'll recall from #2427 that the appropriate standards body is CODATA, and the page that you point to recommends the CODATA 2018 constants.

Look at the URL you just linked to.  Now look at the one in the comment.  Hint: they're the same. :)

jchodera: How does that address any of the three stated actions bullet points above?

That URL hosts pointers to both the latest and historical CODATA sets. The actions I articulate above state that we should *specify which of the sets listed in the URL* we are using. 

We have no control over that site, and it is clear that the newer CODATA version will be posted there when it is available, so simply linking to that site without specifying which version of CODATA there we use is completely unhelpful.

peastman:
Is this ok to merge now?

plugins/drude/tests/TestDrudeNoseHoover.h
```diff
@@ -131,7 +131,7 @@ void testWaterBox() {
     double frequencyDrude = 1000.0;
     int randomSeed = 100;
     DrudeNoseHooverIntegrator integ(temperature, frequency,
-                                    temperatureDrude, frequencyDrude, 0.0005,
+                                    temperatureDrude, frequencyDrude, 0.0004,
```
jchodera, line 134: Did this change accidentally creep in?

plugins/amoeba/serialization/src/AmoebaVdwForceProxy.cpp
```diff
@@ -85,7 +85,7 @@ void AmoebaVdwForceProxy::serialize(const void* object, SerializationNode& node)
             type.setDoubleProperty("sigma", sigma).setDoubleProperty("epsilon", epsilon);
         }
         SerializationNode& pairs = node.createChildNode("TypePairs");
-        for (int i = 0; i < force.getNumParticleTypes(); i++) {
+        for (int i = 0; i < force.getNumTypePairs(); i++) {
```
jchodera, line 88: Is this a bugfix?

wrappers/python/simtk/unit/constants.py
```diff
@@ -41,8 +41,8 @@
 #################
 
 # codata 2006
```
jchodera, line 43: This should be changed to "CODATA 2018"

wrappers/python/simtk/unit/unit_definitions.py
```diff
@@ -157,7 +157,7 @@
 moles = mole = Unit({mole_base_unit: 1.0})
 
 single_item_amount_base_unit = BaseUnit(amount_dimension, "item", "")
-mole_base_unit.define_conversion_factor_to(single_item_amount_base_unit, 6.0221417930e23)
+mole_base_unit.define_conversion_factor_to(single_item_amount_base_unit, 6.02214076e23)
```
jchodera, line 160: Can we add a note in this file that we updated this to CODATA 2018, in case someone ever copies this file?

Also, a number of the other conversion factors in this file have not been updated to CODATA 2018.
* [Planck time](https://physics.nist.gov/cgi-bin/cuu/Value?plkt)
* [Planck length](https://physics.nist.gov/cgi-bin/cuu/Value?plkl|search_for=planck)
* [Bohr radius](https://physics.nist.gov/cgi-bin/cuu/Value?bohrrada0|search_for=bohr)

Hm, it actually looks like _all_ of these conversion factors need updating!

plugins/drude/tests/TestDrudeNoseHoover.h
peastman: That was intentional.  The changes caused this test to start failing, but on investigating I discovered it was just that the test was right on the edge of passing/failing, and any tiny perturbation to it could make it fail.  Reducing the time step slightly makes it much more robust.

plugins/amoeba/serialization/src/AmoebaVdwForceProxy.cpp
peastman: Correct.

jchodera: Great! Let's just document this in the top of the PR so we can list this bugfix when we write release notes?

peastman: It fixed a bug that was just introduced a few weeks ago in #2765.  The release notes only list fixes to bugs that were present in previous releases, not ones that were created and then fixed in the course of a single release.

peastman:
The Python unit definitions are updated.

PR #2818, opened 2020-08-27
CCMA with a small number of constraints uses a single kernel

### LINKED ISSUE #2814: A small number of constraints have extreme impact of CustomIntegrator performance
The most recent batch of Folding@home free energy calculations has run into a surprising issue where the calculations are exceptionally slow. A bit of debugging has identified that a handful of restraints in the small molecule are responsible for this, where ligand-in-solvent goes from 230 ns/day down to 90 ns/day on a GTX 1080 Ti. (The same issues also manifest in the large complex simulation, but are harder to interactively debug there.)

Here is an attached ZIP file and a short benchmark script. While the CustomIntegrator is rather complex (for nonequilibrium cycling free energy calculations), commenting out the constraints for the solute (particles 0-50) takes the simulation from 90 ns/day -> 230 ns/day. 

This may be running into similar issues as observed earlier, where near-colinear constrained bonds can cause pathological problems for CCMA. This may occur frequently in alchemical calculations because old and new atoms are excluded from each other. 

Is it easy to establish if this is the case? If so, can we also support an alternative algorithm (e.g. SHAKE/RATTLE) that is slightly less efficient for normal dynamics but incredibly more efficient than CCMA for free energy calculations?
[ccma-issues.zip](https://github.com/openmm/openmm/files/5121621/ccma-issues.zip)

peastman:
Fixes #2814.  For systems with up to 1024 CCMA constraints, it runs that whole algorithm in a single kernel.  This involves much less overhead than the standard implementation, but also requires it to be run in a single thread block.

I haven't tested this with CUDA yet.  The cutoff of 1024 is roughly where the performance crossover happens on my laptop's GPUs, but higher end GPUs might have different performance.  This will need a lot of testing.

### DIFF
--- a/platforms/common/include/openmm/common/IntegrationUtilities.h
+++ b/platforms/common/include/openmm/common/IntegrationUtilities.h
@@ -148,6 +148,7 @@ protected:
     ComputeArray randomSeed;
     ComputeArray stepSize;
     ComputeArray ccmaAtoms;
+    ComputeArray ccmaConstraintAtoms;
     ComputeArray ccmaDistance;
     ComputeArray ccmaReducedMass;
     ComputeArray ccmaAtomConstraints;
--- a/platforms/common/include/openmm/common/IntegrationUtilities.h
+++ b/platforms/common/include/openmm/common/IntegrationUtilities.h
@@ -136,7 +136,7 @@ protected:
     ComputeKernel settlePosKernel, settleVelKernel;
     ComputeKernel shakePosKernel, shakeVelKernel;
     ComputeKernel ccmaDirectionsKernel, ccmaPosForceKernel, ccmaVelForceKernel;
-    ComputeKernel ccmaMultiplyKernel, ccmaUpdateKernel;
+    ComputeKernel ccmaMultiplyKernel, ccmaUpdateKernel, ccmaFullKernel;
     ComputeKernel vsitePositionKernel, vsiteForceKernel, vsiteSaveForcesKernel;
     ComputeKernel randomKernel, timeShiftKernel;
     ComputeArray posDelta;
--- a/platforms/common/src/IntegrationUtilities.cpp
+++ b/platforms/common/src/IntegrationUtilities.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2009-2019 Stanford University and the Authors.      *
+ * Portions copyright (c) 2009-2020 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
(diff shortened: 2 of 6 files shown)


jchodera:
Awesome! Thanks for putting this together so quickly!

Did you test this on the system in #2814? What is the impact on performance?

peastman:
CUDA tests are now passing.  If you can test it on that system and see how the performance changes, that would be great.

jchodera:
Let me build and test this morning!

jchodera:
Apologies this took all day to test---I don't have a good build environment set up for OpenMM anymore. We should probably invest in defining a conda environment that can be used for rapidly creating a useful OpenMM local build environment.

[Here's an example](https://anaconda.org/openforcefield/qcarchive-worker-openff/files) of how that works with an [environment YAML file](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#).

I built a conda package locally on linux using this recipe:
https://github.com/omnia-md/conda-dev-recipes/pull/197

These tests were all run on a GTX 1080 Ti with OpenCL in mixed precision.

I see no impact from this change:
* openmm 7.5.0b1: 29.126 ns/day
* ccma branch: 29.641 ns/day
* openmm 7.5.0b1 with one constraint removed: 94.887 ns/day
* ccma branch with one constraint removed: 87.324 ns/day

jchodera:
Here is the benchmark script and the test inputs:
[ccma-test.zip](https://github.com/openmm/openmm/files/5139784/ccma-test.zip)

jchodera:
**EDIT:** I just realized these were set up with 2 fs (instead of 4 fs) timesteps, so the impact of constraints is likely less severe here than in the 4 fs example. You can change the timestep manually in `integrator.xml`.
Here are timings with 4 fs timestep:
* openmm 7.5.0b1: 59.793 ns/day
* ccma branch: 150.956 ns/day
* openmm 7.5.0b1 with one constraint removed: 219.966 ns/day
* ccma branch with one constraint removed: 216.928 ns/day

So there _is_ a significant benefit for 4 fs timesteps, though it doesn't completely restore performance.

peastman:
Try it again.  I made some further optimizations, and it should be significantly faster now.

jchodera:
Thanks! Testing this now.

jchodera:
I'm having a bit of trouble building conda packages locally to test this. It would be great if we could make sure our conda packaging requirements also included automated packaging of variants for all branches so we can easily install and test.

jchodera:
I think I've managed to build these. Testing now.

jchodera:
The results with the debug ZIP archive I posted above are very impressive!

4 fs timestep:
* unconstrained: 219 ns/day
* constrained (this PR): 183 ns/day
* constrained (7.4.2): 62 ns/day
2 fs timestep:
* unconstrained: 96 ns/day
* constrained (this PR): 93 ns/day
* constrained (7.4.2): 39 ns/day

Can we get this into 7.5.0?

peastman:
Thanks!

jchodera:
@peastman : This is puzzling: The above numbers were for OpenCL `mixed` precision. I've been trying to benchmark with both OpenCL and CUDA to compare performance. 

On a a GTX 1080Ti, it seems like there may be a performance regression on CUDA:

4 fs timestep:
```
unconstrained OpenMM 7.4.2: 154 ns/day CUDA, 177 ns/day OpenCL
unconstrained OpenMM 7.5.0 ccma branch: 128 ns/day CUDA, 176 ns/day OpenCL
constrained OpenMM 7.4.2: 70 ns/day CUDA, 78 ns/day OpenCL
constrained OpenMM 7.5.0 ccma branch: 124 ns/day CUDA, 162 ns/day OpenCL
```

2 fs timestep:
```
unconstrained OpenMM 7.4.2: 78 ns/day CUDA, 89 ns/day OpenCL
unconstrained OpenMM 7.5.0 ccma branch: 55 ns/day CUDA, 97 ns/day OpenCL
constrained OpenMM 7.4.2: 37 ns/day CUDA, 26 ns/day OpenCL
constrained OpenMM 7.5.0 ccma branch: 76 ns/day, 88 ns/day OpenCL
```

Here, "unconstrained" here means we just remove the one extra constraint that triggers CCMA---apologies for the unclear language.

peastman:
Which is the constraint to remove?  I'll try to reproduce this.

jchodera:
Here's the constraint I commented out in `system.xml`:
```XML
                <!-- <Constraint d=".1020056453864" p1="25" p2="50"/> -->
```
To get this system (which I foolishly set up in OpenMM 7.5.0 dev build) to run with OpenMM 7.4.2, I needed to change the `NonbondedForce` tag version from `4` to `3` in this line:
```
                <Force alpha="0" cutoff="1" dispersionCorrection="1" ewaldTolerance=".00025" exceptionsUsePeriodic="0" forceGroup="0" ljAlpha="0" ljnx="0" ljny="0" ljnz="0" method="4" nx="0" ny="0" nz="0" recipForceGroup="-1" rfDielectric\
="78.3" switchingDistance="-1" type="NonbondedForce" useSwitchingFunction="0" version="4">
```
If this is an issue, I can set up again with 7.4.2.

Here's the ZIP file with the modified `system.xml` with the constraint commented out for you and the version altered:
[ccma-test-2.zip](https://github.com/openmm/openmm/files/5172066/ccma-test-2.zip)

peastman:
I'm not seeing this.  I ran a bunch of simulations with CUDA on a Titan V, both with and without that one constraint.  I increased the number of steps from 500 to 5000, since that produces smaller fluctuations in the results.  And I ran it twice for each set of conditions.  Both numbers are shown below so you can see how big the typical variations are.

First with the current master branch:

without constraint: 344, 342
with constraint: 288, 291

Then with a recent revision from just before this change was merged:

without constraint: 347, 338
with constraint: 172, 175

And finally with 7.4.2:

without constraint: 326, 340
with constraint: 175, 173

jchodera:
Thanks for independently trying to replicate this. I tried again with 500 -> 5000 timesteps and I am now seeing performance much more in line with your findings.

jchodera:
Hm, it also seems that `OPENMM_CUDA_COMPILER` was being set automatically in my case, which seems to have some impact on results.

PR #2823, opened 2020-08-30
added support for NBFIX with 1-4 special term

aakognole:
Just a small addition for the OpenMM users using CHARMM ForceField. One can NBFIX specific 1-4 interactions involving all atoms of the specified types.

e.g.

NBFIX
CD315B  OD30B -0.07376  3.77000  0.00  3.77000

This will zero out all CD315B - OD30B 1-4 interactions without affecting ordinary CD315B  OD30B interactions.

Attached are comparisons with CHARMM (c44b1) energies for before and after introducing the NBFIX terms with special 1-4 parameters.
[ener.txt](https://github.com/openmm/openmm/files/5147436/ener.txt)

Thanks.


jchodera:
Awesome! 

Should this fix go into https://github.com/ParmEd/ParmEd (which shares much of the same PSF infrastructure) too?

aakognole:
> Awesome!
> 
> Should this fix go into https://github.com/ParmEd/ParmEd (which shares much of the same PSF infrastructure) too?

Sure. That will be great! Thanks.

peastman:
Thanks!  That's great!

Would it be possible for you to add a test to verify this code is working correctly?  It can go in [TestCharmmFiles.py](https://github.com/openmm/openmm/blob/master/wrappers/python/tests/TestCharmmFiles.py).

Your code could be simplified slightly: two of the four lines in the two blocks are actually identical.  You could simplify it to

```python
charge_prod = (atom1.charge * atom4.charge)
if str(atom4.type) in atom1.type.nbfix:
    epsilon = (atom1.type.nbfix[str(atom4.type)][3]) * ene_conv
    sigma = (atom1.type.nbfix[str(atom4.type)][2]) * (length_conv * sigma_scale)
else:
    epsilon = sqrt(abs(atom1.type.epsilon_14 * atom4.type.epsilon_14)) * ene_conv
    sigma = (atom1.type.rmin_14 + atom4.type.rmin_14) * (length_conv * sigma_scale)
force.addException(ia1, ia4, charge_prod, sigma, epsilon)
```

aakognole:
Hi @peastman 
Simplified the code and added a test case. Added a new system not to disturb other tests. Let me know if you want any more changes. Thank you!

peastman:
The test you added is failing:

```
>       self.assertAlmostEqual(ene, -1103.11598, delta=1.0)
E       AssertionError: -371.5010107233917 != -1103.11598 within 1.0 delta (731.6149692766083 difference)
```

aakognole:
Yes. I double checked the potential energy with CHARMM (-1112.12) and it comes close to my value.  I don't understand why the energy is so much different here.

aakognole:
The fix is in #3324.

PR #2833, opened 2020-09-04
Added code of conduct

peastman:
This is the one [recommended by MolSSI](https://github.com/MolSSI/cookiecutter-cms).  It's based on the [Contributor Covenant](http://contributor-covenant.org).


jchodera (approved):
Looks great!

Should we add an attribution link to the MolSSI version at the bottom?

peastman:
I think they just copied it straight from the CC site.  The text here is exactly what you would get if you created your site from the cookiecutter template.

PR #2837, opened 2020-09-09
Fixed illegal memory access in nonbonded kernel

### LINKED ISSUE #2830: bug in revised AMOEBA plugin code for Vdw interactions
There is a bug in the recently revised AMOEBA plugin code for Vdw interactions. This would be the code updated under the merged/closed pull request "Added new features to AmoebaVdwForce #2765".

The problem occurs for a system having a number of atoms that is an exact multiple of 32. For those systems, there is an attempt to access an undefined atom off the end of an atom block. This was patched in the earlier pull request we put up some time ago ("AMOEBA Update, 8 May 2020 #2674"), but the patch did not make it into the final revised vdw version.

Code from the original pull request that works correctly can be found at:

https://github.com/jayponder/openmm/blob/AMOEBA-Update/plugins/amoeba/platforms/cuda/src/kernels/amoebaVdwForce3.cu

In particular, note the "if clause" shown below that is added around the bulk of the code in the above link from the original pull request. This if clause is missing or was removed in the revised current version.

if (atom2 < NUM_ATOMS) {
.
.
}

There may be other ways to solve this, perhaps deeper inside OpenMM, but the above patch seems to be a general solution in our hands.

Also, in the same section of the revised current code, the variable "tempForce" is not used and can be removed.

peastman:
Fixes #2830.  This only caused problems in unusual situations.  It came up in the revised AMOEBA vdw kernel.  It's possible there could have been other situations where it would have produced errors, but I don't know of anyone actually encountering any.

### DIFF
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -651,6 +651,21 @@ CUfunction CudaNonbondedUtilities::createInteractionKernel(const string& source,
         }
     }
     replacements["LOAD_ATOM2_PARAMETERS"] = load2j.str();
+    
+    stringstream clearLocal;
+    for (int i = 0; i < (int) params.size(); i++) {
+        if (useShuffle)
+            clearLocal<<"shfl";
+        else
+            clearLocal<<"localData[atom2].";
+        clearLocal<<params[i].getName()<<" = ";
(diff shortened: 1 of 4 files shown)


ponderlab:
Hi Peter, Thanks very much for tracking this down and putting in a fix!

jayponder:
Hi Peter,  Thanks very much for tracking this down and putting in a fix!

PR #2878, opened 2020-10-07
Reduced cutoff for short list sorting kernel

### LINKED ISSUE #2875: TileSize 32 and performance of AMD GCN
This is more a question than an issue. I'm trying to understand why AMD GCN (Radeon VII) on many projects provides low throughput. Browsing through the code I find TileSize set to 32, while GCN has 64 parallel threads in a CU.

Does the tile size of 32 still use the full width of the GCN CUs? Or is the TileSize completely unrelated to simdWidth, which is 64?

Sorry for asking beginners questions, if too stupid, please just do not respond.

peastman:
Fixes #2875.

### DIFF
--- a/platforms/cuda/src/CudaSort.cpp
+++ b/platforms/cuda/src/CudaSort.cpp
@@ -58,7 +58,7 @@ CudaSort::CudaSort(CudaContext& context, SortTrait* trait, unsigned int length) 
     int maxSharedMem;
     cuDeviceGetAttribute(&maxSharedMem, CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK, context.getDevice());
     int maxLocalBuffer = (maxSharedMem/trait->getDataSize())/2;
-    int maxShortList = min(8192, max(maxLocalBuffer, CudaContext::ThreadBlockSize*context.getNumThreadBlocks()));
+    int maxShortList = min(3000, max(maxLocalBuffer, CudaContext::ThreadBlockSize*context.getNumThreadBlocks()));
     isShortList = (length <= maxShortList);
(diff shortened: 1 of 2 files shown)


jchodera:
How often will we have to keep tweaking these hard-coded constants?

bdenhollander:
Is the number of elements to be sorted or the number of bytes what determines the efficiency of sortShortList? Sorting an array of floats vs. a list of x, y, z doubles will use the same cutoff point.

`int maxLocalBuffer = (maxSharedMem/trait->getDataSize())/2;`
This takes element size into account.

`int maxShortList = max(maxLocalBuffer, (int) OpenCLContext::ThreadBlockSize*context.getNumThreadBlocks());` 
This discards memory size in favour of compute elements since the right side dominates the max on even a small GPU like an AMD RX 460.

In the linked issue, what was `trait->getDataSize()` used in the benchmark? If it was 16B, 1024 as a constant would make sense.
`maxSharedMem` is 32KB on AMD (48KB on NVIDIA).
`maxLocalBuffer = (32*1024)/16/2 = 1024 elements`

peastman:
The speed is determined by the number of elements.  The check on memory size is to make sure the whole list can be loaded into shared memory.  But even if we *can* use the short list kernel, that doesn't mean we *should*.  For longer lists it ends up being slower.  That's what this PR is for.

peastman:
> How often will we have to keep tweaking these hard-coded constants?

Hopefully never again, but it's impossible to say for sure.  Hardware keeps changing, and we need to keep adapting the code as it does.  It's equally possible that by then we'll have switched to a different sorting method, or even eliminated the need for the sort altogether.  I'm currently thinking about some ideas for that.

PR #2891, opened 2020-10-18
Update IntegrationUtilities.cpp

informatorius:
* Use pinned buffer for context download
* Use template to reduce duplicate code

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -37,12 +37,12 @@ the results.
 Installing OpenMM
 *****************
 
-OpenMM is installed using the Conda package manager (http://conda.pydata.org).
+OpenMM is installed using the Conda package manager (https://docs.conda.io).
 Conda is included as part of the Anaconda Python distribution, which you can
-download from http://docs.continuum.io/anaconda/install.  This is a Python
+download from https://docs.continuum.io/anaconda/install.  This is a Python
 distribution specifically designed for scientific applications, with many of the
 most popular mathematical and scientific packages preinstalled.  Alternatively
-you can use Miniconda (available from http://conda.pydata.org/miniconda.html),
+you can use Miniconda (available from https://docs.conda.io/en/latest/miniconda.html),
 which includes only Python itself, plus the Conda package manager.  That offers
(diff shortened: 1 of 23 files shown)


peastman:
Looks good.  Since we're so close to releasing 7.5, I'm currently only merging important bug fixes.  I'll leave this open until 7.5 is out, then merge it.

informatorius:
OK. There are several parts in OpenMM code where a check is done if double/float/mixed should be used and then use almost identical code parts except the type. Some of these code parts could remove duplicate code by using templates. 

Would this template pattern be desired? One disadvantage e.g. is MS Visual Studio does not show intellisense for template code (yet).

peastman:
It depends on the situation.  Which leads to a smaller amount of total code?  Github reports that this PR added 18 lines and removed 12 lines.  So by "removing" duplicate code, it actually increased the total amount of code.

informatorius:
The advantage would be if you want to change the code later then it only needs to change in one part and not two or more. And duplicate code has the risk that both parts then are different by mistake.

peastman:
> The advantage would be if you want to change the code later then it only needs to change in one part and not two or more. 

Right, it's always a judgement call.  How likely is it to change in the future?  This code is just a trivial loop over an array, summing the elements, so very unlikely to change.  And there's no possibility of us later deciding we want it to calculate something different, because kinetic energy is precisely defined.  There's one and only one possible thing for it to calculate.  For other routines, the odds of it changing in the future might be much higher, and so the benefit of centralizing the code would be higher.

> And duplicate code has the risk that both parts then are different by mistake.

Any code at all has a risk of containing a mistake.  That's why the total number of lines of code makes a good metric: all else being equal, the more lines of code you have, the higher the risk there's an error in it somewhere.

informatorius:
Then maybe only take the 
*Use pinned buffer for context download
as minor performance improvement

peastman:
7.5 is released, so we're ready to merge backlogged PRs.  Do you want to make the changes you described above?

informatorius:
Pull request updated to only use pinned buffer for minor performance improvement

peastman:
Thanks!

PR #2897, opened 2020-10-22
Added HMR to make ApoA1 benchmark more stable

### LINKED ISSUE #2864: benchmark.py tests occasionally fail
@peastman 

Just in case this is at all helpful I ran the script ten times to see which test it failed on:

Test: pme (cutoff=0.9)
Test: apoa1pme
Finished
Test: apoa1rf
Finished
Test: apoa1rf
Test: apoa1rf
Test: apoa1pme
Test: apoa1rf
Test: apoa1rf

peastman:
Fixes #2864.

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -68,6 +68,7 @@ def runOneTest(testName, options):
                 method = app.CutoffPeriodic
                 cutoff = 1*unit.nanometers
             friction = 1*(1/unit.picoseconds)
+            hydrogenMass = 1.5*unit.amu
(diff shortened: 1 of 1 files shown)


jchodera:
We should probably change the `--heavy` (hydrogens) argument to `--allbonds` instead, since it doesn't accurately describe the use of heavy hydrogens with constraints to bonds to hydrogen only.

peastman:
Yeah, I need to reevaluate that option and figure out what really makes sense for it.

PR #2899, opened 2020-10-24
Lepton: Add three functions that can be used as activation functions in neural network.

HanatoK:
I am implementing the forward propagation of the dense neural network as
a CV in Colvars, and this implementation allows to use a custom
activation function defined by Lepton. However, it's hard to write an
expression doing the same things as ReLU and ELU by combining the
existing functions, so in this commit, three additional functions are
added:

1. select2(x, expr1, expr2): Evaluate expr1 when x>0, otherwise
evaluate expr2. Basically this is a helper operation for ELU and RELU.

2. elu(x, alpha): Exponential linear unit. Evaluate alpha*(exp(x)-1)
when x>0, otherwise evaluate x.

3. relu(x): Rectified linear unit. Evaluate x when x>0, otherwise return
0.

The test cases can be found at https://gist.github.com/HanatoK/0390457acf199b2b5bc0c794f271d59f


peastman:
I think you can easily do all of these with the existing functions.

1. select(step(x), expr1, expr2)
2. select(step(x), alpha*(exp(x)-1), x)
3. max(x, 0)

HanatoK:
@peastman Thanks for your reply! I didn't think of the combination of `select` and `step`. I will close the PR and use the expressions that you suggested instead.

PR #2909, opened 2020-10-31
Fix 4GB memory check

bdenhollander:
4*(1<<30) evaluates to 0 so the check returns true on 2GB cards. Test runs fine on 2GB cards but requires over 4GB of system memory.


peastman:
Thanks!

PR #2926, opened 2020-11-16
cuFFT execution error checking

dmclark17:
Minor changes to avoid silent failure of cuFFT execution if an error occurs. Previously, if a cuFFT execution call returned a non-zero result, it would not trigger an exception. This would cause the affected unit tests to fail, seemingly for correctness issues, giving something like this:

```
101: exception: Assertion failure at TestEwald.h:191.   Expected [-198.058, -224.745, -102.259], found [-276.679, -263.117, -92.902]
1/1 Test #101: TestCudaEwaldSingle ..............***Failed    1.58 sec
```

I added error checks to the execution step similar to those of the plan creation. If the FFT execution returns a non-zero error code, it is now reflected in the failure message: 

```
101: exception: Error executing FFT: 1
1/1 Test #101: TestCudaEwaldSingle ..............***Failed    1.59 sec
```

I did not measure any impact on performance on the Apoa1 PME benchmark.

Thank you very much for your time!

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -1292,10 +1292,15 @@ double CudaCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeF
             cu.executeKernel(pmeFinishSpreadChargeKernel, finishSpreadArgs, gridSizeX*gridSizeY*gridSizeZ, 256);
 
             if (useCudaFFT) {
-                if (cu.getUseDoublePrecision())
-                    cufftExecD2Z(fftForward, (double*) pmeGrid1.getDevicePointer(), (double2*) pmeGrid2.getDevicePointer());
-                else
-                    cufftExecR2C(fftForward, (float*) pmeGrid1.getDevicePointer(), (float2*) pmeGrid2.getDevicePointer());
+                if (cu.getUseDoublePrecision()) {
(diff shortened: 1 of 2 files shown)


peastman:
Thanks!  We have a macro that can be used to simplify this code.  For example,

```c++
cufftResult result = cufftExecD2Z(fftForward, (double*) pmeGrid1.getDevicePointer(), (double2*) pmeGrid2.getDevicePointer());
if (result != CUFFT_SUCCESS)
    throw OpenMMException("Error executing FFT: "+cu.intToString(result));
```

can be simplified to

```c++
CHECK_RESULT(cufftExecD2Z(fftForward, (double*) pmeGrid1.getDevicePointer(), (double2*) pmeGrid2.getDevicePointer()), "Error executing FFT");
```

It looks like some of the indentation is messed up.  Can you make sure you're using spaces instead of tabs everywhere?

dmclark17:
Hi Peter,

Sorry about the tabs—I will replace them with spaces.

Since the cufft calls return a cufftResult type instead of a CUresult, I don't think the CHECK_RESULT macro in CudaKernels.cpp will work. Should I define a new macro that handles cufft result types?

peastman:
> Since the cufft calls return a cufftResult type instead of a CUresult, I don't think the CHECK_RESULT macro in CudaKernels.cpp will work. Should I define a new macro that handles cufft result types?

That makes sense.  Probably not worth the trouble of a macro, since it only gets called four times.

peastman:
Looks good.  I'll merge this once we make the cut for 7.5, which should be some time next week.

peastman:
7.5 is released, so merging!

PR #2944, opened 2020-12-08
Partially Fix Charmm NBThole Calculation Issue (Only for Cpu Platform now)

### LINKED ISSUE #2750: CHARMM Drude NBThole implementation
There is a difference in charmm and openmm energies for drude systems. 
```
Software | Version | Energy (kcal/mol)
         |         | FULL      | Only DNA | Only WATER | Only IONS
Charmm   | c44b2   | -99876.39 | -3054.58 |  -84250.37 | -93.45
OpenMM   | 7.4.2   | -98798.26 | -2245.06 |  -83989.81 | -85.04
``` 
The difference appears to be coming from this modification to add exception parameters into NBThole by @jing-huang

[https://github.com/openmm/openmm/commit/31f7f3559990caa924560118dfdf7f8be835d6fe](url)

Energies after commenting this part are acceptable. However there is a difference in water energy and ions which is another issue to solve.
```
Software | Version | Energy (kcal/mol)
         |         | FULL      | Only DNA | Only WATER | Only IONS
Charmm   | c44b2   | -99876.39 | -3054.58 |  -84250.37 | -93.45
OpenMM   | 7.4.2   | -98798.26 | -2245.06 |  -83989.81 | -85.04
OpenMM   | mod     | -99597.54 | -3044.19 |  -83989.92 | -85.04
``` 
Thanks.

qiuzy:
this commit fixes #2750 

This update fix OpenMM NBThole calculation bug on Cpu Platform.  One error source is 1-4 pair interactions was excluded in NBThole calculation. This leads to the dominant difference from Charmm result. So I add a function **addException(int particle1, int particle2)** into openmmapi/CustomNonbonded and also related change in CpuKernels and CpuCustonNonbonded source file.

Actually there is another error source which is from the cutoff criteria during NBThole calculation. Considering two atom pairs with drude atom, for example, heavy atom **C1** and it drude atom **D1**, heavy atom **C2** and its drude atom **D2**. They form four pairs, that is pair C1-C2, C1-D2, D1-C2, D1-D2. In Charmm, only the distance between C1 and C2 is less than cutoff, then the four pairs will calculate their NBThole interaction. While in OpenMM, whether the four pairs' NBThole interactions are calculated or not are depended on  the distance of the two atom in the pair, rather than decided by the distance of heavy atoms.
For now, this error source is not be fixed. This part is much less than the error led by first error source. One solution is to introduce the same group flag for the heavy atom and it drude atom. Other Suggestions are welcomed !!!!

**Compilation Error** 
This commit has a trobule in compiling when it generates C and Fortran wrapper.  The error information like this:
```
.../build/wrappers/OpenMMCWrapper.h:50:16: error: ‘OpenMM_CustomNonbondedForce’ has not been declared
 typedef struct OpenMM_CustomNonbondedForce::ExceptionInfo_struct OpenMM_CustomNonbondedForce::ExceptionInfo;
...
build/wrappers/OpenMMCWrapper.cpp:3:0:
.../build/wrappers/OpenMMCWrapper.h:569:51: error: invalid use of incomplete type ‘OpenMM_CustomNonbondedForce {aka struct OpenMM_CustomNonbondedForce_struct}’
 extern OPENMM_EXPORT OpenMM_CustomNonbondedForce::ExceptionInfo* OpenMM_CustomNonbondedForce::ExceptionInfo_create();
```
I don't know which I miss to change in files in serization and wrappers/, although I have added something in serialization/src/CustomNonbondedForceProxy.cpp.  My solution is to delete related part in OpenMMCWrapper.h and related files which is really rough. Sorry for this. And HELP is needed for this trouble.

**Another Issue Report**
During searching bug, I found in some case, the improper dihedral term is not corrected which is caused by the way of matching parameters
```
    # Now do the impropers
        for imp in self.improper_list:
            # Store the atoms
            a1, a2, a3, a4 = imp.atom1, imp.atom2, imp.atom3, imp.atom4
            at1, at2, at3, at4 = a1.attype, a2.attype, a3.attype, a4.attype
            key = tuple(sorted([at1, at2, at3, at4]))
            if not key in parmset.improper_types:
                # Check for wild-cards
                for anchor in (at2, at3, at4):
                    key = tuple(sorted([at1, anchor, 'X', 'X']))
                    if key in parmset.improper_types:
                        break # This is the right key
            try:
                imp.improper_type = parmset.improper_types[key]
            except KeyError:
                raise MissingParameter('No improper parameters found for %r' %
                                       imp)
```
This line **key = tuple(sorted([at1, at2, at3, at4]))** may find the incorrect dihedral parameters. For example, in one of my test case, I have a improper dihedral whose atom type is **CD2O4A-ND2R6C-ND2R6B-OD2C1B** whose K value is 80.00. But in the parameter file, there is another improper dihedral type below it: **CD2O4A-ND2R6B-ND2R6C-OD2C1B** whose K value is 110.00. Sort command will make this two type is the same, and the former one is override by the latter one and lead to a wrong value.  I think this issue is needed to be fixed. Please check it.

Now only CPU platform is corrected. I am working on fixing in CUDA platform and try to fixed it in this week.

### DIFF
--- a/openmmapi/include/openmm/CustomNonbondedForce.h
+++ b/openmmapi/include/openmm/CustomNonbondedForce.h
@@ -177,12 +177,6 @@ public:
         return exclusions.size();
     }
     /**
-     * Get the number of extra particle pairs whose interactions were needed to calculate.
-     */
-    int getNumExceptions() const {
-        return exceptions.size();
-    }
-    /**
      * Get the number of per-particle parameters that the interaction depends on.
      */
     int getNumPerParticleParameters() const {
@@ -413,30 +407,6 @@ public:
      */
     void createExclusionsFromBonds(const std::vector<std::pair<int, int> >& bonds, int bondCutoff);
     /**
-     * Add a particle pair to the list of interactions that should be excluded.
-     * 
-     * @param particle1  the index of the first particle in the pair
-     * @param particle2  the index of the second particle in the pair 
-     * @return  the index of the exclusion that was added 
-     */
-    int addException(int particle1, int particle2);
-    /**
-     * Get the particles in a pair whose interaction may be excluded and should be calculated.
-     *
-     * @param index            the index of the exception for which to get particle indices
-     * @param[out] particle1   the index of the first particle in the pair
-     * @param[out] particle2   the index of the second particle in the pair
-     */
-    void getExceptionParticles(int index, int& particle1, int& particle2) const;
-    /**
-     * Set the particles in a pair whose interaction shoud be added into exception list for calculation.
-     * 
-     * @param index           the index of the exception for which to set particle indices
-     * @param particle1       the index of the first particle in the pair
-     * @param particle2       the index of the second particle in the pair
-     */
-    void setExceptionParticles(int index, int particle1, int particle2);
-    /**
      * Add a tabulated function that may appear in the energy expression.
      *
      * @param name           the name of the function as it appears in expressions
@@ -540,7 +510,6 @@ private:
     class PerParticleParameterInfo;
     class GlobalParameterInfo;
     class ExclusionInfo;
-    class ExceptionInfo;
     class FunctionInfo;
     class InteractionGroupInfo;
     NonbondedMethod nonbondedMethod;
@@ -551,8 +520,6 @@ private:
     std::vector<GlobalParameterInfo> globalParameters;
     std::vector<ParticleInfo> particles;
     std::vector<ExclusionInfo> exclusions;
-    std::vector<ExceptionInfo> exceptions;  
-    std::map<std::pair<int, int>, int> exceptionMap;
     std::vector<FunctionInfo> functions;
     std::vector<InteractionGroupInfo> interactionGroups;
     std::vector<int> energyParameterDerivatives;
@@ -640,18 +607,6 @@ public:
         set1(set1), set2(set2) {
     }
 };
-
-
-class CustomNonbondedForce::ExceptionInfo {
-public:
-    int particle1, particle2;
-    ExceptionInfo() {
-    }
-    ExceptionInfo(int particle1, int particle2) :
-        particle1(particle1), particle2(particle2) {
-    }
-};
-
 
 } // namespace OpenMM
 
--- a/openmmapi/include/openmm/CustomNonbondedForce.h
+++ b/openmmapi/include/openmm/CustomNonbondedForce.h
@@ -177,6 +177,12 @@ public:
         return exclusions.size();
     }
     /**
+     * Get the number of extra particle pairs whose interactions were needed to calculate.
+     */
+    int getNumExceptions() const {
+        return exceptions.size();
+    }
+    /**
      * Get the number of per-particle parameters that the interaction depends on.
      */
     int getNumPerParticleParameters() const {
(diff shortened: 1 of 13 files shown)


peastman:
Thanks!  I'm kind of confused.  What exactly are "exceptions"?  How are they different from exclusions?  In CpuCustomNonbondedForce.cpp you add them to `groupInteractions` which is used for interaction groups, and is ignored if you haven't created any interaction groups.

qiuzy:
In my understanding, exclusion list includs those atom pairs which in the same bond, angle and dihedral. Their interaction will not be calculated. But, in genaral, the interactions between atom1 and atom4 in the same dihedral (1-4 interaction) are considered. So this is why 1-4 interaction is considered in NonbondedForce after exclusion list had been created. 

In previous NBThole calculation, these 1-4 pairs were not added after exclusion list set up which means those atoms are labelled as excluded and will not be calculated. In openmmapi/CustomNonbonded, **interactionGroups** includes two set of atom, for example, interactionGroups={{1,2,3,4,5}, {6,7,8,9,10}}, then CpuCustomNonbonded.cpp will check whether they are excluded and pair them in function:
```
void CpuCustomNonbondedForce::setInteractionGroups(const vector<pair<set<int>, set<int> > >& groups) {
    useInteractionGroups = true;
    for (auto& group : groups) {
        const set<int>& set1 = group.first;
        const set<int>& set2 = group.second;
        for (set<int>::const_iterator atom1 = set1.begin(); atom1 != set1.end(); ++atom1) {
            for (set<int>::const_iterator atom2 = set2.begin(); atom2 != set2.end(); ++atom2) {
                if (*atom1 == *atom2 || exclusions[*atom1].find(*atom2) != exclusions[*atom1].end())
                    continue; // This is an excluded interaction.
                if (*atom1 > *atom2 && set1.find(*atom2) != set1.end() && set2.find(*atom1) != set2.end())
                    continue; // Both atoms are in both sets, so skip duplicate interactions.
                groupInteractions.push_back(make_pair(*atom1, *atom2));
            }
        }
    }
}
```
In this function, 1-4 pairs with NBThole parameters will not be paired and also not calculated. So I add a function **addInteractionPairs** to add these 1-4 atom pairs into groupInteractions which has been created. Variable **NumExceptions** is the number of these pairs in exceptions, DIFFERENT FROM variable **NumExceptions**. Because addException is not necessary in other cases and NumExceptions can be zeros, addInteractionPairs will be called only if NumExceptions is larger than 0.

Actually, addException also can be found in NonbondedForce and CpuGayBerneForce.cpp. I admit exceptions and exclusions make people confused. I wish maybe I am wrong. Anyway, it is a name problem, will not affect my fix I think. 

As you see and what I mentioned, I meet compilation problem when generating wrapper files. I don't know where I need to change. **Could you tell me what I need to do to fix the compilation error?**

And the pull request seems wrong. I am not farmiliar with git. **Which branch I should pull request**? 

The last note: do you have better idea to fix the second error source. I am warried that my way will introduce too much change and lead to loss in generality.

peastman:
An exclusion simply means an interaction that shouldn't be computed.  It's totally up to you how to use them.  The current code adds them for 1-2, 1-3, and 1-4 pairs, but that's not required.  If you want an interaction to be computed, just don't add an exclusion for it.

NonbondedForce refers to exceptions, which are closely related to exclusions.  An exception means an interaction that should use different parameters than you would get with the standard combining rules.  If you set both epsilon and charge to 0, that means the interaction is completely excluded.  Many force fields exclude 1-2 and 1-3 interactions and reduce the strength of 1-4 interactions.  That again is specific to the force field.

CustomNonbondedForce doesn't refer to exceptions because they don't make sense.  It doesn't have built in combining rules, so it doesn't know how to use different parameters from the combining rules.  Therefore the only thing you can do is exclude an interaction.  Of course, perhaps you do still want to reduce the strength of 1-4 interactions.  In that case, you mark them as excluded and then use a CustomBondForce to compute the reduced version of them.

Everything you're trying to do only requires changes to the Python code.  No changes to the C++ library are needed.

qiuzy:
Yes. I have tried to solve 1-4 interaction exclusion issue in NBTthole calculation by changing charmmpsffile.py. But as I mentioned, one exception was thrown that **Exception: All Forces must have identical exclusions**.  This problem has been discussed in issue #357. This problem only show on CPU platform. I test Reference and CUDA platform. Both of them works.

In summary, On **CUDA and Reference platform**, NBThole calculation bug can be fixed by changing charmmpsffile.py. While On **CPU platform**, this does't work. It has to change C++ library. So two different ways!!   What I should do?

peastman:
Have a 1:1 match between exceptions in the NonbondedForce and exclusions in the CustomNonbondedForce.  Use a CustomBondForce for any interactions you need to add back in with either the original or altered parameters.

qiuzy:
I have commit my fix on NBThole calculation and improper dihedal parameter match problem in a new commit. #2952.

peastman:
Closing this since it's superseded by another PR.

Issue #2943, opened 2020-12-08
Minor CUDA Changes


dmclark17:
I have some minor changes to submit, which address some changes in an upcoming version of CUDA. 

1. Changing NVRTC Program name to `NULL`

https://github.com/openmm/openmm/blob/b49b82efb5a253a7c891ca084b3370e181de2ea3/plugins/cudacompiler/src/CudaCompilerKernels.cpp#L67

Right now the name of the NVRTC program is `""`, however, `NULL` can be passed as the argument to give the program a default name.

2. Adding launch bounds to `findBlocksWithInteractions`

I'd like to add a `__launch_bounds__(256,1)` to this line:

https://github.com/openmm/openmm/blob/b49b82efb5a253a7c891ca084b3370e181de2ea3/platforms/cuda/src/kernels/findInteractingBlocks.cu#L179

In the upcoming version of CUDA, I found this slightly improved performance for some benchmarks—I understand if you'd like to wait for its release to test out for yourself.

If these sound reasonable, I can submit a pull request with the changes. Thanks!

peastman:
A PR would be great!  findInteractingBlocks defines a constant for the block size, so you can specify the bounds as `__launch_bounds__(GROUP_SIZE,1)`.

peastman:
Fixed in #2947.

PR #2947, opened 2020-12-10
Minor CUDA Changes

dmclark17:
Changes suggested in #2943 with @peastman's suggestion to use `GROUP_SIZE`

Thanks for your time and let me know if you have any other suggestions!


peastman:
Looks good!

PR #2950, opened 2020-12-11
Switch openmmforcefields channel to omnia

rsdefever:
I don't believe the `openmmforcefields` package is available on `conda-forge`. This switches the docs back to `omnia` for `openmmforcefields`.

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -810,7 +810,7 @@ You can install this via conda with:
 
 .. code-block:: bash
 
-    $ conda install -c conda-forge openmmforcefields
+    $ conda install -c omnia openmmforcefields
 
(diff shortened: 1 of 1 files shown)


jchodera:
Good catch, but we should probably just migrate `openmmforcefields` to conda-forge ASAP!
@jaimergp : Can you help?

jchodera:
@rsdefever : I just realized we already have a PR open to handle the migration!
https://github.com/conda-forge/staged-recipes/pull/13448

Thanks so much for catching this and contributing the PR, though! We should have the migration done shortly!

jaimergp:
The package is in that PR, yes! `openmmforcefields` depends on the openff stack though, so it involves more moving parts than `pdbfixer` and `openmm-setup`. I think we need to move carefully there specially if I am going to be on vacations... 😬

Issue #2895, opened 2020-10-21
CHARMM parameter loading logic does not match the code


yaoyic:
Recently I met with a situation where the input CHARMM parameter file contains multiple DIHEDRAL entries with the same set of atoms and multiplicities but different energy coefficient as well as different equilibrium dihedral angle. According to the warning message, the original dihedral entry should be overwritten by the new entry, but actually the code did not do anything and the original dihedral entry just remains unchanged.

Corresponding code is here:
In [L388 of app/charmmparameterset.py](https://github.com/openmm/openmm/blob/418855e640d30f72110d339c509cd3aa25faa6b8/wrappers/python/simtk/openmm/app/charmmparameterset.py#L388), the warning says "Replacing dihedral (old) with (new)". But then in [L390 of app/charmmparameterset.py](https://github.com/openmm/openmm/blob/418855e640d30f72110d339c509cd3aa25faa6b8/wrappers/python/simtk/openmm/app/charmmparameterset.py#L390), nothing was changed. In my opinion, L390 should be changed to `self.dihedral_types[key][i] = dihedral` in order to match what was warned above.

Although I don't know how in the original CHARMM program they handle this behavior, the current implementation seems not self-consistent.

qiuzy:
Jing Huang have told me this issue. I have run some tests and will fix this issue in tomorrow.

qiuzy:
Thanks for reporing this bug!!

In Charmm, if flex key word is added when reading parameters, then parameter repeat checking will be triggered. This part code can be found in parmio.F90 in Charmm source code.

I already fix this issue as you suggested.

qiuzy:
I have fixed this bug in #2952 which is already merged into openmm-master. Please Check and test it. Thx!

yaoyic:
Thanks for the bug fix! It looks good.

PR #2952, opened 2020-12-12
fix Charmm NBThole calculation bug and Improper dihedral parameter match issue

### LINKED ISSUE #2750: CHARMM Drude NBThole implementation

qiuzy:
this commit fixes #2750 and #2895 

In the commit, NBThole interaction calculation bug is fixed. 
And improper dihedral match problem is fixed.
More discussion on the two issues can be found in #2944 

And also multiple dihedral with the same periodicity check problem is also fixed.


peastman:
Thanks!  Can you add a test case that verifies this fixes the problem?  This simplest way is to create a minimal system (just a few atoms) that gives the wrong energy without this fix but the right energy with it.  The files can go in `wrappers/python/tests/systems` and the test method in TestCharmmFiles.py.

qiuzy:
I have added a test system of CYT-GUA-CYT short DNA segnment to test NBThole.

peastman:
Thanks!  Looks good.

mdpoleto:
Are those fixes included in the current conda package (v7.5)? If not, are there any plans on this?

peastman:
7.5 was released on Dec. 9, and this was fixed on Dec. 20.  It will be in the next major release, or if you think it's urgent we could try to include it in a patch release.

mdpoleto:
I believe this is a critical fix for Drude force field users since it directly impacts potential energy and nonbonded calibrations. Currently, v7.5 does not issue any warning on this, so, in my opinion, a patch release is much needed.

mdpoleto:
I see this was added to 7.5.1 milestones and they are all closed now. Thank you @peastman. Should we expect a release as a conda package soon?

peastman:
We're hoping to get a release candidate posted this week or next week.

Issue #2919, opened 2020-11-12
Changes in Travis CI pricing model


jaimergp:
See https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

OSX will be a paid feature after 1000 mins, and Linux will be deprioritized. I don't think this is an urgent change, but we should think of moving them to Azure?

peastman:
Thanks for bringing this up.  It's actually much worse, though they try hard to obscure it.  The blog post starts by talking a lot about how they're making all these changes just to help you.  Then they bring up the more obscure cases, like building on Macs or running more than 10 concurrent builds.  Then they finally get to the main point, which is written in an intentionally confusing way: they're getting rid of free support for open source projects.  If you have an open source project that is currently using travis for free, they will "upgrade" you to a trial plan that gives only a very small amount of build time.  Once you've used that up, you need to "consider which of our plans will meet your needs."

I don't think Azure supports either ARM or PPC, so we need to figure out what to do with them.

jaimergp:
Azure can provide ARM / PPC through emulation, as conda-forge does. It's way slower but hopefully it still runs within 6h. I think it requires some Docker trickery but it's definitely doable.

jaimergp:
I might give it a try in the next weeks. Would you be ok with having all the CI running on GH Actions? I know Windows is running on Azure now, but I am more comfortable with GHA (and has a faster dev cycle, less auth/login issues), and I can migrate Windows to GHA if you want!

peastman:
Github Actions would be fine too.  We used to run Windows CI on Appveyor, but it was really slow, so we eventually switched to Azure.  Whatever works and is easiest to maintain.

jchodera:
Now that we are moving toward conda-forge, do we want OpenMM's CI infrastructure to more closely resemble a conda package build/deploy? I'm just wondering if there's a way to reduce the entropy of the different bits we have to maintain.

jaimergp:
At least one of the things I'd do is to get rid of `omnia-dev` and have nightlies built and uploaded using `openmm/openmm-feedstock`. The CI pipelines are not really covered by the conda-forge infrastructure, but since we are going to use GHA, we can delegate the fragile parts to established _actions_ (e.g. installing Miniconda with `conda-incubator/setup-miniconda`).

swails:
I could also add a Raspberry Pi 4 to the Jenkins setup 😆

peastman:
[...]

PR #2964, opened 2020-12-28
Migrate CI to GH Actions

jaimergp:
Phew, this has been fun 😄! Hopefully it provides a solid start to end up with a CI that is useful, comprehensive and extensible to future needs. Sorry about the insane amount of commits, but it took a bit to get all the pieces in place!

This will close https://github.com/openmm/openmm/issues/2919

## Summary

This PR provides CI for the following OS / architectures:

* Linux x64
* MacOS
* Windows
* Linux PPC (emulated with QEMU/Docker)
* Linux ARM (emulated with QEMU/Docker)

The matrix gets expanded with:

* CUDA versions from 10.0 and above are provided. 
* Python 3.6 to 3.9, when applicable (only CPU builds for now, to speed things up).
* Docs only job that does not run tests.
* Compilers, when applicable. I made sure to use system compilers whenever possible, but there are entries that cover conda-forge compilers too, since we want to guarantee those pass the full suite.

Other features:

* CCache _should be_ enabled in all runs
* Steps have been isolated and/or grouped as much as possible to easily identify the failing parts.

Cron jobs run nightly, by default on all matrix entries. This can be changed if needed.

## Comments

* Secrets:
  * `CACHE_VERSION`: needed in case cache needs to be cleared. Do not set to something like `0` because GHA will mask it believing it's sensitive material, making the output very difficult to read. Better use something like `plkghj`; short, meaningless and hard to find in a useful message. Cache expires after a week anyway, so no need to keep a correlative scheme here.
  * S3 login details for docs deployment must be defined as Github secrets. I haven't been able to test this part due to lack of access. 
* Docs:
  * This PR also fixes PDF builds for the documentation (check changes in `developer.rst`).
  * There are some 404s and dead links that need to be updated or removed.
* PowerPC:
  * Several critical failures with GCC 9 (conda-forge). All OK with GCC7.
  * Some pytest warnings with division by zero.

## Before merge

- [x] Update README badges
- [ ] Tests targeting `conda-forge` should use `conda-forge-pinning` version restraints
- [ ] Provide needed secrets (see above)
- [x] Check cache is working (can also be done / fixed in separate PR if needed)
- [x] Check results are consistent with Azure / Travis
- [ ] Remove Azure and Travis from repo
- [ ] Some tests fail with recent Gromacs, which is why it is pinned to `2018.*`. This should be addressed in a different PR so we can unpin it.


peastman:
That's awesome!  I'll need to spend some serious time studying this.

I wonder if this is going a bit too far in terms of the number of builds?  You have 42 of them.  For example, we probably don't need to test every Python version separately on every platform.  If we just test the oldest and newest versions we support, it's pretty likely it also works on all the intermediate ones.  Same for CUDA.  Testing multiple Python versions on ARM and PPC is also probably unnecessary, since there's nothing in the Python code that's ARM or PPC specific.

jchodera:
This is awesome, @jaimergp! Thanks so much for putting this together!

A few quick comments from looking this over:
* Would it be possible to add (1) a few comments in the CI scripts to help us understand each of the sections, and (2) a few comments in a `CI-README.md` to help us understand what we would do to extend this to new architectures? The better we understand it, the less we would need to rely on you to fix small things or make extensions, which will free up more of your time in future! Your PR notes here are a good start at the README already.
* We should clarify that "Mac OSX" is "Mac OSX Intel", and we'll eventually need to sort out how to do CI on Mac OSX ARM (ugh). Looks like there is some attempt to figure this out [here](https://github.com/actions/runner/issues/805).

jchodera:
> I wonder if this is going a bit too far in terms of the number of builds? You have 42 of them. For example, we probably don't need to test every Python version separately on every platform. If we just test the oldest and newest versions we support, it's pretty likely it also works on all the intermediate ones. Same for CUDA. Testing multiple Python versions on ARM and PPC is also probably unnecessary, since there's nothing in the Python code that's ARM or PPC specific.

Is this actually a bottleneck right now? Seems like the actual slow part is the PowerPC tests, which take >200 min.

Notably, win *did* specifically [fail on Python 3.9](https://github.com/openmm/openmm/pull/2964/checks?check_run_id=1619270910#step:11:289), but not 3.8. It appears to a stochastic Langevin test failure, but our re-running didn't behave as expected:
```
0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.08 sec

The following tests FAILED:
	 26 - TestReferenceLangevinIntegrator (Failed)
Errors while running CTest
Rerunning failing tests...

D:\a\openmm\openmm\build>if errorlevel 1 exit 1 
Error: Process completed with exit code 1.
```

This also turned up some interesting failures on [OpenCL](https://github.com/openmm/openmm/pull/2964/checks?check_run_id=1619271067#step:14:360):
```
98% tests passed, 3 tests failed out of 157

Total Test time (real) = 2677.27 sec

The following tests FAILED:
	 97 - TestOpenCLNonbondedForceSingle (Failed)
	141 - TestOpenCLDrudeLangevinIntegratorDouble (Failed)
	145 - TestOpenCLDrudeSCFIntegratorSingle (SEGFAULT)
Errors while running CTest
    Start  97: TestOpenCLNonbondedForceSingle
    Start 141: TestOpenCLDrudeLangevinIntegratorDouble
1/3 Test #141: TestOpenCLDrudeLangevinIntegratorDouble ...   Passed   28.18 sec
    Start 145: TestOpenCLDrudeSCFIntegratorSingle
2/3 Test #145: TestOpenCLDrudeSCFIntegratorSingle ........Subprocess aborted***Exception:  36.30 sec
double free or corruption (!prev)

3/3 Test  #97: TestOpenCLNonbondedForceSingle ............***Failed  140.20 sec
exception: Assertion failure at TestNonbondedForce.h:593.  Expected -93295.6, found -93292.9


33% tests passed, 2 tests failed out of 3

Total Test time (real) = 140.21 sec

The following tests FAILED:
	 97 - TestOpenCLNonbondedForceSingle (Failed)
	145 - TestOpenCLDrudeSCFIntegratorSingle (Subprocess aborted)
Errors while running CTest
Rerunning failing tests...
Error: Process completed with exit code 8.
```

jaimergp:
> I wonder if this is going a bit too far in terms of the number of builds? You have 42 of them. 


Agreed, and notice that I only expanded Python versions for CPU builds 😄 Imagine populating the CUDA branch with those too... Anyway, adding or removing Python versions is a matter of commenting out the corresponding matrix entries. I do agree that testing _all_ of the Python versions might be unnecessary, but it's also _free_ as they run in parallel anyway. Maybe it's enough if all of them are tested in one OS (Linux 64), while the rest only get the oldest + newest combo. Also, you can check which jobs are "required" in the GitHub settings if you want to visually prioritize some of them.

> Would it be possible to add (1) a few comments in the CI scripts to help us understand each of the sections, and (2) a few comments in a CI-README.md.

Noted! Where do you think this CI-README should be placed? Under `devtools`, or repo root? Or maybe in the repo wiki?

> Seems like the actual slow part is the PowerPC tests, which take >200 min.

PPC / ARM will be very slow due the QEMU emulation. There's no way around that, but once the CCache warms up it should speed up considerably. This also applies to Linux, MacOS and Windows.

> It appears to a stochastic Langevin test failure, but our re-running didn't behave as expected:

I have seen stochastic errors in several places during the implementation phase. They are, well, stochastic. I didn't see any consistent effect on any specific matrix entry (Windows is more prone for some reason, though). I tried modifying the `ctest` wrapper to do more than one attempt (you can see how the stochastic failures do get solved after one retry in other jobs), but I guess it's not as easy as looping over that function. I'll need to take a look at that script more closely.

jaimergp:
> This also turned up some interesting failures on OpenCL:

Oh, take into account this is only AMD's OpenCL. The CUDA entries also build for Nvidia's OpenCL, but they can't be tested here. Jenkins CI does I think, and that one does work.

jaimergp:
> Noted! Where do you think this CI-README should be placed? Under devtools, or repo root? Or maybe in the repo wiki?

Or, thinking about it, should all this "internal" documentation live within the published docs too? I am worried we forget about them as we make changes to the infrastructure (e.g. the release guidelines).

peastman:
> Notably, win did specifically fail on Python 3.9, but not 3.8.

It was a failure in a C++ test, so nothing to do with the Python version.  When testing multiple Python versions, we only need to run the C++ tests on one of those builds.  No need to repeat them on each build.

> I have seen stochastic errors in several places during the implementation phase. They are, well, stochastic.

Right, those are expected.  That's why run-ctest.py keeps track of failures and reruns them to see if they'll pass the second time.  Of course, the more builds we do, the higher the chance that some test somewhere will randomly happen to fail twice in a row.

jchodera:
> I have seen stochastic errors in several places during the implementation phase. They are, well, stochastic. I didn't see any consistent effect on any specific matrix entry (Windows is more prone for some reason, though). I tried modifying the ctest wrapper to do more than one attempt (you can see how the stochastic failures do get solved after one retry in other jobs), but I guess it's not as easy as looping over that function. I'll need to take a look at that script more closely.

I believe the test logic is to re-run tests that failed, and only exit with an error if the re-run also fails. Two failures in a row likely indicates a problem. It looked like this logic may not be working correctly?

jchodera:
> Noted! Where do you think this CI-README should be placed? Under devtools, or repo root? Or maybe in the repo wiki?

Under `devtools/` sounds optimal! Normal users wouldn't need it.

jchodera:
> PPC / ARM will be very slow due the QEMU emulation. There's no way around that, but once the CCache warms up it should speed up considerably. This also applies to Linux, MacOS and Windows.

I'm not 100% sure I fully understand if this will help, but it looks like hardware that supports virtualization could speed up execution if the appropriate libraries are installed---see [this issue](https://github.com/actions/virtual-environments/issues/114).

jchodera:
> Oh, take into account this is only AMD's OpenCL. The CUDA entries also build for Nvidia's OpenCL, but they can't be tested here. Jenkins CI does I think, and that one does work.

Shouldn't our OpenCL platform still work with AMD's OpenCL? If we think it is fundamentally broken and OpenMM should fail on it, should we test on it? Or should we take the failures seriously? This test failed twice and does not appear to be stochastic:
```
3/3 Test  #97: TestOpenCLNonbondedForceSingle ............***Failed  140.20 sec
exception: Assertion failure at TestNonbondedForce.h:593.  Expected -93295.6, found -93292.9
```

peastman:
Is it running on an actual AMD GPU?  Or is it using AMD's OpenCL on a CPU?

jaimergp:
> Is it running on an actual AMD GPU?  Or is it using AMD's OpenCL on a CPU?

Most certainly on the CI CPU.

jaimergp:
> > PPC / ARM will be very slow due the QEMU emulation. There's no way around that, but once the CCache warms up it should speed up considerably. This also applies to Linux, MacOS and Windows.
> 
> I'm not 100% sure I fully understand if this will help, but it looks like hardware that supports virtualization could speed up execution if the appropriate libraries are installed---see [this issue](https://github.com/actions/virtual-environments/issues/114).

This link has sent me through a rabbit hole and might be able to simplify some bits of the CI using a couple of tricks! Thanks for sharing! In the meantime maybe don't pay too much attention at _how_ things are implemented but _what_ I am trying to implement. Let's see if we can agree on a matrix of things to try, for example!

jaimergp:
> Under `devtools/` sounds optimal! Normal users wouldn't need it.

Added a draft! Let me know if that's useful!

jaimergp:
Also, [this entry](https://github.com/github/roadmap/issues/95) in the GH roadmap might grant some wishes like GPU-enabled tests! I don't know if that's going to be free but we'll see!

jchodera:
> Added a draft! Let me know if that's useful!

This is great! Thanks!

> Also, this entry in the GH roadmap might grant some wishes like GPU-enabled tests! I don't know if that's going to be free but we'll see!

image

jaimergp:
Is there anything else you want me to add / remove / change here?

peastman:
Can we narrow it down to a smaller number of builds?  With this many of them, it's almost guaranteed we'll get a failure somewhere through stochastic tests.  Here's what I think we need:

- At least one build for each combination of OS and processor architecture.
- At least one build each on the oldest and newest Python versions we support.
- At least one build each on the oldest and newest CUDA versions we support.
- At least one build running OpenCL tests.
- At least one build that builds the documentation.

For comparison, Travis does roughly all of that in only eight builds.  (It only has one CUDA build, but on the other hand it also tests building static libraries, which probably isn't too important now that Folding@home has moved to dynamic libraries.)

jaimergp:
What about this config?

peastman:
That's looking a lot better.  What's the difference between `Linux CPU Python 3.6` and `Linux CPU Python 3.6 with condaforge compilers`?  What compilers does the first one use?

I notice the AMD OpenCL build failed after 49 minutes, and the last test failed with the message "Subprocess aborted".  Might that indicate it reached a timeout on the build?  Do we need to make it run faster (like by running fewer tests)?

It reports a failure in the docs build, but it isn't clear to me what the error was.  I don't see anything in the log that looks like an error message.

jaimergp:
> What's the difference between Linux CPU Python 3.6 and Linux CPU Python 3.6 with condaforge compilers? What compilers does the first one use?

First one uses whatever is default on Ubuntu. Second uses the ones pinned by condaforge. This hasn't been relevant for Linux x64 so far, but it bit us for ppc64le so I'd rather make sure the packages we build on conda forge can pass the test suite. That said, we can also drop the system compilers build if you feel we don't need that. My point was trying to support those users that might compiler from source on Ubuntu, since that's the most popular distro.

peastman:
The AMD OpenCL build is now giving a segfault on one test, and an assertion failure on a different one.  Let me see if I can reproduce them locally.

> First one uses whatever is default on Ubuntu. Second uses the ones pinned by condaforge.

That makes sense.

> Look for links marked as [404]. Those are dead links in our documentation that should get updated and/or removed.

Cool, I didn't know it did that.  Sounds like I should fix the links.

peastman:
I can reproduce the assertion failure, but not the segfault, when running locally.  I'm looking into it now.

peastman:
Naturally, that test failed the first time I ran it, and I haven't been able to get it to fail again since.

jaimergp:
I am going to re-run this to see if it can be reproduced in the CI at least (but there's no GPU, so don't know how reliable these OpenCL tests are).

We could also use [`act`](https://github.com/nektos/act) to run the pipeline locally but with access to a GPU, but I haven't experimented with that yet.

peastman:
Still getting the same errors on CPU based OpenCL. :(

The Windows build got a stochastic failure in one test.  It then printed, "Rerunning failing tests...", but I don't see any indication that it actually reran anything.

The PPC build also had a few failures, which more concerningly included a segfault.  Again I don't see any indication that it reran them.

And we still have the errors in docs.  I'll see if I can address those today.

jaimergp:
> The PPC build also had a few failures

Note it's using [GCC9](https://github.com/openmm/openmm/pull/2964/checks?check_run_id=1761855316#step:6:609) (on purpose). Using GCC>7 had caused some issues on PCC before (see #2781). Maybe it's related?

jaimergp:
> The Windows build got a stochastic failure in one test. It then printed, "Rerunning failing tests...", but I don't see any indication that it actually reran anything.

That message is being printed out of order (no clue why). 

Note how it [had more failures](https://github.com/openmm/openmm/pull/2964/checks?check_run_id=1761855070#step:11:317), and then some of them disappeared after the [actual re-run](https://github.com/openmm/openmm/pull/2964/checks?check_run_id=1761855070#step:11:332).

peastman:
Ok, I see.  It was just a bit confusing.

Since the PPC tests are being run with an emulator, we ought to be able to reproduce them locally.  Do you have instructions on how to do that?  I'm a little suspicious of them, since those errors don't happen on travis which uses an actual PPC machine.  It might be a flaw in the emulator.  But let's investigate and try to figure it out.

jaimergp:
AFAIK Travis is using [GCC7](https://travis-ci.org/github/openmm/openmm/jobs/756083133#L3346), which passed the tests with QEmu just fine.

I'll add instructions on how to debug PPC locally to the CI README.

jaimergp:
Check https://github.com/jaimergp/openmm/blob/gh-actions/devtools/CI-README.md#how-to-debug-powerpc--arm-locally for instructions.

peastman:
When I ran `start_docker_locally.sh`, it failed on the line

```
ls /proc/sys/fs/binfmt_misc/
```

But I don't think that line is actually used for anything?  I commented it out, and the script ran without problem.

I then tried to run `run_steps_inside_docker_image.sh`, but it failed with this output:

```
[conda@2c847235556f ~]$ bash /home/conda/workspace/devtools/ci/gh-actions/scripts/run_steps_inside_docker_image.sh
::endgroup::
::group::Prepare build environment...
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - conda-forge
```

jaimergp:
Is this running on an `osx` host? I will only work on Linux hosts, I am afraid.

jaimergp:
Ah no, that's not the problem, sorry. I defined `COMPILERS` incorrectly. It needs to be `compilers` (the conda metapackage for their compilers), not `conda-forge`. Script is updated now.

peastman:
Thanks!  That fixed it.  It seems to be working fine on a MacOS host.  I just had to remove that one line.

jaimergp:
> It seems to be working fine on a MacOS host

Hmmm, that's interesting! It might give us a slight performance boost if I manage to get Docker running on a MacOS-based CI host, since they seem to use a higher tier VM.

peastman:
It finally finished!  Building under emulation is *really* slow.  I got four test failures.  I'll see if I can figure them out.

How can I execute individual tests?  If I just try to run them directly I get an error:

```
[conda@4c05a335dce0 build]$ ./TestCpuNonbondedForce 
bash: ./TestCpuNonbondedForce: cannot execute binary file
```

peastman:
It looks like there are two copies of the compiled tests: one in ~/workspace/build which can't be executed, and one in ~/build which can.  I assume this is something to do with the emulation?

jaimergp:
I am assuming you are running the tests within the Docker image, right?

In that case, that might be emulation issues, yes. There's a chance the QEmu configuration is not working on MacOS (since you had to comment the `binfmt_misc` stuff), and that CMake is magically cross-compiling or something like that? 

I can try and do the same steps on my Linux host and see if it can be reproduced.

peastman:
I think I was just getting confused between the two different build directories.  Next question is how I can recompile after making a change to add debugging code?  Just typing `make` from the build directory doesn't work.

jaimergp:
Assuming you haven't shut down the Docker image:

1. Create a new `bash` session
2. `source /opt/conda/etc/profile.d/conda.sh`
3. `conda activate build`
4. `make`

peastman:
That works, thanks!

peastman:
The code

```c++
        for (int i = 0; i < 3; i++)
            for (int j = 0; j < 3; j++)
                periodicBoxVectors[i][j] = (float) boxVectors[i][j];
        for (int i = 0; i < 3; i++)
            for (int j = 0; j < 3; j++)
                printf("%g %g\n", boxVectors[i][j], periodicBoxVectors[i][j]);
```

prints out

```
2.5 0
0 0
0 0
0.5 0
3 0
0 0
0.7 2.72008e+23
0.9 -1.07374e+08
3.5 3.5
```

Is there any way this could possibly be something other than a compiler bug?  I'm not seeing it.  `periodicBoxVectors` is declared as

```
float periodicBoxVectors[3][3];
```

peastman:
Definitely a compiler bug.  I can work around it by rewriting the loop as

```c++
        for (int i = 0; i < 3; i++)
            for (int j = 0; j < 3; j++) {
                volatile float temp = (float) boxVectors[i][j];
                periodicBoxVectors[i][j] = temp;
            }
```

Using the volatile temporary variable prevents the compiler from doing most optimizations.  That fixes three of the four test failures.  Only TestCpuNonbondedForce is still failing.  I'll look into it next.

peastman:
I think that test may just be running out of memory.  It gets killed in `testHugeSystem()`, which creates a system with over 3 million particles.  Checking the memory use on my Mac with

```
/usr/bin/time -l ./TestCpuNonbondedForce 
```

reports

```
2851094528  maximum resident set size
```

So it uses 2.8 GB.  But running `top` inside the container reports

```
KiB Mem :  2037988 total,  1671816 free,   188868 used,   177304 buff/cache
```

It only has 2 GB of memory.  If I reduce the number of particles to only 1 million then it passes.

jaimergp:
Hm, does that one time out on the GHA CI too? We allegedly have [7GB RAM](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) to spare.

peastman:
Good point, it looks like that one passed on GHA.  It only had the other three failures, which are fixed by my change.  I'll make a PR for it.

peastman:
The PR is merged.  Hopefully this means the PPC and docs errors are now fixed.  Just the OpenCL ones left to figure out.

jaimergp:
Let's see!

peastman:
Any idea why the docs build is still failing?  It's reporting the following as a broken link:

```
    [0] https://support.amd.com/
```

This is strange for two reasons.  First, because https://support.amd.com/ works fine when I paste it into a browser.  And second, because 0 is not a standard HTTP response code!

jaimergp:
Some websites reject requests like these, or the CI interferes somehow. I normally add those to the exclusion list after checking it works on a browser. I don't  think AMD is going anywhere so I'll add it!

jaimergp:
This one is actually broken: http://nihroadmap.nih.gov/bioinformatics

swails:
> Any idea why the docs build is still failing? It's reporting the following as a broken link:

Per @jaimergp, I've observed websites that have blacklisted IP ranges for CI services like Travis.  The PDB is actually one of them, I think.

peastman:
I'm getting back to the OpenCL tests.  I think the error in TestOpenCLNonbondedForce is just an overly strict tolerance.  The accuracy of energy accumulation depends on the number of threads.  For CPU OpenCL, it only uses a very small number of threads.  When that's combined with single precision, the accuracy comes out less than what the test is expecting.

So that just leaves TestOpenCLDrudeSCFIntegrator to figure out.

peastman:
Figured it out.  The error didn't happen on computers with more than two cores, which is what made it hard to reproduce.  The fixes for both are in #3003.

devtools/ci/gh-actions/scripts/run_steps_inside_docker_image.sh
```diff
@@ -0,0 +1,95 @@
...
+test -f ${CONDA_PREFIX}/lib/plugins/libOpenMMPME.so
+if [[ ! -z ${CUDA_VER} ]]; then
+    test -f ${CONDA_PREFIX}/lib/plugins/libOpenMMCUDA.so
+    test -f ${CONDA_PREFIX}/lib/plugins/libOpenMMOpenCL.so
+fi
+echo "::endgroup::"
+
+# Python tests
+echo "::group::Run Python tests..."
+python -m simtk.testInstallation
```
peastman, line 86: This should be changed to `openmm.testInstallation` now.

peastman:
The fix is merged.

jaimergp:
> This one is actually broken: http://nihroadmap.nih.gov/bioinformatics

I think this is now https://commonfund.nih.gov/bioinformatics, as hinted by [archive.org](https://web.archive.org/web/20160402035153/http://nihroadmap.nih.gov/bioinformatics).

jaimergp:
Ohhh it's all green! 😍

peastman:
Yay!  Shall I merge this?

jaimergp:
Well, depends on whether you want to deal with this now or in the future:

* Docs deployment ([this block](https://github.com/jaimergp/openmm/blob/gh-actions/.github/workflows/CI.yml#L584-L595)). We haven't checked if that works because I set it up so it only runs when `master` has received a push. I don't know if you deploy to S3 as PRs are merged, or only on releases.
* Gromacs is pinned to [v2018](https://github.com/openmm/openmm/blob/53c1851d215cf541a2d221e0723ee311ad01fb4b/devtools/ci/gh-actions/conda-envs/build-ubuntu-latest.yml#L22) because 2019 and 2020 fail. I guess we can deal with that in a different PR.
* I didn't implement any automatic way to track conda-forge pinnings in our environment file, but this is not as critical now that I figured out how to build packages that run the full CI, so we can always ensure everything is ok on the conda-forge side during the RC phase.

jaimergp:
Oh, and I guess we should also delete Azure and Travis from here, or do you want to run them side by side for a while?

peastman:
> I don't know if you deploy to S3 as PRs are merged, or only on releases.

They should get pushed to the `development` folder on every merged PR, and then we copy them to `latest` when we do a release.

> Gromacs is pinned to v2018 because 2019 and 2020 fail. I guess we can deal with that in a different PR.

That seems fine.

> I didn't implement any automatic way to track conda-forge pinnings in our environment file, but this is not as critical now that I figured out how to build packages that run the full CI, so we can always ensure everything is ok on the conda-forge side during the RC phase.

Up to you what you think is best.

peastman:
I don't think we need to run Travis and Azure in parallel for long, but let's wait until this is merged and working before removing them.

jaimergp:
> They should get pushed to the `development` folder on every merged PR, and then we copy them to `latest` when we do a release.

Ok, that's the behavior I implemented, perfect! I'll need you to add [these GitHub secrets](https://github.com/jaimergp/openmm/blob/gh-actions/.github/workflows/CI.yml#L590-L592) (S3 authentication) for this to work. They are added [here](https://github.com/openmm/openmm/settings/secrets/actions) as plain text, then GitHub will take care of the rest.

> Up to you what you think is best.

I'll give it a thought, but for sure this is not blocking now. I'll open an issue so I can have it assigned and I'll revisit it in the future. Same for Gromacs.

peastman:
Where exactly do I get those values from?  I only have limited experience with AWS.  I *think* `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are supposed to be from a security credential created for the openmm.org user?  I can see the access key ID, but it says there's no way to retrieve the secret key if you don't already have it.  @jchodera do you have it?  If not we can delete the existing credential and create a new one, but I don't know what I'll break by doing that.

Is `AWS_S3_BUCKET` just the name of the bucket containing the docs?  I don't think that's secret, is it?

jchodera:
@peastman : I'll work with @jaimergp to handle this for you tomorrow.

jchodera:
@peastman @jaimergp : I've just populated `AWS_S3_BUCKET`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY` as repository secrets. You should be good to go!

peastman:
Thanks!  I'll go ahead and merge this, and we can see what happens on the next PR.

jaimergp:
Yay! 🎉

jaimergp:
Looks like S3 deployed without errors, but there are two link checks that are rejected:

```
    [0] https://docs.conda.io/
    [0] https://docs.continuum.io/anaconda/install
```

They should be added to the ignore list. The Continuum one might need to be replaced with byhttps://docs.anaconda.org/anaconda/install/.

peastman:
It looks like this worked correctly on #3018, including pushing the docs to AWS when it was merged.  Shall I go ahead and turn off the Travis and Azure builds?

jaimergp:
I think so! We can always roll back from history if needed.

jaimergp:
I guess that the `[0]` link errors were only temporary due to some network shenanigans.

peastman:
Done!  For the moment I haven't changed any files.  I just disabled the Travis webhook and the Azure app.  If we need to reenable them, it's just a few clicks.

Issue #2966, opened 2020-12-29
Error in simtk.unit conversion


dwhswenson:
There seems to be an error when calculating conversion factors in `simtk.unit`:

```python
from simtk import unit
J_per_mol = unit.joules / unit.mole
J_per_mol.conversion_factor_to(unit.kilojoules_per_mole)
# 0.0009999999999999998
```

Small difference, but this does cause real problems for reproducibility, since the value you get depends (slightly) on which unit system you used. See test failures in OpenMMTools: https://github.com/choderalab/openmmtools/issues/490. The workaround I proposed there is hideous, so an upstream fix would be better.

(Tested in macOS, Python 3.7, OpenMM 7.5.0)

dwhswenson:
A little more detail: this appears to be a result of rounding errors in floating point math when dealing with multiplying large floats, as used to adjust between SI prefixes. The issues are already present in `BaseUnit._conversion_factor_to_by_name` after import.

```python
from simtk import unit
base_unit = list(list(unit.nanometer._top_base_units.values())[0].keys())[0]
base_unit
# BaseUnit(base_dim=BaseDimension("length"), name="nanometer", symbol="nm")

base_unit._conversion_factor_to_by_name
# {'nanometer': 1.0,
# 'meter': 1e-09,
# 'yottometer': 1000000000000000.1,
# 'zeptometer': 1000000000000.0002,
# 'attometer': 999999999.9999999,
# 'femtometer': 999999.9999999999,
# ... 
#  'millimeter': 1.0000000000000002e-06, ... }
```

The conversion for, e.g., nm to mm is being calculated something like this:

```python
# mm / nm = 1e-9 m / nm * 1e3 mm / m
1e-9 * 1e3
# 1.0000000000000002e-06
```

Issues with associativity/commutativity also show up here: 

```python
(1.0  # J/mol = kg * m**2 / s**2 / mol; convert to g * nm**2 / ps**2 / mol
 * 1e3  # g / kg
 * (1e9)**2  # nm**2 / m**2
 * (1e-12)**2  # s**2 / ps**2
)
# 0.001

# but:
1.0  * 1e3 * (1e-12)**2 * (1e9)**2
# 0.0009999999999999998

# while:
1.0  * 1e3 * ((1e-12)**2 * (1e9)**2)
# 0.001
```

[...]

PR #2968, opened 2020-12-29
Reduce numerical error in computing conversion factors between units

peastman:
This is an attempt at fixing #2966.


peastman:
Since it's been a couple of weeks and I still haven't gotten any feedback from @dwhswenson, I'm going to go ahead and merge this.  It should at least help, even if it doesn't completely solve the problem.

PR #2970, opened 2021-01-02
Add benchmarks from Amber20 benchmark suite to standard benchmark script

aizvorski:
This pull request makes the standard OpenMM benchmark script be able to run tests from the Amber suite (as well as all the original tests).  The suite is downloaded and unpacked automatically (into the current directory) if it is not already there. The Amber20 benchmark suite is available here: https://ambermd.org/GPUPerformance.php

Rationale: this has some significantly larger systems than the benchmarks already included in OpenMM, up to the 1 million atom STMV system.  It also allows direct comparisons with published Amber benchmarks.  The intention is to make this as close to apples-to-apples as possible (but there are some remaining small issues, see below).

Questions / caveats:

- [x] Is it okay to download the .tar.gz directly from ambermd.org?  Including this here probably won't immediately cause a huge increase in downloads, but it may be good to give them a heads up, and/or mirror it somewhere.  (Note: these tests are not included if running benchmark.py with no arguments - the user needs to select one of the tests specifically before it triggers a download)
- [x] Is it okay to assume wget and tar are available?  (There are pure python replacements if preferred) => Replaced with pure python urllib and tarfile
- [x] The original runs NPT and NVE simulations; the simulation here is NVT (which would probably have a slight speed advantage over NPT).  Can we add a MonteCarloBarostat for NPT to make it an exact comparison? (none of the other tests use that)  How would we do a 4fs time step NVE? (maybe VelocityVerlet? sorry, I'm not familiar with it) => Added --ensemble option, NPT is default for amber tests
- [x] Hydrogen mass: these benchmarks are all intended to use HMR, but I couldn't figure out what the hydrogen mass is supposed to be.  Is this already included in the .prmtop file? (and if yes, is it read in correctly?)  The tleap default is 3amu, so I just set it explicitly to that. => Fixed; don't change hydrogen mass from inputs
- [x] Adding `context.setVelocitiesToTemperature()` causes "particle position is nan" pretty quickly if running one of these systems.  Not sure why; they may have velocities included in the inpcrd file.  That's why this is skipped for these.
- [x] What should the names of the tests be?  => Lowercase, amber20- prefix

Impact: should have no effect on any of the existing benchmarks; just adds new ones.

Testing:
- Tested on T4, V100, A100 GPUs on AWS (will post results as an issue shortly)

Todo:
- Add the implicit solvent tests from the same suite

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -19,15 +19,18 @@ def timeIntegration(context, steps, initialSteps):
     return elapsed.seconds + elapsed.microseconds*1e-6
 
 def downloadAmberSuite():
-    dir = 'Amber20_Benchmark_Suite'
-    if not os.path.exists(dir):
-        os.mkdir(dir)
-        os.chdir(dir)
-        import subprocess
-        subprocess.run('wget https://ambermd.org/Amber20_Benchmark_Suite.tar.gz', shell=True, check=True)
-        subprocess.run('tar xzvf Amber20_Benchmark_Suite.tar.gz', shell=True, check=True)
-        os.chdir('..')
-    return dir
+    """Download and extract Amber benchmark to Amber20_Benchmark_Suite/ in current directory."""
+    dirname = 'Amber20_Benchmark_Suite'
+    url = 'https://ambermd.org/Amber20_Benchmark_Suite.tar.gz'
+    if not os.path.exists(dirname):
+        import urllib.request
+        print('Downloading', url)
+        filename, headers = urllib.request.urlretrieve(url, filename='Amber20_Benchmark_Suite.tar.gz')
+        import tarfile
+        print('Extracting', filename)
+        tarfh = tarfile.open(filename, 'r:gz')
+        tarfh.extractall(path=dirname)
+    return dirname
 
 def runOneTest(testName, options):
     """Perform a single benchmarking simulation."""
@@ -68,9 +71,9 @@ def runOneTest(testName, options):
         dt = 0.002*unit.picoseconds
         integ = mm.MTSIntegrator(dt, [(0,2), (1,1)])
     elif amber:
-        dir = downloadAmberSuite()
-        prmtop = app.AmberPrmtopFile(os.path.join(dir, f'PME/Topologies/{testName}.prmtop'))
-        inpcrd = app.AmberInpcrdFile(os.path.join(dir, f'PME/Coordinates/{testName}.inpcrd'))
+        dirname = downloadAmberSuite()
+        prmtop = app.AmberPrmtopFile(os.path.join(dirname, f'PME/Topologies/{testName}.prmtop'))
+        inpcrd = app.AmberInpcrdFile(os.path.join(dirname, f'PME/Coordinates/{testName}.inpcrd'))
         topology = prmtop.topology
         positions = inpcrd.positions
         dt = 0.004*unit.picoseconds
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -37,7 +37,7 @@ def runOneTest(testName, options):
     explicit = (testName in ('rf', 'pme', 'amoebapme'))
     amoeba = (testName in ('amoebagk', 'amoebapme'))
     apoa1 = testName.startswith('apoa1')
-    amber = (testName in ('JAC', 'Cellulose', 'FactorIX', 'STMV'))
+    amber = (testName.startswith('amber'))
     hydrogenMass = None
     print()
     if amoeba:
@@ -72,8 +72,10 @@ def runOneTest(testName, options):
         integ = mm.MTSIntegrator(dt, [(0,2), (1,1)])
     elif amber:
         dirname = downloadAmberSuite()
-        prmtop = app.AmberPrmtopFile(os.path.join(dirname, f'PME/Topologies/{testName}.prmtop'))
-        inpcrd = app.AmberInpcrdFile(os.path.join(dirname, f'PME/Coordinates/{testName}.inpcrd'))
+        names = {'amber20-dhfr':'JAC',  'amber20-factorix':'FactorIX', 'amber20-cellulose':'Cellulose', 'amber20-stmv':'STMV'}
+        fileName = names[testName]
+        prmtop = app.AmberPrmtopFile(os.path.join(dirname, f'PME/Topologies/{fileName}.prmtop'))
+        inpcrd = app.AmberInpcrdFile(os.path.join(dirname, f'PME/Coordinates/{fileName}.inpcrd'))
         topology = prmtop.topology
         positions = inpcrd.positions
         dt = 0.004*unit.picoseconds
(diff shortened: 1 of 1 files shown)


jchodera:
Thanks! This is super useful!

> Is it okay to download the .tar.gz directly from ambermd.org? Including this here probably won't immediately cause a huge increase in downloads, but it may be good to give them a heads up, and/or mirror it somewhere. (Note: these tests are not included if running benchmark.py with no arguments - the user needs to select one of the tests specifically before it triggers a download)

This sounds like a reasonable way to (1) not increase the size of the repo or conda packages, and (2) ensure the AMBER benchmarks are always current (by just deleting the directory if you want to redownload it).

> Is it okay to assume wget and tar are available? (There are pure python replacements if preferred)

I've used plenty of systems using wget, so the pure Python variants would be preferred.

> The original runs NPT and NVE simulations; the simulation here is NVT (which would probably have a slight speed advantage over NPT). Can we add a MonteCarloBarostat for NPT to make it an exact comparison? (none of the other tests use that) 

Yes, it seems `MonteCarloBarostat` would be appropriate to add for NPT systems.

> How would we do a 4fs time step NVE? (maybe VelocityVerlet? sorry, I'm not familiar with it)

I believe AMBER uses the equivalent of the OpenMM `VerletIntegrator` for that.

> Hydrogen mass: these benchmarks are all intended to use HMR, but I couldn't figure out what the hydrogen mass is supposed to be. Is this already included in the .prmtop file? (and if yes, is it read in correctly?) The tleap default is 3amu, so I just set it explicitly to that.

It appears that heavy hydrogens are already specified in the `%FLAG MASS` section of the Amber topologies, so you should not need to specify an argument for `hydrogenMass`---if you did, you would be moving more mass around, which would cause problems.

> Adding context.setVelocitiesToTemperature() causes "particle position is nan" pretty quickly if running one of these systems. Not sure why; they may have velocities included in the inpcrd file. That's why this is skipped for these.

That's odd---it might be the hydrogen mass issues? 
In any case, the `inpcrd` files are indeed NetCDF restart files containing velocities, so it should be unnecessary to reset the velocities.

examples/benchmark.py
```diff
@@ -54,6 +67,20 @@ def runOneTest(testName, options):
...
+        dir = downloadAmberSuite()
+        prmtop = app.AmberPrmtopFile(os.path.join(dir, f'PME/Topologies/{testName}.prmtop'))
+        inpcrd = app.AmberInpcrdFile(os.path.join(dir, f'PME/Coordinates/{testName}.inpcrd'))
+        topology = prmtop.topology
+        positions = inpcrd.positions
+        dt = 0.004*unit.picoseconds
+        method = app.PME
+        cutoff = options.cutoff
+        constraints = app.HBonds
+        hydrogenMass = 3*unit.amu # NOTE do the inputs already have this included?  3.0 is standard for tleap
```
jchodera, line 80: You shouldn't need to specify this!

examples/benchmark.py
```diff
@@ -54,6 +67,20 @@ def runOneTest(testName, options):
...
+        friction = 1*(1/unit.picoseconds)
+        integ = mm.LangevinMiddleIntegrator(300*unit.kelvin, friction, dt)
+        system = prmtop.createSystem(nonbondedMethod=method, nonbondedCutoff=cutoff, constraints=constraints, hydrogenMass=hydrogenMass)
```
jchodera, line 83: You shouldn't specify `hydrogenMass` since the `prmtop` file already contains heavy hydrogens. This would further alter hydrogen masses, leading to problems.

examples/benchmark.py
```diff
@@ -131,7 +164,7 @@ def runOneTest(testName, options):
 parser = ArgumentParser()
 platformNames = [mm.Platform.getPlatform(i).getName() for i in range(mm.Platform.getNumPlatforms())]
 parser.add_argument('--platform', dest='platform', choices=platformNames, help='name of the platform to benchmark')
-parser.add_argument('--test', dest='test', choices=('gbsa', 'rf', 'pme', 'apoa1rf', 'apoa1pme', 'apoa1ljpme', 'amoebagk', 'amoebapme'), help='the test to perform: gbsa, rf, pme, apoa1rf, apoa1pme, apoa1ljpme, amoebagk, or amoebapme [default: all]')
+parser.add_argument('--test', dest='test', choices=('gbsa', 'rf', 'pme', 'apoa1rf', 'apoa1pme', 'apoa1ljpme', 'amoebagk', 'amoebapme', 'JAC',  'FactorIX', 'Cellulose', 'STMV'), help='the test to perform: gbsa, rf, pme, apoa1rf, apoa1pme, apoa1ljpme, amoebagk, or amoebapme [default: all]')
```
jchodera, line 167: Instead of just `JAC`, `FactorIX`, etc., would it make sense to call these `Amber20-JAC`, `Amber20-FactorIX` to emphasize they come from the hard-coded Amber20 benchmark suite URL?

peastman: Good idea.  Let's also make them all lower case: `amber20-cellulose`, etc.  That's consistent with the other values, and also easier to type!  Also update the help text to list all values.

peastman: Also, does 'JAC' mean 'DHFR'?  If so, let's call it that.

peastman:
`wget` doesn't exist on MacOS, and I don't think either one exists on Windows.  We should use the Python versions.

examples/benchmark.py
aizvorski: Okay - fixed to just use the masses from the file.

examples/benchmark.py
aizvorski: @jchodera @peastman Sounds good, renamed to the amber20-cellulose style.  

JAC is DHFR; it has the same number of atoms as the "pme" test, but is consistently slightly faster.

aizvorski:
> `wget` doesn't exist on MacOS, and I don't think either one exists on Windows. We should use the Python versions.

Alright, I thought that might be better - it's done.

aizvorski:
Opps, closed accidentally, didn't mean to do that

aizvorski:
@jchodera Thank you!  I think I've fixed up everything, this should be good to go.

> I've used plenty of systems using wget, so the pure Python variants would be preferred.

Replaced wget and tar with pure python urllib and tarfile versions.

> Yes, it seems `MonteCarloBarostat` would be appropriate to add for NPT systems.
> 
> > How would we do a 4fs time step NVE? (maybe VelocityVerlet? sorry, I'm not familiar with it)
> 
> I believe AMBER uses the equivalent of the OpenMM `VerletIntegrator` for that.

Added an --ensemble={NPT,NVT,NVE} option.  NPT is LangevinMiddleIntegrator+MonteCarloBarostat, NVT is LangevinMiddleIntegrator, NVE is VerletIntegrator. 

It may be slightly confusing that this option only applies to Amber tests, but at least the help message says so explicitly :)

> It appears that heavy hydrogens are already specified in the `%FLAG MASS` section of the Amber topologies, so you should not need to specify an argument for `hydrogenMass`---if you did, you would be moving more mass around, which would cause problems.

Good to know - I was concerned about that.  Okay, fixed to just use the mass in the file :+1: 

> > Adding context.setVelocitiesToTemperature() causes "particle position is nan" pretty quickly if running one of these systems. Not sure why; they may have velocities included in the inpcrd file. That's why this is skipped for these.
> 
> That's odd---it might be the hydrogen mass issues?
> In any case, the `inpcrd` files are indeed NetCDF restart files containing velocities, so it should be unnecessary to reset the velocities.

Somewhat mysteriously, this happens with the original hydrogen mass as well.  Probably okay to leave as is?

jchodera:
> I tried, this happens with the original hydrogen mass as well; and also with small steps (1fs fails, 0.1fs works). Probably okay to leave as is?

Yes, but it's worth @peastman taking a closer look at what is going on there to see if something more sinister is going on!

jchodera:
And thanks again, @aizvorski! This is a great new feature.

swails:
> so you should not need to specify an argument for `hydrogenMass`---if you did, you would be moving more mass around, which would cause problems.

You shouldn't have to, but this will not cause problems.  The code sets the hydrogens to the specified mass then subtracts the difference between the current mass and the desired mass from the attached heteroatom.

So if you set `hydrogenMass` to 3, but the prmtop already has them set to 3, then there will effectively be no difference as it'll add 0 to the hydrogens and substract 0 from their attached heteroatoms.  It would be clear to the reader, however, that hydrogen mass repartitioning was performed (arguably a good thing).

peastman:
> It may be slightly confusing that this option only applies to Amber tests, but at least the help message says so explicitly :)

Or we could make it apply to all the tests.  I think the same integrator/barostat combinations should work for all the non-AMOEBA tests.  The AMOEBA ones should use MTSIntegrator for NVE and MTSLangevinIntegrator for NVT or NPT.  We probably should make the default NVT, since that what these benchmarks currently use.

peastman:
I'll go ahead and merge this, then fix the `--ensemble` handling in a separate PR.

PR #2989, opened 2021-01-20
Implements findBlocksWithInteractions with matrix multiplication

dmclark17:
This changes findBlocksWithInteractions to compute the interacting atoms between two given sets using the following matrix multiplication framing for the pairwise distance calculation. 

```
d_{ij}^2 = (a_x - b_x)^2 + (a_y - b_y)^2 + (a_z - b_z)^2
         = (a_x^2 - 2a_xb_x + b_x^2) + (a_y^2 - 2a_yb_y + b_y^2) + (a_z^2 - 2a_zb_z + b_z^2)
         = (a_x^2 + a_y^2 + a_z^2) + (1) * (b_x^2 + b_y^2 + b_z^2) - 2*(a_xb_x) - 2*(a_yb_y) - 2*(a_zb_z)
```
		 
The squared norms of the position vectors are precomputed and stored in shared memory. Letting `a=a_x^2 + a_y^2 + a_z^2` and `b=b_x^2 + b_y^2 + b_z^2` gives:

```
0.5 d_{ij}^2 = 0.5a + (0.5)(b) + (a_x)(-b_x) + (a_y)(-b_y) + (a_z)(-b_z)
```

The pairwise distance can be computed using NxMx4 matrix multiplication where C is packed with `0.5*a`, the A matrix contains `(-a_x, -a_y, -a_z, 0.5)` and the B matrix contains `(b_x, b_y, b_z, b)`. The resulting D matrix will contain the pairwise distance divided by 2

By my count, this reduces the number of floating point operations by about 30%. A single pairwise distance calculation is 6 operations. In the matrix framing, assuming the norm calculations can be sufficiently amortized, a single distance calculation is 4 FMA operations. 

On a Titan V with CUDA11.0, I measured the following performance changes—Apoa1pme: a 2.5 % increase in the ns/day and a 26% speedup in the findBlocksWithInteractions kernel. Apoa1rf: a 3.4% increase in the ns/day and a 25% speedup in the kernel. I measured similar speed ups on A100 and P100 cards, however I did not see a speedup on a T4. The new algorithm does increase the shared memory and register usage of the kernel which might hurt performance on older architectures.

The matrix multiplication framing is currently an all or nothing approach, unlike the previous version which could skip unnecessary interactions by iterating from `first` to `last`. Are there any systems which you'd expect this to have a particularly large effect? I also transposed the shared memory buffers to improve the access patterns in the matrix multiplication, however this eliminates vectorized shared memory loads for the non-single periodic copy calculation. For Apoa1, it seems like most of the interactions can be done with a single periodic copy, are there any systems for which this would not be the case?

Please let me know what you think of the change and thanks for your time!


proteneer:
This is awesome - I may borrow this for another project...

proteneer:
Also - cheaper neighborlist computations also mean that we may be able to change the default padding from 1A to say 0.9A, which will indirectly speedup the actual nonbonded calculation itself.

peastman:
Very nice! I'll take look when I'm able to. (My power is out right now, so
it might be a bit.)

peastman:
I don't understand why the code is structured the way it is.  I feel like it's a lot more complicated than it needs to be.  We have `pos1` and `pos2`, which are now four component vectors with the squared magnitude divided by 2 in the `w` component.  Why not just write

```c++
halfDist2 = pos1.w + pos2.w - pos1.x*pos2.x - pos1.y*pos2.y - pos1.z*pos2.z;
```

and be done with it?  That looks to me like just one addition followed by three FMAs.  What is gained by all the complicated loops and copies to separate arrays?

> The matrix multiplication framing is currently an all or nothing approach, unlike the previous version which could skip unnecessary interactions by iterating from first to last. Are there any systems which you'd expect this to have a particularly large effect?

It depends on how dense the system is, or to put it differently, the typical size of a block's bounding box compared to the cutoff distance.  If that's large, there may be many atoms with no interactions, so we gain a lot by skipping them.  If that's small, fewer atoms will have no interactions so it gains less.

> For Apoa1, it seems like most of the interactions can be done with a single periodic copy, are there any systems for which this would not be the case?

That's the normal case.  Usually only a small fraction of them can't be handled by the single periodic copy case.  The exception would be relatively small systems where the box isn't that big compared to the cutoff distance.

dmclark17:
@proteneer Thanks—I hope it can be useful!

@peastman Thanks for taking the time to review the code, and I am glad you power is presumably back on!

That is a good point about directly using the expanded distance formula—I had missed that simplification. However, I believe the current implementation will have more efficient shared memory access. In the direct method, each thread needs to read 32 float4 values from shared memory. The current method uses a 4x4 thread block: 4 values from A and 4 values from B are read into registers and used to compute a 4x4 contribution of the output matrix. Each thread is responsible for two 4x4 tiles and therefore reads in 2*(4+4)*k float values from shared memory. The reads can be vectorized, which brings it down to 16 float4 values. 

I will do some experimentation to see how the suggested method performs. I am a little biased, but the current method also has the advantage of easier Tensor Cores integration in the future.

peastman:
I thought reads from shared memory were as fast as registers?  That's what I was always told.

dmclark17:
I don't think that that is the case—while shared memory has much higher bandwidth and lower latency than local and global memory, accessing a register generally consumes zero extra clock cycles per instruction.

jchodera:
> I am a little biased, but the current method also has the advantage of easier Tensor Cores integration in the future.

Since this function is used to identify potentially interacting atoms rather than compute energy or force contributions, can we get by with TF32 precision here and make use of tensor cores now?

dmclark17:
I spoke too soon, switching to the direct version further improved the performance—I believe this is because it removes the shared memory atomics I introduced. My preliminary new measurements now show a 5% increase for Apoa1rf and 3.5% for Apoa1pme. I will do some more extensive benchmarks.

peastman:
Even better!

dmclark17:
I updated the PR with the suggestion from @peastman. For apoa1, I found that it was faster to execute the inner most loop 32 times as opposed to using the `first` and `last` variables—executing the loop for a fixed number of unrolled iterations reduces the number of logic and control operations. The difference in ns/day was about 1.5% for apoa1rf and 0.5% for apoa1pme.

I also noticed that some of the other loops were being unrolled beyond their maximum number of iterations. To fix this, I added some explicit unroll pragmas. This improved the performance by about a percent for apoa1rf and apoa1pme in CUDA 11.2. 

As with the initial code change, I measured similar performance improvements on the A100 and P100 (roughly 5% overall for apoa1rf and 3.5% overall for apoa1pme), and a less than a percent improvement on the T4.

Regarding the TF32 version—I can open a new PR with a Tensor Core enabled version for further discussion about its accuracy and performance implications.

peastman:
That looks great!  I'll try it out and see how it affects performance on my computer.

peastman:
This works great.  I ran a bunch of benchmarks of different sizes on a Titan V.  For each one I tested three versions:

1. The original version ("Original")
2. The version in this PR ("PR")
3. One where I switched the loop back to `first` to `last` instead of 0 to 32 ("Loop").

Here are the speeds in ns/day.

| Test | Atoms | Original | PR | Loop |
| --- | --- | --- | --- | --- |
| pme | 23558 | 845.0 | 901.4 | 871.3 |
| amber20-factorix | 90906 | 291.3 | 294.2 | 299.3 |
| apoa1pme | 92224 | 250.6 | 260.3 | 258.4 |
| amber20-cellulose | 408609 | 46.5 | 46.8 | 47.3 |
| amber20-stmv | 1067095 | 14.4 | 14.5 | 14.5 |

In all cases the version in this PR is faster than the original one.  The impact is quite large for small systems, but gets smaller as the system size increases.  The handling of the loop is less clear.  Sometimes one version is faster, sometimes the other one.

I'll merge this, then try making the same change to the OpenCL version and see how it affects that.  I'll also experiment with changing the padding radius to see if a smaller value is now faster.

peastman:
Here's the effect of changing the amount of padding.  The current code adds 10% padding to the cutoff distance.  I tried values down to 6%.

| Test | 10% | 9% | 8% | 7% | 6% |
| --- | --- | --- | --- | --- | --- |
| pme | 901.4 | 910.1 | 903.8 | 916.8 | 890.7 |
| amber20-factorix | 294.2 | 296.2 | 298.5 | 300.7 | 279.3 |
| apoa1pme | 260.3 | 259.7 | 262.1 | 261.9 | 262.0 |
| amber20-cellulose | 46.8 | 47.2 | 47.5 | 46.8 | 41.5 |
| amber20-stmv | 14.5 | 14.4 | 14.7 | 14.2 | 12.3 |

So what do we make of this?  6% seems to generally be too low.  It has much worse performance for all tests except ApoA1, which actually has its second highest performance there.  DHFR and FactorIX have their best performance at 7%.  ApoA1, Cellulose, and STMV are all fastest at 8%.  In all cases the speedups are quite small.  Let's just compare how much faster each test is at 8% than at 10%:

| Test | Speedup |
| --- | --- |
| pme | 1.003 |
| amber20-factorix | 1.015 |
| apoa1pme | 1.007 |
| amber20-cellulose | 1.015 |
| amber20-stmv | 1.014 |

So we're looking at speedups between 0.3% and 1.5%.  Still, I guess it's more than nothing!

philipturner:
Regarding https://github.com/openmm/openmm/issues/2489#issuecomment-1374149297, `findBlocksWithInteractions` takes the longest on M1 chips. If we can implement this as a matrix multiplication, the Metal backend could utilize `simdgroup_matrix` or `simd_sum` operations to boost performance. This would boost ALU utilization from 25% to 80%. M1 Pro would execute steps ~1.37x faster (30 ns/day on ApoA1 RF).

Apple's API goes like this. A hardware instruction takes pairs of FP32 registers across an entire SIMD-group. That means 64 FP32 values in total - an 8x8 matrix. `simdgroup_multiply_accumulate` takes three pairs of registers per thread, performing an 8x8x8 MAC in 20 cycles. Even if the last component is just 4 (8x8x4) and the other half zeroed, we'd still boost ALU from 25% to 40%. Another approach might use explicit FMA with quad shuffles/broadcasts (2 cycles) and quad sums (6 cycles).

> Since this function is used to identify potentially interacting atoms rather than compute energy or force contributions, can we get by with TF32 precision here and make use of tensor cores now?

Have you made any progress accelerating this with tensor cores?

philipturner:
I quickly realized that `simdgroup_matrix` probably won't help much here - ALU is already utilized quite well. You're coalescing `__local` memory reads; threadgroup bandwidth likely isn't a bottleneck (unlike SGEMM). Sorry for sending an extra ping!

PR #2992, opened 2021-01-22
Made ordering of impropers more deterministic

peastman:
See https://github.com/conda-forge/openmm-feedstock/pull/46.  This accounts for a difference between CPython and Pypy in how they order the elements of sets.

### DIFF
--- a/wrappers/python/simtk/openmm/app/forcefield.py
+++ b/wrappers/python/simtk/openmm/app/forcefield.py
@@ -608,6 +608,7 @@ class ForceField(object):
                 self.bondedToAtom[bond.atom2].add(bond.atom1)
                 self.atomBonds[bond.atom1].append(i)
                 self.atomBonds[bond.atom2].append(i)
+            self.bondedToAtom = [sorted(b) for b in self.bondedToAtom]
 
         def addConstraint(self, system, atom1, atom2, distance):
             """Add a constraint to the system, avoiding duplicate constraints."""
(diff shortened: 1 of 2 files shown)


wrappers/python/simtk/openmm/app/forcefield.py
```diff
@@ -989,8 +990,8 @@ def _buildBondedToAtomList(self, topology):
 
         Returns
         -------
-        bondedToAtom : list of set of int
-            bondedToAtom[index] is the set of atom indices bonded to atom `index`
+        bondedToAtom : list of set list int
```
jaimergp, line 993: Shouldn't this be

```suggestion
        bondedToAtom : list of list of int
```

?

jaimergp:
From https://github.com/conda-forge/openmm-feedstock/pull/46#issuecomment-765682957:

> However, it does mean that some impropers will end up in a different order than they did in previous OpenMM versions.

The iteration order of that set was arbitrary before this patch anyway (since it depended on the value of the random seed set in each interpreter session). If you want to save the `sorted` call, an alternative could rely on the keys of a `dict` (`{key: None}`), which preserve insertion order since Python 3.6 in both implementations.

wrappers/python/simtk/openmm/app/forcefield.py
peastman: Thanks, fixed.

peastman:
That would also work, but the order would still be different from what it was in previous releases.  The new order is neither better nor worse than the old one.  We just need to be aware that it will sometimes be different.

Issue #2997, opened 2021-01-26
Rename python module?


peastman:
This question came up recently in #2991.  I've thought for a long time it would be nice if we could drop the `simtk` prefix from the Python modules.  It's only there for historical reasons, and it creates extra clutter.  `simtk.openmm` would become just `openmm.`   `simtk.openmm.app` would become `openmm.app`.  `simtk.unit` would become `openmm.unit`.

It would be essential to do this in a way that doesn't break compatibility with existing code.  I believe we could do that by keeping around the existing packages, but making them just be aliases for the new ones.  For example, `simtk.openmm` would contain an `__init__.py` file that reads

```python
from openmm import *
```

If we made this change, we could also update the module names for plugins that get installed separately.  `openmmtorch` could become `openmm.torch`, `openmmplumed` could become `openmm.plumed`, etc.

proteneer:
This is a fantastic idea! Note that for ```simtk.unit``` there's actually a stand alone package in pypi https://pypi.org/project/simtk.unit/ that was initially maintained by @rmcgibbo.

We actually use the pypi package in some non-openmm projects as well and I was wondering if it was possible to keep that library decoupled still.

jchodera:
Would love to hear thoughts from others on whether this is an opportunity to create a more Pythonic Python API that breaks free of legacy naming and other choices from the C++ API and harmonizes with modern efforts such as the [Open Force Field Initiative](http://openforcefield.org) standards, such as:
* PEP8 naming conventions
* Pythonic idioms (e.g. iterators)
* Modern unit systems that can check dimensionality (e.g. [pint](https://pint.readthedocs.io/en/stable/), [unyt](https://unyt.readthedocs.io/en/stable/))

raimis:
@peastman, I agree!

The import names should be consistent with the package names.

`openmm` package provides:
```python
from openmm import *
```
`openmm-plumed` package provides:
```python
from openmm.plumed import *
```
and so on.

peastman:
> Note that for simtk.unit there's actually a stand alone package in pypi https://pypi.org/project/simtk.unit/ that was initially maintained by @rmcgibbo.

It looks like that was last updated in 2014, so it's pretty far out of date.

> Would love to hear thoughts from others on whether this is an opportunity to create a more Pythonic Python API

I think that's its own issue.  Let's keep this one just to the package names.  Feel free to open a separate issue for other questions.

mattwthompson:
[...]

PR #3000, opened 2021-01-30
Top level Python module is now "openmm"

peastman:
Implements #2997.  `simtk.openmm` is now just `openmm`.  `simtk.unit` is now `openmm.unit`.  `simtk.testInstallation` is now `openmm.testInstallation`.  The old names are still supported for backward compatibility.


jchodera:
@peastman : What about adding in a deprecation warning for importing the legacy `simtk.*` paths? It would be useful to warn people that we may at some point in the future decide to deprecate the legacy paths.

peastman:
Added.

PR #3003, opened 2021-02-02
Fixed test failures on CPU OpenCL

peastman:
See https://github.com/openmm/openmm/pull/2964.


jchodera:
@peastman Good catch!

Does this just make the error more deterministic, or was there actually a race condition that could have impacted the results to a significant degree?

peastman:
Neither one.  If you were using a CPU implementation of OpenCL, and your CPU only had two cores, the pinned buffer was too small which led to illegal memory access.  In any other case, the buffer was large enough so there was no problem.

Issue #3002, opened 2021-02-01
Discussion about OpenMM units


peastman:
This is to continue the discussion that was started in #2997 about the OpenMM units module.  @jchodera proposed switching to a different package such as pint or unyt.  Here are the particular arguments for and against that have been raised.

He pointed out that most functions from the C++ library don't check the dimensionality of their inputs, so you could pass a length to an argument that expects a time.  Being able to check this would help to prevent errors.  I pointed out that fixing this is independent of what units package we use.  We just need to fill in the expected argument types in swigInputConfig.py.  That would be true regardless of what package we used.

He asserted the OpenMM units package is much slower than the other ones.  We don't yet have data to support this.  He presented an example to show units were slow, but it just compared units to no units, rather than comparing to pint or unyt.  @swails presented a benchmark in which OpenMM units were 5x faster than pint.  @jchodera pointed out there is no way to prevent OpenMM from adding units to the values it returns, and this creates overhead.  @swails pointed out that adding units is much cheaper than removing units, and you can avoid the cost of the latter by accessing `._value` directly.

@jchodera pointed out a case where OpenMM's units don't interoperate well with numpy: the `prod()` method.  Improved numpy interoperability would be a good thing.

Other arguments raised against the proposal are that it 1) would be a lot of work, and 2) would break compatibility with existing code that uses OpenMM.  We should investigate whether there's any way the change could be made while still preserving backward compatibility.

peastman:
A couple of further thoughts on the subject of speed.  We need to evaluate this with realistic benchmarks rather than artificial microbenchmarks.  The examples so far compared `system.addParticle(1)` to `system.addParticle(1*dalton)`.  But `addParticle()` is a completely trivial method that takes negligible time.  The overhead of adding and removing units is much higher than the cost of the method itself, but even with that overhead it's still probably a negligible fraction of the time in any real world application.

The other thing is that I recently learned from https://github.com/conda-forge/openmm-feedstock/pull/45 that OpenMM can now work under Pypy.  It would be interesting to see how that affects this.  Adding and removing units isn't really very complicated, but it's all implemented in Python, and that makes it slow.  How does a JIT change the speed?

mattwthompson:
A couple minor comments on the other unit packages (sidestepping completely the questions of if OpenMM should use them, how feasible such a transition would be, etc.)


### [`unyt`](https://github.com/yt-project/unyt) 
[...]

PR #3010, opened 2021-02-06
Improved units checking for function arguments

peastman:
I updated swigInputConfig.py to specify the expected units for lots of methods in the C++ API.  This allows allows it to check that you're passing a compatible value.  This isn't comprehensive, but I tried to get most of the commonly used classes.

See #3002.


jchodera:
This is a great safety improvement! Thanks!

PR #3017, opened 2021-02-09
Improve selection of architecture to compile for

peastman:
In 7.5 we changed the behavior of compiling CUDA kernels, so it would prefer the runtime compiler over the command line compiler.  This was supposed avoid problems, but it ended up creating some new problems.  One of them is that conda would sometimes make a bad choice of what toolkit to install, and you would get a compiler too old to support your GPU's architecture.

This PR addresses the problem by querying the runtime compiler to find the maximum architecture it supports.  It then tells the compiler to target the highest architecture that is supported by both the compiler and the GPU.

### DIFF
--- a/platforms/cuda/include/CudaContext.h
+++ b/platforms/cuda/include/CudaContext.h
@@ -530,9 +530,10 @@ private:
     int contextIndex;
     int numAtomBlocks;
     int numThreadBlocks;
+    int gpuArchitecture;
     bool useBlockingSync, useDoublePrecision, useMixedPrecision, contextIsValid, boxIsTriclinic, hasCompilerKernel, isNvccAvailable, hasAssignedPosqCharges;
     bool isLinkedContext;
-    std::string compiler, tempDir, cacheDir, gpuArchitecture;
+    std::string compiler, tempDir, cacheDir;
     float4 periodicBoxVecXFloat, periodicBoxVecYFloat, periodicBoxVecZFloat, periodicBoxSizeFloat, invPeriodicBoxSizeFloat;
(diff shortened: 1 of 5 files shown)


peastman:
Apparently the ability to query the compiler for what architectures it supports was only added in CUDA 11. :(

jchodera:
Can we still try this if CUDA 11 is available? Catch failures with an exception? It sounds like a useful improvement if CUDA 11 is available.

peastman:
The code won't compile with any CUDA version before 11.  I should be able to use a `#ifdef` to include it based on what version you're compiling against.

jaimergp:
Nightlies for CUDA 11 won't compile, @peastman. Check logs [here](https://dev.azure.com/OmniaMD/conda-dev-recipes/_build/results?buildId=565&view=logs&j=09c6ee00-92ab-5751-f80f-c049ce8ce0ed&t=480de774-f455-59ec-d10f-fc6e4f2ba338&l=6080).

jchodera:
Is there a way to get alerts about that?

jaimergp:
Azure emails the team on failures. Check your account configuration here: https://dev.azure.com/OmniaMD/conda-dev-recipes/_settings/projectOverview

jaimergp:
I readded you in case that fixes some missing permissions. I guess it will fail tonight again, so you should get a notification tomorrow morning.

peastman:
Thanks!  I fixed it in #3036.

jaimergp:
This made me realize the nightlies need to be updated to include 11.1 and 11.2. I'll open an issue.

PR #3018, opened 2021-02-09
Use command line compiler if it is explicitly specified

peastman:
In 7.5 we changed the behavior of compiling CUDA kernels, so it would prefer the runtime compiler over the command line compiler. This was supposed to avoid problems, but it ended up creating some new problems. One of them is that if anything went wrong with the runtime compiler, it was hard to fix.  There was no straightforward way to force it to use the command line compiler instead.

This PR changes the behavior so that if you explicitly specify a command line compiler, either with the `'CudaCompiler'` property or the `OPENMM_CUDA_COMPILER` environment variable, that forces it to use the compiler you specify.  It will never use the runtime compiler instead.

### DIFF
--- a/devtools/Jenkinsfile
+++ b/devtools/Jenkinsfile
@@ -15,10 +15,8 @@ pipeline {
                     }
                     steps {
                         sh "git clean -fxd && git checkout ."
-                        withEnv(["OPENMM_CUDA_COMPILER=/usr/local/cuda/bin/nvcc"]) {
-                            sh "devtools/ci/jenkins/install.sh"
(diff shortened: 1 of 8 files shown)


peastman:
run jenkins

peastman:
@swails Jenkins is failing to start up:

```
+ docker inspect -f . swails/openmm-all:latest
.
Failed to run image 'swails/openmm-all:latest'. Error: docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]].
```

swails:
> @swails Jenkins is failing to start up:

I was bringing another agent into the Jenkins array that needed to be rebooted.  The PR job has now passed.

peastman:
Thanks!

PR #3025, opened 2021-02-17
Reduced padding on cutoff

peastman:
This is a very minor optimization.  See https://github.com/openmm/openmm/pull/2989.


jchodera:
Is this a value that could be tuned on the fly to deliver optimal performance?

peastman:
It's not worth the trouble.  See the benchmarks at the link above.  Any speedup from further tuning would probably be no more than about 1%.

PR #3037, opened 2021-02-27
Point based geometric functions for custom forces

peastman:
This adds three new functions that can be used in CustomCompoundBondForce, CustomCentroidBondForce, and CustomManyParticleForce: `pointdistance()`, `pointangle()`, and `pointdihedral()`.  These are similar to the `distance()`, `angle()`, and `dihedral()` functions, but instead of taking particle identifiers they take the coordinates of the points to base the calculation on.  This allows lots of computations that weren't possible before.  As an example, you can calculate the distance from p1 to the midpoint between p2 and p3 with `pointdistance(x1, y1, z1, (x2+x3)/2, (y2+y3)/2, (z2+z3)/2)`.

As a bonus, I was able to use this feature to simplify the implementation of custom forces, so it actually leads to a large reduction in total lines of code!

My immediate reason for implementing this is that it will let me use custom forces for more of the AMOEBA terms.  It should be useful for a lot of other things too.

### DIFF
--- a/openmmapi/include/openmm/CustomCentroidBondForce.h
+++ b/openmmapi/include/openmm/CustomCentroidBondForce.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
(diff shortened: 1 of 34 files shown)


swails:
Nice!

PR #3038, opened 2021-02-28
Read E14FAC Coulomb 1-4 scaling factor in CHARMM files

agiliopadua:
Read the E14FAC Coulomb 1-4 scaling factor from CHARMM parameter files (NONBONDED line).
This would enable support of different force field specifications in the CHARMM input format.

### DIFF
--- a/wrappers/python/openmm/app/charmmparameterset.py
+++ b/wrappers/python/openmm/app/charmmparameterset.py
@@ -116,6 +116,7 @@ class CharmmParameterSet(object):
         self.nbthole_types = dict()
         self.parametersets = []
         self.nbxmod = 5
+        self.e14fac = 1.0
 
         # Load all of the files
         tops, pars, strs = [], [], []
@@ -273,6 +274,8 @@ class CharmmParameterSet(object):
                     if nbxmod not in list(range(-5, 6)):
                         raise CharmmFileError('Unsupported value for NBXMOD: %d' % nbxmod)
                     self.nbxmod = nbxmod
+                if 'E14FAC' in fields:
+                    self.e14fac = float(fields[fields.index('E14FAC')+1])
                 continue
             if line.startswith('NBFIX'):
                 section = 'NBFIX'
--- a/wrappers/python/openmm/app/charmmpsffile.py
+++ b/wrappers/python/openmm/app/charmmpsffile.py
@@ -1384,7 +1384,7 @@ class CharmmPsfFile(object):
             for ia1, ia4 in self.pair_14_list:
                 atom1 = self.atom_list[ia1]
                 atom4 = self.atom_list[ia4]
-                charge_prod = (atom1.charge * atom4.charge)
+                charge_prod = (atom1.charge * atom4.charge) * params.e14fac
                 epsilon = sqrt(abs(atom1.type.epsilon_14 * atom4.type.epsilon_14)) * ene_conv
                 sigma = (atom1.type.rmin_14 + atom4.type.rmin_14) * (length_conv * sigma_scale)
                 force.addException(ia1, ia4, charge_prod, sigma, epsilon)


jchodera:
Thanks so much for this contribution, @agiliopadua!

@peastman @swails : I'm not sure how to wrap my head around the divergence between the implementations of the CHARMM PSF and parameter set readers in OpenMM and [ParmEd](https://github.com/ParmEd/ParmEd/blob/master/parmed/charmm/parameters.py) at this point (which already handles `E14FAC`). Is there a way we could avoid the code duplication here?

agiliopadua:
Sorry for bothering you with this small thing... I am trying to simulate PF6-, which has 90 and 180 degree F-P-F angles. Only the 90 degree angles should have terms in the force field.
Using xml, which I prefer, OpenMM guesses the angles from the bond topology and the 180 degree angles will be in the system (bad), but one can specify the 1-4 Coulomb factor (good).
Using CHARMM files, the topology is strictly taken from the psf, which my tool builds without the 180 degree angles (good) only here the 1-4 Coulomb factor is not read (bad)...
Thanks!

swails:
The code in OpenMM is from an incarnation of ParmEd that predated the unified data model.  That version had very few advantages in ParmEd, but made it easy to transplant it into OpenMM, which is what I did.  Much more of ParmEd would need to be transplanted to OpenMM to bring them two versions of the code into line.

One option is to vendor a copy of ParmEd (say inside `app/internal`) and pull in the components we want to expose in the app layer.

One upside of vendoring ParmEd is that OpenMM could also cut out its Amber topology file parsers and just expose ParmEd's, which are more complete (for instance, OpenMM's AmberPrmtopFile doesn't actually support the latest Amber force fields because they employ CMAPs, but OpenMM's parsers ignore those sections -- I will open this in a new issue shortly).

The same could be done for GROMACS file support, although I suspect that's the one set of parsers more exercised in OpenMM than ParmEd.

jchodera:
I like the idea of 'circling the wagons' and maintaining a single set of tools for reading these various formats. Introducing `numpy` and `parmed` dependencies should be no problem at this point given this is the point of using `conda-forge` in the first place: Creating an interoperable ecosystem of tools.

My worry is that the other approach of vendoring a copy of `parmed` within OpenMM would enable them to get out of sync, and certainly delivering `parmed` as an accessible package within OpenMM would lead us to the same trouble we had with AmberTools doing this. If vendored a copy rather than simply introduced a package dependency, we would want to make sure `parmed` was hidden and internal, rather than had the potential to conflict with a conda-installed `parmed`. But there really is no point to vendoring a copy anymore---it's just more developer effort, more potential for issues, with no upside.

Longer term, the Open Force Field Initiative is creating an interoperable and extensible `System` object that could make it very easy to translate between many formats, and I'd love to see us be able to work with that in the future. I believe they've already reached out to @swails to see how easy it might be to adapt the reader components of ParmEd to populate this object---that might be a better long-term solution.

swails:
> Longer term, the Open Force Field Initiative is creating an interoperable and extensible `System` object that could make it very easy to translate between many formats, and I'd love to see us be able to work with that in the future. I believe they've already reached out to @swails to see how easy it might be to adapt the reader components of ParmEd to populate this object---that might be a better long-term solution.

These are ongoing discussions (that started very recently).  I'm pushing to explore the feasibility of morphing ParmEd into that solution (abandoning strict adherence to backwards-compatibility, either adding co-leads or new leads, whatever is needed).  But that may not be the route that is taken.

This was part of the reason I suggested vendoring, so the dependencies don't flip around.

> My worry is that the other approach of vendoring a copy of `parmed` within OpenMM would enable them to get out of sync,

I don't actually see this as a problem.  I don't think they'll get out-of-sync, I think the ParmEd version vendored into OpenMM would simply fall behind, and there are very simple mechanisms (subtree merges, for example) of updating a vendored copy when the need arises.

But who cares if it falls behind as long as the functionality that it's "missing" is unrelated to creating OpenMM objects from amber, charmm, and gromacs files?

> and certainly delivering `parmed` as an accessible package within OpenMM would lead us to the same trouble we had with AmberTools

This is different - ParmEd is vendored into AmberTools as a first-class package importable at the top-level.  The plan is to remove ParmEd from the AmberTools conda package and introduce it as a dependency for precisely this reason.  And ParmEd was part of AmberTools far before it was made into a standalone package, and it was made into a standalone package far before AmberTools was built into a conda package.

My suggestion was to vendor ParmEd inside `app/internal` and use it from there.  It won't collide with, or use, any other ParmEd package and will avoid collisions entirely.  I've written the imports precisely to make this kind of private, internal vendoring possible.

I'm certainly fine if the dependency route is chosen instead of vendoring, but I see the downside of vendoring to be fairly minimal.  There's also the compromise approach of vendoring ParmEd only until there is clarity and maturity in the OpenFF-sponsored System, then turn that into a dependency (be it `parmed` or something else).

swails:
@peastman, do you have thoughts about this discussion?

peastman:
I see advantages to both approaches.  Code duplication is generally a bad thing.  It's very hard to prevent the duplicated code from diverging, and you miss out on bug fixes made to the upstream version.  On the other hand, dependencies lead to your code spontaneously breaking because of a change to the package you depend on, and then your users start screaming and you have to rush out a patch.  (Just in the last few weeks, DeepChem has had to deal with a whole string of things breaking due to changes in upstream packages.  This is a sensitive subject for me at the moment!)

In theory you can avoid those problems by pinning to a particular version of the library you use.  In practice, due to the way Python and conda work, version pins are a nightmare.  They quickly lead to situations where it's impossible to install two packages in the same environment because they pin to different versions of some third package.

For small amounts of code, my preference is to duplicate it.  For larger amounts of code (and ParmEd counts as large), it's better if you can make it a dependency, but only if the package makes robust guarantees about forward compatibility and gets tested against all the packages that depend on it before new versions are released.

If there's a project underway to create a new representation for translating between formats, and the plan is to eventually adopt that, it might be best to just wait until it's available.

swails:
> If there's a project underway to create a new representation for translating between formats, and the plan is to eventually adopt that, it might be best to just wait until it's available.

This *could* be ParmEd or it could be something different (if it's ParmEd, the interface that OpenMM would utilize would be stable, but that's just `parse`, `createSystem`, and `topology`).  But even an optimistic plan measures the time frame of replacing ParmEd in years.  And if it's *not* ParmEd, it still doesn't target all of the functionality that ParmEd provides for the OMM ecosystem (e.g., translating Amber and CHARMM force fields, for instance).

Waiting for this situation to resolve likely amounts to waiting forever and will then force OpenMM to make parallel fixes to ParmEd.

If you're comfortable making ParmEd a dependency then that's probably the better route than vendoring, but you'd expressed resistance to introducing dependencies in the past (ParmEd would also introduce a numpy dependency, but everything else is optional).

peastman:
Regarding this PR, the change looks very simple.  My knowledge of CHARMM file formats is limited though.  Can someone confirm it's correct?  Do we also need to take it into account when handling 1-4 interactions between Drude particles?

https://github.com/openmm/openmm/blob/ac754c45a05c498e20624e09647b1947d77cb677/wrappers/python/openmm/app/charmmpsffile.py#L1493-L1497?

swails:
The parsers already handle the explicit 1-4 interactions.  Where the exceptions are added, we just need to add the 1-4 scaling divisor.

Your proposed change would start double-counting the 1-4 interactions.  It could work that way, but would then require converting the existing exceptions into exclusions.

swails:
Note that ParmEd already handles this case correctly.

swails:
Oh I misread your comment. That might be all that’s necessary. I had forgotten that was how 1-4s were implemented in OpenMMs charmm parser...

agiliopadua:
Normally the Thole function handles the short-range damping between Drude dipoles, so the 1-4 factor should not affect the charges making the induce dipole.

jchodera:
> If there's a project underway to create a new representation for translating between formats, and the plan is to eventually adopt that, it might be best to just wait until it's available.

@peastman: We would guarantee, as part of the design, that the code changes required to integrate ParmEd's speculated successor to read files and create an OpenMM `System` would be extremely minimal. However, the timeframe for this replacement is uncertain. It could be a year or two.

peastman:
On the immediate question, is it ok with everyone if I merge this?

On the slightly longer term, the question seems to be whether we should leave the code alone until the eventual library is ready, or convert it now to use ParmEd.  If the current code is basically working, I don't see any urgent reason to change it.  It's been pretty low maintenance code, so continuing to maintain it for another year or two isn't a big deal.  Are there any major benefits to replacing it immediately?

swails:
>  Are there any major benefits to replacing it immediately?

With Amber in particular, you get immediate support for Amber FFs with CMAPs (which is the currently-recommended family of force fields), Chamber-style topology files (converted CHARMM systems), NetCDF-formatted restart files, and additional reporters (most useful is probably NetCDF Trajectory reporter).

You also get to jettison a very large portion of the app layer code responsible for Amber file parsing.  And the OpenMM System creation code is centralized in ParmEd, meaning that the CHARMM and AMBER file parsers pass through the same code.  ParmEd also has utilities for doing energy decomposition (very helpful for investigating conversions and comparing against references from other programs).

So I guess the benefits are that you can stop maintaining the CHARMM file parsers (which are just an old incarnation of ParmEd, anyway) and you get to close #3040 as fixed.

peastman:
That does sound like enough benefit to justify it!  And ParmEd doesn't bring in any dependencies beyond numpy?  That sounds worth doing.

We should keep ParmEd as a soft dependency.  If someone doesn't have it installed, `import openmm` should still work, but creating an AmberPrmtopFile will throw an exception.

In the mean time, I'm going to merge this PR.  There's no reason not to, even if it will become irrelevant once the CHARMM implementation gets replaced.

swails:
Awesome! I'll get it done!

PR #3041, opened 2021-02-28
Trap Amber CMAPs

### LINKED ISSUE #3040: AmberPrmtopFile support is broken for ff19SB
The new generation of Amber force fields have adopted CMAP terms.  OpenMM doesn't parse these sections, but also doesn't fail if they're present, meaning that any Amber topology file that has CMAP terms is silently and non-obviously doing the wrong thing.

swails:
Fixes #3040 

Just adds a trap - I really don't have an interest in re-implementing this from ParmEd, but we should at least prevent people from running the wrong force field.


peastman:
Thanks!  I'll merge this, but also keep #3040 open since we should support them in the future.

Issue #3039, opened 2021-02-28
CUDA platform produces INF energy or crashes when launched on multiple A100 GPUs


Dann239:
I've been running some simulations on a system with two A100 gpus and encountered the following issue: when i launch a simulation in multi-gpu mode, it produces incorrect results on small systems or crashes when the system size exceeds approximately 300000 particles. Here's the code that reproduces the issue for me:
```cpp
#include <vector>
#include <iostream>
#include <cstdio>
#include <cmath>
#include <map>
#include <string>
#include "OpenMM.h"

using properties_t = std::map<std::string, std::string>;

double len(const OpenMM::Vec3& v) {
    return sqrt(v.dot(v));
}
double gen_rand(double max) {
    return rand() * max / RAND_MAX;
}

int lj(int amount, std::string platform1, properties_t properties1, std::string platform2, properties_t
properties2, double eps) {
    double size = cbrt(amount) + 2;

    std::vector<OpenMM::Vec3> pos(amount);

    // this generates some configuration of particles. what exactly happens here is ultimately unimportant
    int grid_dim = size;
    std::vector<std::vector<std::vector<int>>> grid(grid_dim,
        std::vector<std::vector<int>>(grid_dim,
        std::vector<int>(grid_dim, 0)));
    srand(0);
    for(auto& p : pos) {
        int* occ = nullptr;
        do {
            int grid_pos[3] = {
                rand() % grid_dim,
                rand() % grid_dim,
                rand() % grid_dim};
            p = {
                static_cast<double>(grid_pos[0]) + gen_rand(0.1),
                static_cast<double>(grid_pos[1]) + gen_rand(0.1),
                static_cast<double>(grid_pos[2]) + gen_rand(0.1)
            };
            occ = &grid[grid_pos[0]][grid_pos[1]][grid_pos[2]];
        } while(*occ);
        *occ = 1;
    }
    // end of the unimportant part

    OpenMM::System sys;
    sys.setDefaultPeriodicBoxVectors(
        {size, 0, 0},
        {0, size, 0},
        {0, 0, size}
    );

    // i create a simple nonbondedforce. lj units are used here for convenience
    auto lj = new OpenMM::NonbondedForce();
    lj->setNonbondedMethod(OpenMM::NonbondedForce::CutoffPeriodic);
    lj->setUseSwitchingFunction(true);
    lj->setSwitchingDistance(2);
    lj->setCutoffDistance(4);
    for(int i = 0; i < amount; i++) {
[...]

PR #3042, opened 2021-02-28
Made ComputeContext worker threads rethrow encountered exceptions into the main thread.

Dann239:
While trying to debug the #3039 i ran into a problem: when i tried to catch crashes from main, the exception just "ignored" the catch statement. It puzzled me for a while and after some thinking i realized what the problem was: when an exception arises inside of a worker thread, it crashes the entire thing. So i made it stash the exception until the `flush()` is called and then rethrow it into the main thread.

### DIFF
--- a/platforms/common/include/openmm/common/ComputeContext.h
+++ b/platforms/common/include/openmm/common/ComputeContext.h
@@ -515,7 +515,8 @@ public:
     void flush();
 private:
     std::queue<ComputeContext::WorkTask*> tasks;
-    bool waiting, finished;
+    bool waiting, finished, crashed;
+    OpenMM::OpenMMException crash;
     pthread_mutex_t queueLock;
     pthread_cond_t waitForTaskCondition, queueEmptyCondition;
     pthread_t thread;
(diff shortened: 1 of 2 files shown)


peastman:
I like this idea.  We need to think about the implementation a bit.  First I need to clarify something: the title of the PR says it's a change to ThreadPool, but actually you changed the ComputeContext worker thread.  That's not the same thing, and it's used for a lot fewer purposes than ThreadPool.  Which one did you intend to change?

Dann239:
My bad, got my terminology mixed up. I amended the commit message and the PR title, i indeed only intended to change the ComputeContext worker threads.

peastman:
That simplifies things a lot.  The only thing that thread is ever used for is parallelizing a computation across multiple GPUs.  There's a separate ComputeContext for each GPU, and each one has its own worker thread.

This code looks good.  The one thing I'd change is the terminology, which is misleading.  Throwing an exception is not a crash.  It's just an exception, which can potentially be caught and handled property.  So I would rename "crashed" to "threwException" and "crash" to "exception".

Also one very minor bit of formatting.  When a language keyword (if, while, etc.) is followed by an expression in parentheses, there should be a space in between.  So `catch (const OpenMMException& e)`, not `catch(const OpenMMException& e)`, and so on.

Dann239:
@peastman : i did what you suggested, please verify that i haven't missed anything.

peastman:
Looks good.  Thanks so much for this change, and congratulations on your first merged PR!

Issue #2700, opened 2020-05-26
Overhaul AMOEBA code


peastman:
cc @jayponder @jppiquem @pren

The AMOEBA plugin has some of the oldest, messiest code in OpenMM.  The whole architecture is badly out of date, working quite differently from how the rest of OpenMM does.  This causes many problems.  For one thing, it's just really hard to maintain.  I dread having to make any changes to it (such as support for [splitting the multipole interactions into short and long range parts](https://github.com/openmm/openmm/issues/2520#issuecomment-633699140)).  It also is slower than it needs to be, since it has missed out an many optimizations made to the rest of the library over the years.  Some of the convoluted architecture was required to work around limitations of circa-2008 GPUs.  We've never supported AMOEBA on the OpenCL platform, because the thought of cloning all that code, converting it to OpenCL, and maintaining two copies of it was just too daunting.

I think the time has come for a major cleanup.  I suggest proceeding through the following steps.

First, eliminate most of the specialized bonded forces.  These classes only exist because AMOEBA predates custom forces.  That isn't how we implement new force fields anymore.  Most of them could be replaced by just a few lines of code using a CustomCompoundBondForce.  Eliminating these classes would remove thousands of lines of unnecessary code.

For backward compatibility, ForceField would still accept the old tags in XML files.  It would just create appropriate custom forces instead.  We could even do the same with serialization.  If you had previously saved a System to a file, and it included an AmoebaStretchBendForce, it could still be loaded.  That force would just get replaced by an equivalent custom force.

Second, I want to completely rework the code for AmoebaMultipoleForce, especially the CUDA implementation.  This should both simplify it and improve the performance.  I saw the effect of that while I was implementing HIPPO.  Instead of starting from the existing CUDA code I mostly wrote it from scratch in a more modern way, and that led to a substantial performance gain.  I believe the same should be possible with AMOEBA.

Third, I'll convert the CUDA platform to be based on the [common compute](http://docs.openmm.org/development/developerguide/developer.html#common-compute) framework.  This will finally allow us to support OpenCL, so AMOEBA simulations can be run on AMD and Intel GPUs.

peastman:
I'd like to start moving forward on this.  Before I do, it would be really helpful to get feedback, especially from @jayponder, @jppiquem, and @pren.  Doing this will lead to a big improvement in code maintainability and hopefully also a significant gain in speed.  Does it sound to you like a good idea?

jayponder:
[...]

PR #3046, opened 2021-03-02
Replaced several AMOEBA bonded forces with custom forces

peastman:
This is the first piece of #2700.  It eliminates AmoebaBondForce, AmoebaAngleForce, AmoebaInPlaneAngleForce, AmoebaOutOfPlaneBendForce, and AmoebaStretchBendForce.  I suspect that AmoebaPiTorsion and perhaps AmoebaWcaDispersionForce could also be replaced, but I couldn't find clear documentation on those forces, and the code that implements them is confusing.  If someone can provide the functional forms for them, that would be great.

Mostly this PR just deletes a whole lot of code (over 10,000 lines!).  The only significant changes beyond that are in forcefield.py, where it creates custom forces to implement the AMOEBA terms.  For example,

```python
force = mm.AmoebaBondForce()
force.setAmoebaGlobalBondCubic(self.cubic)
force.setAmoebaGlobalBondQuartic(self.quartic) 
```

is replaced by

```python
energy = "k*(d^2 + %s*d^3 + %s*d^4); d=r-r0" % (self.cubic, self.quartic)
force = mm.CustomBondForce(energy)
force.addPerBondParameter('r0')
force.addPerBondParameter('k')
```

Of course, some of them are a lot more complicated.  Here's the energy expression for in-plane angles.

```python
energy = """k*(d^2 + %s*d^3 + %s*d^4 + %s*d^5 + %s*d^6); d=theta-theta0;
            theta = %.15g*pointangle(x1, y1, z1, projx, projy, projz, x3, y3, z3);
            projx = x2-nx*dot; projy = y2-ny*dot; projz = z2-nz*dot;
            dot = nx*(x2-x3) + ny*(y2-y3) + nz*(z2-z3);
            nx = px/norm; ny = py/norm; nz = pz/norm;
            norm = sqrt(px*px + py*py + pz*pz);
            px = (d1y*d2z-d1z*d2y); py = (d1z*d2x-d1x*d2z); pz = (d1x*d2y-d1y*d2x);
            d1x = x1-x4; d1y = y1-y4; d1z = z1-z4;
            d2x = x3-x4; d2y = y3-y4; d2z = z3-z4""" % (self.cubic, self.quartic, self.pentic, self.sextic, 180/math.pi)
```
One disadvantage to this change is that context creation takes longer, due to the time needed to process the expressions and generate code from them.  Thanks to #3044, I've gotten that down to only about six seconds.


peastman:
@jayponder and @pren if you could review this that would be great!

jayponder:
Hi Peter, This is probably fine for the terms you list above as the ones you have already converted. Of course it breaks our Tinker-OpenMM interface, as it pulls code out of the plugin to replace with these custom forces in the Python interface. So we now have to switch to the custom force method in our interface code to Tinker :)  I guess we can probably mostly "borrow" the code you've put into Python and just use it in our interface. We would like to keep the Tinker-OpenMM interface working (see github.com/TinkerTools/Tinker-Openmm), as there are people who actually use it.

As you mention, some of the terms such as in-plane angles and out-of-plane bending are fairly complex. And the other two terms you mention, Torsion-Torsion and Pi-Torsion will be still worse. The Torsion-Torsion is essentially the Tinker version of the CHARMM CMAP mechanism. It adds a two-torsion coupling energy which is derived from a spline fit to a parameter-supplied 2D grid of energy values as a function of the two torsions (for example, a phi-psi coupling term for the peptide backbone). There are two forms, using different spline formulations- one for torsions that can rotate over the full 360 degrees, and another for torsions that have only a limited, fixed range (for example, torsions in rings). Even if you already have CHARMM CMAP, you can't just directly use it, as Tinker does the splines differently, and there are other differences. It is next to impossible (at least for me...) to imagine putting that term into a single "custom force".

Converting the PI-Torsion term would also be messy, though perhaps not quite as bad. It essentially computes a p-orbital axis direction at trigonal sp2 atoms, and then uses a standard trigonometric torsion function to compute a "pi-torsion" energy between such p-orbitals on adjacent atoms. We use it in places for aromatic systems in AMOEBA-like models (though it's really little to no different from just an average of all torsions across an aromatic bond).

peastman:
Is there a reference that gives the exact functional form of the pi-torsion force?

The other one I mentioned was WCA, not torsion-torsion.

> So we now have to switch to the custom force method in our interface code to Tinker :)

Sorry!  The changes will be very minimal, though, just a few lines for each force.  And you can copy the expressions directly from the Python code.  This does mean people need to be sure they have the correct OpenMM version to match the Tinker version.

jayponder:
Hi again, The WCA term is a Weeks-Chandler-Andersen term that is only used by the Generalized Kirkwood implicit solvation model, which is sort of a polarizable AMOEBA analog of Generalized Born with a fancy nonpolar term. WCA is used to estimation of the implicit dispersion energy component. The method used by the WCA code involves something similar to the HCT method for computing Born radii in GB models. This would also be hard to implement as a custom force.

There is no literature reference for the pi-torsion term, but I can describe it exactly. Consider the case of two bonded trigonal sp2 atoms, C and D:

A               E
   \           /
    C----D
   /           \
B               F

Then the pi-torsion term finds the p-orbital axis at C as the normal to the A-B-C plane passing through C, and the p-orbital axis at D as the normal to the D-E-F plane passing through D. It then computes the dihedral angle between these two p-orbital axes and computes a "standard" 2-fold trigonometric torsional term for that dihedral angle value. That's the why to describe it as orbital chemistry, but of course it's just the angle between the A-B-C and D-E-F planes...

And as you will further realize, the above pi-torsion is just a simple arithmetic average of the 2-fold A-C-D-E, A-C-D-F, B-C-D-E and B-C-D-F torsions. So I think we could in principle do away with this entirely as a separate energy term, but we have thusfar chosen not to.

peastman:
Perfect!  That sounds very easy to implement.

peastman:
A correction to the above, mainly for the sake of posterity.  The pi torsion energy is based on the angle between the A-B-D and C-E-F planes.  Here's the energy expression for it.

```python
energy = """2*k*sin(phi)^2;
            phi = pointdihedral(x3+c1x, y3+c1y, z3+c1z, x3, y3, z3, x4, y4, z4, x4+c2x, y4+c2y, z4+c2z);
            c1x = (d14y*d24z-d14z*d24y); c1y = (d14z*d24x-d14x*d24z); c1z = (d14x*d24y-d14y*d24x);
            c2x = (d53y*d63z-d53z*d63y); c2y = (d53z*d63x-d53x*d63z); c2z = (d53x*d63y-d53y*d63x);
            d14x = x1-x4; d14y = y1-y4; d14z = z1-z4;
            d24x = x2-x4; d24y = y2-y4; d24z = z2-z4;
            d53x = x5-x3; d53y = y5-y3; d53z = z5-z3;
            d63x = x6-x3; d63y = y6-y3; d63z = z6-z3"""
```

I still don't understand some of the code that calculates the force and how it's equivalent this.  (I think [this line](https://github.com/TinkerTools/tinker/blob/7d1d2a66af87e2410edd94711ed9cef87aaa26e9/source/epitors.f#L167) deserves an award for... something.  I'm not sure what, but definitely something!)  But the energies match to seven significant digits, so I'm pretty sure it's correct.

jayponder:
The great thing about trigonometry is with so many interesting identities, there is an almost unlimited opportunity to obfuscate things... Cheers!

peastman:
AmoebaPiTorsionForce is replaced.  Another 2000 lines of code deleted, replaced with 10 lines. :)

Issue #3055, opened 2021-03-10
OMM metadynamics vs PLUMED discrepancy


dlukauskis:
I've been comparing OpenMM's metadynamics behaviour to Amber and Gromacs PLUMED metadynamics. Using the same input files, I ran 10 alanine dipeptide simulations and have found that OpenMM metaD converges to a different deltaF value when compared to Amber + PLUMED. 

[Here's a link the files + plots](https://drive.google.com/file/d/1R9bRcXeq5RQqW3ke06ekOOG4OwlTxHNk/view?usp=sharing)

I changed metadynamics.py to print the hill height (+ a few other minor things) and when you plot the hill height vs time for PLUMED and OpenMM, you can see that the initial hills in OpenMM aren't scaled the same way their are scaled with PLUMED. Could this be the issue?

peastman:
@raimis any thoughts on this?

raimis:
@dlukauskis, I believe, you are missing a scaling factor. To convert from bias to free energy, you need to multiply by `-(T + delta T)/T` (https://www.plumed.org/doc-v2.5/user-doc/html/belfast-6.html), which is -1.2 in your case.

*PLUMED*, by default, already scales the gaussians in `HILLS` file, so you observer the maximum height of `1.44 kJ/mol/ = 1.2 kJ/mol * 1.2`. In case of *OpenMM*, the scaling is done `openmm.app.Metadyanmics.getFreeEnergy`, but if your read a bias file directly you miss it.

Alternatively, you can use *PLUMED* with *OpenMM* via a plugin (https://github.com/openmm/openmm-plumed).

dlukauskis:
Thank you, @raimis! That was indeed the issue and scaling the bias to get the free energy produces a FES identical to one produced by PLUMED.

PR #3060, opened 2021-03-11
Correctly use the force group for the metadynamics force

raimis:
The current code would produce incorrect results, if the force group `31` is already assigned and the metadynamics force is added.


raimis:
Discovered this while debugging https://github.com/openmm/openmm/issues/3055

peastman:
Thanks for catching this!

peastman:
Is this ready to merge?

raimis:
Yes!

Issue #3064, opened 2021-03-15
OpenCL Duplicate Definitions


dmclark17:
It appears that for some of the OpenCL kernels there are macro definitions that are defined twice. Specifically, `WORK_GROUP_SIZE` is defined twice in the source generated by `TestOpenCLCustomIntegrator`. One definition seems to be coming from:

https://github.com/openmm/openmm/blob/59b8afb1b0f01fbdd85d050e093f4e4d62c93daf/platforms/opencl/src/OpenCLContext.cpp#L203

and the other from:

https://github.com/openmm/openmm/blob/59b8afb1b0f01fbdd85d050e093f4e4d62c93daf/platforms/common/src/CommonKernels.cpp#L6366

Would it be okay to remove this behavior? I assume the value in the `defines` mapping should be favored over `compilationDefines`, and I was thinking of querying the `defines` structure before adding definitions from `compilationDefines`. If that sounds reasonable, I can submit a PR with the code change. Thanks!

peastman:
That sounds very reasonable.  When the same symbol is defined in both, `defines` should take precedence over `compilationDefines`.  We should change it both in the OpenCL platform at

https://github.com/openmm/openmm/blob/59b8afb1b0f01fbdd85d050e093f4e4d62c93daf/platforms/opencl/src/OpenCLContext.cpp#L574-L579

and in the CUDA platform at

https://github.com/openmm/openmm/blob/59b8afb1b0f01fbdd85d050e093f4e4d62c93daf/platforms/cuda/src/CudaContext.cpp#L491-L496

This won't actually change the behavior.  The later `#define` already overrides the earlier one.  But it's good to eliminate the warning, and perhaps it will save someone a little confusion some day.

peastman:
Fixed by #3065.

PR #3065, opened 2021-03-16
OpenCL Duplicate Definitions

dmclark17:
Queries the `defines` map before adding each `compilationDefines` variable to the kernel source code in order to avoid duplicate definitions. See #3064.

Thanks for your time!


peastman:
Thank you!

PR #3072, opened 2021-03-17
Fix PSF setBox function for Triclinic case

andysim:
When viewing a PDB generated from some CHARMM files earlier, I noticed that the unitcell was orthorhombic even after setting a triclinic box up via the `setBox` function.  Unless I'm misunderstanding, the problem is that the orthorhombic `setUnitCellDimensions` was being called on the underlying topology object so I switched it.  The test case in this PR provides a MWE.

### DIFF
--- a/wrappers/python/openmm/app/charmmpsffile.py
+++ b/wrappers/python/openmm/app/charmmpsffile.py
@@ -786,7 +786,7 @@ class CharmmPsfFile(object):
 
         # Add the periodic box if there is one
         if self.box_vectors is not None:
-            topology.setUnitCellDimensions(self.boxLengths)
+            topology.setPeriodicBoxVectors(self.box_vectors)
 
         return topology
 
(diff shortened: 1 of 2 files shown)


peastman:
Thanks!  This looks good.  I'll merge it once CI finishes running.

swails:
Aaaaand that's because when the CHARMM parsers were written OpenMM didn't support triclinic cells.

ParmEd already does the right thing here.  The sticking point to switching to ParmEd's parsers are that I'm backporting the Drude and lone pair support that were added here so we can use ParmEd's parsers as a replacement for both CHARMM and Amber (since the CHARMM parsers here are really just a now-diverged ancestor of what's in ParmEd).

Much of that work is done (ParmEd/ParmEd#1157), and I'm using the test cases here to validate the changes.

Once complete, I'll cut a new ParmEd release and raise a PR here replacing the Amber and CHARMM implementations with what's in ParmEd (as an optional dependency) per the discussion in #3038

peastman:
That will be great to have too!

PR #3084, opened 2021-03-30
Update openforcefield namespace to new conda-forge namespace in User guide

jchodera:
When the [OpenFF toolkit](http://github.com/openforcefield/openforcefield) migrated from omnia to conda-forge, it changed namespace from `openforcefield` to `openff.toolkit`. The [openmmforcefields README](https://github.com/openmm/openmmforcefields) was updated, but the corresponding section in the [OpenMM User guide](http://docs.openmm.org/latest/userguide/application.html#small-molecule-parameters) was not. This PR corrects that oversight.

cc: https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=161&t=13201&p=37924

### DIFF
--- a/docs-source/usersguide/application.rst
+++ b/docs-source/usersguide/application.rst
@@ -817,14 +817,14 @@ Field Initiative small molecule force fields using the following example:
 
 ::
 
-    # Create an openforcefield Molecule object for benzene from SMILES
-    from openforcefield.topology import Molecule
+    # Create an OpenFF Molecule object for benzene from SMILES
+    from openff.toolkit.topology import Molecule
     molecule = Molecule.from_smiles('c1ccccc1')
-    # Create the SMIRNOFF template generator with the most up to date Open Force Field Initiative force field
+    # Create the SMIRNOFF template generator with the default installed force field (openff-1.0.0)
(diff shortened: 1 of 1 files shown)


peastman:
You changed all the imports back to using the `simtk` prefix again?

jchodera:
Whoops, I copy-pasted from the [openmmforcefields README](https://github.com/openmm/openmmforcefields)!

I've fixed this now, but for 7.5.1, we should back-port this back to the docs for the bugfix with the old `simtk.openmm` namespace.

peastman:
You still have `from simtk import unit`.

jchodera:
Whoops! Fixed?

jchodera:
@peastman Don't forget to undo the namespace changes for OpenMM when you backport into the `7.5_branch` for 7.5.1!

peastman:
In the past we've never bothered building new docs for patch releases, since they don't involve any changes to the API or installation process.

jchodera:
> In the past we've never bothered building new docs for patch releases, since they don't involve any changes to the API or installation process.

Considering we've already had a question about the incorrect documentation, don't you think this merits a change in strategy?

PR #3086, opened 2021-04-05
Added a Github Actions build for Pypy

peastman:


peastman:
@jaimergp any idea why it isn't working?  Following the instructions at https://github.com/marketplace/actions/setup-python, I specified

```yaml
python-version: "pypy-3.7"
```

But it fails with the error

```
ResolvePackageNotFound: 
  - python=pypy-3.7
  - ```

**Comment by jaimergp** [comment]
We are using [`conda-incubator/setup-miniconda`](https://github.com/conda-incubator/setup-miniconda), which has a different args API. You can use the [Miniforge PyPy](https://github.com/conda-forge/miniforge/releases) builds to provide `pypy`, I think. See [this example](https://github.com/conda-incubator/setup-miniconda#example-5-custom-installer).

**Comment by peastman** [comment]
Currently we use the following step to install it:

https://github.com/openmm/openmm/blob/0607d41c18ee1b52efe872b7b951e097f670ad4e/.github/workflows/CI.yml#L145-L151

If I understand you correctly, that won't work for Pypy.  So I need to create a second version of that step that removes `python-version` and replaces it with `installer-url` and the other properties in the example.  And then I should add `if` clauses to both steps so only the correct one gets executed for each build.  Is that right?

**Comment by jaimergp** [comment]
That would be one solution, yes, but I don't think we need to go that far.

I think you can use `installer-url` with no `if`s. The default value is [an empty string](https://github.com/conda-incubator/setup-miniconda/blob/master/action.yml#L31), and if the matrix doesn't specify a key, GHA will render it as an empty string, I believe.

I'll add a couple suggestions to the diffs.

File: .github/workflows/CI.yml [hunk 1]
```diff
@@ -130,6 +130,15 @@ jobs:
               -DOPENMM_BUILD_OPENCL_LIB=OFF \
               -DOPENMM_BUILD_OPENCL_TESTS=OFF \
 
+          - name: Linux CPU Pypy 3.7
+            python-version: "pypy-3.7"
```
jaimergp, line 134:
```suggestion
            python-version: "3.7"
            installer-url: https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge-pypy3-Linux-x86_64.sh
```

jaimergp:
```diff
      - uses: conda-incubator/setup-miniconda@v2
        name: "Prepare base dependencies"
        with:
          python-version: ${{ matrix.python-version }}
          activate-environment: build
          environment-file: devtools/ci/gh-actions/conda-envs/build-${{ matrix.os }}.yml
          auto-activate-base: false
+         installer-url: ${{ matrix.installer-url }}
```

jaimergp:
If that doesn't work, my next attempts would be:

* Option A: edit the environment file to specifies `python_impl=*=pypy` and `python_impl=*=cpython`, as needed. The file name would need to be parameterized in some way to get the correct variant.
* Option B: add an extra step after setup miniconda to replace cpython with pypy (`conda install pypy`) would suffice. This is simpler but it would require one extra conda solve, adding to the CI time. This step would need to be guarded with an `if` clause.

jaimergp:
Looks like `python-version` is upgrading pypy 3.6 to CPython 3.7, undoing the PyPy installation provided by `installer-url`. I don't think pypy 3.7 is already out of beta so...

Things to try in matrix entry for pypy:

* Removing `python-version` and letting the one bundled in `installer-url` be the default ([pypy36](https://github.com/openmm/openmm/pull/3086/checks?check_run_id=2282172907#step:3:131)).
* `python-version: "3.6=*pypy*"` (to be a bit more explicit, but I don't know if this will happily pass through `setup-miniconda`)

jaimergp:
Option A didn't work, since it releases the constraints for Python and `conda` updates happily all the way to 3.9. I guess we need to try the Option B workaround and cross our fingers.

If that doesn't work, we'll need to prepare a new environment file to constrain the Python implementation.

.github/workflows/CI.yml
```diff
@@ -130,6 +130,15 @@ jobs:
               -DOPENMM_BUILD_OPENCL_LIB=OFF \
               -DOPENMM_BUILD_OPENCL_TESTS=OFF \
 
+          - name: Linux CPU Pypy 3
+            installer-url: https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge-pypy3-Linux-x86_64.sh
```
jaimergp, line 134:
```suggestion
            installer-url: https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge-pypy3-Linux-x86_64.sh
            python-version: "3.6=*pypy*"
```

peastman:
That didn't work:

```
  ResolvePackageNotFound: 
    - python==3.6[build=*pypy*]
```

What is different between our script and the setup-miniconda example that causes it to work for them but fail for us?

peastman:
Attempting to answer my own question, is it because we list `python` as a dependency in build-ubuntu-latest.yml?

jaimergp:
Hm, I tried locally and I think we were missing an asterisk. Try with `python-version: "3.6.*=*pypy*"`

jaimergp:
I am afraid we'll need a new environment file then, and no `python-version` hacks (it should just have "3.6").  The environment file should add `- python_abi * *pypy*`. We'll need to add a suffix to the environment filename with `-pypy` or something, that is blank when we use CPython.

peastman:
Let me make sure I have everything correct.  I should do the following:

- Clone build-ubuntu-latest.yml.  The new version (build-ubuntu-latest-pypy.yml?) should be identical except for adding a single line:

```
- python_abi * *pypy*
```

- The matrix entry for this build should specify `python-version: "3.6"`, should *not* include the `installer-url` property, and should use the new environment file.

Is that correct?

jaimergp:
First point: correct. Second point: both things should work. You'll also need a new variable in the Pypy matrix to add the suffix to the env file name.

peastman:
Same result as in https://github.com/openmm/openmm/pull/3086#issuecomment-815136110.  It crashes with an exception deep inside conda.  Is conda somehow incompatible with Pypy?  Do we need it to switch to a different version?

jaimergp:
Hm, I [ran some simplified tests](https://github.com/jaimergp/setup-miniconda-pypy-tests/pull/1) and this block is working:

```yaml
jobs:
  pypy:
    ...
    strategy:
      fail-fast: false
      matrix:
        python-version:
          - "3.6.*=*pypy*"
          - ...
    steps:
      - uses: actions/checkout@v2

      - uses: conda-incubator/setup-miniconda@v2
        with:
          python-version: ${{ matrix.python-version }}
          environment-file: devtools/ci/gh-actions/conda-envs/build-${{ matrix.os }}.yml
```

So, I think it should work if we use:

* `python-version: "3.6.*=*pypy*"`
* No `installer-url` field.
* No separate environment file for the pypy config.

This will use the GHA bundled conda. I think last Miniconda uses a more recent `conda`, which might have a bug with version strings.

peastman:
I think that worked!  Does everything look ok to you?

jaimergp:
I see mentions to pypy in the logs!

* [Here](https://github.com/openmm/openmm/runs/2326978568#step:3:481)
* [And here](https://github.com/openmm/openmm/runs/2326978568#step:12:381)
* [And here](https://github.com/openmm/openmm/runs/2326978568#step:15:41)

So it looks we are in the green!

peastman:
And after all of that, it turned out to be really simple: just specifying the right value for `python-version`.  I'll go ahead and merge this.  Thanks!

PR #3089, opened 2021-04-09
Updated plugin path lookup to use OPENMM_INSTALL_PREFIX.

wilkie:
If one changes `OPENMM_INSTALL_PREFIX` to install in the non-default location, the plugins might silently fail to be loaded. This causes a slowdown when no CPU plugin is loaded. It's a little baffling to discover the reason when this occurs as one expects the plugins to be loaded relatively, but the workaround is to always set `OPENMM_PLUGIN_DIR`. However, that should not be necessary. This patch sets the default lookup to be based on the install prefix.


peastman:
There's no CMake variable called `OPENMM_INSTALL_PREFIX`.  Did you mean to use `CMAKE_INSTALL_PREFIX`?  That's the one that determines where it gets installed.

I have mixed feelings about this.  On one hand, it means that when building locally your plugins will always be found, no matter where you install.  On the other hand, it makes all builds non-portable.  Every compiled version of OpenMM will have a hardcoded reference to wherever it was installed on the computer it was built on.

peastman:
Also note that this is only an issue for programs that use the C++ library directly.  The Python module already stores a reference to where it's installed.

wilkie:
@peastman there is a `OPENMM_INSTALL_PREFIX` and `CMAKE_INSTALL_PREFIX` inherits it. https://github.com/openmm/openmm/blob/master/CMakeLists.txt#L62

I'm OK with it being the more traditional `CMAKE_INSTALL_PREFIX` and, yes, the Python should likely be changed, too.

| it makes all builds non-portable

how so? the default is the same. it's a hard-coded path when built

peastman:
> and, yes, the Python should likely be changed, too.

The Python doesn't need to be changed.  It already records the install location.

> how so? the default is the same. it's a hard-coded path when built

If you build on one computer and install to a non-default location, the binary will use that non-default location on every computer instead of the default one.

wilkie:
> If you build on one computer and install to a non-default location, the binary will use that non-default location on every computer instead of the default one.

Gotcha. But that's the expected behavior of an install prefix. It is already true that the binary will look in the hard-coded path for which it is built. What is the purpose of having a loose non-default install path for the base libraries but a strict plugins path? That feels like a much more unexpected behavior. Either way, you can still override it with the `OPENMM_PLUGINS_DIR` environment variable.

It should probably, at any rate, warn when the plugins path cannot be found. It took me forever to realize what my issue was.

peastman:
> What is the purpose of having a loose non-default install path for the base libraries but a strict plugins path?

It's a good question.  This really relates to the history of how our installation methods have evolved.  We used to have installers that used /usr/local/openmm as the default installation location, though they allowed you to customize it.  Then we moved to installing with conda, which meant libraries got installed inside your miniconda folder, which could be anywhere.  It almost certainly isn't the same location as on the VMs where builds get done.  At that point we made the Python module record the install location so it would be found automatically.

So that leaves people who are using the C++/C/Fortran APIs.  We can further divide those people based on how they get OpenMM.

1. They build it from source.
2. They use the conda package.
3. They get it by installing some other package that bundles OpenMM, like ChimeraX or Tinker-OpenMM.

Your change would make things slightly easier for people in case 1.  On the other hand, it might cause confusion since they might think they could send their program (bundled with OpenMM libraries) to other people, and wouldn't realize it had a nonportable path hardcoded in it.  In case 2 people will have to manually specify a path regardless, since conda packages will be in an unpredictable location.  Case 3 will depend on where that other package gets installed, and whether or not it uses a standard location.

> It should probably, at any rate, warn when the plugins path cannot be found. It took me forever to realize what my issue was.

I like that idea.  `loadPluginsFromDirectory()` could warn you (or perhaps even throw an exception) if you specify a directory that doesn't exist.

Issue #2993, opened 2021-01-25
Wishlist for 7.5.1


jaimergp:
Writing down what I'd need to for the conda-forge packaging:

- #2978
- #2992 
- #2998

jchodera:
@peastman @jaimergp : How about we use the "Milestone" tagging feature to tag the issues listed above for the "7.5.1" release milestone? That will make it very easy to read off what we expect to go into 7.5.1 from the [Milestone dashboard])https://github.com/openmm/openmm/milestones).

jchodera:
It's on the right-hand side of issues and PRs:
![Uploading image.png…]()

peastman:
The above issues can basically be summarized as

- Support ARM based Macs
- Support Pypy
- Support PPC

I'm not too worried about the PPC support.  We've had people using it for a while, and the fix was just to deal with a bug in a particular compiler.  The Pypy issue wasn't even a bug, just making a test less sensitive to a subtle difference in behavior.  It would be good if all of us spent a little while trying out Pypy for our day to day work to see if any real problems turn up.

For ARM based Macs, it would be good if at least one person had spent a while doing regular work on one.  Any update on when your new computer is going to come?

jchodera:
> For ARM based Macs, it would be good if at least one person had spent a while doing regular work on one. Any update on when your new computer is going to come?

Another couple of months, according to MSK.

peastman:
Anyone else on here have an ARM based Mac?

jaimergp:
If it helps, these are current downloads for `osx-arm64` packages. Not too many:

![image](https://user-images.githubusercontent.com/2559438/109388422-29537780-7907-11eb-881f-6b9db789cb0b.png)

jchodera:
@peastman : This may have been missed:
> How about we use the "Milestone" tagging feature to tag the issues listed above for the "7.5.1" release milestone? That will make it very easy to read off what we expect to go into 7.5.1 from the Milestone dashboard.

Can we do this? Is there a good reason not to?

peastman:
It's fine with me.  First we need to make sure there's agreement about what changes should be included in it.

peastman:
Check the [list of changes](https://github.com/openmm/openmm/milestone/2?closed=1) currently marked for 7.5.1.  Do all of those look ok?  Anything else we should add?

jaimergp:
My particular wishlist, as you summed it up before in the thread:

> The above issues can basically be summarized as
> 
[...]

PR #3091, opened 2021-04-12
add docs for conda forge releases

jaimergp:
Addresses https://github.com/openmm/openmm/issues/2993#issuecomment-813472836

### DIFF
--- a/devtools/CONDA_FORGE_RELEASES.md
+++ b/devtools/CONDA_FORGE_RELEASES.md
@@ -2,17 +2,14 @@
 
 ## Final releases
 
-Create a new version tag on `openmm/openmm` and publish a new GitHub release. The Conda Forge bots will auto submit a PR with the new version within a couple hours. Go to step 8 below.
-
-> If the automated PR doesn't happen for any reason (wait 24h just in case), follow the manual instructions below.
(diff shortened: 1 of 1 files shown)


peastman:
Thanks!  I'm making my way through the instructions, trying to build 7.5.1rc2.  I'll post any problems or confusing instructions I run into.

Here are the first two issues I've found.

> Make sure source.url is correct.

There is no `url` property in the `source` block.  Do you mean `git_url`?

> Update source.sha256 with the output of openssl sha256 downloaded_file_from_source_url.tar.gz

There's no `sha256` property either.

jaimergp:
Oh, right, I forgot we are using the full git repo instead of a tarball snapshot. I'll update the docs.

peastman:
The PR is at https://github.com/conda-forge/openmm-feedstock/pull/52.  Let me know if it looks ok!

jaimergp:
Looks ok! We need to fix a couple of unrelated issues that popped up there, but looks like the checklist is working!

peastman:
Ok, I'll merge this then.  Thanks!

peastman:
> Once merged, check the CI status on the rc branch.

I'm not sure what this instruction means.  Where do I go to check the CI status of a branch?

jaimergp:
In the feedstock, switch to the rc branch:

image


Then, click on the status icon for the last commit (it should be green):

image

peastman:
Thanks!  The status is green, so I think that means we have a release candidate.  I'll post the announcement.

Issue #2892, opened 2020-10-20
Metadynamics walkers failure


dlukauskis:
I'm testing out the metadynamics walkers implementation and I'm seeing odd behaviour. When walker simulations are running simultaneously, the deposited biases are huge, in the order of 10^6 kJ/mol, even just after a few ns. Very soon all but one of the simulation crash. 

What's the logic of sharing biases between simulations? Does each simultation just add the total bias from other sims? Because that's what it looks like. It makes more sense for each sim to find the difference between it own deposited bias and other bias files and add that difference to its own bias grid. Or just use one bias file that each walker will read and write to - that's what PLUMED WALKERS_MPI does.

Attaching a [directory with input files](https://github.com/openmm/openmm/files/5407180/walkers_test.zip) to replicate what I'm seeing.

peastman:
I'm cleaning up old issues and see this never got answered.

In your case it sounds like you probably want to make them save/load biases more frequently, so each one won't accumulate so much bias before discovering the others have been exploring exactly the same area.

Using multiple walkers is most useful when they're all exploring different regions.  With only a single process, the bias is based on everything it has explored up to that point.  With multiple processes, that becomes everything it *or any other process* has explored up to that point.

PR #3092, opened 2021-04-14
Changes to metadynamics

dlukauskis:
This adds the ability to get hill height at any point, which allows the analysis step of metaD sims to be done using PLUMED. Also, it addresses the issue I had with walkers sharing the bias, see issue #2892 .

### DIFF
--- a/wrappers/python/openmm/app/metadynamics.py
+++ b/wrappers/python/openmm/app/metadynamics.py
@@ -192,6 +192,12 @@ class Metadynamics(object):
         """Get the current values of all collective variables in a Simulation."""
         return self._force.getCollectiveVariableValues(simulation.context)
 
+    def getHillHeight(self, simulation):
+        """Get the current height of the Gaussian hill in kJ/mol"""
+        energy = simulation.context.getState(getEnergy=True, groups={31}).getPotentialEnergy()
+        currentHillHeight = self.height*np.exp(-energy/(unit.MOLAR_GAS_CONSTANT_R*self._deltaT))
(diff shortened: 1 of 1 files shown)


wrappers/python/openmm/app/metadynamics.py
```diff
@@ -192,6 +192,12 @@ def getCollectiveVariables(self, simulation):
         """Get the current values of all collective variables in a Simulation."""
         return self._force.getCollectiveVariableValues(simulation.context)
 
+    def getHillHeight(self, simulation):
+        """Get the current height of the Gaussian hill in kJ/mol"""
+        energy = simulation.context.getState(getEnergy=True, groups={31}).getPotentialEnergy()
```
raimis, line 197: The force group should not be hard-coded:
```python
forceGroup = self._force.getForceGroup()
energy = simulation.context.getState(getEnergy=True, groups={forceGroup}).getPotentialEnergy()
```
There was already a similar bug (https://github.com/openmm/openmm/pull/3060).

raimis (changes requested):

wrappers/python/openmm/app/metadynamics.py
dlukauskis: Yeah, that's a good catch.

raimis:
@dlukauskis are you still working on this?

dlukauskis:
I haven't made any more changes to the code, if that's what you're asking.

raimis:
Is https://github.com/openmm/openmm/issues/3105 relevant to your changes?

PR #3112, opened 2021-05-11
Remove hard-coded default path to nvcc

jlmaccal:
The current version of the code hardcodes the default value of `CudaCompiler` to `/usr/local/cuda/bin/nvcc`, but it is commonly not installed at that location on HPC systems. This means that nvcc compilation cannot be forced by simply setting `CudaCompiler` to the value from `getPropertyDefaultValue("CudaCompiler")`. Instead, the user must explicitly set `OPENMM_CUDA_COMPILER` which adds undesirable complexity.

This changes the default value of `CudaCompiler` to simply be `nvcc` rather than a hardcoded path. This should use whatever version of nvcc is in the user's path. This will make it easier to force the compilation of a kernel with nvcc rather than nvrtc without requiring the user to set `OPENMM_CUDA_COMPILER`.

### DIFF
--- a/docs-source/usersguide/library.rst
+++ b/docs-source/usersguide/library.rst
@@ -1908,7 +1908,7 @@ The CUDA Platform recognizes the following Platform-specific properties:
 
   * Mac/Linux: It first looks for an environment variable called
     OPENMM_CUDA_COMPILER.  If that is set, its value is used.  Otherwise, the
-    default location is set to /usr/local/cuda/bin/nvcc.
+    default location is set to nvcc.
   * Windows: It looks for an environment variable called CUDA_BIN_PATH, then
     appends \nvcc.exe to it.  That environment variable is set by the CUDA
     installer, so it usually is present.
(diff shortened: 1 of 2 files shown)


peastman:
Thanks!  I'm trying to work through all the possible scenarios to see how this will affect each of them.

1. The CUDA toolkit was installed with conda.  That version of the toolkit doesn't include nvcc.  So in that case, this would neither help nor hurt.

2. They used an installer downloaded from NVIDIA.  Those installers by default create a link at `/usr/local/cuda`, but do not add it to the `PATH`.  In that case, this change would break people's installs.

3. The toolkit was preinstalled on a cluster.  In that case, it could be in any location and might or might not have been added to the `PATH`.  Depending on where the toolkit is installed and whether it's added to the `PATH`, this could help, hurt, or do neither.

All of the above assumes that for some reason the runtime compiler isn't available.  If it is, that will get used and the default path to nvcc is irrelevant.  It will only be used if the user explicitly specifies a path for it.

jlmaccal:
Ah… I didn’t realize point 2. I assumed nvcc would always be in the user’s path.

For what it’s worth, I think the best solution for my specific issue is to extend `createModule` with another overload that allows for arrays of header file names and their contents to be passed, which would then be passed to rtc.

peastman:
We definitely want to come up with some solution for `#include` files with nvrtc.  I'm just not sure yet what it will be.  In any case, this PR wouldn't have helped with that problem.  It would still default to nvrtc unless you explicitly specified a path to the compiler.

I'm going to close this, since it would create problems for some users and not fix the main problem it was meant to tfix.

PR #3143, opened 2021-06-10
Support of Amber FF19SB CMAP

kyw220:
Dear,

We have modified “amber_file_parser.py” to handle ff19SB properly by adding CMAP torsions. The previous version ignored CMAP information in Amber parm7 file.

The attached file below is a comparison of the CMAP energy of the PDB ID: 6O0Z system calculated via Amber and OpenMM. We prepared parm7 and rst7 files for simulations using CHARMM-GUI (http://charmm-gui.org/).

[CMAP_energy-Amber-OpenMM.txt](https://github.com/openmm/openmm/files/6634512/CMAP_energy-Amber-OpenMM.txt)

Thank you.


peastman:
Looks good as far as I can tell.  @swails if you have time to look this over, that would be great.

Could you add a test case?  The easiest way is to create a small system (just a few residues is probably enough) and add the input files to `wrappers/python/tests/systems`.  Then you can add a method to TestAmberPrmtopFile.py whose name begins with "test".  It should load the files, compute the energy, and see if it matches the value computed by Amber.

Could you include docstrings for the new methods you've added?

wrappers/python/openmm/app/internal/amber_file_parser.py
```diff
@@ -461,6 +459,67 @@ def getDihedrals(self):
...
+    def getNumMaps(self):
+        """ """
+        try:
+            return self._numCMAP
+        except AttributeError:
+            pass
+        if "CMAP_COUNT" in self._raw_data.keys():
+            self._numCMAP=int(self._raw_data["CMAP_COUNT"][1])
+            return self._numCMAP
+        return False
```
peastman, line 471: Wouldn't it make more sense to return 0 in this case?

wrappers/python/openmm/app/internal/amber_file_parser.py
```diff
@@ -461,6 +459,67 @@ def getDihedrals(self):
...
+    def getCMAPResolutions(self):
+        """ """
+        try:
+            return self._cmapResolution
+        except AttributeError:
+            pass
+        if "CMAP_RESOLUTION" in self._raw_data.keys():
+            self._cmapResolution=self._raw_data["CMAP_RESOLUTION"]
+            return self._cmapResolution
+        return False
```
peastman, line 482: Likewise an empty list would make sense here.

wrappers/python/openmm/app/internal/amber_file_parser.py
```diff
@@ -742,6 +801,34 @@ def readAmberSystem(topology, prmtop_filename=None, prmtop_loader=None, shake=No
         force.addTorsion(iAtom, jAtom, kAtom, lAtom, periodicity, phase, forceConstant)
     system.addForce(force)
 
+    # Add CMAP info.
+    ## Get mapSize and Resolutions
+    numMap = prmtop.getNumMaps()
+    mapSize = prmtop.getCMAPResolutions()
+
+    if numMap and mapSize and numMap == len(mapSize):
```
peastman, line 809: This would become

```python
if numMap > 0 and numMap == len(mapSize):
```

However, does `numMap != len(mapSize)` indicate an error in the file?  If so, we should probably detect that case and throw an exception.

wrappers/python/openmm/app/internal/amber_file_parser.py
kyw220: Returning 0 makes sense.

wrappers/python/openmm/app/internal/amber_file_parser.py
kyw220: Also, returning 0 makes sense.

wrappers/python/openmm/app/internal/amber_file_parser.py
kyw220: At first, I inserted the part for detecting the CMAP parameters in parm7 file, and just tried to add a simple error checking. However, on second thought, there will be no error as far as parm7 files are created by CHARMM-GUI or leaf program. Thus, it would become 
```suggestion
    if numMap > 0:
```

swails: No, just push it to your branch and it automatically updates your pull request.

wrappers/python/openmm/app/internal/amber_file_parser.py
```diff
@@ -461,6 +459,67 @@ def getDihedrals(self):
...
+        try:
+            return self._cmapList
+        except AttributeError:
+            pass
+        cmapPointers = self._raw_data["CMAP_INDEX"]
+        self._cmapList=[]
+        forceConstConversionFactor = (units.kilocalorie_per_mole).conversion_factor_to(units.kilojoule_per_mole)
+        for ii in range(0,len(cmapPointers),6):
+            if any([int(cmapPointers[ii+jj])<0 for jj in range(5)]):
+                raise Exception("Found negative cmap atom pointers %s"
```
swails, line 505:
```suggestion
                raise ValueError("Found negative cmap atom pointers %s"
```

wrappers/python/openmm/app/internal/amber_file_parser.py
```diff
@@ -461,6 +459,67 @@ def getDihedrals(self):
...
+
+    def getCMAPParameters(self, index):
+        num_str=str(index)
+        zero_index = num_str.zfill(2)
+        name="CMAP_PARAMETER_" + zero_index
+        cmapPointers = self._raw_data[name]
+        _cmapParameterList=[]
+        for ii in range(0,len(cmapPointers)):
+            _cmapParameterList.append(float(cmapPointers[ii]))
+        return _cmapParameterList
```
swails, line 492:
```suggestion
    def getCMAPParameters(self, index):
        flag = "CMAP_PARAMETER_{:02d}".format(index)
        return [float(pointer) for pointer in self._raw_data[flag]]
```

swails (commented):
Looks good (as long as it reproduces identical results to Amber/ParmEd, and as @peastman a unit test would be critical to ensuring this capability remains unbroken).

peastman:
I see you checked in some changes.  Is this ready for another review?

kyw220:
> I see you checked in some changes. Is this ready for another review?

Yes, please review it. Thank you.

peastman:
This looks good.  Let's wait for the tests to finish running, and then I think we can merge it.

PR #3147, opened 2021-06-14
fixing a small bug in langevin integrator delta

samarjeet:


jchodera:
Oh, my, this appears to be a critical bug. Thanks so much for spotting this!

@peastman: Does this appear in any released code, or just the development branch? And does it impact integrators, or just `LangevinIntegrator`?

peastman:
Good catch!

This only affects LangevinIntegrator and VariableLangevinIntegrator.  The corresponding code for VerletIntegrator is correct, and LangevinMiddleIntegrator doesn't have corresponding code.  This code is used on low end GPUs that don't support double precision, to try to get slightly better than single precision accuracy in the result.  That's generally not important in Langevin integration, since the correction is tiny compared to the random force.  That's probably why this was never noticed.  For Verlet integration, it significantly improves the energy conservation.

jchodera:
@peastman: Can you write a brief description (based on the above) we can include in the next release that this bugfix is included within? Sounds like this will be in the next major release if it's not critical.

peastman:
The docs build failed, but that's for unrelated reasons that we already know about.  Everything else passed, so I'll merge this.

PR #3161, opened 2021-06-29
Allow custom template matchers for residues

peastman:
This implements a new feature described in https://github.com/openmm/openmmforcefields/pull/156#issuecomment-870799618.  It allows overriding the standard logic for matching templates to residues.


peastman:
I'm going to merge this so it doesn't block @zhang-ivy from working on GLYCAM.  If anyone has comments, we can still make changes.

Issue #3134, opened 2021-06-01
Unify the documentation build process


Yoshanuikabundi:
The user's guide would be greatly improved by including links to the API reference wherever a class or object is described. This is currently somewhere between impractical and impossible, as the two exist in parallel Sphinx trees and are processed independently. Unifying them would allow linking between them, simplify building the docs, unify the themes, and allow moving the top-level documentation page into Sphinx. It might also facilitate moving the documentation to a service like ReadTheDocs in the future.

Issue #3132, opened 2021-06-01
Improvements to API reference search


Yoshanuikabundi:
At the moment, the [API reference](http://docs.openmm.org/latest/api-c++/library.html)'s search function is a bit rudimentary. I would like to make the following improvements:

1. Pressing enter after typing a query should bring up a page of search results. While the drop down that currently appears while typing is neat and not should not be removed, it is not sufficient because:
    - Results are often obscured by the browser's suggestions popup: 
 ![search_obscured](https://user-images.githubusercontent.com/28590748/120283185-42046c80-c2fe-11eb-8930-a20b3c2d8d70.png)
    - Most search engines use a drop down for suggestions and results appear after pressing enter. Users expecting this behaviour may be confused (I thought the search function was broken for a long time)
    - The context provided is insufficient to actually choose a search result
2. Fuzzy searching should be done by default, so that searching "force" displays entries that end in force, rather than only those that begin with it. Ditto "thermostat", "nonbonded", etc. This is essential for discoverability.
3. If possible, the drop down results should display more context. Typing "force" and getting the results "Force", "Force", "~Force", "force", and "force" is not very helpful. Providing the full paths/namespaces of the results would be beneficial.

PR #3165, opened 2021-07-02
Reorganise documentation

### LINKED ISSUE #3133: Improve navigation of User's and Developer's Guides
At the moment, the User's and Developer's Guides both have intended layouts that do not match their layouts in source or on the web page. The intended layout as far as I can tell is:

- Documentation
    - User's Guide
        - Introduction
        - Part I: Application layer
             - Chapters 2-6
        - Part II:  Library layer
            - Chapters 7-16
        - Part III: Theory
            - Chapters 17-22
        - Bibliography
    - Dev's Guide
        - Chapters 1-8
    - Python API ref
    - C++ API ref

This matches the layout of files on disk, with single files associated with each part of the user's guide. It does not match the semantics of the source or the built documentation itself, which looks more like this: 

- User's Guide
    - Chapters 1-23
- Dev's Guide
    - Chapters 1-8
- Python API ref
- C++ API ref

Note that the chapters in the User's Guide are flat, and are not delineated into parts. As Sphinx builds the table of contents and navigation by looking at the semantics of the source, but builds the actual HTML files based on the files in the source, this makes navigation very confusing as the user is unsure whether clicking a link will take them to a new page or not.

In addition, pages are extremely long, as entire parts of 5+ chapters are compiled into single web pages. This makes navigation frustrating as the user must scroll to the top of the page to find the navigation menu on the left hand side.

The following changes to the source would very quickly and easily make a massive improvement to the navigability of the User's Guide in particular, but also the developer's guide:

 - Insert new section levels corresponding to parts. This will clearly separate out which chapters belong to which parts and make it easier for the user to find themselves in the navigation tree.
 - Break the source files up so that each chapter appears in its own source file, and therefore its own web page. This will help users to find the navigation menu with minimum scrolling.

The following changes to the template would facilitate the above:

- "Next Chapter" and "Previous Chapter" links should be added to the template at the top and bottom of every page.
- Parts should have a distinct visual design in the navigation menu.
- The navigation menu on the left should not go deeper than the chapter level, except for the current chapter where it should go to the subsection level (this is actually the current behaviour, but it will be much more usable with pages-as-chapters)
- A search function should be added to make up for the loss of ctrl+F (which is already compromised by breaking the pages up into parts)

Yoshanuikabundi:
This PR:
 - Breaks the user and developer guides down so that each chapter is on its own page. This makes navigation much simpler and easier, and also makes the layout of especially the user guide much clearer (resolves #3133)
 - Adds prev/next page links to the bottom of each page of the dev and user guide to facilitate reading through the manuals (ibid.)
 - Unifies and simplifies styling across the dev guide, user guide, and API references. (mitigates #3134)
 - Adds a very basic `.gitignore` file to support out-of-tree builds in the directories `build` and `build0` through `build9` (Is there a reason this wasn't here before? How did people cope!?)
 - Fixes a bug in autonumber.py that caused autonumbers in a file's top level section to fail, and otherwise updates autonumber.py to work with the new documentation layout
 - Prevents the browser from trying to suggest search items with lunrsearch, which previously obscured the results (mitigates #3132)
 - Formats all four `conf.py` files with isort and black
 - Fixes typos as I've found them


Yoshanuikabundi:
AFAICT, the "broken link" is http://www.ks.uiuc.edu/Research/vmd/ (which I did not add). It's loading very slowly for me, so it must have timed out or had an intermittent error.

peastman:
Thanks!  The may take a *little* while to review (52 files changed). :)

peastman:
> (Is there a reason this wasn't here before? How did people cope!?)

I'm not sure what you mean?  Documentation is built entirely in the build directory, outside the source tree.

peastman:
I'm making my way through the user guide.  So far it's looking really good!  I'll post comments here as I come across things.  So far just a few very minor ones.

I'm not sure about the nested tables of contents.  The main TOC only shows two levels, which means no details finer than what were chapters in the previous organization.  So I click on the link for "The OpenMM Application Layer" and get... another table of contents!  What about showing the third level directly in the main TOC, and getting rid of the secondary ones?

The captions under examples (e.g. "Example 2.2-1" in section 2.2.1) should be centered.

It has switched to referencing chapters and sections by name instead of number, which leads to some awkward language.  For example, section 2.2.1 (previously 3.1) used to contain the sentence, "More information on how to add missing atoms and residues using OpenMM tools can be found in Chapter 4."  Now that sentence has changed to, "More information on how to add missing atoms and residues using OpenMM tools can be found in Chapter Model Building and Editing."

peastman:
I'm not sold on the way it's been grouped into just five "chapters": Introduction, Application Layer, Library, Theory, and Bibliography.  This creates a few issues.

- It leads to section numbers with a kind of excessive number of parts.  Even relatively major sections like "Force Fields" or "Integrators" now have four part section numbers.  And the truly minor sections are up to six!  For example, "2.2.6.5.1.2. Specifying the Polarization Method".

- It doesn't reflect the actual organization of the manual.  The introduction, which is only a single page, is put on a level with the three major sections.  Whereas "Running Simulations", an important chapter that's about 20 pages long, has been demoted a subsection.

Yoshanuikabundi:
> > (Is there a reason this wasn't here before? How did people cope!?)
> 
> I'm not sure what you mean? Documentation is built entirely in the build directory, outside the source tree.

I just mean without a .gitignore, you can't use patterns like `git add ..`, because you have to explicitly exclude the build directory.

> what were chapters in the previous organization

I'm still thinking of those as chapters. I'm thinking of the higher levels as parts. So we have an introduction as frontmatter, then three parts that each have chapters, sections etc, then a backmatter bibliography. I don't *think* Sphinx ever interprets any level as a chapter or section or whatever (it certainly [claims not to](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections))

> I click on the link for "The OpenMM Application Layer" and get... another table of contents! What about showing the third level directly in the main TOC, and getting rid of the secondary ones?

I've added the third level to the main TOC and added styling to make it easier to process visually in 7aa838f. I think this styling makes up for the added complexity, and the extra information is definitely a win, so I'm happy with this.

I don't think Sphinx will let us get rid of the secondary TOC. I also don't think its necessary or even particularly desirable- clicking on the chapter titles already sends the reader to the right place, and they're useful for navigating individual parts.

> The captions under examples (e.g. "Example 2.2-1" in section 2.2.1) should be centered.

Fixed in 6e40c39. Thanks!

> It has switched to referencing chapters and sections by name instead of number, which leads to some awkward language.

This is the current behavior [on master](http://docs.openmm.org/development/userguide/application.html#a-first-example), so I'm guessing the update to Sphinx 4 silently broke the `numsec.py` plugin. But I'm happy to fix it in this PR (or in a future one)! I think the best pattern would be to use the [`numref`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-numref) role when we want section numbers, rather than repurpose the `ref` role. That also lets us get rid of the `numsec.py` plugin.

> I'm not sold on the way it's been grouped into just five "chapters": Introduction, Application Layer, Library, Theory, and Bibliography. This creates a few issues.
> 
>     * It leads to section numbers with a kind of excessive number of parts.  Even relatively major sections like "Force Fields" or "Integrators" now have four part section numbers.  And the truly minor sections are up to six!  For example, "2.2.6.5.1.2. Specifying the Polarization Method".
> 
>     * It doesn't reflect the actual organization of the manual.  The introduction, which is only a single page, is put on a level with the three major sections.  Whereas "Running Simulations", an important chapter that's about 20 pages long, has been demoted a subsection.

The documentation was already organised into these headings, it just wasn't reflected in the TOC tree. The part headings (Application layer, Library, Theory) are important for organising the wealth of information in the user manual. So I don't think we should get rid of them as a layer. If you have a proposition to break one or more of them up into further parts, I'd be interested to hear it.

However, I completely agree that six part section numbers are way too much. We should aim never to go deeper than a subsection of a chapter (four levels - part.chapter.section.subsection). I also agree that the importance of the information in "Running Simulations" is not reflected in its placement, and I note that this chapter is the only place where the section depth gets this bad.

I think the solution is to break "Running Simulations" into multiple chapters. It's already the longest chapter by a good margin, as well as being one of the most important. I think promoting the "Simulation parameters" section to a chapter would be a good start.

peastman:
> I've added the third level to the main TOC and added styling to make it easier to process visually in 7aa838f.

Looks good!

> I think the best pattern would be to use the numref role when we want section numbers

Sounds good.  I think that role may not have existed yet when we first wrote it.

> I think promoting the "Simulation parameters" section to a chapter would be a good start.

That would be reasonable.

> So I've got a branch where the parts themselves are not numbered, but all the chapters are, including the introduction and bibliography.

I like that!  It does a nice job of showing the high level grouping without obscuring the organization.

peastman:
CI is reporting a bunch of errors like this in the docs build:

```
  npm ERR! code E403
  npm ERR! 403 403 Forbidden - GET https://registry.npmjs.org/normalize-package-data
  npm ERR! 403 In most cases, you or one of your dependencies are requesting
  npm ERR! 403 a package version that is forbidden by your security policy.
  npm ERR! 403 
  npm ERR! 403 It was specified as a dependency of 'meow'
  npm ERR! 403 
  
  npm ERR! A complete log of this run can be found in:
  npm ERR!     /home/runner/.npm/_logs/2021-07-06T09_59_51_471Z-debug.log
  Install for [ 'linkinator@latest' ] failed with code 1
```

I'm not sure what those mean.

peastman:
The CI errors seem to have gone away on their own.  Probably something intermittent.

I looked into using relative URLs for the links in the API docs sidebar like we discussed.  For example, change

```
        "C++ API reference": "http://docs.openmm.org/latest/api-c++/",
```

to

```
        "C++ API reference": "../api-c++/index.html",
```

Note that "index.html" is needed when loading local files, since it won't automatically redirect to it.  On either the top or secondary levels (index.html, app.html, library.html), it works fine.  When I click down to a specific class, though, it fails since there's now another level of directory in the URL.  The most obvious solution would be for the sidebar to be a separate frame.  I don't know if that's hard to manage, but it would also be a better UI, since you don't really want the sidebar to scroll when you scroll the page.

Yoshanuikabundi:
Separate frame as in an <iframe> is a no-go for a bunch of reasons, but I had a brain wave overnight and implemented both of those features in CSS and a bit of templating. Sidebar will remain fixed on the left when you scroll the document (but is still scrollable itself as needed), and the relative links will always be relative to the root directory of the sphinx project. Note that the `html` directory is deployed as userguide or developerguide etc, so I'm pretty sure the current links are correct, even though they're broken on local builds. Hopefully the link checker in CI will bear that out.

Yoshanuikabundi:
Hmmm... The 404 check failed because of the relative links, so I added another level of `..`, and the 404 check looked at the same place again. Locally the link changes, but this change seems to not be reflected in the check. I think it might be a bug in the 404 checker. Tomorrow I'll try and run the workflow locally so I can inspect the output and see if I can debug but in the meantime I've reverted the second level of `..`.

peastman:
If you want I can go ahead and merge this.  The relative links are a minor point.  We don't need to hold up the rest of the PR for them.  We just need to make sure that one way or another, the links work by the time we release 7.6.

Yoshanuikabundi:
Yeah that makes sense I think. Thanks!

peastman:
Ok, great.  We can clean up the sidebar in a separate PR.

PR #3167, opened 2021-07-03
Support initialization scripts in force fields

peastman:
This allows force fields to include an `<InitializationScript>` tag that gets executed when the file is loaded.  This is needed to let GLYCAM be fully self contained.  See https://github.com/openmm/openmmforcefields/pull/156#issuecomment-870799618 for details.

For the moment, I'm leaving this as an undocumented feature and treating it as part of the internal implementation of GLYCAM.  That way we can change it in the future without breaking users' code.  If we decide later we want to make it a fully supported feature, we can document it then.

For GLYCAM, add the following tag to the XML file:

```xml
  <InitializationScript>
class GlycamTemplateMatcher(object):
  def __init__(self, glycam_residues):
    self.glycam_residues = glycam_residues
  def __call__(self, ff, residue):
    if residue.name in self.glycam_residues:
      return ff._templates[residue.name]
    return None

glycam_residues = set()
for residue in tree.getroot().find('Residues').findall('Residue'):
  glycam_residues.add(residue.get('name'))
self.registerTemplateMatcher(GlycamTemplateMatcher(glycam_residues))
  </InitializationScript>
```

cc @zhang-ivy

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -330,7 +330,7 @@ class ForceField(object):
                             self.registerTemplatePatch(resName, patchName, 0)
                     self.registerResidueTemplate(template)
 
-        # Load the patch defintions.
+        # Load the patch definitions.
 
         for tree in trees:
             if tree.getroot().find('Patches') is not None:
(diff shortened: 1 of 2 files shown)


zhang-ivy:
@peastman Just to clarify, the purpose of this feature is so that users wanting to parametrize glycosylated systems do not have to run the template matcher code each time they want to parametrize, correct? 

If so, after generating the converted glycam ffxml, I should add the above XML code to the ffxml, after which anyone who uses that ffxml can parametrize the system with the standard parametrization code (see below), correct?
```python
# Load the ffxml
ffxml = ['../ffxml/protein.ff14SB.xml', "../ffxml/GLYCAM_06j-1_trunc_modified.xml"]
ff = app.ForceField(*ffxml)
system_omm = ff.createSystem(topology)
```

peastman:
Exactly.  It will call `registerTemplateMatcher()` automatically so the user doesn't have to.

PR #3184, opened 2021-07-21
Deleted the default copy-constuctor for ContextImpl.

Dann239:
Whenever a `ContextImpl` object is destroyed, it calls `contextDestroyed` which deletes the `platformData`, thus corrupting all copies of the object that might have been created, because all the `platformData` pointers in copies are now dangling. The same happens to pointers in `forceImpls`. Therefore, it should be impossible to copy `ContextImpl`.

I got bitten by this after accidentally passing a `ContextImpl` object by value into a function, which deleted the `platformData` after it returned. It wasn't trivial to track down, hence the motivation to propose this change.

### DIFF
--- a/openmmapi/include/openmm/internal/ContextImpl.h
+++ b/openmmapi/include/openmm/internal/ContextImpl.h
@@ -57,6 +57,7 @@ public:
      */
     ContextImpl(Context& owner, const System& system, Integrator& integrator, Platform* platform, const std::map<std::string, std::string>& properties,
             ContextImpl* originalContext=NULL);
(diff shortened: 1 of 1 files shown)


peastman:
Looks good.  Do you want to also block the copy assignment operator?

```c++
ContextImpl& operator=(const ContextImpl&) = delete;
```

Dann239:
Sure, why not. Although it's highly unlikely that someone would try to copy-assign a `ContextImpl` object (either intentionally or accidentally), it wouldn't hurt to add it. I pushed.

peastman:
Thanks!

PR #3186, opened 2021-07-22
Update instructions on compiling from source

peastman:
This is a major reorganization and rewrite of the chapter on compiling from source.  It hopefully should now be simpler and easier to follow, as well as being accurate and up to date.

### DIFF
--- a/docs-source/usersguide/library/02_compiling.rst
+++ b/docs-source/usersguide/library/02_compiling.rst
@@ -3,281 +3,134 @@
 Compiling OpenMM from Source Code
 #################################
 
-This chapter describes the procedure for building and installing OpenMM
-libraries from source code.  It is recommended that you use binary OpenMM
-libraries, if possible.  If there are not suitable binary libraries for your
-system, consider building OpenMM from source code by following these
-instructions.
+This chapter describes the procedure for building and installing OpenMM from
+source code.  In most cases, it is best to use the pre-built versions installed
+with conda.  Sometimes you might need to build from source, though, such as if
+you want to modify OpenMM, or if conda does not provide packages compatible with
+your environment.
+
+We first describe how to build on Linux or Mac.  We then describe how to build
+on Windows, where the process is slightly different.
+
+Compiling on Linux and Mac
+**************************
 
 Prerequisites
-*************
-
-Before building OpenMM from source, you will need the following:
-
-* A C++ compiler
-
-* CMake
-
-* OpenMM source code
-
-
-See the sections below for specific instructions for the different platforms.
-
-Get a C++ compiler
-==================
-
-You must have a C++ compiler installed before attempting to build OpenMM from
-source.
-
-Mac and Linux: clang or gcc
----------------------------
-
-Use clang or gcc on Mac/Linux.  OpenMM should compile correctly with all recent
-versions of these compilers.  We recommend clang since it produces faster code,
-especially when using the CPU platform.  If you do not already have a compiler
-installed, you will need to download and install it.  On Mac OS X, this means
-downloading the Xcode Tools from the App Store.
-
-Windows: Visual Studio
-----------------------
-
-On Windows systems, use the C++ compiler in Visual Studio 2015 or later.  You
-can download a free version of the Visual Studio C++ build tools from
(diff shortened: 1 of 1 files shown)


jchodera:
Should we ask @Yoshanuikabundi to review?

peastman:
I'm certainly hoping he'll have comments on it!

jchodera:
You can request his review by using this gear icon!
image

docs-source/usersguide/library/02_compiling.rst
```diff
@@ -325,69 +325,50 @@ You should see a series of test results like this:
...
-================================
+The documentation that you're currently reading, as well as the Developer Guide and API
+documentation, can be built through CMake.  To do that, you need to install a few
+additional tools.  The easiest way is to use :code:`pip` to install them into
+your Python environment.  The following command installs everything needed to
+build documentation in HTML format.
+::
 
-The following dependencies are required to build the Python and C++ API documentation.
+    pip install sphinx sphinxcontrib-bibtex sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen jinja2
```
Yoshanuikabundi, line 347: I'm not sure how I feel about suggesting users install things with pip that can be installed with conda. It's probably not the end of the world but we should at least make a note somewhere to change this to conda once the pip-only deps are no longer required (which I'm working on now), and I would be in favour of having two lines here so that the dependencies available on conda-forge are installed with it.

docs-source/usersguide/library/02_compiling.rst
```diff
@@ -325,69 +325,50 @@ You should see a series of test results like this:
...
 
-These dependencies may not be available in your system package manager, but should
-be installable through Python's ``pip`` package manager. ::
+* :code:`sphinxpdf`: Build the User Guide and Developer Guide in PDF format.
 
-   pip install sphinx sphinxcontrib-lunrsearch sphinxcontrib-autodoc_doxygen
+* :code:`C++ApiDocs`: Build the C++ API documentation.
 
-To build the C++ API documentation, type: ::
+* :code:`PythonApiDocs`: Build the Python API documentation.
```
Yoshanuikabundi, line 363: These instructions work for me for all targets except `PythonApiDocs`, which needs the `install` target to have been run previously. We should either fix this dependency or document it here.

```suggestion
* :code:`PythonApiDocs`: Build the Python API documentation. This requires that ``make install`` has already been run.
```

Yoshanuikabundi (commented):
I've variously tested different parts of these instructions on VPSs running Debian, Ubuntu, Arch and Windows. They're a big improvement! These instructions are clear and complete, and they look correct and clear when built. I got a few test failures here and there but I'm happy to chalk them up to weirdnesses that would go away on rebuild. I've made some notes on small items we could improve on.

The big problem I have is that it is still impossible to build the `PythonApiDocs` target without previously running `make install`. The other documentation targets work. (this is a clarification from previously; I didn't realise `install` was just a regular target. It's the install target that is the missing dependency) I tried it again following these instructions given our last discussion about it, but even with the system C++ compiler it doesn't work on Arch or Ubuntu. I have tried this many times over the last few months and have never seen the desired behavior. The reason seems clear from my understanding of cmake: Python looks for the openmm library in the installed location, not in the build directory, and `make install` puts it there. `PythonInstall`, upon which `PythonApiDocs` depends, does not. I think needing to install the library in order to compile the docs or work on the Python library is definitely an anti-pattern, but if we're going to stick with it we definitely need to document it.

peastman:
> The big problem I have is that it is still impossible to build the PythonApiDocs target without previously running make install.

I'll see if I can get that working.  It shouldn't be hard.

docs-source/usersguide/library/02_compiling.rst
peastman: I'll be happy to switch the instructions to using conda as soon as the packages are available on conda! :)

peastman:
Apparently there's no way to make a custom target (like `PythonApiDocs`) depend on a built in target (like `install`).  There's an open feature request for it.  https://gitlab.kitware.com/cmake/cmake/-/issues/8438

peastman:
I made the changes.  Thanks!

PR #3188, opened 2021-07-23
Accept relative links between documentation sites

Yoshanuikabundi:
Previously, the 404 checker run on the docs in CI would fail when a document has a relative link to a page in one of the other documentation sites. This is because linkinator assumed the web server was running from each site's root directory, so relative paths further up the file hierarchy were errors. This PR fixes this behaviour and should hopefully fix the documentation check failures.

The failure is confirmed spurious by the fact that the sidebar links work correctly on the [development documentation](http://docs.openmm.org/development/userguide/)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -613,7 +613,7 @@ jobs:
           set +e
           for d in api-c++ api-python developerguide userguide; do
             echo "::group:: Check ${d}"
-            npx linkinator ${d}/ --recurse --timeout=5000 --skip "$SKIP_URL" --server-root ./build/api-docs/
+            npx linkinator ./${d}/ --recurse --timeout=5000 --skip "$SKIP_URL" --server-root ./build/api-docs
             ((exitcode+=$?))
             echo "::endgroup::"
           done
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -613,7 +613,7 @@ jobs:
           set +e
           for d in api-c++ api-python developerguide userguide; do
             echo "::group:: Check ${d}"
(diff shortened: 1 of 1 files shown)


peastman:
The docs build is failing:

```
  Error: The provided glob "./build/api-docs/api-c++/" returned 0 results. The current working directory is "/home/runner/work/openmm/openmm".
      at Object.processOptions (/home/runner/.npm/_npx/8829/lib/node_modules/linkinator/build/src/options.js:59:23)
      at async LinkChecker.check (/home/runner/.npm/_npx/8829/lib/node_modules/linkinator/build/src/index.js:32:25)
      at async main (/home/runner/.npm/_npx/8829/lib/node_modules/linkinator/build/src/cli.js:162:20)
```

I tried rerunning it to see if it was just an intermittent problem, but it still failed.

peastman:
Thanks!  Any idea why it reports a failure for that URL?  I just tested it and it worked.  Should we tell it to skip testing that particular URL so we can get the build to pass?

Yoshanuikabundi:
It seems to just be timing out. Increasing the timeout to 20s seems to fix it. https://img.shields.io/conda/v/conda-forge/openmm.svg is still suffering some intermittent failures.

peastman:
What is the empty .confirm_shortcut_change file for?

Yoshanuikabundi:
Looks like I installed something unrelated from a terminal in my openmm directory which for some reason made a file which I then accidentally committed. Fixed in 771f0f9. Once upon a time I thought this would be a simple, easy PR...

peastman:
> Once upon a time I thought this would be a simple, easy PR...

Yes, those were simpler times, weren't they?

I see all tests are passing.  I'll merge this before something else can break!

PR #3201, opened 2021-08-05
Move to actively maintained dependencies for API docs

### LINKED ISSUE #3132: Improvements to API reference search

### LINKED ISSUE #3149: Replace unmaintained documentation dependencies
We should replace `sphinxcontrib-lunrsearch` and `sphinxcontrib-autodoc_doxygen` with modern, actively maintained alternatives. The latter should be possible to replace with [`breathe`](https://breathe.readthedocs.io/en/latest/), and we may be able to replace the former with Sphinx's built in search functionality.

How committed are we to the search-as-you-type feature? As I've discussed in #3132, I think its idiosyncratic at best, and at the very least needs to be augmented with full page search on pressing enter. I'd be in favour of removing the search-as-you-type functionality altogether and sticking with the built in search, especially if it lets us remove an unmaintained dependency.

Are we aware of any other unmaintained dependencies?

Yoshanuikabundi:
This PR replaces `sphinxcontrib-autodoc_doxygen` with [`breathe`](https://github.com/michaeljones/breathe) and replaces `sphinxcontrib-lunrsearch` with the default Sphinx search. Fixes #3149 and fixes #3132.

Moving to Breathe required some pretty major surgery for the C++ API docs. Previously, all the classes were on a single big page, separated into sections for Core, Forces, Integrators, and basically every thing else. I've reorganised this so that Core classes are linked from the index page, where they are also put into context with other classes. Forces and integrators are broken down further into groups, and the extras is broken down into subsections as well. I think this describes the logic of the API better, and will help people find what they're looking for. (I'm growing to really like the design of the C++ API by the way, it's very elegant)

Breaking it down like this obviously doesn't work with automatic generation, so when new classes are added the author will have to manually add them to the appropriate section. So it's trading off automation for clarity. To make sure this step isn't forgotten, I've added the remaining undocumented classes to Doxygen's exclude setting, and set sphinx to fail if it encounters warnings (just for the C++ API docs).

If we don't think this trade off is worth it, or we don't like the way I've organised things, I can implement something more automatic. I haven't reorganised the Python docs, mostly because I wanted feedback before I jump down that rabbit hole.

Sorry this has taken so long, I had to learn a LOT of different software to put this together.


peastman:
This looks really nice.  I like the organization.

To make sure I understand, whenever we add a new C++ class we now need to manually add it to the documentation index.  But if we forget to do that, building the documentation will fail with a clear error message?  If so, that seems reasonable.

In the current system, we mark internal classes that shouldn't be documented with `@private`.  I think that's cleaner and easier to maintain than a separate list of exclusions in the Doxyfile.  Would it be possible to keep it working that way?

One minor change I would suggest: on the pages on forces and integrators, I would list the standard forces/integrators before the custom ones.

The new search function doesn't produce very clean results.  For example, I searched for "addparticle".  It finds every `addParticle()` method, as expected, but it also lists every argument to every one of those methods as a separate result.  Likewise, I searched for "periodictorsion".  In addition to finding the PeriodicTorsionForce class, it reported every method of that class and every argument of every method as a separate result.

jchodera:
> To make sure I understand, whenever we add a new C++ class we now need to manually add it to the documentation index. 

Pull request templates with checklists are also a good way to make sure we correctly address issues. For example:
- [ ] tagging the issue being addressed
- [ ] adding a test for the failure, and ensuring that it fails before the issue is fixed
- [ ] making changes to fix the issue
- [ ] updating the API documentation
- [ ] updating the changelog

Yoshanuikabundi:
Thanks!

> To make sure I understand, whenever we add a new C++ class we now need to manually add it to the documentation index. But if we forget to do that, building the documentation will fail with a clear error message? If so, that seems reasonable.

Exactly!

> In the current system, we mark internal classes that shouldn't be documented with @private. I think that's cleaner and easier to maintain than a separate list of exclusions in the Doxyfile. Would it be possible to keep it working that way?

I agree; I tried to make this work but couldn't figure it out. The classes I added to the Doxyfile are all already marked as @private, but Doxygen creates XML for them anyway. Maybe the `public` keyword in the source overrides the comment? I tried switching off EXTRACT_LOCAL_CLASSES, but it didn't work. There were already entries in the EXCLUDE_SYMBOLS variable so I figured that the next best thing would be to just extend it. All the classes I added follow the pattern `*Force::*Info`, so instead of the long list I could just exclude that?

> One minor change I would suggest: on the pages on forces and integrators, I would list the standard forces/integrators before the custom ones.

Done in 4974cd1

> The new search function doesn't produce very clean results. For example, I searched for "addparticle". It finds every addParticle() method, as expected, but it also lists every argument to every one of those methods as a separate result. Likewise, I searched for "periodictorsion". In addition to finding the PeriodicTorsionForce class, it reported every method of that class and every argument of every method as a separate result.

I agree that it's not ideal; I think it's an improvement over Lunr, which would just show a bunch of copies of "addParticle" with no additional context. Unfortunately Sphinx's search functionality is not very configurable. The only alternative I'm aware of would be [ReadTheDocs](https://docs.readthedocs.io/en/stable/guides/searching-with-readthedocs.html), but that means running a special HTTP server. It might be possible to implement logic to remove redundant results in the [template](https://github.com/sphinx-doc/sphinx/blob/6ac326e019db949c2c8d58f523c2534be36d4e62/sphinx/themes/basic/search.html#L52:L62)... but that means implementing it in Jinja2, which is fine for templating but very limited for actual logic.

> Pull request templates with checklists are also a good way to make sure we correctly address issues. For example:

I agree; we should do this too!

peastman:
> The classes I added to the Doxyfile are all already marked as @Private, but Doxygen creates XML for them anyway.

Doxygen processes the classes and records their visibility.  We make use of that to only document public classes:

https://github.com/openmm/openmm/blob/f5f2773bc79789df88448d799ff3d841f0e3ddd9/docs-source/api-c%2B%2B/render.py#L39-L41

Yoshanuikabundi:
Cool, I'll have a crack at supporting that in `breathe-apidoc.py`.

Looks like Linkinator is complaining because of a bug. I've reported it [here](https://github.com/JustinBeckwith/linkinator/issues/323). I think I can work around it.

Yoshanuikabundi:
OK, I've incorporated that check for publicity into the script that generates stub files, and I've (hopefully) fixed linkinator by stripping the `OpenMM::` prefix from generated files. I've also neatened up the index page by using actual headings for the core classes subheadings, rather than single element toctrees.

While we're here, one thing I would definitely like to add to the `Platform` docs is a sentence describing how to get a list of kernel names, or even better a table describing the different kernels. Is there an easy way to do that?

peastman:
> While we're here, one thing I would definitely like to add to the Platform docs is a sentence describing how to get a list of kernel names, or even better a table describing the different kernels. Is there an easy way to do that?

Not really.  There's no global list of kernel names.  Every force or integrator is free to define whatever kernels it wants.  There's a bunch of them in https://github.com/openmm/openmm/blob/master/olla/include/openmm/kernels.h, and each plugin defines additional ones, such as https://github.com/openmm/openmm/blob/master/plugins/amoeba/openmmapi/include/openmm/amoebaKernels.h.

peastman:
We also need to update the instructions on building documentation in the manual, since this changes what dependencies you need to install.

Yoshanuikabundi:
Ok. I've revised the Platform docs a little, as they seemed geared to users of the lower level API rather than the library layer. Let me know if I should break that out into its own PR.

I've also updated the manual. Is there anything else to do here before merging? It turns out that Breathe makes it a fair bit easier to write docs...

olla/include/openmm/Platform.h
```diff
@@ -45,23 +45,42 @@ class Kernel;
...
+ * set of KernelFactory objects which together implement the kernels.  The
+ * Platform class, in turn, provides a static registry of all available
+ * Platform objects.
  *
- * To get a Platform object, call
+ * OpenMM ships with platforms that provide high-performance simulation on a CPU
+ * or GPU. They are described in more detail in the User Manual. Usually, the
+ * platform is chosen automatically when a Context is constructed.
  *
+ * There are three main ways to get a Platform object:
```
peastman, line 58: There's also another way: use `Platform::getPlatform()` to get it by index.  For example, if you want a list of all available Platforms, you would call `getNumPlatforms()` and then loop over them.

olla/include/openmm/Platform.h
```diff
@@ -45,23 +45,42 @@ class Kernel;
...
+ * 2. If a Context has already been constructed, call Context::getPlatform():
  * \verbatim embed:rst:leading-asterisk
- * .. code-block:: cpp
+ *     .. code-block:: cpp
+ *
+ *        Platform& platform = context.getPlatform();
+ * \endverbatim
  *
- *    Platform& platform = Platform::findPlatform(kernelNames);
+ * 3. If the required kernels are known, call findPlatform():
```
peastman, line 74: In practice there is almost never a reason for someone to call this method.  It exists for use by ContextImpl, to automatically select a Platform if the user didn't specify one.  It's still ok to document it, just mention that people are unlikely to need it.  (Although it probably isn't accurate to describe it as one of the "main ways" to get a Platform.)

olla/include/openmm/Platform.h
```diff
@@ -45,23 +45,42 @@ class Kernel;
...
  *
+ *        Platform& platform = Platform::findPlatform(kernelNames);
  * \endverbatim
  *
- * passing in the names of all kernels that will be required for the calculation you plan to perform.  It
- * will return the fastest available Platform which provides implementations of all the specified kernels.
- * You can then call createKernel() to construct particular kernels as needed.
+ * The latter two methods returns the fastest available Platform which provides
+ * implementations of all the specified kernels. You can then call createKernel()
+ * to construct particular kernels as needed.
```
peastman, line 83: Method 2 returns whatever Platform the Context is using.  That might be the fastest available, or it might be one that the user explicitly specified when creating the Context.

peastman:
The manual changes look good.  A few changes needed to the Platform docs.  It might be best to move them to a separate PR, since they're unrelated to everything else, though if you prefer to keep them here it's not a big deal.

Yoshanuikabundi:
Yeah, let's merge this now, I can do a new PR with the platform changes and a few others from the issue tracker.

olla/include/openmm/Platform.h
Yoshanuikabundi: Prior to these changes, this was the only method documented. I think what might have happened is that Platform was originally documented for internal use, and was later made part of the public API and the documentation was never updated. I'll remove it in the upcoming PR

peastman:
Looks good!

PR #3202, opened 2021-08-05
Serialization of unit.Quantities

dwhswenson:
`openmm.unit.Quantity` instances do not have any built-in serialization. This means that other packages that build on OpenMM need to implement serialization methods. As a result, different packages use different approaches to serialize instances of `unit.Quantity`, leading duplicated developer effort and incompatible approaches.

This PR proposes an `openmm.unit.serialization ` module, which contains functions to facilitate serialization of `openmm.unit.Quantity` instances. This involves converting the `Quantity` to a `dict` that is suitable for serialization to JSON/YAML/TOML/etc. This also includes some conveniences for working with the Python's `json` standard library module.

Since there is no standard JSON serialization for NumPy arrays (https://github.com/numpy/numpy/issues/12481), I give the user an option to either serialize NumPy arrays as a list or as bytes.

I'll probably want to clean up some docstrings (at least), but I wanted to put up the PR so far for conceptual feedback before I go any further. It should currently work with any value that is natively JSON-serializable, as well as with NumPy arrays.

### DIFF
--- a/wrappers/python/openmm/unit/__init__.py
+++ b/wrappers/python/openmm/unit/__init__.py
@@ -16,3 +16,4 @@ from .quantity import Quantity, is_quantity
 from .unit_math import *
 from .unit_definitions import *
 from .constants import *
+from . import serialization
--- a/wrappers/python/openmm/unit/serialization.py
+++ b/wrappers/python/openmm/unit/serialization.py
@@ -1,0 +1,104 @@
+"""
+Module openmm.unit.serialization
+
+Contains tools for converting Quantities to dicts that can be serialized by
+various methods, included JSON and YAML, as well as several tools to
+facilitate integration with the Python standard library package json.
+"""
+import json
+import numpy as np
+from .unit import Unit
+from . import unit_definitions
+from .quantity import Quantity, is_quantity
+
+# These are implemented as separate functions so that they can be used by
+# external developers in their own serialization routines. They can also
+# dump to other formats (YAML, TOML, etc.)
+
+def to_dict(quantity, as_bytes=True):
+    """Convert a Quantity to a dict suitable for serialization.
+
+    Parameters
+    ----------
+    quantity : Quantity
+    """
+    units = {p.name: int(power)
+             for p, power in quantity.unit.iter_base_or_scaled_units()}
+
+    if isinstance(quantity._value, np.ndarray):
+        # treat numpy as a special case
+        v = quantity._value
+        value = {'shape': list(v.shape),
+                 'dtype': str(v.dtype)}
+        if as_bytes:
+            value.update({
+                'ndarray': 'bytes',
+                'value': quantity._value.tobytes().decode('latin-1'),
+            })
+            ndarray = 'bytes'
+        else:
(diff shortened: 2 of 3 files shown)


peastman:
This is part of a larger feature (#1543), which is to make objects in the python layer serializable.  We should come up with a coherent design for the whole thing, rather than implementing it piecemeal.  Let's figure out how we want serialization to work in general.

Note that Quantity already has `__getstate__()` and `__setstate__()` methods.  Those do essentially the same thing as your `to_dict()` and `from_dict()` methods, but they're part of a standard Python convention.  That also means they can be pickled, which isn't a great serialization method, but it's good enough for some purposes.

jchodera:
Might be a good opportunity to also build in compatibility with [pint](https://pint.readthedocs.io/en/stable/) or [unyt](https://unyt.readthedocs.io/en/stable/), which are also serializable. If we had the ability to transform from/to these units for convenience for our users, we could use that feature to serialize units.

j-wags:
OpenFF poked around a bit in the "simtk/pint interoperability" direction. Our approach is still evolving but you can check out our code here: https://github.com/openforcefield/openff-units

dwhswenson:
> Note that Quantity already has `__getstate__()` and `__setstate__()` methods.

I'd overlooked those. Sorry about that. The code here is largely stuff we were already doing in OPS, and it came up again in choderalab/openmmtools#518 that there was a need to replicate that behavior. If multiple downstream packages are implementing different approaches for handling the same problem, then it makes sense to try to put a solution in the upstream package.

The fundamental problem is that NumPy arrays don't have a "blessed" (as that NumPy issue phrased it) serialization compatible with standard library JSON. (Since so many other tools interact via JSON, I think it would be good to create a serialization intermediate with JSON in mind.) For OpenMM, it seems to me that the options are:

1. Roll your own, using whichever of NumPy's several serialization methods feels is best. That's what I was doing here.
2. Provide JSON encoding/decoding for `Quantity` and `Unit` (building on `__getstate__`/`__setstate__`), and tell users they need to use a JSON package that natively supports NumPy (e.g., `nujson`).
3. Use/require users to use an external package like [`jsonpickle`](https://jsonpickle.github.io/). `jsonpickle` can already serialize quantities.
4. Build an internal package to handle this. You can see the one I wrote for OPS [here](https://github.com/openpathsampling/openpathsampling/blob/master/openpathsampling/experimental/simstore/custom_json.py), with it's `simtk.unit` support [here](https://github.com/openpathsampling/openpathsampling/blob/dd9b5e1159e64f541857e26798af66acf90d3aca/openpathsampling/experimental/storage/simtk_unit.py#L18-L45).

If there's an interest in having this PR evolve along the direction of option 1 or option 2, I'm happy to move it either of those directions. Otherwise, feel free to close this, @peastman. (If option 3, this PR isn't needed; if option 4, that's a completely different PR -- from your comment above, it sounds like you're looking at one of these.)

peastman:
I'm going to close this PR, but let's continue the discussion at #1543.  I'll post a summary of the issues there.

Issue #3179, opened 2021-07-15
Disuflide bond potential issue in OpenMM app


jchodera:
@peastman: I just noticed this [openmm patch](https://github.com/deepmind/alphafold/blob/main/docker/openmm.patch) in the recent GitHub release of the [DeepMind AlphaFold2 code](https://github.com/deepmind/alphafold). Seems like the patch prevents a potential bug where more than one disulfide bond can be assigned to a Cys `SG` atom.

Perhaps we should see if this needs to be incorporated as a bugfix?

peastman:
Yes, that does look useful.  @augustinzidek is it ok if we adopt your fix directly?  We would need permission to use it under the MIT license, since that's what the OpenMM Python layer uses.

And I'm very glad to see it's released.  I'm sure I'm not the only person who has been eagerly waiting for it!

jchodera:
Tagging @Augustin-Zidek (since @augustinzidek appears to be a typo)!

peastman:
Oops, sorry about that!

Augustin-Zidek:
Sorry for the late reply. We are happy with the patch being included in OpenMM.

Note however, that our code is licensed under Apache 2.0, so that applies also to the patch as well. If that is compatible with OpenMM's license then feel free to apply the patch. If not, I am happy to send a separate PR to OpenMM with the patch which we would license under CC0.

peastman:
Thanks!  A separate PR would be great.  The OpenMM Python code uses the MIT license, so one way or another we need permission to distribute the patch under those terms (which CC0 would do).

Augustin-Zidek:
Hello, I sent a PR: https://github.com/openmm/openmm/pull/3203 with this patch.

jchodera:
Thanks so much @Augustin-Zidek!

peastman:
Fixed by #3203.

PR #3203, opened 2021-08-09
Remove tri-sulfite bond formation.

Augustin-Zidek:
Pdbfixer attempts to add di-sulfide bonds to any pair of CYS "S" atoms that are less than 3 A. This procedure fails when multiple such bonds are assigned to a given atom. 

Here we rectify this issue by a greedy algorithm: Atom I is chosen to be bonded to `argmin(distance([:I-1, I])` over all candidate di-sulfide bonds. This procedure will not generally result in a set of minimal distance di-sulfide bonds, but will fix the "tri-sulfide bond" problem. Fixing the more general case would require a solution to global constraint satisfaction problem.

See https://github.com/openmm/openmm/issues/3179 for more context.

**Legal**
This patch is provided under the CC0 license (https://creativecommons.org/share-your-work/public-domain/cc0/).
This is not an officially-supported Google product. It is provided ‘as-is’ without any warranty of any kind, whether expressed or implied.

### DIFF
--- a/wrappers/python/openmm/app/topology.py
+++ b/wrappers/python/openmm/app/topology.py
@@ -354,7 +354,7 @@ class Topology(object):
             names = [atom.name for atom in res._atoms]
             return 'SG' in names and 'HG' not in names
         # This function is used to prevent multiple di-sulfide bonds from being
-        # assigned to a given atom. This is a DeepMind modification.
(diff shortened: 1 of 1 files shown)


peastman:
This looks great.  Thank you so much!

Issue #3170, opened 2021-07-07
Intermittent errors with OpenMM 7.5.1 benchmarks


dmclark17:
I am noticing some intermittent errors when running OpenMM 7.5.1 benchmarks; specifically, with the default timestep, the pme test sometimes throws “Particle coordinate is nan” exceptions. It does seem to be quite rare—the error happened in two of 100 trials. This was done on a Titan V and CUDA 11.2. I am using the following benchmark command:

```
python benchmark.py --test=pme --platform=CUDA --seconds=30
```

Thanks!

peastman:
Thanks!  My guess is that we just need to add `hydrogenMass = 1.5*unit.amu` for DHFR, the same way we do for ApoA1.

dmclark17:
Sorry for the delay. Increasing the hydrogen mass seemed to fix the issue—the benchmark successfully finished all of the 100 trials that I ran. I can submit a PR with the change.

peastman:
That would be great.  Let's just make that our standard for most benchmarks.  The logic would be:

```python
if amoeba:
  hydrogenMass = None
elif options.heavy:
  hydrogenMass = 4*unit.amu
else:
  hydrogenMass = 1.5*unit.amu
```

peastman:
Fixed by #3208.

PR #3208, opened 2021-08-11
Increases hydrogen mass for benchmarks

dmclark17:
Increases the mass of hydrogen for the DHFR benchmark to avoid intermittent errors in simulation, see #3170.


peastman:
It also needs to get set in the `elif amber:` block.

peastman:
Good question.  In principle yes it should support that option.  But really what we should do is restructure that function to unify the Amber tests with the other ones.  For the moment maybe just add the `hydrogenMass` setting and leave it at that.  I'll do a more thorough cleanup later.

The heavy hydrogens options also isn't that useful anymore, now that use 4 fs even without it.  At some point I might just remove it altogether.

peastman:
Looks good.  Thank you!

PR #3209, opened 2021-08-12
Fix for incorrect behavior of OpenCL platform on multiple GPUs.

Dann239:
This addresses the "OpenCL is broken too" part of #3039 

The `VerletIntegrator` provides incorrect results on OpenCL platform when launched on multiple GPUs. This is because it uses 64-bit fixed point `longForceBuffer` array which isn't updated properly by `OpenCLParallelCalcForcesAndEnergyKernel`.

I believe this was introduced by #2488 so everyone who might've used OpenCL on multiple GPUs since then would have been affected by this.


peastman:
Looks good.  Thanks so much for figuring this out!

Issue #3131, opened 2021-05-31
Deprecate PyPI openmm package


jchodera:
Long ago, @rmcgibbo created a [PyPI `openmm` package](https://pypi.org/project/openmm/) that simply distributes a useful Python CLI for an older version of OpenMM. @peastman is also a co-owner of this project.

We should deprecate or rename this project until we can identify a path to distributing OpenMM itself through PyPI.

For future reference in potentially building OpenMM for pip distribution, we should check out [`cibuildwheel`](https://github.com/pypa/cibuildwheel).

jchodera:
For motivation, it looks like 129 people/month still make the mistake of trying to install `openmm` via `pip`:
https://pypistats.org/packages/openmm
![image](https://user-images.githubusercontent.com/3656088/120225605-88c96800-c213-11eb-8b58-45bfabddd422.png)

rmcgibbo:
@peastman: feel free to remove me as a co-owner. It doesn't appear that I can remove myself.

jchodera:
@peastman : Given #3211, perhaps we should see if there is a way to delete the `openmm` pip package?

cclauss:
Whoever has rights to the PyPI record can delete the package but that leaves the space open for someone else to take over the name which might not be a good idea.  Instead it might be useful to see the PyPI message to point to the current install instructions.

giadefa:
maybe just remove the files and add just a README pointing to the openmm.org
page for installation instructions

peastman:
It looks like Robert added me as a maintainer.  Presumably that means I have the permissions needed to do that.  I've never built PyPI packages before, so I don't know much about what it involves.  Ideally if someone tries to `pip install openmm` it should give an error message telling them the correct way to install it and then not install anything.  Is there a way to do that?

cclauss:
Yes.  I can help with that but I need a few days before I can focus on that.

peastman:
Fantastic, thanks!

mikemhenry:
I am happy to make `pip install openmm` install a package that says "this isn't the openmm you are looking for, install it via conda, pip distribution is in progress"

swails:
Why not just get rid of that package on pypi and then push one when it's ready?

mikemhenry:
If pypi lets you do that and keep the name squatted, than that would be even better IMHO.

mattwthompson:
Name-squatting in general is not permitted and in the case of a deleted package it seems that the name is released. DI:

[...]

PR #3211, opened 2021-08-16
Create README.md to provide installation notes

cclauss:


jchodera:
> `python3 -m pip install openmm` does not work because the [PyPI package](https://pypi.org/project/openmm) has not been updated since early 2014.

The creation of the [openmm PyPI package](https://pypi.org/project/openmm) was due to some confusion, and we've been meaning to address that. See https://github.com/openmm/openmm/issues/3131

Perhaps we could continue any discussion of what should happen to that package in https://github.com/openmm/openmm/issues/3131?

We've tried to collect current installation information in the [user guide](http://docs.openmm.org/latest/userguide/application.html#installing-openmm). Do you think we should link to that explicitly in the README?

Issue #2675, opened 2020-05-11
Introduce new parameters to amber99sb_obc.xml


ghost:
Hello, 
My protein has about 4 NAG (N-Acetyl-beta-D-glucosamine) molecules. I need to optimize the protein with an implicit water model.

I edited the amber99sb.xml  to include the parameters of the NAG (angles, bonds, ..etc)  and optimization in vacuum worked fine. 

but when using the implicit water model (amber99sb_obc.xml) I get this error 

`ValueError: GBSAOBCForce: No parameters defined for atom type 1978`
I am new to openMM and have no idea from where could I get those parameters needed for NAG.

F-AH:
more specifically I need to know how to get the scale parameter in the amber99sb_obc.xml file for the new atom types introduced to the ff file.

peastman:
I'm not sure how Amber typically selects implicit solvent parameters for small molecules.  Where did you get the rest of the force field parameters from?

jchodera:
@peastman: This description is relevant: https://github.com/openmm/openmmforcefields/issues/84#issuecomment-593110961

F-AH:
> I'm not sure how Amber typically selects implicit solvent parameters for small molecules.  Where did you get the rest of the force field parameters from?

I got the rest of the parameters from this website 
[https://atb.uq.edu.au/molecule.py?molid=465793&fbclid=IwAR301Pt3fU-9Vtp9YOc8Ce72646ckXqU42bB-5Sh_T9jPX4j_7YM0Rt_NH4#panel-md](url)

jchodera:
> I'm not sure how Amber typically selects implicit solvent parameters for small molecules.

The LEaP tool from AmberTools is responsible for [assigning GBSA parameters during `prmtop` writing](https://github.com/choderalab/ambermini/blob/8af922f71376aad6b01c876e4cb8e3fe3639e88c/leap/src/leap/unitio.c#L6061-L6293). This code was ported to [openmm](https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/customgbforces.py), but for some reason I'm unsure of, we only use it to _check_ parameters, not assign them.

It should be relatively straightforward just to extend `ForceField` to allow GBSA parameters to be assigned using this code via an optional argument. I'm not quite sure why we haven't done that before.

peastman:
It could even be done in the XML file: include a script that calls that code to assign parameters.  I'll mark this as a feature request.

[...]

PR #3214, opened 2021-08-17
Implicit solvent for modern force fields

### LINKED ISSUE #3104: AMBER14 or CHARMM36 force field with implicit water model?
I have successfully done MD simulations with both AMBER14 and CHARMM36 force fields using the explicit water models (`amber14/tip3pfb.xml` and `charmm36/water.xml`, respectively).

However, from reading the documentation and examples I have been unable to use AMBER14 or CHARMM36 force fields with implicit water models.

For AMBER99 I can successfully use an implicit water model as

```python
   forcefield = mm.app.ForceField("amber99sbildn.xml", "amber99_obc.xml")
```

Please could you let me know how I would use an implicit water model with AMBER14 or CHARMM36?  There does not seem to be an implicit water model XML file in the respective directories.

peastman:
This provides an implementation of GBSA implicit solvent that works with current force fields.  Like the old Amber99 implementation you just include an extra XML file.  But instead of containing parameters, it contains a script that invokes existing code to select parameters.

So far I've written the OBC2 implementation.  Modifying it for other versions will be trivial.

A question is where to put the files and how to present them.  Right now I have them in a folder that's just called `implicit`, indicating that they aren't specific to any force field.  That's kind of true.  They should work with any force field as long as we have parameters for all necessary elements.  But they came from Amber, and I don't know how valid it is to use them with other force fields.  If you used the Amber GB parameters with a CHARMM force field, would you reasonably expect to get accurate results?

Fixes #3104 and #2675.

### DIFF
--- a/docs-source/usersguide/application/02_running_sims.rst
+++ b/docs-source/usersguide/application/02_running_sims.rst
@@ -593,6 +593,40 @@ such as :file:`charmm36/water.xml`, which specifies the default CHARMM water mod
 The converted parameter sets come from the `CHARMM36 July 2017 update <http://mackerell.umaryland.edu/charmm_ff.shtml>`_
 and were converted using the `openmm-forcefields <https://github.com/choderalab/openmm-forcefields>`_ package and `parmed <https://github.com/parmed/parmed>`_.
 
+Implicit Solvent
+----------------
+
+The Amber and CHARMM force fields described above can be used with any of the Generalized
+Born implicit solvent models from AMBER.  To use them, include an extra file when
+creating the ForceField.
+
+.. tabularcolumns:: |l|L|
+
+======================  ==================================================================================================================================
+File                    Implicit Solvent Model
+======================  ==================================================================================================================================
+:file:`implicit/hct`    Hawkins-Cramer-Truhlar GBSA model\ :cite:`Hawkins1995` (corresponds to igb=1 in AMBER)
+:file:`implicit/obc1`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ I parameters (corresponds to igb=2 in AMBER).
+:file:`implicit/obc2`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ II parameters (corresponds to igb=5 in AMBER).
+:file:`implicit/gbn`    GBn solvation model\ :cite:`Mongan2007` (corresponds to igb=7 in AMBER).
+:file:`implicit/gbn2`   GBn2 solvation model\ :cite:`Nguyen2013` (corresponds to igb=8 in AMBER).
+======================  ==================================================================================================================================
+
+You can further control the solvation model in a few ways.  First, you can
+specify the dielectric constants to use for the solute and solvent:
+::
+
+    system = forcefield.createSystem(topology, soluteDielectric=1.0, solventDielectric=80.0)
+
+If they are not specified, the solute and solvent dielectrics default to 1.0 and
+78.5, respectively.
+
+You also can model the effect of a non-zero salt concentration by specifying the
+Debye-Huckel screening parameter\ :cite:`Srinivasan1999`:
+::
+
+    system = forcefield.createSystem(topology, implicitSolventKappa=1.0/nanometer)
+
 AMOEBA
 ------
 
--- a/docs-source/usersguide/application/02_running_sims.rst
+++ b/docs-source/usersguide/application/02_running_sims.rst
@@ -598,19 +598,30 @@ Implicit Solvent
 
 The Amber and CHARMM force fields described above can be used with any of the Generalized
 Born implicit solvent models from AMBER.  To use them, include an extra file when
-creating the ForceField.
+creating the ForceField.  For example,
+::
+
+    forcefield = ForceField('amber14-all.xml', 'implicit/gbn2.xml')
 
 .. tabularcolumns:: |l|L|
 
-======================  ==================================================================================================================================
-File                    Implicit Solvent Model
-======================  ==================================================================================================================================
-:file:`implicit/hct`    Hawkins-Cramer-Truhlar GBSA model\ :cite:`Hawkins1995` (corresponds to igb=1 in AMBER)
-:file:`implicit/obc1`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ I parameters (corresponds to igb=2 in AMBER).
-:file:`implicit/obc2`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ II parameters (corresponds to igb=5 in AMBER).
-:file:`implicit/gbn`    GBn solvation model\ :cite:`Mongan2007` (corresponds to igb=7 in AMBER).
-:file:`implicit/gbn2`   GBn2 solvation model\ :cite:`Nguyen2013` (corresponds to igb=8 in AMBER).
-======================  ==================================================================================================================================
+==========================  ==================================================================================================================================
+File                        Implicit Solvent Model
+==========================  ==================================================================================================================================
+:file:`implicit/hct.xml`    Hawkins-Cramer-Truhlar GBSA model\ :cite:`Hawkins1995` (corresponds to igb=1 in AMBER)
+:file:`implicit/obc1.xml`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ I parameters (corresponds to igb=2 in AMBER).
+:file:`implicit/obc2.xml`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ II parameters (corresponds to igb=5 in AMBER).
+:file:`implicit/gbn.xml`    GBn solvation model\ :cite:`Mongan2007` (corresponds to igb=7 in AMBER).
+:file:`implicit/gbn2.xml`   GBn2 solvation model\ :cite:`Nguyen2013` (corresponds to igb=8 in AMBER).
+==========================  ==================================================================================================================================
+
+The only nonbonded methods that are supported with implicit solvent are :code:`NoCutoff` (the default),
+:code:`CutoffNonPeriodic`, and :code:`CutoffPeriodic.`  If you choose to use a nonbonded cutoff with
+implicit solvent, it is usually best to set the cutoff distance larger than is typical with explicit solvent.
+A cutoff of 2 nm gives good results in most cases.  Periodic boundary conditions are not usually used
+with implicit solvent.  In fact, the lack of need for periodicity and the artifacts it creates is one
+of the advantages of implicit solvent.  The option is still offered, since it could be useful in some
+unusual situations.
 
 You can further control the solvation model in a few ways.  First, you can
 specify the dielectric constants to use for the solute and solvent:
@@ -618,7 +629,7 @@ specify the dielectric constants to use for the solute and solvent:
 
     system = forcefield.createSystem(topology, soluteDielectric=1.0, solventDielectric=80.0)
 
-If they are not specified, the solute and solvent dielectrics default to 1.0 and
+If they are not specified, the solute and solvent dielectric constants default to 1.0 and
 78.5, respectively.
 
 You also can model the effect of a non-zero salt concentration by specifying the
--- a/docs-source/usersguide/application/02_running_sims.rst
+++ b/docs-source/usersguide/application/02_running_sims.rst
@@ -638,6 +638,14 @@ Debye-Huckel screening parameter\ :cite:`Srinivasan1999`:
 
     system = forcefield.createSystem(topology, implicitSolventKappa=1.0/nanometer)
 
+The screening parameter can be calculated as
+
+.. math::
+  \kappa = 367.434915 \sqrt{\frac{I}{\epsilon T}}
+
+where :math:`I` is the ionic strength in moles/liter, :math:`\epsilon` is the solvent
+dielectric constant, and :math:`T` is the temperature in Kelvin.
+
 AMOEBA
 ------
 
--- a/wrappers/python/openmm/app/data/implicit/gbn.xml
+++ b/wrappers/python/openmm/app/data/implicit/gbn.xml
@@ -1,0 +1,46 @@
+<ForceField>
+ <Script>
+import openmm
+import openmm.app as app
+import openmm.unit as unit
+
+# Find the NonbondedForce.  We need it to look up charges, and also to set the reaction field dielectric to 1.
+
+nonbonded = [f for f in sys.getForces() if isinstance(f, openmm.NonbondedForce)]
+if len(nonbonded) != 1:
+    raise ValueError('Implicit solvent requires the System to contain a single NonbondedForce')
+nonbonded = nonbonded[0]
+nonbonded.setReactionFieldDielectric(1)
+
+# Construct the CustomGBForce.
+
+from openmm.app.internal.customgbforces import GBSAGBnForce
+argMap = {'soluteDielectric':'soluteDielectric', 'solventDielectric':'solventDielectric', 'implicitSolventKappa':'kappa'}
+solventArgs = {'SA':'ACE'}
+for key in argMap:
+    if key in args:
+        solventArgs[argMap[key]] = args[key]
+if nonbondedMethod != app.NoCutoff:
+    solventArgs['cutoff'] = nonbondedCutoff.value_in_unit(unit.nanometers)
+force = GBSAGBnForce(**solventArgs)
+params = GBSAGBnForce.getStandardParameters(topology)
+for i, p in enumerate(params):
+    charge, sigma, epsilon = nonbonded.getParticleParameters(i)
+    force.addParticle([charge, p[0], p[1]])
+force.finalize()
+
+# Set the nonbonded method and cutoff distance.
+
+if nonbondedMethod == app.NoCutoff:
+    force.setNonbondedMethod(openmm.CustomNonbondedForce.NoCutoff)
+elif nonbondedMethod == app.CutoffNonPeriodic:
+    force.setNonbondedMethod(openmm.CustomNonbondedForce.CutoffNonPeriodic)
+    force.setCutoffDistance(nonbondedCutoff)
+elif nonbondedMethod == app.CutoffPeriodic:
+    force.setNonbondedMethod(openmm.CustomNonbondedForce.CutoffPeriodic)
+    force.setCutoffDistance(nonbondedCutoff)
+else:
+    raise ValueError("Illegal nonbonded method for use with implicit solvent")
+sys.addForce(force)
+ </Script>
+</ForceField>
--- a/wrappers/python/tests/TestForceField.py
+++ b/wrappers/python/tests/TestForceField.py
@@ -303,11 +303,11 @@ class TestForceField(unittest.TestCase):
     def test_ImplicitSolventForces(self):
         """Compute forces for different implicit solvent types, and compare them to ones generated with AmberPrmtopFile."""
 
-        solventType = ['hct', 'obc1', 'obc2', 'gbn', 'gbn2']
-        nonbondedMethod = [NoCutoff, CutoffNonPeriodic, CutoffNonPeriodic, NoCutoff, NoCutoff]
-        kappa = [0.0, 0.0, 1.698295227342757, 1.698295227342757, 0.0]
-        file = ['HCT_NoCutoff', 'OBC1_NonPeriodic', 'OBC2_NonPeriodic_Salt', 'GBn_NoCutoff_Salt', 'GBn2_NoCutoff']
-        for i in [2]:#range(5):
+        solventType = ['obc1', 'obc2', 'gbn2']
+        nonbondedMethod = [CutoffNonPeriodic, CutoffNonPeriodic, NoCutoff]
+        kappa = [0.0, 1.698295227342757, 0.0]
+        file = ['OBC1_NonPeriodic', 'OBC2_NonPeriodic_Salt', 'GBn2_NoCutoff']
+        for i in range(len(file)):
             forcefield = ForceField('amber96.xml', f'implicit/{solventType[i]}.xml')
             system = forcefield.createSystem(self.pdb2.topology, nonbondedMethod=nonbondedMethod[i], implicitSolventKappa=kappa[i])
             integrator = VerletIntegrator(0.001)
--- a/wrappers/python/tests/TestForceField.py
+++ b/wrappers/python/tests/TestForceField.py
@@ -300,6 +300,26 @@ class TestForceField(unittest.TestCase):
                 numDifferences += 1
         self.assertTrue(numDifferences < system.getNumParticles()/20) # Tolerate occasional differences from numerical error
 
+    def test_ImplicitSolventForces(self):
+        """Compute forces for different implicit solvent types, and compare them to ones generated with AmberPrmtopFile."""
+
+        solventType = ['hct', 'obc1', 'obc2', 'gbn', 'gbn2']
+        nonbondedMethod = [NoCutoff, CutoffNonPeriodic, CutoffNonPeriodic, NoCutoff, NoCutoff]
+        kappa = [0.0, 0.0, 1.698295227342757, 1.698295227342757, 0.0]
+        file = ['HCT_NoCutoff', 'OBC1_NonPeriodic', 'OBC2_NonPeriodic_Salt', 'GBn_NoCutoff_Salt', 'GBn2_NoCutoff']
+        for i in [2]:#range(5):
+            forcefield = ForceField('amber96.xml', f'implicit/{solventType[i]}.xml')
+            system = forcefield.createSystem(self.pdb2.topology, nonbondedMethod=nonbondedMethod[i], implicitSolventKappa=kappa[i])
+            integrator = VerletIntegrator(0.001)
+            context = Context(system, integrator, Platform.getPlatformByName("Reference"))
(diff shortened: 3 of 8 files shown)


peastman:
@swails I'd love to get your thoughts on this question.  If we use GBSA with the parameters from Amber, is it reasonable to combine that with a CHARMM force field?  Or should we treat all of the Amber GB models as specific to Amber force fields?

jchodera:
> @swails I'd love to get your thoughts on this question. If we use GBSA with the parameters from Amber, is it reasonable to combine that with a CHARMM force field? Or should we treat all of the Amber GB models as specific to Amber force fields?

[Here's a paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3142295/) in which Charlie Brooks considered that question by using both AMBER and CHARMM GB models/parameters with GAFF small molecule parameters, and found they were all about equivalent. See [Table 3](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3142295/table/T3/?report=objectonly).

peastman:
Thanks!  It looks like they tested the opposite combination: CHARMM implicit solvent models with GAFF?  But it does suggest the solvent models are mostly independent of force field.

jchodera:
Yep, my conclusions as well. I couldn't find a study of the opposite combination.

swails:
Sorry for my late arrival here - I was on vacation last week.

I don't have anything better than what @jchodera provided here.  The Amber GB models should use the Amber GB parameters (screening and intrinsic radii parameters), but they're tuned to reproduce PB results and should be largely FF independent.

Notably the Force Fields are *not* trained for GB performance, suggesting that the GB models should be more transferrable than one that *is* trained in tandem with a force field.

docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -593,6 +593,40 @@ such as :file:`charmm36/water.xml`, which specifies the default CHARMM water mod
...
+The Amber and CHARMM force fields described above can be used with any of the Generalized
+Born implicit solvent models from AMBER.  To use them, include an extra file when
+creating the ForceField.
+
+.. tabularcolumns:: |l|L|
+
+======================  ==================================================================================================================================
+File                    Implicit Solvent Model
+======================  ==================================================================================================================================
+:file:`implicit/hct`    Hawkins-Cramer-Truhlar GBSA model\ :cite:`Hawkins1995` (corresponds to igb=1 in AMBER)
```
jchodera, line 608: The ffxml files we list in tables above end in `.xml`---e.g. `charmm36.xml` on line 539. Should we do the same here for consistency?

docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -593,6 +593,40 @@ such as :file:`charmm36/water.xml`, which specifies the default CHARMM water mod
 The converted parameter sets come from the `CHARMM36 July 2017 update <http://mackerell.umaryland.edu/charmm_ff.shtml>`_
 and were converted using the `openmm-forcefields <https://github.com/choderalab/openmm-forcefields>`_ package and `parmed <https://github.com/parmed/parmed>`_.
 
+Implicit Solvent
+----------------
+
+The Amber and CHARMM force fields described above can be used with any of the Generalized
+Born implicit solvent models from AMBER.  To use them, include an extra file when
+creating the ForceField.
```
jchodera, line 601: Can we give an explicit example of this, such as
```python
# Initialize a force field using the OBC2 GBSA parameters with the AMBER14SB protein force field
forcefield = ForceField('amber14sb.xml', 'implicit/obc2.xml')
```

docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -593,6 +593,40 @@ such as :file:`charmm36/water.xml`, which specifies the default CHARMM water mod
...
+======================  ==================================================================================================================================
+:file:`implicit/hct`    Hawkins-Cramer-Truhlar GBSA model\ :cite:`Hawkins1995` (corresponds to igb=1 in AMBER)
+:file:`implicit/obc1`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ I parameters (corresponds to igb=2 in AMBER).
+:file:`implicit/obc2`   Onufriev-Bashford-Case GBSA model\ :cite:`Onufriev2004` using the GB\ :sup:`OBC`\ II parameters (corresponds to igb=5 in AMBER).
+:file:`implicit/gbn`    GBn solvation model\ :cite:`Mongan2007` (corresponds to igb=7 in AMBER).
+:file:`implicit/gbn2`   GBn2 solvation model\ :cite:`Nguyen2013` (corresponds to igb=8 in AMBER).
+======================  ==================================================================================================================================
+
+You can further control the solvation model in a few ways.  First, you can
+specify the dielectric constants to use for the solute and solvent:
```
jchodera, line 616: Add "when creating a `System`".

docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -593,6 +593,40 @@ such as :file:`charmm36/water.xml`, which specifies the default CHARMM water mod
...
+::
+
+    system = forcefield.createSystem(topology, soluteDielectric=1.0, solventDielectric=80.0)
+
+If they are not specified, the solute and solvent dielectrics default to 1.0 and
```
jchodera, line 621: "dielectric constants".

"dielectrics" refers to materials, not dielectric constants.

docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -593,6 +593,40 @@ such as :file:`charmm36/water.xml`, which specifies the default CHARMM water mod
...
+78.5, respectively.
+
+You also can model the effect of a non-zero salt concentration by specifying the
+Debye-Huckel screening parameter\ :cite:`Srinivasan1999`:
```
jchodera, line 625: Can we give any guidance on how one might identify the appropriate kappa from a more familiar ionic strength?

wrappers/python/openmm/app/data/implicit/gbn.xml
```diff
@@ -0,0 +1,46 @@
...
+    charge, sigma, epsilon = nonbonded.getParticleParameters(i)
+    force.addParticle([charge, p[0], p[1]])
+force.finalize()
+
+# Set the nonbonded method and cutoff distance.
+
+if nonbondedMethod == app.NoCutoff:
+    force.setNonbondedMethod(openmm.CustomNonbondedForce.NoCutoff)
+elif nonbondedMethod == app.CutoffNonPeriodic:
+    force.setNonbondedMethod(openmm.CustomNonbondedForce.CutoffNonPeriodic)
```
jchodera, line 37: I'm not sure anyone really understands how to run GBSA with cutoffs, so it may be sensible to issue a warning here.

wrappers/python/openmm/app/data/implicit/gbn.xml
```diff
@@ -0,0 +1,46 @@
...
+    force.setCutoffDistance(nonbondedCutoff)
+elif nonbondedMethod == app.CutoffPeriodic:
+    force.setNonbondedMethod(openmm.CustomNonbondedForce.CutoffPeriodic)
```
jchodera, line 40: Does it seem wise to allow users to apply GBSA models in periodic cutoff systems? I've really never heard of this before. Perhaps at least a warning is in order?

wrappers/python/tests/TestForceField.py
```diff
@@ -300,6 +300,26 @@ def test_Forces(self):
                 numDifferences += 1
         self.assertTrue(numDifferences < system.getNumParticles()/20) # Tolerate occasional differences from numerical error
 
+    def test_ImplicitSolventForces(self):
+        """Compute forces for different implicit solvent types, and compare them to ones generated with AmberPrmtopFile."""
+
+        solventType = ['obc1', 'obc2', 'gbn2']
```
jchodera, line 306: This only tests 3 out of 5 models, does it not? Why not include complete test coverage?

jchodera (changes requested):
This is going to be a great addition!

docs-source/usersguide/application/02_running_sims.rst
peastman: Good question.  Here's the code in the Amber file parser to compute kappa:

https://github.com/openmm/openmm/blob/ad113a0cb37991a2de67a08026cf3b91616bafbe/wrappers/python/openmm/app/amberprmtopfile.py#L267-L274

All of that seems straightforward, except for the factor of 0.73 "to account for ion exclusions".  I have no idea what that is talking about.  I looked in the Amber manual, but couldn't find any description of it there.

docs-source/usersguide/application/02_running_sims.rst
peastman: Oops, thanks for catching that.

docs-source/usersguide/application/02_running_sims.rst
peastman: Yes, I'll add that.

wrappers/python/openmm/app/data/implicit/gbn.xml
peastman: Here's what the Amber manual says about it:

> If the nonbonded cutoff is used in GB calculations, it should be greater than that for PME calculations, perhaps cut=16.

I'll add a note recommending a larger cutoff.  The theory of applying a cutoff to GB is straightforward, since it's closely related to reaction field.  (Reaction field assumes everything beyond the cutoff distance is a uniform dielectric.  GB assumes everything beyond the Born radius is a uniform dielectric.)

wrappers/python/openmm/app/data/implicit/gbn.xml
peastman: There's nothing "wrong" with applying periodic boundary conditions to a GB calculation, although I'm not sure why someone would want to.  Maybe if you wanted to simulate a bulk solution of molecules dissolved in water, but avoid having to represent the water molecules explicitly?

wrappers/python/tests/TestForceField.py
peastman: For HCT and GBn, the forces don't quite match because the new version uses different radii than the old one (mbondi for HCT, bondi for GBn).  That's to match the recommendations in the Amber manual.  I figured testing three of the files here was good enough, since the only difference between them is the class name and all the Python code is already tested elsewhere.

wrappers/python/openmm/app/data/implicit/gbn.xml
jchodera: It doesn't make any sense to do so, and there is certainly no precedent in the literature.

OpenMM does a number of thins to keep users from accidentally making mistakes that are easy to make. For example, we even [warn users when they have non-optimal GBSA parameters](https://github.com/openmm/openmm/blob/0118afa0ae6febd7804feff7867622702cfb22c7/wrappers/python/openmm/app/internal/amber_file_parser.py#L1103). 

A warning here is an easy way to help users avoid simple mistakes. Programs can always disable the warning if they think this is the right thing to do.

wrappers/python/tests/TestForceField.py
jchodera: Can we at lest test _applying_ the parameters to make sure a typo doesn't creep into the `xml` files? There is no way to ensure they remain valid otherwise.

wrappers/python/openmm/app/data/implicit/gbn.xml
jchodera: I wasn't aware of these recommendations! Sounds like a great opportunity to emit a warning if the cutoff is too small.

docs-source/usersguide/application/02_running_sims.rst
jchodera: I wonder if it would make sense to accept either `implicitSolventSaltConc` or `implicitSolventKappa` as arguments? I imagine the former will be much more convenient!

wrappers/python/openmm/app/data/implicit/gbn.xml
peastman: > It doesn't make any sense to do so, and there is certainly no precedent in the literature.

Why do you think that?  The choice of periodic boundary conditions is completely unrelated to the use of implicit solvent.  It isn't like PME, where the periodicity actually affects the form of the potential function.  As long as the cutoff is less than half the box width (which we enforce), the periodicity has no effect at all on the GB calculation.  It just means we consider there to be more molecules beyond the edge of the box.

peastman:
Since you suggested having it print warning messages in a few places, let me describe my general philosophy about warning messages (and why OpenMM almost never uses them).  It's based on a few principles.

- Use sensible defaults to steer people toward the behavior you think is correct.  But if they choose to override those defaults, trust that they know what they're doing.
- Messages printed to a console are not an effective way of alerting people to things.  They likely won't see it.
- An inappropriate warning about something that's actually correct is much worse than no warning at all.  It trains users to ignore warning messages.

That's why the standard pattern in OpenMM is that if something is *definitely* wrong, we throw an exception and force them to deal with it, but if we can't be certain something is wrong, we trust that the user knows what they're doing.

As an example, you suggested printing a warning if the user specifies too small a cutoff with GB.  What counts as too small?  For a large globular protein, I would usually suggest a cutoff of around 2 nm.  But for small molecules, a much shorter cutoff would be fine.  In fact, if all Born radii are less than the cutoff distance, GB can actually be *more* accurate than reaction field at the same cutoff.  Yet we don't warn people about short cutoffs with reaction field.

jchodera:
My rationale for issuing a warning, if `CutoffPeriodic` was used with GBSA is
* We already [warn if the GBSA parameters differ from published values](https://github.com/openmm/openmm/blob/0118afa0ae6febd7804feff7867622702cfb22c7/wrappers/python/openmm/app/internal/amber_file_parser.py#L1103)
* Using GBSA in a periodic system is, to my knowledge, without literature precedent. (Have you seen _any_ examples?)
* `CutoffPeriodic` also uses a modified reaction field electrostatics functional form unless the `solventDielectric` is set to 1, but we use this argument to specify the GBSA dielectric instead. I'm not sure how we reconcile these.

However, if the philosophy behind warnings is as you describe above, perhaps the better path is to open an issue to review all use of warnings against this philosophy, and perhaps also document this philosophy in the user guide? The default for `ForceField.createSystem()` is indeed already `NoCutoff`, so we would be conforming to that philosophy in this PR (but not necessarily elsewhere in the existing codebase).

For example, it seems like we should [remove the warning regarding non-optimal GBSA parameters](https://github.com/openmm/openmm/blob/0118afa0ae6febd7804feff7867622702cfb22c7/wrappers/python/openmm/app/internal/amber_file_parser.py#L1103).

peastman:
> We already warn if the GBSA parameters differ from published values

Thanks for pointing that out.  It only does that because I didn't write it. :)  That actually seems like an especially inappropriate warning, since the Amber radii were never really properly optimized.  The differences between the sets mostly just come from someone looking at them and saying, "I think that value is a little too big so let's try reducing it."  Their recommendations aren't even that strong.  For OBC2, for example, the manual says, "With this option, you should use the command "set default PBradii mbondi2" in setting up the prmtop file, although "set default PBradii bondi" is also OK."  If either one is OK, we really shouldn't print a warning about it.

> Using GBSA in a periodic system is, to my knowledge, without literature precedent. (Have you seen any examples?)

Here are a couple of the top hits in a search for "implicit solvent for periodic system".

https://pubs.acs.org/doi/10.1021/acs.jctc.7b00169
https://pubs.rsc.org/en/content/articlelanding/2016/cp/c6cp04094b

The first one computes solvation free energies of organic molecules in organic solvents.  (They concluded that GB and PB are both inaccurate for this case, but at least they were able to run the calculation and see!)  The second one is modelling periodic surfaces.

> CutoffPeriodic also uses a modified reaction field electrostatics functional form unless the solventDielectric is set to 1, but we use this argument to specify the GBSA dielectric instead.

The script includes the line

```python
nonbonded.setReactionFieldDielectric(1)
```

jchodera:
> Here are a couple of the top hits in a search for "implicit solvent for periodic system".

> https://pubs.acs.org/doi/10.1021/acs.jctc.7b00169
> https://pubs.rsc.org/en/content/articlelanding/2016/cp/c6cp04094b
> 
> The first one computes solvation free energies of organic molecules in organic solvents. (They concluded that GB and PB are both inaccurate for this case, but at least they were able to run the calculation and see!) The second one is modelling periodic surfaces.

I can't find any evidence that the first paper actually uses GB in a periodic system. In fact, it specifically says "For GB, the simulation setup is consistent with the vacuum simulation", which was not periodic.

The second one is exclusively about QM/MM calculations with PCM (not GB) solvent model for a palladium catalyst system, which seems about as far from our use cases as you can get since we support none of these things. The MM part of this work is carried out in TIP3P water.

> The script includes the line
> nonbonded.setReactionFieldDielectric(1)

Thanks---I had missed this!

peastman:
I just checked in a bunch of updates based on comments.

@Yoshanuikabundi while working on this, I noticed that the numbering of references in the manual seems to have gotten messed up.  Presumably a consequence of #3165?  For example, the user guide contains the sentence

> You also can model the effect of a non-zero salt concentration by specifying the Debye-Huckel screening parameter[16]:

When I click the "16", it takes me to the correct reference in the bibliography, but it's actually number 27, not 16.  This isn't that big a deal, since it does link to the correct reference, but it would be good to fix it.

peastman:
@jaimergp the CUDA builds are failing with an error during setup:

```
Err:36 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  Packages
  404  Not Found [IP: 152.195.19.142 443]
```

Is that something we need to fix, or is it a problem on the NVIDIA and/or conda-forge side?

Yoshanuikabundi:
@peastman: Looks correct on latest and on development, and when I build this branch locally it's correct too. I suspect sphinx just hasn't rebuilt zbibliography.html in your build - try deleting it and rebuilding?

jaimergp:
> Is that something we need to fix, or is it a problem on the NVIDIA and/or conda-forge side?

Looks like an issue on Nvidia's side. Maybe it was just a temporary network problem?

[Instructions have not changed on their end](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#ubuntu-installation).

peastman:
Rerunning the build doesn't fix the CUDA problem, so I don't think it's just a network error.  I'll see if I can figure it out.

docs-source/usersguide/application/02_running_sims.rst
swails: > All of that seems straightforward, except for the factor of 0.73 "to account for ion exclusions". I have no idea what that is talking about. I looked in the Amber manual, but couldn't find any description of it there.

Caution, *highly* precise language ahead:

I think I recall seeing this in one of the original GB papers that introduced the approximation.  If I recall correctly, it's an empirical factor that improves performance of the model, explained as accounting for ions actually taking up space (and not being point-charges).  The comment is equally vague in sander:

```Fortran
      !       ---scale kappa by 0.73 to account(?) for lack of ion exclusions:

      kappa = 0.73d0* kappa
```

wrappers/python/openmm/app/data/implicit/gbn.xml
swails: I was always taught to run GB with no cutoffs.  The cut=16 recommendation seems to be a relic of the days when computers were painfully slow.

wrappers/python/openmm/app/data/implicit/gbn.xml
swails: It's still worth a warning.  We're usually taught that it's either-or, and other codes (at least AMBER) explicitly prohibits GB with PBC.  So while the math may work out, and the physics can even be hand-waved, my estimation is that most uses of GB with PBC are unintended.  `UserWarning` is noisy without blocking users from doing something they really *want* to do.

swails:
> Thanks for pointing that out. It only does that because I didn't write it. :) That actually seems like an especially inappropriate warning, since the Amber radii were never really properly optimized. The differences between the sets mostly just come from someone looking at them and saying, "I think that value is a little too big so let's try reducing it."

I don't think this is entirely accurate.  There's also a difference between the ForceField and the Amber file parsers.  In the former, you can assign the correct radii when the user chooses a GB model - you can use the same input files and change the GB model you want freely.  You can't do this with Amber topology files, where the GB radii is built into the topology file.  In this particular case, it's *impossible* to provide sensible defaults, since OpenMM has no control over where those radii come from.

I've also learned that the actual values of these radii are *incredibly* important, and small changes can yield huge differences in FE calculations (a difference of 0.1 Angstroms in carboxylate oxygens was the difference between excellent pKa prediction performance and horrific performance).

When you say "the Amber radii were never really properly optimized", that's not accurate either.  Some radii were highly optimized (like the GBn2 model) to reproduce experimental properties.  Others were used to fit GB model parameters themselves (GBOBC1/2 and GBn1).

If you can make it hard to do the wrong thing accidentally, then I agree warnings are unnecessary.  But if you can't, then I think warnings are absolutely justified.

> For OBC2, for example, the manual says, "With this option, you should use the command "set default PBradii mbondi2" in setting up the prmtop file, although "set default PBradii bondi" is also OK."

This is a bug in the manual - it should recommend mbondi2 and only mbondi2.  That's also one of the oldest supported GB models.  The new ones were highly fitted to a particular radius set and you should expect poor behavior using the wrong one.

docs-source/usersguide/application/02_running_sims.rst
peastman: Good to know I had a reason for being confused!  The comment in the Amber code is actually much more informative.  The (?) tells you that the person who wrote it didn't understand it either. :)  I guess I'll just roll it into the prefactor and give a formula with a magic constant in the manual.

peastman:
Here is the current text I have about cutoffs and periodic boundary conditions.

> The only nonbonded methods that are supported with implicit solvent are `NoCutoff` (the default), `CutoffNonPeriodic`, and `CutoffPeriodic.`  If you choose to use a nonbonded cutoff with implicit solvent, it is usually best to set the cutoff distance larger than is typical with explicit solvent. A cutoff of 2 nm gives good results in most cases.  Periodic boundary conditions are not usually used with implicit solvent.  In fact, the lack of need for periodicity and the artifacts it creates is one of the advantages of implicit solvent.  The option is still offered, since it could be useful in some
unusual situations.

Does that sound accurate?

peastman:
Other people have been reporting the CUDA problem too.

https://forums.developer.nvidia.com/t/apt-get-not-working-on-ubuntu-18-04/188312

I restarted the builds, and it seems to be working now.

peastman:
> I suspect sphinx just hasn't rebuilt zbibliography.html in your build - try deleting it and rebuilding?

You're right, thanks!

I added the formula for the screening parameter.

jchodera:
> I added the formula for the screening parameter.

Thanks! But don't we want to let the user specify the ionic strength as an optional parameter as an alternative to the screening parameter? I imagine it will be used by most novice users, while the kappa will be used only by those wanting to reproduce previously published work.

peastman:
It also depends on the temperature.

jchodera:
We appear to support the specification of temperature and salt concentration as optional arguments (in addition to allowing kappa to be specified instead) in the [AmberPrmtopFile](https://github.com/openmm/openmm/blob/ad113a0cb37991a2de67a08026cf3b91616bafbe/wrappers/python/openmm/app/amberprmtopfile.py#L184-L190) class. Maybe we could be consistent?

peastman:
That's another API that I didn't write. :(  Having a `temperature` argument to `createSystem()` is begging for confusion.  Anyone seeing it will naturally assume that if they specify a temperature there, a thermostat will be added to their system and their simulation will be run at the temperature they specified.  It violates the principle of least surprise on multiple levels.

swails:
> Anyone seeing it will naturally assume that if they specify a temperature there, a thermostat will be added to their system and their simulation will be run at the temperature they specified.

The main blocker here is that for some constant temperature simulations, constant T is maintained by the integrator, and AndersenThermostat is only needed when it's not.

You could say something similar about specifying the temperature on the MC barostat - it *doesn't* maintain constant T without the right thermostat or integrator, but is needed to set up the barostat force.  You specify it for the same reason here.

peastman:
Yeah, I've never been entirely happy with the dual temperature specifications in the integrator/thermostat and the barostat.  We've discussed various ways of trying to unify them, but haven't come up with a design we were entirely happy with.

A difference here is that the integrator and barostat temperatures can change with time, but the one used in computing kappa can't.  It gets baked into the energy expression.  If someone specifies kappa, they'll have no expectation of any value ever being used except the one they specified.  But if they specify a salt concentration, they might reasonably expect that concentration to remain fixed even if they later change the simulation temperature.

swails:
Can we make temperature something that can be inserted into any energy expression that the lepton parser knows how to insert?  Is that already a thing?

peastman:
Right now `kappa` is hardcoded into the expression:

https://github.com/openmm/openmm/blob/ad113a0cb37991a2de67a08026cf3b91616bafbe/wrappers/python/openmm/app/internal/customgbforces.py#L360

We would change it into a global parameter.  That would allow it to be changed, but it would also have a performance cost.

Going further, we could make the temperature and salt concentration into global parameters and include the calculation to compute kappa from them as part of the energy expression.  That would have an even larger performance cost.  It would mean recomputing kappa for every interaction.

jchodera:
Perhaps we could leave the choice of "which parameters should be global parameters that can be changed on the fly?" to the user? We could have an optional dict arg, for example, that maps the available parameters into preferred global parameter names, and anything in this list would be allowed to be a global parameter. By default, these would all just be fixed for maximum performance.

peastman:
That's possible.  I think we're getting quite far beyond the scope of this PR though.  "Support GB with modern force fields" is a high priority feature.  "Allow the GB salt concentration to vary during a simulation" is a much lower priority one.

I do think it would be good to revisit the question of standardizing the mechanism for setting temperature.  It's discussed in #1440.  It's a hard problem, especially if we need to maintain backward compatibility, but it would have clear benefits.

peastman:
Ok, I'll go ahead and merge this.  We can continue discussing it, especially the temperature question, and make further changes in another PR if needed.

PR #3238, opened 2021-09-08
Make openmm package install in empty virtualenv

### LINKED ISSUE #3237: Python package does not install cleanly in an empty virtualenv
https://github.com/openmm/openmm/blob/master/wrappers/python/setup.py contains a top-level import of numpy and Cython which means that you cannot install the OpenMM Python package cleanly in an empty virtualenv.

```python
$ python3 -m venv .env
$ source .env/bin/activate
$ python3 setup.py build_ext bdist_wheel
  File "setup.py", line 9, in <module>
    import numpy
```
The solution is to create a custom build extension in setup.py and add a `setup_requires=["Cython", "numpy"]` argument to `setup()`. See example in e.g. https://github.com/scikit-garden/scikit-garden/pull/86

ecederstrand:
Pushes import of numpy and Cython into a custom extension class and thus to a point in the package install where setuptools has had the possibility to honor the `setup_requires` directive which now installs numpy and Cython.

Fixes #3237

### DIFF
--- a/wrappers/python/setup.py
+++ b/wrappers/python/setup.py
@@ -6,9 +6,8 @@ import re
 import os
 import sys
 import platform
-import numpy
-from distutils.core import setup
-from Cython.Build import cythonize
+from setuptools import Extension, setup
+from setuptools.command.build_ext import build_ext
 
 MAJOR_VERSION_NUM='@OPENMM_MAJOR_VERSION@'
 MINOR_VERSION_NUM='@OPENMM_MINOR_VERSION@'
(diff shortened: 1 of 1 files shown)


peastman:
See my comments on #3237.  By the time it gets to executing setup.py, it's already quite far into the build process and has already made use of lots of other dependencies (SWIG, Doxygen, CMake, etc.).  You can't just create a new environment, execute setup.py, and expect it to work.

ecederstrand:
Ok. Thanks for the explanation!

PR #3239, opened 2021-09-08
Linux PyPI-compatible build, proof of principle

tristanic:
Not necessarily ready for merging yet, but it works! :)

I've attached the Singularity recipe I've been using to create a controlled build environment. While the result should be fully compliant with the `manylinux2014` tag, it should really be migrated to base on the `manylinux2014` Docker image instead. Key changes:

- Add an extra `OPENMM_BUILD_UNVERSIONED` flag to the `CMakeLists.txt`, because Python's `setuptools` turns all symbolic links into actual files when copying.
- Replace `from distutils.core import setup` with `from setuptools import setup` in `setup.py` - this is necessary to support the `bdist_wheel` build target, and should be fully backward compatible.
- When building a wheel, the compiled library files and headers will go into `site-packages/openmm/lib` and `site-packages/openmm/include` respectively, and the `version.py` will point to the correct locations. I know `version.py` previously didn't point to headers, but having these bundled in with a queryable location will make building against it much easier.

I've only tested this in Linux so far. There may be a few more tweaks needed to successfully build in Windows and Mac environments, but I don't think there will be anything show-stopping.


jchodera:
@tristanic: It would be great to be able to support a PyPI package for OpenMM as well. Are you folks now in a position where you could help us with this, since none of us are experts in this process? If so, we may be able to distribute OpenMM 8 via both conda-forge and pip, which would be fantastic.

tristanic:
I won't be officially starting in the new role until mid-late Aug... but I'm certainly hoping this will be possible!

peastman:
OpenMM 8.2 added PyPI builds.

PR #3242, opened 2021-09-09
Check for keyword args to createSystem() that are never used

### LINKED ISSUE #97: Warn if unused flags are present.
So currently, users can specify arbitrary flags when creating a system:

```
system = forcefield.createSystem(pdb.topology, stuff=1.0, typo=10.0)
```

Perhaps we should print a warning (or raise an exception) if there are flags that are unused?  I think this might help users identify issues that arise from misnamed arguments.  I suspect this might be harder to implement than I imagine, as we have to someone keep track of which flags eventually get used.

cc @tjlane

peastman:
Fixes #97.

The main benefit of this is to detect typos.  In a few cases it can also detect standard arguments that were provided when they didn't make sense, but most of the time it can't do that.

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -46,6 +46,7 @@ import openmm.unit as unit
 from . import element as elem
 from openmm.app.internal.singleton import Singleton
 from openmm.app.internal import compiled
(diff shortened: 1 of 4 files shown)


peastman:
It even caught a few incorrect argument names in some test cases.

jchodera:
Awesome!

PR #3244, opened 2021-09-10
add setAtomIndex() method to ComputeContext

egallicc:
We discovered that the current implementation of atom reordering in the [ATMMetaForce](https://github.com/Gallicchio-Lab/openmm-atmmetaforce-plugin) plugin based on calling reorderAtoms() on each inner context did not lead to good performance of the non-bonded energy calculation. While we did not fully investigate the exact origin of the problem, one suspicion is that, because the inner contexts lack a full complement of Forces, reordering does not lead to an efficient neighbor list.

In any case, we obtained significantly better performance (30% or so) by synchronizing in the ReorderListener the atom index of the inner contexts with those of the main context. To do so we added a setAtomIndex() method to ComputeContext. Hence the present pull request.

### DIFF
--- a/platforms/common/include/openmm/common/ComputeContext.h
+++ b/platforms/common/include/openmm/common/ComputeContext.h
@@ -357,6 +357,16 @@ public:
         return atomIndex;
     }
     /**
+     * Set the reordered index of an atom.
+     *
+     * It is used only in special circumstances to sync the ordering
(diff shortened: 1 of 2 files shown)


egallicc:
This PR is actually premature. We will look at the issue in greater depth first.

PR #3248, opened 2021-09-16
Allow querying current step count

### LINKED ISSUE #2655: Provide a method getCurrentStep() for all platform?
When restarting from a checkpoint, it's useful to know which step it is at so that trajectories can be appended at the correct step, which is necessary for analyzing time correlation properties.

Cuda and OpenCL platform record the current step in the checkpoint file. The CudaContext and OpenCLContext have a getStepCount() method, but it's not exposed in python wrapper. For Reference and CPU platform, there's no such method, and the step is not saved in the checkpoint.

Do you think it's reasonable to provide such a method for all platforms, and also save step information in the checkpoint for Reference platform?

peastman:
Fixes #2655.

All of the platforms already tracked the step count internally, but there wasn't a public way to get it.  I exposed it through `getStepCount()` and `setStepCount()` on Context, and `getStepCount()` on State.  I also changed the type from `int` to `long long`, because running more than 2 billion steps isn't totally absurd.

Simulation already had a `currentStep` field that it kept updated independently.  I turned it into a property that invokes `getStepCount()` and `setStepCount()`.


peastman:
It's unrelated to this PR, but the docs build is failing because the URL https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html doesn't exist.  It looks like we need to change it to https://open-forcefield-toolkit.readthedocs.io/en/latest/users/smirnoff.html.

PR #3256, opened 2021-09-23
Use the primary CUDA context

raimis:
The CUDA documentation (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#context) recommends that the applications which create their own CUDA context should properly push and pop the context from the context stack.

Meanwhile OpenMM uses `cuCtxSetCurrent` which overwrites an existing CUDA context. This causes problems with the other software, e.g. PyTorch in https://github.com/openmm/openmm-torch/.

There are two solutions:
1.  OpenMM should properly use `cuCtxPushCurrent()` and `cuCtxPopCurrent()` to managed its CUDA context.
2. OpenMM should use the primary CUDA context (https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PRIMARY__CTX.html) and don't create its own.

This PR implements the 2nd solution. PyTorch uses the primary CUDA context too, so this is the most convenient to fix the interoperability issues (https://github.com/openmm/openmm-torch/issues/42).


raimis:
This was deprecated by https://github.com/openmm/openmm/pull/3258.

Issue #1910, opened 2017-10-02
Using a barostat ignores enforcePeriodicBox


stefdoerr:
So the documentation of getState shows a flag for disabling the wrapping:
http://docs.openmm.org/6.3.0/api-c++/classOpenMM_1_1Context.html#a5c53803407b6c22454813f668c42a33e

I noticed that if I simulate my system with a barostat I get wrapped coordinates.
```C++
OpenMM::Force *barostat = new OpenMM::MonteCarloBarostat(sim->barostatpressure, sim->thermostattemp);
system->addForce(barostat);
[...]
OpenMM::State state = sim->context->getState( OpenMM::State::Positions | OpenMM::State::Energy, false );
```
If I don't add the barostat I get them unwrapped. Is this expected behaviour?

swails:
Yes

swails:
Several trajectory analysis programs (such as cpptraj) have an "unwrap" command that will undo the wrapping.

stefdoerr:
Thing is I want to do it in C++ while running the sim. So I will have to look at the source code of cpptraj since there doesn't seem to be a C++ API.

Why does the barostat wrap though? (out of curiosity)

swails:
@peastman will have to answer definitively, but the barostat requires coordinate scaling, which means it is a Force that modifies coordinates. Since there is not really any restriction on how those coordinates can be modified by a Force, I suspect that the Context has to basically discard the information on each atom that tracks what periodic cell the "original" image is in.

For reasons of numerical precision all coordinates are represented in their "wrapped" state and then unrolled prior to fetching from a Context if enforcePeriodicBox is set to False. If this unrolling information is invalidated or lost, then you're stuck with the "wrapped" representation. 

But like I said, @peastman would have the definitive answer. I'm just speculating.

swails:
Why not just post process the trajectory after the simulation is over?

stefdoerr:
Trying to work on acemd3 which uses openMM as backend.
We want it done automatically by acemd3 to keep compatibility with acemd2 and to not force the user to do it.
Also you might ctrl+c the simulation at some point and then you'd end up with a wrapped sim. Would be breaking the consistency.

If it's not too complicated I would prefer it done in openMM than to use cpptraj. The way you describe it it sounds more like an inconsistency of openMM than a necessity. Let's see if @peastman can chime in on it :)

peastman:
[...]

PR #3260, opened 2021-09-29
Barostat does not translate into a single periodic box

### LINKED ISSUE #2898: Potential enhancement: Add option to MonteCarloBarostat to not wrap coordinates into box
Currently, the `MonteCarloBarostat` will wrap all positions into the main simulation box each time it attempts to resize the box.

When simulating protein-ligand or protein-protein systems, this can be inconvenient, since it requires expensive operations (like [`mdtraj.Trajectory.image_molecules()](http://mdtraj.org/development/api/generated/mdtraj.Trajectory.html) to re-image and make complexes whole, and even this process is heuristic and not fool-proof. 

It would be useful to have an optional mode for `MonteCarloBarostat` to re-image the molecules back by applying the negative scaled displacement vector after the Monte Carlo scaling operation.

Alternatively, as a work-around, would adding dummy harmonic bonds (with zero spring constants) between pairs of molecules work for ensuring they are imaged together, or would this cause havoc with the barostat were they ever to dissociate?

peastman:
Fixes #2898.

Previously every time a barostat attempted a step, it would begin by translating all the molecules into a single periodic box.  I'm not sure why I originally wrote it that way.  I thought it was necessary, but looking at it now I can't see any reason it would be.  I removed the code to do the translation and all the tests still pass.

### DIFF
--- a/platforms/common/src/kernels/monteCarloBarostat.cc
+++ b/platforms/common/src/kernels/monteCarloBarostat.cc
@@ -24,17 +24,12 @@ KERNEL void scalePositions(float scaleX, float scaleY, float scaleZ, int numMole
         center.y *= invNumAtoms;
         center.z *= invNumAtoms;
 
-        // Move it into the first periodic box.
-
-        real3 oldCenter = center;
-        APPLY_PERIODIC_TO_POS(center)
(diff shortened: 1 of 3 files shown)


stefdoerr:
Great! Related in case this explains why you originally did it https://github.com/openmm/openmm/issues/1910#issuecomment-333575147

PR #3273, opened 2021-10-06
Atom types in CHARMM FF are non-case-sensitive

charmm-gui:
Atom types in CHARMM FF are non-case-sensitive, but current charmmparameterset.py considers case of the atom type name.
This caused an atom type name mismatch between the PSF and the read parameter.


charmm-gui:
Done.
Now all tests are passed.

jchodera:
Thanks so much, @charmm-gui!

charmm-gui:
@swails A test case is attached.
[test_case.tar.gz](https://github.com/openmm/openmm/files/7349202/test_case.tar.gz)

PR #3275, opened 2021-10-06
Further optimization to updating long range correction

peastman:
After a global parameter changes, it doesn't recompute the long range correction coefficient until the next time it needs to compute energy.  In some cases this can save a lot of work.


zhang-ivy:
@peastman : This PR provides a very significant speed up for the test code you sent [here](https://github.com/openmm/openmm/pull/3236#issuecomment-929580822) 

As a reminder, here are the simulation times I was seeing before with #3236 (old experiments):
With #3236 (calling `context.getParameter()`): 3.3659 s
With #3236 (not calling `context.getParameter()`): 0.664 s
Without #3236 (calling `context.getParameter()`): 21.133 s
Without #3236 (not calling `context.getParameter()`): 0.646 s

With this PR (new experiments): 
With #3275 (calling `context.getParameter()`): 0.862 s
With #3275 (not calling `context.getParameter()`): 0.784 s

I'm testing this on our local HPC, so I'm using different GPUs in the old vs new experiments, which explains the discrepancies. 

However, I just checked, and in the nonequilibrium simulations I am running, we unfortunately do retrieve the potential energy every step, so this PR won't help with speeding those simulations up.

Issue #2514, opened 2020-01-07
Fully flexible unit cells during NPT simulations


sandvdnh:
In molecular dynamics simulations at constant temperature and pressure of certain porous crystals such as metal/covalent organic frameworks -- and probably also other solids -- it is necessary to use barostat algorithms that allow arbitrary shape fluctuations in the unit cell. As an example, a single solid can exhibit two phases in which the angles between the box vectors (as well as their lengths) are different; transformations between those phases cannot occur unless the full six degrees of freedom in the unit cell (three lengths and three angles) are allowed to vary during the simulation. The current <code>MonteCarloAnisotropicBarostat</code> only allows changes in the length of each box vector, but does not allow the angles in between them to change. I presume that this feature is not implemented in OpenMM because it is usually not important in biomolecular simulations.

Other MD codes have implemented this feature (see [LAMMPS](https://lammps.sandia.gov/doc/Intro_features.html#ensembles-constraints-and-boundary-conditions) or [YAFF](https://github.com/molmod/yaff)) but are obviously inferior to OpenMM in many other ways.

My question, therefore, is three-fold:

1. **Is there an intrinsic limitation in OpenMM that would prevent box vectors from rotating?** (e.g. some neighbor list algorithm that would break) As far as I understand, the <code>MonteCarloAnisotropicBarostat</code> already requires to check at every step in the simulation if the box vectors are still in reduced form. Perhaps the existence of a limitation is dependent on the platform (i.e. it would be possible on the Reference platform, but not on either CPU or CUDA/OpenCL platforms)? Obviously, angles between box vectors don't need to go between 0 and 180 degrees. Extreme cases are, to the best of my knowledge, not really allowed in any (high-performance) MD code.

2. (assuming there IS NOT an intrinsic limitation) **Is it possible to implement a new** <code>MonteCarloFullBarostat</code> **that allows changes in the unit cell in all six degrees of freedom?** It is somewhat nontrivial to extend the existing Monte Carlo trial moves to also include angle changes, especially if one desires to rigorously prove that the scheme will generate the correct ensemble. AFAIK, this has also not been done in literature yet -- all examples of MD codes that I've come across that implement full flexibility do so through virial-based barostats. 
[...]

PR #3284, opened 2021-10-13
Created MonteCarloFlexibleBarostat

peastman:
Implements #2514.

This is a barostat that simulates a completely flexible periodic box.  All three lengths and all three angles are free to change.  It is mainly intended for materials simulations.


peastman:
Thanks!  I added the citation.

PR #3303, opened 2021-10-27
Added GLYCAM

### LINKED ISSUE #3175: addHydrogens() fails to add hydrogens to topology with glycam residues
I am trying to call `modeller.addHydrogens()` on a topology that has glycam residues, however, it fails to add hydrogens to to glycam residues. I think this is because [hydrogens.xml](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/data/hydrogens.xml) does not include glycam residues. 

How was hydrogens.xml generated? If there's existing code that takes a forcefield and generates a hydrogens.xml file, could you point me to it? 

Tagging @peastman  for feedback!

peastman:
I think I have everything working correctly.  @zhang-ivy @tristanic if you have time to look this over, that would be great.  I have a test case that computes the energy of a simple glycopeptide, but I don't have anything to compare it to.  With enough effort I can probably figure out how to compute a reference value with Amber, but if someone who is more of an Amber user than me can do it, that would also be very much appreciated!

I had to make a small change to the API for template matchers.  It was already marked as experimental and subject to change, so I think that's acceptable, but be aware that this breaks the existing GLYCAM version in openmmforcefields.  That version is already broken in other ways, though, such as the incorrect `<ExternalBond>` tags for terminal residues.

Fixes #3175.


zhang-ivy:
Yes, agreed with Tristan -- I'm not sure why they included HYP and related variants in the GLYCAM force field.

I've just reviewed everything else in the PR and it looks good to me!

PR #3321, opened 2021-11-09
Removed incorrect checks for offset scale factors

### LINKED ISSUE #3319: NonbondedForce parameter offsets: Negative scale factors should be allowed
#3286 added a number of helpful checks for valid parameter values, but it also appears to have [restricted `sigmaScale` and `epsilonScale` `NonbondedForce` parameter offset scaling factors to be nonnegative](https://github.com/peastman/openmm/blob/44d10d6608639193c0acc44bded0abfe28841bba/openmmapi/src/NonbondedForceImpl.cpp#L112-L115). This does not appear to be necessary, since it should be perfectly reasonable to reduce the initial parameter values (provided the final value doesn't end up negative, but this cannot be checked ahead of time). We in fact use this in perses, since this is necessary for relative alchemical free energy calculations. (cc https://github.com/choderalab/perses/issues/890).

peastman:
Fixes #3319.

### DIFF
--- a/openmmapi/src/NonbondedForceImpl.cpp
+++ b/openmmapi/src/NonbondedForceImpl.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2008-2018 Stanford University and the Authors.      *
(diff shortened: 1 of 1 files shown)


zhang-ivy:
@peastman : Thanks for this PR! 

It looks like one of the tests is failing (not sure if this is a showstopper), but could you merge this asap? I'm trying to use the dev version of openmm with the GLYCAM changes, but I cannot do so because in my perses hybrid systems, I have offsets that are < 0.

PR #3326, opened 2021-11-12
Removed more unnecessarily restrictive parameter checks

### LINKED ISSUE #3325: Exception raised when creating exception with zero sigma and epsilon
Thanks for packaging betas - it's a huge help to be able to test against upstream changes without needing to build locally.

One of the additions in https://github.com/openmm/openmm/pull/3286 was a requirement that sigma values in exceptions must be non-negative. Forbidding negative sigma makes sense as that's non-physical. But there are some cases in which programmatically generating exceptions can hit the case of sigma being zero (along with epsilon also being zero). Five-site water models are one example; it's hard to imagine what a sigma should be for the interactions that have zero epsilon, so a zero sigma can sneak in and not cause too much trouble.

The exception is raised [here](https://github.com/openmm/openmm/blob/7.7.0beta/openmmapi/src/NonbondedForceImpl.cpp#L96-L97) and is new since version 7.6. More or less I'm proposing that it still throws on negative sigma but only throws on zero sigma if epsilon in non-zero. This could (and maybe should) be handled downstream, moving that logic to where we add virtual sites and their interactions to the force object. But it could also be a small patch directly in OpenMM. What do you think?

peastman:
Fixes #3325.

### DIFF
--- a/openmmapi/src/HarmonicAngleForceImpl.cpp
+++ b/openmmapi/src/HarmonicAngleForceImpl.cpp
@@ -64,8 +64,6 @@ void HarmonicAngleForceImpl::initialize(ContextImpl& context) {
         }
         if (angle < 0 || angle > M_PI*1.000001)
             throw OpenMMException("HarmonicAngleForce: angle must be between 0 and pi");
-        if (k < 0)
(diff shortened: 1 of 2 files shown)


mattwthompson:
Thanks!

PR #3332, opened 2021-11-16
More detailed instructions for using local builds.

ijpulidos:
I added some instructions on how to test local build and how to use it with other conda packages, which I believe must be a common case scenario.

Specifically, I changed the order of `make test` to test the local build before install (which I think it is the correct logical order) and added some instructions on how to use the local build of `openmm` alongside conda tools/packages that need it and could potentially overwrite the local build installation.

### DIFF
--- a/docs-source/usersguide/library/02_compiling.rst
+++ b/docs-source/usersguide/library/02_compiling.rst
@@ -12,6 +12,8 @@ your environment.
 We first describe how to build on Linux or Mac.  We then describe how to build
 on Windows, where the process is slightly different.
 
+.. _compiling-openmm-from-source-linux:
+
 Compiling on Linux and Mac
 **************************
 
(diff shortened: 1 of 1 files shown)


docs-source/usersguide/library/02_compiling.rst
```diff
@@ -115,36 +118,24 @@ Configure (press “c”) again.  Adjust any variables that cause an error.
...
+    make
 
 .. _test-your-build:
 
-Step 4: Test your build
+Step 3: Test your build
 =======================
 
+This step is optional but recommended. Tests can take up to several minutes depending on your
+hardware configuration.
```
peastman, line 134: That's a bit of an underestimate!  When building all platforms, it can easily take half an hour or more to run all tests.

peastman:
Looks good!  I just started the tests running.

peastman:
Thanks!

Issue #3293, opened 2021-10-19
amoebagk and amoebapme bencmarks fail on apple m1 chips


mikemhenry:
When running the `benchmark.py` script in the `examples` folder, I get these errors for the `amoebagk` and `amoebapme` benchmark.

`amoebagk`
<details>

```
Test: amoebagk (epsilon=1e-05)
Ensemble: NVT
Step Size: 2 fs
UNSUPPORTED (log once): buildComputeProgram: cl2Metal failed
OpenCL internal error: [CL_DEVICE_NOT_AVAILABLE] : OpenCL Error : Error: Build Program driver returned (10014)
OpenCL internal error: OpenCL Warning : clBuildProgram failed: could not build program for 0x1027f00 (Apple M1) (err:-2)
OpenCL internal error: [CL_BUILD_ERROR] : OpenCL Build Error : Compiler build log:
Compilation failed: 

program_source:138:13: error: call to 'atom_add' is ambiguous
            ATOMIC_ADD(&tempForceBuffers[atom2], (mm_ulong) fx2);
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
program_source:65:33: note: expanded from macro 'ATOMIC_ADD'
#define ATOMIC_ADD(dest, value) atom_add(dest, value)
                                ^~~~~~~~
/System/Library/PrivateFrameworks/GPUCompiler.framework/Versions/31001/Libraries/lib/clang/31001.184/include/opencl-c.h:13370:12: note: candidate function
int __ovld atom_add(volatile __global int *p, int val);
           ^
/System/Library/PrivateFrameworks/GPUCompiler.framework/Versions/31001/Libraries/lib/clang/31001.184/include/opencl-c.h:13371:21: note: candidate function
unsigned int __ovld atom_add(volatile __global unsigned int *p, unsigned int val);
                    ^
/System/Library/PrivateFrameworks/GPUCompiler.framework/Versions/31001/Libraries/lib/clang/31001.184/include/opencl-c.h:13374:12: note: candidate function
int __ovld atom_add(volatile __local int *p, int val);
           ^
/System/Library/PrivateFrameworks/GPUCompiler.framework/Versions/31001/Libraries/lib/clang/31001.184/include/opencl-c.h:13375:21: note: candidate function
unsigned int __ovld atom_add(volatile __local unsigned int *p, unsigned int val);
                    ^
[...]

Issue #101, opened 2013-08-12
OpenMM cannot load a larger inpcrd/prmtop


leeping:
Hi there,

When building examples for the OpenMM command line program, I found that it can't load one of the larger AMBER example files.  First, it was unable to load the coordinates because the .inpcrd file had floats that were running together.  I was able to work around this by replacing the last decimal of precision with a space.  

Next, it wasn't able to load the .prmtop file with a more cryptic error message, which I wasn't able to solve.

The files are here:

https://dl.dropboxusercontent.com/u/5381783/dppc_128.inpcrd
https://dl.dropboxusercontent.com/u/5381783/dppc_128.prmtop

First error:
```
>>> inpcrd = app.AmberInpcrdFile('dppc_128.inpcrd')
Traceback (most recent call last):
  File "/home/leeping/local/bin/openmm", line 5, in <module>
    pkg_resources.run_script('openmm-cmd==0.1', 'openmm')
  File "/home/leeping/.local/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 489, in run_script
  File "/home/leeping/.local/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 1207, in run_script
  File "/home/leeping/local/lib/python2.7/site-packages/openmm_cmd-0.1-py2.7.egg/EGG-INFO/scripts/openmm", line 1018, in <module>
    openmm.start()
  File "/home/leeping/local/lib/python2.7/site-packages/openmm_cmd-0.1-py2.7.egg/EGG-INFO/scripts/openmm", line 766, in start
    topology = self.general.get_topology()
  File "/home/leeping/local/lib/python2.7/site-packages/openmm_cmd-0.1-py2.7.egg/EGG-INFO/scripts/openmm", line 253, in get_topology
    self.load_coords()
  File "/home/leeping/local/lib/python2.7/site-packages/openmm_cmd-0.1-py2.7.egg/EGG-INFO/scripts/openmm", line 134, in load_coords
    self.inpcrd_file = app.AmberInpcrdFile(self.coords)
  File "/home/leeping/.local/lib/python2.7/site-packages/simtk/openmm/app/amberinpcrdfile.py", line 57, in __init__
    results = amber_file_parser.readAmberCoordinates(file, read_velocities=loadVelocities, read_box=loadBoxVectors)
[...]

PR #3335, opened 2021-11-17
Add M1 build testing to GHA

mikemhenry:
I'll need access to the org or the repo (we could also just do a quick zoom and I could get the token I need) to set up the runner, but I've got a mac m1 mini setup to run a github actions runner. This should allow us to test building openmm on the new mac m1 chip arm64 arch.


mikemhenry:
:rofl: I should have squashed my commits even before making this PR but as you can see it took some trial and error...

To see the results of the action, you can check my PR here:
https://github.com/mikemhenry/openmm/pull/1
and scroll down to the checks to see that everything is working :smile:

peastman:
Looking at your output, it appears none of the tests actually ran.  They all have messages like

> Could not find executable /Users/lilmac/Projects/GH-Runner/actions-runner/_work/openmm/openmm/build/TestSerializePeriodicTorsionForce

Then a long list of places it looked for the executable and didn't find it, and then finally

> 101/114 Test #101: TestSerializePeriodicTorsionForce .............***Not Run   0.00 sec

mikemhenry:
@peastman Sorry about that! At the end I simplified some logic but forgot to remove an `if` now I think it looks good https://github.com/mikemhenry/openmm/runs/4260087324?check_suite_focus=true

peastman:
Can we make it build the OpenCL platform as well, like on Intel?  Even better would be if we could run the OpenCL tests too.  I'm not sure why we don't run them on the Intel Mac.  Maybe it just took too long?  The M1 is completing the entire build and test in only five minutes, so we can afford to run a lot more tests.

devtools/ci/gh-actions/conda-envs/build-M1-arm64.yml
```diff
@@ -0,0 +1,19 @@
...
+# build
+- cmake
+- ccache
+# host
+- python
+- cython
+- swig
+- fftw
+- numpy
+- doxygen
```
peastman, line 15: Let's pin doxygen to 1.9.1, which is the current latest version.  Otherwise there's a high chance of the build spontaneously breaking when a new version changes something in the output format.

peastman:
Actually, I think the reason we don't run OpenCL tests for Intel Mac is that it runs inside a VM and can't access the GPU.  That shouldn't be a problem for this build.

mikemhenry:
I got an error when building the OpenCL code that was the same in this issue https://github.com/openmm/openmm/issues/3293
I'll add it back into the build and we can try and troubleshoot it, we might need add an macro to skip that part of the OpenCL build on M1 hardware (or hardware that lacks those OpenCL capabilities).

devtools/ci/gh-actions/conda-envs/build-M1-arm64.yml
mikemhenry: I'll give that a try, I think I had an issue with environment resolving but that was at a phase of this project where I had lots of things broken at the same time. I might just need to look into how to get conda-forge to cross-compile/rebuld an arm64 build of an older doxygen version

jchodera:
@mikemhenry : Didn't https://github.com/openmm/openmm/pull/3336 fix the doxygen 1.9.2 issue? If not, should we fix that rather than pin?

peastman:
> I got an error when building the OpenCL code that was the same in this issue #3293

That should only impact running the tests.  It shouldn't cause any problems for building the code.  But you're right, we'll get exceptions when trying to test any force that requires 64 bit atomics.  Maybe best to skip running them for the moment.  Eventually I'd like to try to figure out a way to run all the tests that are expected to work.

mikemhenry:
@jchodera I misread the doxygen version that Peter suggested, it builds fine with that pin!

jchodera:
> That should only impact running the tests. It shouldn't cause any problems for building the code. But you're right, we'll get exceptions when trying to test any force that requires 64 bit atomics. Maybe best to skip running them for the moment. Eventually I'd like to try to figure out a way to run all the tests that are expected to work.

Are we going to be unable to use a number of forces on M1 hardware as a result? Or does this just prevent us from using `mixed` or `double` precision?

peastman:
> @mikemhenry : Didn't #3336 fix the doxygen 1.9.2 issue? If not, should we fix that rather than pin?

The code should now work with 1.9.2, but 1.9.1 is the latest version that's available through conda-forge.  But if we don't pin, the build will break the next time another Doxygen version changes the output format in some way.

jchodera:
> The code should now work with 1.9.2, but 1.9.1 is the latest version that's available through conda-forge. But if we don't pin, the build will break the next time another Doxygen version changes the output format in some way.

Don't we want to catch that?

peastman:
> Are we going to be unable to use a number of forces on M1 hardware as a result?

Correct.  The following forces require 64 bit atomics:

- CustomCentroidBondForce
- CustomManyParticleForce
- GayBerneForce
- AmoebaMultipoleForce
- HippoNonbondedForce

In addition, if you want to have two virtual sites whose positions both depend on the same atom (such as in TIP5P), that requires 64 bit atomics.

This only applies to the GPU, of course.  If you use the CPU platform all of these should work.

peastman:
> Don't we want to catch that?

Since Doxygen is only used at build time, we've never tried to supported arbitrary versions.  We just pin the builds to a particular version and only update what version we use very infrequently.

mikemhenry:
Okay @peastman I turned back on the OpenCL build and got doxygen pinned, looks like it works https://github.com/mikemhenry/openmm/runs/4292551292?check_suite_focus=true

```
OpenMM Version: 7.6
Git Revision: a9322c2a9376fa6709c7c0f9944b2150a60f79f5

3 OpenCL - Successfully computed forces

Reference vs. CPU: 6.28287e-06
Reference vs. OpenCL: 6.73789e-06
CPU vs. OpenCL: 7.37492e-07
```

peastman:
Where does that git revision come from?

```
Git Revision: a9322c2a9376fa6709c7c0f9944b2150a60f79f5
```

That doesn't correspond to any revision in the project history.  Once it's merged, will it correctly build the expected branch?

mikemhenry:
RE the git revision, I'm not sure how that is calculated as I also am also unable to find it:
`fatal: reference is not a tree: a9322c2a9376fa6709c7c0f9944b2150a60f79f5`

Before merging this PR, I'd like to setup the runner to work with the openmm org, so that we can make sure everything is working as we would expect. 

If you go here: https://github.com/organizations/openmm/settings/actions/runners/new?arch=x64&os=osx
Under "configure" there is a line that looks like this:
`./config.sh --url https://github.com/samplchallenges --token CAZYRXTGWHRLUB89FAQTW7LBTQVMY`

If you could DM me the token on the openmm slack (or email, whatever is easiest) then I can setup the runner and it should work for the entire `openmm` org, which will be nice if we want to use it for future repositories.

mikemhenry:
@peastman looks like the runner now is working as part of the openmm org! Any more changes you want to see before we merge it in?

peastman:
Excellent!  And although it doesn't run the OpenCL tests, it does at least run `testInstallation`, which reports that OpenCL is available and produced correct results.

Thanks!

jchodera:
Would it be possible to get the OpenCL tests to run as well?

peastman:
We just discussed that.  https://github.com/openmm/openmm/pull/3335#issuecomment-975952841

PR #3338, opened 2021-11-18
 Final HIP Platform implementation for AMD GPUs on ROCm

AJcodes:
Hi Dr. Eastman and co.,

This MR continues the work that was done [here](https://github.com/openmm/openmm/pull/2736), where we made sure that the current code is up to date with the latest release of the OpenMM master branch, and supports all the same functionality as the current CUDA platform (this includes all of the CUDA-specific plugins too, such as amoeba).

Additionally, further optimisations have been made in this MR to ensure the best possible performance on AMD GPUs. Such optimisations include load-balancing, the use of shuffle operations, etc that are specific to the HIP platform.

As for testing, we focused on two sets of tests; the unit tests within the repository and the tests from `openmmtools`. On our side, the unit tests within the repository all pass. I've attached a log from the tests we ran from `openmmtools` which we ran on a Mi100 (64 tests pass, and 4 fail, but this is what we also got when running the tests on a RTX3090).

[2021-11-10-openmmtools-0-mi100-fixed-amoeba.log](https://github.com/openmm/openmm/files/7561882/2021-11-10-openmmtools-0-mi100-fixed-amoeba.log)

Given that the scope of this PR is extensive, I'd like to discuss the next steps for this (e.g. how can we help to ease the reviewing process)


peastman:
Thanks for your work on this!  Let's discuss some options for moving forward.

We had a lot of discussion on the first PR.  See especially my comments at https://github.com/openmm/openmm/pull/2736#issuecomment-642881505.  Merging this into the main OpenMM code base isn't going to happen in the foreseeable future.  It's a huge amount of new code (over 43,000 lines added) which I would be accepting responsibility for maintaining, and I simply don't have the bandwidth to do it.  Given the limited number of our users with AMD GPUs, and the fact that we already support them with OpenCL, I just can't justify adding it.

If this is something you're willing to maintain long term, a much better option would be to keep it in its own repository as a separate plugin.  If we can make it conda installable, it will be very easy for users to get, and it will automatically be detected and used at runtime.  We'll be happy to work with you in getting all of that set up.

giadefa:
out of curiosity, what is the speed difference of HIP against the OpenCL
implementation and what against a RTX3090?

jchodera:
Does the forthcoming [AMD MI250](https://www.tomshardware.com/news/amd-throws-down-gauntlet-to-nvidia-with-instinct-mi250-benchmarks) change the equation any?

AJcodes:
> If this is something you're willing to maintain long term, a much better option would be to keep it in its own repository as a separate plugin. If we can make it conda installable, it will be very easy for users to get, and it will automatically be detected and used at runtime. We'll be happy to work with you in getting all of that set up.

@peastman Thanks for your reply, perhaps we could get started with making this conda installable. What are the next steps for this?

AJcodes:
> Does the forthcoming [AMD MI250](https://www.tomshardware.com/news/amd-throws-down-gauntlet-to-nvidia-with-instinct-mi250-benchmarks) change the equation any?

@jchodera The optimisations in this MR do take the Mi250 into account

peastman:
Great!  Let's do that.

The first thing you'll want to do is bring this up to date with the current OpenMM code.  It looks like it branched off a while ago, and github reports conflicts that would prevent merging it.  So merge the latest changes from the master branch, fix conflicts, and get it working again.

The next step is to move it into its own repository and allow it to be built independently.  Mostly that should be straightforward.  It's already an independent plugin, so you probably won't need to make any code changes.  But the CMake scripts will require some reworking to separate it out.  You might want to look at the [example plugin](https://github.com/openmm/openmmexampleplugin), which can serve as an example for doing that.

Once that is done, the final step is to create a feedstock on conda-forge for packaging it.  We can help with that once you get to that point.

peastman:
Since you'll be maintaining it, it would be best to have it in a repository you own.  That repository won't include all of OpenMM.  It will include only the code for the new plugin.  This is similar to how we handle other plugins that are distributed separately, like https://github.com/openmm/openmm-torch and https://github.com/openmm/openmm-plumed/.  Those repositories can also serve as useful examples.

AJcodes:
@peastman One last question on the topic; would it be possible to have a repository in the OpenMM Github group (e.g. `openmm/openmm-hip`)?

peastman:
I'd need to discuss it with all the PIs.  Putting it in the openmm organization could imply it's code we maintain and support.

None of this will matter to users, of course.  They'll just type

```
conda install -c conda-forge openmm-hip
```

jchodera:
@AJcodes : Huge thanks for all the effort in getting this fully working!

A few questions that will help us out when the OpenMM team meets soon:
* Have you folks run the full set of OpenMM benchmark suite on any hardware? Especially interesting is the comparison of OpenCL and HIP
* If we did bring this into the OpenMM GitHub org as a separate repo for the plugin, what kind of commitment to help maintain the code could be provided? If the commitment was limited, could we have a warning that this is experimental?
* Is there any interest in having the HIP platform also available on the OpenMM core deployed across Folding@home, or is the primary interest in achieving top performance on high-end datacenter-grade GPUs?

If it would be easier to discuss some of these topics in a call, just reach out [by email](https://www.choderalab.org/members) and I can help set something up.

muziqaz:
I also would like to thank @AJcodes for incredible work done regarding this. On top of that I would like to add few arguments for adding HIP support to OpenMM.
While OpenCL support is there for AMD GPUs, performance is really poor to put it politely. Moreover, AMD seem to have started to distance themselves from OpenCL support in past year. We are seeing up to 50% of Folding@Home (and other distributed computing projects) performance degradation going from driver version 21.3.2 (Windows) to anything after that. Even with 21.3.2 drivers AMD is far behind nVidia CUDA initiative, so add another 50% drop in performance, and we have no choice, but to exclude all the AMD hardware from future F@H projects. AMD seem to be completely oblivious to our cries about it. So I figure they are now fully concentrating on HIP instead of OpenCL. Having HIP support on OpenMM would save the existing base of AMD users at F@H (and other distributed computing projects), and it would not put off new users, which are plenty, due to AMD being very competitive across the board nowadays. We see incredible inroads done by AMD in various GPU markets, which haven't happened for very long time, if ever. It would be beneficial to OpenMM project to have HIP support, otherwise we end up looking stagnant in regards to AMD hardware by settling for OpenCL.
I know AMD themselves are big part of the issue here.
If there is a need of AMD hardware to do tests on it should be possible to make arrangements to get access to Vega64, Radeon 7, 5700xt and 6900xt (Windows only, though) remotely.
Regards

Asghan86:
If it helps, I have a AMD RX 6900XT (directly from AMD, no AOB) and can offer Remote Access for testing stuff if needed.

xCaradhras:
No updates in almost 3 month...is this initiative dead?

ex-rzr:
> No updates in almost 3 month...is this initiative dead?

No, it's not dead. You can check it: https://github.com/StreamHPC/openmm-hip

We discussed with OpenMM team: the new approach is to split HIP-related changes in two parts: the HIP backend as a plugin and some changes in OpenMM (required for HIP - in common kernels etc.).

I'll merge the recent changes to https://github.com/StreamHPC/openmm and resolve conflicts, then we'll update this PR (or perhaps create a new one).

AJcodes:
@peastman @jchodera As discussed, @ex-rzr has completed the split. This PR now contains changes in OpenMM (in the common kernels, etc) that are required for HIP, and is ready for review

peastman:
Thanks!  I'll start going through it and making comments.  It may take me a little while--there's a lot to look through!

Could you comment on your changes to the benchmarking script?  Some of them are obvious changes to support HIP, but others seemed to be adding new features, and it wasn't clear what the reason for them was.

peastman:
One note on naming.  We use the word "warp" to refer to a block of 32 threads, even when that doesn't match the SIMD width of the processor.  It's kind of an abuse of notation, but we use it pretty widely.  For example, the `SYNC_WARPS` macro ensures that groups of 32 threads are synchronized.  On Intel GPUs where the SIMD width is only 16, that means synchronizing the entire block.  It's a bit more than strictly required, and possibly hurts performance a bit.  But not very much, and anyway, we aren't too concerned about getting the best possible performance on very low end GPUs.

platforms/common/src/kernels/customCentroidBond.cc
```diff
@@ -3,14 +3,15 @@
...
+    //
+    LOCAL volatile real3 temp[BLOCK_SIZE];
     for (int group = GROUP_ID; group < numParticleGroups; group += NUM_GROUPS) {
         // The threads in this block work together to compute the center one group.
 
         int firstIndex = groupOffsets[group];
         int lastIndex = groupOffsets[group+1];
         real3 center = make_real3(0);
-        for (int index = LOCAL_ID; index < lastIndex-firstIndex; index += LOCAL_SIZE) {
+        for (int index = LOCAL_ID; index < lastIndex-firstIndex; index += BLOCK_SIZE) {
```
peastman, line 14: How is `BLOCK_SIZE` different from `LOCAL_SIZE`?

In other kernels where we need to define a macro for the thread block size, we call it `THREAD_BLOCK_SIZE` to be clear.

platforms/common/include/openmm/common/CommonKernels.h
```diff
@@ -570,6 +570,9 @@ class CommonCalcCustomCentroidBondForceKernel : public CalcCustomCentroidBondFor
 private:
     class ForceInfo;
     int numGroups, numBonds;
+    // note -- we assume a blocksize of 64 here, i.e., 2 warps on NV, and a single wavefront on AMD
+    //         this seems to match the behaviour in the CUDA version.
+    const int blockSize;
```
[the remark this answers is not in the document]
peastman: Never mind.  Github had just collapsed the file where it was being used.

platforms/common/src/CommonKernels.cpp
```diff
@@ -1664,6 +1665,8 @@ void CommonCalcCustomCentroidBondForceKernel::initialize(const System& system, c
     replacements["COMPUTE_FORCE"] = compute.str();
     replacements["INIT_PARAM_DERIVS"] = initParamDerivs.str();
     replacements["SAVE_PARAM_DERIVS"] = saveParamDerivs.str();
+    replacements["BLOCK_SIZE"] = cc.intToString(this->blockSize);
+    replacements["WARP_SIZE"] = cc.intToString(cc.getSIMDWidth());
```
peastman, line 1669: Following on my earlier comments, "warp size" is always defined as 32, regardless of the hardware.  Be aware that `getSIMDWidth()` is not entirely reliable, since there's no standard mechanism for determining it in OpenCL.  If we aren't sure what the width is, that method returns 1.

gunnarre:
> We discussed with OpenMM team: the new approach is to split HIP-related changes in two parts: the HIP backend as a plugin and some changes in OpenMM (required for HIP - in common kernels etc.).
Has it been decided to integrate HIP support into the Folding at Home GPU core too?

platforms/common/src/CommonKernels.cpp
```diff
@@ -2083,16 +2086,35 @@ void CommonCalcCustomNonbondedForceKernel::initialize(const System& system, cons
...
+    return (operand<<operant);
+}
+
+
+template <typename tileflags, typename tileflags4>
 void CommonCalcCustomNonbondedForceKernel::initInteractionGroups(const CustomNonbondedForce& force, const string& interactionSource, const vector<string>& tableTypes) {
     // Process groups to form tiles.
-    
+
+    const int tileSize = cc.getTileSize();
```
peastman, line 2109: The changes in this class are based on the assumption that the tile size used for computing interaction groups is the same as the size used for standard nonbonded interactions.  They both happen to be 32, but there's no reason they need to be the same.  They're computed with different code using different neighbor lists.  I think it's best to leave it as it is, with the size just hardcoded to 32.  Some day we might (or might not) decide it would be useful to make that configurable, but either way it shouldn't be required to match the main nonbonded kernel.

platforms/common/src/CommonKernels.cpp
```diff
@@ -3011,7 +3048,8 @@ void CommonCalcCustomGBForceKernel::initialize(const System& system, const Custo
         pairValueDefines["NUM_ATOMS"] = cc.intToString(cc.getNumAtoms());
         pairValueDefines["PADDED_NUM_ATOMS"] = cc.intToString(cc.getPaddedNumAtoms());
         pairValueDefines["NUM_BLOCKS"] = cc.intToString(numAtomBlocks);
-        pairValueDefines["TILE_SIZE"] = "32";
+        pairValueDefines["TILE_SIZE"] = cc.intToString(tileSize);
+        pairValueDefines["tileflags"] = (tileSize > 32 ? "unsigned long" : "unsigned int");
```
peastman, line 3052: That should be `mm_ulong` rather than `unsigned long`.  `long` is 32 bits in CUDA, 64 bits in OpenCL.

peastman: And likewise in similar code below.

platforms/common/src/CommonKernels.cpp
```diff
@@ -4390,17 +4430,19 @@ CommonCalcCustomManyParticleForceKernel::~CommonCalcCustomManyParticleForceKerne
 
 void CommonCalcCustomManyParticleForceKernel::initialize(const System& system, const CustomManyParticleForce& force) {
     ContextSelector selector(cc);
+    const int tileSize = cc.getTileSize();
```
peastman, line 4433: This is another class that's computed with completely different code from standard nonbonded interactions.  It doesn't need to use the same tile size they do.

platforms/common/src/CommonKernels.cpp
```diff
@@ -4994,27 +5040,28 @@ void CommonCalcGayBerneForceKernel::initialize(const System& system, const GayBe
     exceptionParams.initialize<mm_float2>(cc, max(1, numExceptions), "exceptionParams");
     if (numExceptions > 0)
         exceptionParams.upload(exceptionParamsVec);
-    
+
     // Create data structures used for the neighbor list.
 
-    int numAtomBlocks = (numRealParticles+31)/32;
+    const int tileSize = cc.getTileSize();
```
peastman, line 5046: Another one that shouldn't depend on the tile size used by the main nonbonded kernel.

platforms/common/src/kernels/gbsaObc.cc
```diff
@@ -29,13 +41,23 @@ KERNEL void computeBornSum(
...
+    LOCAL volatile int skipTiles[FORCE_WORK_GROUP_SIZE];
+#else
+    AtomData1 localData;
+    AtomData1 data;
+    int skipTiles;
+#endif
 
     // First loop: process tiles that contain exclusions.
-    
+#if !defined(USE_HIP)
```
peastman, line 54: Can you explain what this change is about?  I don't understand it.

platforms/common/src/kernels/removeCM.cc
```diff
@@ -19,22 +19,26 @@ KERNEL void calcCenterOfMassMomentum(int numAtoms, GLOBAL const mixed4* RESTRICT
 
     int thread = LOCAL_ID;
     temp[thread] = cm;
+
+    // Easier to cope with varying block / wavefront sizes w/o perf. penalty if
+    // expressed as a constexpr reduction
```
peastman, line 24: Keep in mind that CM motion removal takes truly negligible time.  It's not worth adding any complexity to optimize it, because there will be no practical benefit.

platforms/cuda/include/CudaContext.h
```diff
@@ -356,6 +356,12 @@ class OPENMM_EXPORT_COMMON CudaContext : public ComputeContext {
     bool getIsCPU() const {
         return false;
     }
+    /**
+     * Get the tile size of the device being used.
```
peastman, line 360: That isn't what `TILE_SIZE` means.  It's the size of the tiles used in computing nonbonded interactions.  It doesn't have anything to do with a particular device.

jchodera:
> Has it been decided to integrate HIP support into the Folding at Home GPU core too?

@gunnarre : We would be excited to integrate HIP support into the Folding@home GPU core (core22)! Once the plugin is available on conda-forge, it should be very easy to incorporate it into a new core22 release.

platforms/cuda/src/kernels/common.cu
```diff
@@ -24,3 +24,17 @@ typedef unsigned long long mm_ulong;
...
+    return static_cast<long long>(x * 0x100000000);
+}
+
+#else
+
+__device__ inline long long realToFixedPoint(float x) {
+    return static_cast<long long>(x * 0x100000000);
+}
+
+#endif
```
peastman, line 40: We don't need separate versions.  They can be combined to just

```c++
__device__ inline long long realToFixedPoint(real x) {
    return static_cast<long long>(x * 0x100000000);
}
```

plugins/amoeba/platforms/common/src/kernels/hippoNonbonded.cc
```diff
@@ -19,11 +19,21 @@ typedef struct {
...
+#else
+
 //support for 64 bit shuffles
 static __inline__ __device__ float real_shfl(float var, int srcLane) {
     return SHFL(var, srcLane);
 }
 
+static __inline__ __device__ int real_shfl(int var, int srcLane) {
+    return SHFL(var, srcLane);
+}
```
peastman, line 35: I don't understand this function.  `real_shfl()` operates on `real`.  If we want to shuffle an `int`, that's what `SHFL()` already does.

peastman:
I made a number of comments about tiles while going through the code.  I'd like to add a bit of background information to explain what tiles are all about.

Suppose we computed every nonbonded interaction independently.  For each one we would need to load data on two atoms, compute the interaction, and save forces for two atoms.  That's really inefficient in terms of memory.  Two atoms loaded and saved for each interaction.

Instead we compute them in tiles: 32 atoms interacting with 32 other atoms.  We load data on 64 atoms, compute 1024 interactions, and save data on 64 interactions.  That gives a much higher ratio of compute to memory.

There's a problem though: the neighbor list is coarse grained, which leads to computing unnecessary interactions.  If an atom in the second set of 32 interacts with any one of the atoms in the first set of 32, we end up having to compute its interaction with all 32 of them.  A large fraction of computed interactions end up being wasted.  We know they're beyond the cutoff distance, but the coarse grained neighbor list forces us to compute them anyway.

The tile size is a tradeoff between those two factors.  Increasing it makes memory access more efficient but makes compute less efficient.  We've found that 32 gives a good tradeoff between them.  It happens to be the same as the SIMD width on NVIDIA GPUs, but that's mostly a coincidence.  We use the same tile size on GPUs where the SIMD width is 16 (Intel) or 64 (some AMD).

This relates to the particular way calculations are done in a particular kernel.  Other kernels work differently and involve different tradeoffs.  For example, CustomManyParticleForce structures its calculation in a totally different way.  The tile size used for standard nonbonded interactions doesn't have anything to do with them.

ex-rzr:
@peastman 

Thank you very much for reviewing. I'll start answer step by step as there are a lot of things to cover.

plugins/amoeba/platforms/common/src/kernels/hippoNonbonded.cc
ex-rzr: Good point. 
Then, I guess, `real_shfl` should be renamed/removed everywhere.
In this file you already have its overload for `int` (I don't think it's used though).
I just made it consistent with nonbonded.cu. where real_shfl is used with, for example, with `int shflatomType` and `mm_long shflobc4_bornForce`.

ex-rzr:
Let's focus on tile size only for now and leave less important things for later.

The main goal of adding tile size as a method instead of hardcoded value (32) is to compare how AMD GPUs with warps ("wavefronts" in AMD terms) of 64 threads (CDNA) perform with 32x32 and 64x64 atoms in tiles.
It turned out in most cases we get better performance on MI cards when we use tileSize = warpSize = 64, especially for larger simulations, like amber20-cellulose (17% faster) and amber20-stmv (22% faster).

So our question is what do you suggest?

1) Revert all tile size-related changes to hard-coded 32?
2) Keep them for the "standard nonbonded interactions" but revert for other cases?
3) Implement it but somehow differently? How? Add different tile size variables for each case where it's used?

You said:

> This relates to the particular way calculations are done in a particular kernel. Other kernels work differently and involve different tradeoffs. For example, CustomManyParticleForce structures its calculation in a totally different way. The tile size used for standard nonbonded interactions doesn't have anything to do with them.

> The changes in this class are based on the assumption that the tile size used for computing interaction groups is the same as the size used for standard nonbonded interactions. They both happen to be 32, but there's no reason they need to be the same. They're computed with different code using different neighbor lists. I think it's best to leave it as it is, with the size just hardcoded to 32. Some day we might (or might not) decide it would be useful to make that configurable, but either way it shouldn't be required to match the main nonbonded kernel.

But what will happen if you decide that another value of tile size is better for one of many `*Kernels` from `platforms/common/src/CommonKernels.cpp`? There are quite many hidden dependencies on the "standard nonbonded" tile size, for example: getPaddedNumAtoms, getExclusionTiles.

muziqaz:
> Let's focus on tile size only for now and leave less important things for later.
> 
> The main goal of adding tile size as a method instead of hardcoded value (32) is to compare how AMD GPUs with warps ("wavefronts" in AMD terms) of 64 threads (CDNA) perform with 32x32 and 64x64 atoms in tiles. It turned out in most cases we get better performance on MI cards when we use tileSize = warpSize = 64, especially for larger simulations, like amber20-cellulose (17% faster) and amber20-stmv (22% faster).
> 

If I understand the concept of the tile correctly, this is similar to what Blender (or any other 3D Renderer) is doing when rendering the scene. They split the scene into tiles which are certain size. And then CPU core/GPU is rendering those tiles(chunks) one by one. It has been known that top end GCN GPUs would benefit from increased tile size compared to default smaller tile sizes. I am not sure how RDNA+ arch behaves in regards to that, since I haven't rendered anything on new cards yet.

How complicated would the addition of manually adjustable tile size option would be?
For example: in F@H we have 700k atom project, which is only being run on high end GPUs, so project owner could set tile size to 64 or more to see if that benefits the performance.
Obviously this is just thinking out loud, as these things might require a lot of work, and who knows maybe RDNA arch cards are not influenced too much by difference of the tile sizes.

peastman:
> Keep them for the "standard nonbonded interactions" but revert for other cases?

For right now, I think that would be the best approach.  All the other kernels need to be treated independently.  They're separate code with separate considerations.

Even for the standard nonbonded interactions (or perhaps, especially for those ones) any of this could change in the future.  For example, I'm currently thinking about possibilities for using non-square tiles, such as 16x64 instead of 32x32.  That might end up giving a better balance.

> It turned out in most cases we get better performance on MI cards when we use tileSize = warpSize = 64

You based your code on the CUDA platform, right?  There's one important change you should copy over from the OpenCL platform.  We sort the tiles with interactions differently depending on the SIMD size:

https://github.com/openmm/openmm/blob/4358f0719ba0d150af2bd814541d0de1750aebe7/platforms/opencl/src/OpenCLNonbondedUtilities.cpp#L230

Normally we sort them to maximize cache effectiveness.  But when the SIMD width is >32, that leads to thread divergence since often some threads within a wavefront will be processing a diagonal tile and others an off-diagonal tile.  So instead we sort them to put all diagonal tiles before all off-diagonal ones.  That may account for some of the speed difference you saw.

ex-rzr:
@muziqaz 
> How complicated would the addition of manually adjustable tile size option would be?

It should be easy for "standard nonbonded interactions".

> and who knows maybe RDNA arch cards are not influenced too much by difference of the tile sizes.

RDNA uses tile size = 32, just as NVidia GPUs.

ex-rzr:
@peastman 

Yes, we know about `compareInt2LargeSIMD`, it makes a minor difference.

Besides, I have a suspicion that it can make some cases incorrect. Perhaps I should open a new issue.

peastman:
Would it make things easier for you to split this into multiple PRs that can be merged individually?  The longer it remains open, the more merge conflicts will appear that have to be dealt with.  For example, a lot of the changes are just from the calls to `realToFixedPoint()`.  If we can get them merged, it will help a lot to reduce the number of affected files.  Smaller, more focused PRs will also be easier and faster to review.

ex-rzr:
@peastman
Sorry for the delay. It's a good idea. The change has been extracted: #3504
(I'll remove the change from this PR once the new PR is merged)

ex-rzr:
@peastman

Could you check the PR again?
I removed a lot of small unnecessary changes: changes in benchmark.py we used during optimizing, that WARP_SIZE-related code, etc.
I also reordered and combined commits to make changes more structured.

Regarding tile size in various common kernels. I reverted tile size to the hard-coded 32 in these kernels:
* CommonCalcCustomNonbondedForceKernel, customNonbondedGroups.cc
* CommonCalcCustomManyParticleForceKernel, customManyParticle.cc
* CommonCalcGayBerneForceKernel

But other kernels (CommonCalcGBSAOBCForceKernel, CommonCalcCustomGBForceKernel, AMOEBA) rely on the tile size used in NonbondedUtilities and call getExclusionTiles and getInteractingTiles.
What do you think, is it enough?

I even have an idea that getTileSize should be moved to NonbondedUtilities, but I'm not sure about this.

Perhaps, we can schedule a call for discussing all the changes.

ex-rzr:
Also I need your advice about conda-forge packages, if you can help.
I created conda recipe and built the packages manually:
https://anaconda.org/streamhpc/openmm (this branch, built using openmm-feedstock/build-locally.py)
https://anaconda.org/streamhpc/openmm-hip (the HIP platform only)

But there is a problem: ROCm doesn't have conda packages similar to cudatoolkit. This means that I can build the openmm-hip package only on a computer with ROCm installed from Ubuntu or CentOS packages (or inside a docker container rocm-terminal).

Do you know if there is a way to use a docker container with conda-forge? It'll be great to use conda-forge infrastructure for building openmm-hip automatically just like the main openmm package and plugins.

(I guess it's a separate discussion and I should not use this PR for such questions)

peastman:
I'll take a look, thanks.

> I even have an idea that getTileSize should be moved to NonbondedUtilities, but I'm not sure about this.

The whole concept of having the `TileSize` variable and `tileflags` type is mostly a remnant of an old design.  Once upon a time it was possible to change the type and have it run with a different tile size, but that hasn't been true in years.  For example, all the kernel code using shuffle operations would need to be rewritten to work with a different tile size now.

It's also quite likely to become even less meaningful in the future.  For example, I've been thinking about restructuring the computation to use non-square tiles (for example 32 by 16), or to allow different tiles to have different sizes.

So yes, those variables are currently used inconsistently.  But the longer term resolution may be to use them even less, or get rid of them altogether, rather than using them more.

> But there is a problem: ROCm doesn't have conda packages similar to cudatoolkit.

Conda-forge is generally pretty strict about not allowing any dependencies on anything outside.  Would it be possible to create a conda package for ROCm?

ex-rzr:
Small update regarding the conda package:

I've managed to render my recipe with conda-smithy and build it locally using build-locally.py in one of conda-forge docker container (linux-anvil-cos7-x86_64 instead of rocm-terminal as I planned previously). To do this I need to install external ROCm packages (with `yum install`): 
https://github.com/StreamHPC/openmm-hip/blob/master/recipe/build.sh#L7-L12

I'm not sure this is allowed to do, I need to ask conda-forge developers.

> Conda-forge is generally pretty strict about not allowing any dependencies on anything outside. Would it be possible to create a conda package for ROCm?

I think it's possible. I see two ways:
1) build from sources, this way will cause HUGE loads of conda-forge infrastructure, because there are heavy packages like llvm;
2) extract rpm/deb packages and "repack" them, similarly to what cudatoolkit does.
Both ways seem more complicated than what I do in the current recipe.

ex-rzr:
@peastman, could you check the current version?
I removed everything related to tile sizes and some other not very important code.

There are still some changes that are not mandatory:
* https://github.com/openmm/openmm/pull/3338/commits/70dab08ab337eee5b16d9cc054af31b00f9b881c - a more convenient way to install python packages for the current user only (we use it when multiple developers need to work on OpenMM without conflicts)
* https://github.com/openmm/openmm/pull/3338/commits/f34057e3b949cdfeb03b3c339d29d52d3cb6c5d8 - the comment says what the change does. I added it after I spent an hour trying to understand why a changed kernel worked incorrectly. It turned out that during development I copied some code with commands like `SHUFFLE_WARP_DATA` and commented one copy with `//` which in the final code make only the first line commented. This check handles such situations.

We think it's ok to remove these changes if you don't like them, just say a word. I also can split them in separate PRs (though they are very small).

Thanks!

peastman:
Thanks!  I'll take a look.  Could you also remove the changes to benchmark.py for the moment, just because we have another open PR that rewrites large parts of that script?  Once that's merged we can make any other changes to it in a separate PR.

platforms/common/src/kernels/gbsaObc.cc
```diff
@@ -1,12 +1,25 @@
 #define WARPS_PER_GROUP (FORCE_WORK_GROUP_SIZE/TILE_SIZE)
 
+#if defined(USE_HIP)
+
+typedef struct alignas(16) {
```
peastman, line 5: Since the structures are identical other than alignment, what about writing it like this?

```c++
#if defined(USE_HIP)
    #define ALIGN alignas(16)
#else
    #define ALIGN
#endif

typedef struct ALIGN {
```

That way we don't repeat the structure definition, which risks someone changing it in one place but not the other.

peastman (commented):
This looks great.  If we can avoid duplicating all the AtomData struct definitions, that would be preferable.  Otherwise, I think it's ready to merge.

platforms/common/src/kernels/gbsaObc.cc
ex-rzr: You are right! Done.

Regarding other structs: I can use the same approach with defines to hide the differences (alignment, paddings) but they also have a different order of fields. I'm worried to rearrange fields for all platforms because I'm not sure that it won't affect performance. I can profile GBSA and Amoeba benchmarks on a few Nvidia GPUs with CUDA: 3090, Titan V, 2060, but definitely this can't cover all devices where OpenMM is used.
In my opinion, it seems safer to keep separate definitions. Also it's likely that we'll be able to remove at least some of this commit in the future because (as far as I know) developers of the HIP compiler continue to work on improving code generation of shared memory accesses.

What do you think?

peastman:
So far as I know, the order of fields in a shared memory struct should never affect performance on an NVIDIA GPU.  It's not like global memory, where the order could affect cache performance.  If you test a few GPUs and confirm it doesn't change performance, that's good enough for me.

ex-rzr:
Ok, I've combined structure definitions and profiled gbsa (computeBornSum, computeGBSAForce1), amoebagk and amoebapme (computeGKForces, computeInducedField, computeEDiffForce, computeElectrostatics) on those 3 devices: I don't see any performance drops caused by fields reordering.

peastman:
That looks good.  Which means it's ready to merge!

DanielWicz:
In which version of OpenMM, HIP is expected to be used ?

ex-rzr:
@DanielWicz

> In which version of OpenMM, HIP is expected to be used ?

This PR contains changes in common files we needed for HIP support.
The HIP platform is a plugin which you can find here: https://github.com/amd/openmm-hip

There are instructions how to build it from sources or install a package from conda (I've noticed that OpenMM 8.0.0 has been released, I'll need to rebuild conda package for this version).

If you have any questions or problems, please open an issue in that repository (and ping me, just to be sure).

egallicc:
We are testing 8.0.0rc1 with openmm-hip building from sources downloaded from [https://github.com/StreamHPC/openmm-hip](https://github.com/StreamHPC/openmm-hip). All tests are passing without changes.

egallicc:
All openmm-8.0.0rc1 + openmm-hip tests passed except `TestHipFFTImplHipFFT{Single,Mixed,Double}` because the executable could not be found.

ex-rzr:
Thanks, @egallicc

You're using RDNA GPUs, correct? What ROCm version, OS? (I'm just gathering statistics, as it's impossible to test everywhere)

> TestHipFFTImplHipFFT{Single,Mixed,Double} because the executable could not be found

That's interesting. These tests had failures on older ROCm version, but they shouldn't fail because of missing executables. Can you post the log?

egallicc:
Yes, @ex-rzr 
GPU: RX 6750 XT (with HSA_OVERRIDE_GFX_VERSION=10.3.0)
Host: Ubuntu 20.04
ROCm: 5.2.0

Test log: https://www.dropbox.com/s/k86vbmjo9hezo8q/LastTest-1.log?dl=0
(actually, it wasn't missing executables)

ex-rzr:
Ok, this is a known issue with rocFFT (a specific problem size): https://github.com/ROCmSoftwarePlatform/hipFFT/issues/26. It's fixed in recent releases of ROCm. But anyway, VkFT is used by default, so it shouldn't be a problem.

muziqaz:
https://docs.google.com/spreadsheets/d/1_FMU8mKlWb4LEp3mOwsHwiwgSMKQWn2WxNxdp7QZsfg/edit?usp=sharing
OpenMM 8.0.0 performance comparison between OpenCL and HIP on Linux AMD. RX 550 is no longer supported neither with OpenCL nor HIP in Linux
Depending on time availability, I will try compiling newer OpenMM version with newer HIP versions (if available) to see if something was done to sort out 7900xtx performance (HIP side). Also Windows HIP SDK seems to be available, will have to check that out. These tests were run from standard conda env set up.
As you can see performance increase is beyond unbelievable.

muziqaz:
One more update for you guys:
We are testing 1.2m atom project on F@H.
I ran it through HIP on 6900xt Linux Openmm 8.0:
OpenCL 8.2ns/day
HIP - 22.3ns/day
For comparison 4070ti/3080ti on CUDA - 17ns/day

Issue #3333, opened 2021-11-16
Optimize multi-GPU communication


aizvorski:
When using multi-GPU, positions get copied from one GPU to all GPUs, and after the kernel is calculated, forces get copied back and summed on one GPU.  This seems like a perfect use for the very fast [ncclBroadcast()](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#broadcast) and [ncclReduce()](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#reduce) calls in the [NCCL](https://developer.nvidia.com/nccl) library.

Looking at the code, it seems most of the work now happens here:

https://github.com/openmm/openmm/blob/f477b106fc946c4cb3b4ae7d61da43889e568c9e/platforms/cuda/src/CudaParallelKernels.cpp#L188-L201

and 

https://github.com/openmm/openmm/blob/f477b106fc946c4cb3b4ae7d61da43889e568c9e/platforms/cuda/src/CudaParallelKernels.cpp#L221-L230

It seems there are two code paths depending on whether peerAccessSupported is true or false.  If it's false, then there is a call to `CudaArray.download()` the inputs and then `CudaArray.upload()` the outputs.  If it is true, then the data is accessed using unified addressing from one GPU to another (without explicit copy).  I think on most recent GPUs, the peerAccessSupported=true code would be used (both on systems with NVLink/NVSwitch and plain PCIe).  I'm not sure of the performance of using unified memory in this way, but it's probably not optimal (especially without prefetch).  Empirically, the multi-GPU code now only reaches ~41 Gbyte/s on a 8x A100 system, vs 8x 600 Gbyte/s hardware limit; there's only 26% speedup going from 2x to 4x GPU, and no speedup at all going from 4x to 8x.

The NCCL calls use more sophisticated tree or ring [routing algorithms](https://developer.nvidia.com/blog/fast-multi-gpu-collectives-nccl/) which are aware of the topology of the links and not equivalent to using n sequential copies, and also can sum during the copy when doing reduce.  This can readily reach near-hardware-limited performance when doing operations on data >= a few megabytes.  So, I would like to request adding NCCL support to the multi-GPU code and/or possibly switching over entirely.

peastman:
[...]

PR #3347, opened 2021-11-22
Improve performance running on multiple GPUs

peastman:
cc @aizvorski @dmclark17

I'm trying to make the improvements discussed in #3333.  So far I've done the first part, using multiple streams to broadcast the positions.  Unfortunately, I don't have any hardware to properly test it on.  The best I can do is "parallelize" the computation across multiple contexts on a single GPU.  That's hopefully enough to verify I haven't broken anything, but it doesn't tell us anything about performance.

Can you try this out?  If you can check the following it would be a big help.

- Does it affect performance?  If so, how?
- Try running it in the profiler.  How does it look?  Hopefully we should now see multiple memcpy operations happening in parallel, and also each GPU should begin computing as soon as it has the positions, even if the transfer to another GPU is still in progress.
- Try running the test cases on multiple GPUs, just to verify it really still works.  You can do that by passing the `DeviceIndex` value as the second argument on the command line:

```
./TestCudaNonbondedForce single 0,1,2,3
```

### DIFF
--- a/platforms/cuda/include/CudaParallelKernels.h
+++ b/platforms/cuda/include/CudaParallelKernels.h
@@ -91,6 +91,7 @@ private:
     CUfunction sumKernel;
     CUevent event;
     std::vector<CUevent> peerCopyEvent;
+    std::vector<CUevent> peerCopyEventLocal;
     std::vector<CUstream> peerCopyStream;
 };
 
--- a/platforms/cuda/include/CudaParallelKernels.h
+++ b/platforms/cuda/include/CudaParallelKernels.h
@@ -90,7 +90,8 @@ private:
     long long* pinnedForceBuffer;
     CUfunction sumKernel;
     CUevent event;
-    CUstream peerCopyStream;
+    std::vector<CUevent> peerCopyEvent;
+    std::vector<CUstream> peerCopyStream;
 };
 
 /**
--- a/platforms/cuda/src/CudaParallelKernels.cpp
+++ b/platforms/cuda/src/CudaParallelKernels.cpp
@@ -95,10 +95,10 @@ class CudaParallelCalcForcesAndEnergyKernel::FinishComputationTask : public Cuda
 public:
     FinishComputationTask(ContextImpl& context, CudaContext& cu, CudaCalcForcesAndEnergyKernel& kernel,
             bool includeForce, bool includeEnergy, int groups, double& energy, long long& completionTime, long long* pinnedMemory, CudaArray& contextForces,
-            bool& valid, int2& interactionCount, CUstream stream, CUevent event) :
+            bool& valid, int2& interactionCount, CUstream stream, CUevent event, CUevent local_event) :
             context(context), cu(cu), kernel(kernel), includeForce(includeForce), includeEnergy(includeEnergy), groups(groups), energy(energy),
             completionTime(completionTime), pinnedMemory(pinnedMemory), contextForces(contextForces), valid(valid), interactionCount(interactionCount),
-            stream(stream), event(event) {
+            stream(stream), event(event), local_event(local_event) {
     }
     void execute() {
         // Execute the kernel, then download forces.
@@ -113,6 +113,9 @@ public:
         }
         if (includeForce) {
             if (cu.getContextIndex() > 0) {
+                cu.restoreDefaultStream();
+                cuEventRecord(local_event, cu.getCurrentStream());
+                cuStreamWaitEvent(stream, local_event, 0);
                 int numAtoms = cu.getPaddedNumAtoms();
                 if (cu.getPlatformData().peerAccessSupported) {
                     int numBytes = numAtoms*3*sizeof(long long);
@@ -145,6 +148,7 @@ private:
     int2& interactionCount;
     CUstream stream;
     CUevent event;
+    CUevent local_event;
 };
 
 CudaParallelCalcForcesAndEnergyKernel::CudaParallelCalcForcesAndEnergyKernel(string name, const Platform& platform, CudaPlatform::PlatformData& data) :
@@ -163,6 +167,8 @@ CudaParallelCalcForcesAndEnergyKernel::~CudaParallelCalcForcesAndEnergyKernel() 
     cuEventDestroy(event);
     for (int i = 0; i < peerCopyEvent.size(); i++)
         cuEventDestroy(peerCopyEvent[i]);
+    for (int i = 0; i < peerCopyEventLocal.size(); i++)
+        cuEventDestroy(peerCopyEventLocal[i]);
     for (int i = 0; i < peerCopyStream.size(); i++)
         cuStreamDestroy(peerCopyStream[i]);
     if (interactionCounts != NULL)
@@ -181,10 +187,16 @@ void CudaParallelCalcForcesAndEnergyKernel::initialize(const System& system) {
         contextNonbondedFractions[i] = 1/(double) numContexts;
     CHECK_RESULT(cuEventCreate(&event, 0), "Error creating event");
     peerCopyEvent.resize(numContexts);
+    peerCopyEventLocal.resize(numContexts);
     peerCopyStream.resize(numContexts);
     for (int i = 0; i < numContexts; i++) {
         CHECK_RESULT(cuEventCreate(&peerCopyEvent[i], 0), "Error creating event");
         CHECK_RESULT(cuStreamCreate(&peerCopyStream[i], CU_STREAM_NON_BLOCKING), "Error creating stream");
+    }
+    for (int i = 0; i < numContexts; i++) {
+        CudaContext& cu_local = *data.contexts[i];
+        ContextSelector selector_local(cu_local);
+        CHECK_RESULT(cuEventCreate(&peerCopyEventLocal[i], 0), "Error creating event");
     }
     CHECK_RESULT(cuMemHostAlloc((void**) &interactionCounts, numContexts*sizeof(int2), 0), "Error creating interaction counts buffer");
 }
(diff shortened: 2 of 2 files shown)


aizvorski:
@peastman Thanks!  I'd love to try it, but I've never built openmm from git before.  What's the recommended method?

I tried this:

```
git clone https://github.com/peastman/openmm/
cd openmm

export export CUDA=true
export OPENCL=false
export CUDA_VERSION="11.5"
export CMAKE_FLAGS="
    -DOPENMM_BUILD_CUDA_TESTS=ON
    -DOPENMM_BUILD_OPENCL_TESTS=ON
    -DOPENMM_BUILD_PYTHON_WRAPPERS=ON
    -DOPENMM_BUILD_REFERENCE_TESTS=OFF
    -DOPENMM_BUILD_SERIALIZATION_TESTS=OFF
    -DOPENMM_BUILD_C_AND_FORTRAN_WRAPPERS=OFF
    -DOPENMM_BUILD_EXAMPLES=OFF
    -DOPENCL_LIBRARY=/usr/local/cuda-11.5/lib64/libOpenCL.so
    -DCUDA_CUDART_LIBRARY=/usr/local/cuda-11.5/lib64/libcudart.so
    -DCUDA_NVCC_EXECUTABLE=/usr/local/cuda-11.5/bin/nvcc
    -DCUDA_SDK_ROOT_DIR=/usr/local/cuda-11.5/
    -DCUDA_TOOLKIT_INCLUDE=/usr/local/cuda-11.5/include
    -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-11.5/"
cmake . $CMAKE_FLAGS -DCMAKE_INSTALL_PREFIX=$HOME/OpenMM
make -j32 install
... builds and installs without errors

./TestCudaNonbondedForce single 0,1,2,3
exception: Error initializing CUDA: CUDA_ERROR_SYSTEM_DRIVER_MISMATCH (803) at /home/ubuntu/openmm/platforms/cuda/src/CudaContext.cpp:139
```

ubuntu=20.04.3, cuda=11.5 from https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/

peastman:
Instructions on building from source are at http://docs.openmm.org/latest/userguide/library/02_compiling.html.  That's a really weird error to be getting.  Here's what the documentation says about it.

> This error indicates that there is a mismatch between the versions of the display driver and the CUDA driver. Refer to the compatibility documentation for supported versions.

But I assume you have no trouble running other CUDA software (including the standard OpenMM package) on this computer?  What happens if you just run the test case normally, not specifying multiple devices?

```
./TestCudaNonbondedForce
```

dmclark17:
I was able to build and run on a DGX V100. The profile indicates that the force-reduction memcpys are being launched in different streams, but they are still being serialized. I'm not sure what is going on here, and I will investigate further.

peastman:
I haven't done anything with force reduction yet.  The current changes in this PR only affect the position broadcast.  Once I know the approach used here works, I can apply it to forces to.

dmclark17:
I had the wrong environment activated when I did the initial profile—sorry about that. It looks like the position broadcast is being overlapped now!

I am seeing a performance improvement from ~32.5 ns/day to ~36.8ns/day for 4 V100. The `./TestCudaNonbondedForce single 0,1,2,3` passes as well.

image

peastman:
Excellent!  I'll try optimizing the force reduction next.

peastman:
I made the changes to use multiple streams for force reduction.  Can you try it out?

aizvorski:
@peastman Nice - this is already a big speedup!  

I followed the build instructions and now ./TestCudaNonbondedForce works, the driver mismatch error is gone.  

After `make pythonInstall`, the individual python tests work (from within the build dir), but testInstallation.py doesn't:

```
~/openmm/build/python/tests$ python TestIntegrators.py 
...
OK
~/openmm/build/python$ python openmm/testInstallation.py 
Failed to import OpenMM packages; OpenMM will not work.
Make sure OpenMM is installed and the library path is set correctly.

Error message: No module named '_openmm'
```

Also I noticed the libraries in `miniconda3/lib/libOpenMM*` are still from the old version, while the python files in `miniconda3/lib/python3.9/site-packages/openmm/` are from the new install.  Should I set CMAKE_INSTALL_PREFIX=/path/to/miniconda3 to get the new install to replace the conda package completely?  

(Sorry I have such elementary build questions)

peastman:
>  Should I set CMAKE_INSTALL_PREFIX=/path/to/miniconda3 to get the new install to replace the conda package completely?

Correct.  Or if you were using a conda environment you would set it to the root directory of that environment (such as /path/to/miniconda3/envs/openmm).

dmclark17:
I'm seeing an error in the new version of the code with multiple devices:

```
openmm.OpenMMException: Error copying forces: CUDA_ERROR_ILLEGAL_ADDRESS (700) at /home/scratch.daclark_sw/OpenMM/change/openmm/platforms/cuda/src/CudaParallelKernels.cpp:121
  what():  Error deleting array bondParams: CUDA_ERROR_ILLEGAL_ADDRESS (700)
Aborted
```

I am actually not sure if we'll be able to overlap these memory operations and, if it is possible, if we'll see a performance improvement—I will look into this more.

peastman:
This might be related to the multiple contexts involved.  I've never been able to find a clear explanation of how this really works.  When you create a CUstream or CUevent, you do that within a particular context.  But when you're copying data between devices, there are two different contexts involved, one on each device.  Should the stream you pass to `cuMemcpyAsyc()` belong to the context for the device you're copying from, or the one you're copying to?  And how do you use an event to synchronize between things happening on two different devices, and hence in different contexts?

dmclark17:
I believe the memcpy should work with either stream, and the event-stream synchronization should be okay across different contexts. I will look into this more though.

It appears that the error is happening before the memcpy operation—the error gets picked up before the `cuMemcpyAsyc` if I insert an error check. I also don't think `cuMemcpyAsyc` can return that error code.

dmclark17:
I believe the issue is that we need a synchronization between the default stream of the non-integration GPUs and the peer copy stream—the peer copy stream appears to exist on both GPUs. This can be achieved by adding an event to the default stream and then waiting on it in the peer copy stream. I believe we need to introduce a new set of events that belong to the non-integration GPU's context to do this.

Doing this seems to resolve the correctness issue. I am now seeing overlap with the force reduction operations and am measuring about 43.4ns/day on 4 V100:
image

peastman:
Very nice!  Can you post your code modifications?

peastman:
In the profile shown above, I notice the computation isn't balanced very well between the GPUs.  Was that step near the start of the run, or was it after 200 or more time steps?  If the latter, that suggests the load balancing code isn't working very well.  At some point I'd like to experiment with other load balancing strategies to see if I can further improve the performance.

dmclark17:
I believe that was after more than 200 time steps, but I am not 100% sure. I have attached a new screenshot that is after more than 200 timesteps and includes both cases for `findBlocksWithInteractions`:

image

peastman:
The fixes are merged.  Can you verify that it works correctly?  I think it's ready to merge then.  There's probably some room to improve the load balancing, but that can wait until later.  This is already a huge improvement from before.

peastman:
I also need to see if I can do something about `sortBuckets()`.  When running on just one GPU, it only takes a minor amount of time.  But when `findBlocksWithInteractions()` and `computeNonbonded()` are getting divided between four GPUs, it becomes a much bigger fraction.

dmclark17:
I found that `./TestCudaNonbondedForce single 0,1,2,3` passes, and the amber20-stmv benchmark runs without issue. 

I also tested on a DGX A100 and saw the following results. The improvement isn't quite as significant compared to the V100 system; I believe this is because of the improved memory bandwidth between GPUs. 

GPUs | ns/day
-- | --
1 | 33.07
2 | 53.19
4 | 73.05
8 | 81.92


Here is a screenshot of the timeline:

image

peastman:
Thanks!  That's still a nice improvement over what it was before.  And based on the profile, I think there's room for further improvement.  If I can get access to a good multi-GPU system, I'll see what I can do with it.

aizvorski:
@peastman @dmclark17 This is great, thank you both!

This is a really nice improvement, but also even better it removes a lot of the bottlenecks when running on very large number of GPUs (8-16 or so).  Any load balancing improvements, sortBuckets etc would make an even bigger difference with this patch in.

@peastman I'll contact you separately to see if I can help set up access to a multi-GPU dev system for you.

peastman:
Thanks, that would be great.  I'm also hoping NVIDIA can provide access to a cluster I could use for it.

peastman:
I'm kind of confused about why `sortBuckets()` takes so long in that profile.  Running amber20-stmv on a single Titan V, it takes around 370 us, while `findBlocksWithInteractions()` and `computeNonbonded()` each take around 15 ms.  Even dividing them by four, they should still take many times longer than `sortBuckets()`.  So I'm confused.  Maybe it's something about Ampere?

Your screenshot of the profile didn't include the axis at the top, so I can't tell how long the kernels took.  Was `sortBuckets()` much slower than I expect, or were the others much faster?

peastman:
You want to look only at the `sortBuckets()` that comes just before `findBlocksWithInteractions()`.  That kernel is used by CudaSort for sorting lists.  It also gets called on the PME stream, but that's to sort a different list of different length.

dmclark17:
I see—thanks! 

I did a little more digging, and it seems like the kernel is being slowed down by `(length > blockDim.x)` code path during sorting on the nonbonded stream: https://github.com/openmm/openmm/blob/0c82c2647de98da5c6dab7bf7a7b8b19705aadc0/platforms/cuda/src/kernels/sort.cu#L223-L228

It looked like there were some buckets that had 1000s of elements—I can experiment with tuning the number of buckets to try and avoid this case.

PR #3363, opened 2021-12-03
Try to fix errors setting up CI builds

peastman:
The AMD OpenCL build has been failing for a while.  I'm trying to get it working again.

The M1 Mac build is also failing for reasons I don't understand.  Here's the relevant part of the log:

```
Ensuring installer...
  Can we use bundled Miniconda?
  Can we download a custom installer by URL?
  ... will download a custom installer by URL.
  Ensuring Installer...
  Checking for cached Miniforge3-MacOSX-arm64.sh@0.0.0-a2d8cad428e2d23deaa456f3311b6d36ac7f3da6dde71365e519b8e85c586e5f...
  Found Miniforge3-MacOSX-arm64.sh cache at /Users/lilmac/Projects/GH-Runner/actions-runner/_work/_tool/Miniforge3-MacOSX-arm64.sh/0.0.0-a2d8cad428e2d23deaa456f3311b6d36ac7f3da6dde71365e519b8e85c586e5f/x64!
Running installer...
Error: Unknown installer extension: 
```

@mikemhenry do you have any idea what the problem is?

### DIFF
--- a/devtools/ci/gh-actions/scripts/install_amd_opencl.sh
+++ b/devtools/ci/gh-actions/scripts/install_amd_opencl.sh
@@ -18,6 +18,7 @@ echo libamdocl64.so > ${OPENCL_VENDOR_PATH}/amdocl64.icd
 export LD_LIBRARY_PATH=${AMDAPPSDK}/lib/x86_64:${LD_LIBRARY_PATH:-}
 chmod +x ${AMDAPPSDK}/bin/x86_64/clinfo
 ${AMDAPPSDK}/bin/x86_64/clinfo
+sudo apt-get update
 sudo apt-get install -y libgl1-mesa-dev
 
 echo "OPENCL_VENDOR_PATH=${OPENCL_VENDOR_PATH}" >> ${GITHUB_ENV}


mikemhenry:
I will take a look -- I ran into this problem when developing this and that issue was related to the wrong installer getting downloaded OR the tool cache getting messed up 

The M1 is just sitting in my office and not a data center so we could run into network issues since it is on a consumer ISP. I've got it on a UPS but I don't really have any monitoring setup if the GHA runner fails or crashes, so ping me if you run into issues (like this one)

mikemhenry:
It looks like the issue now is differnt:

```
  Encountered problems while solving:
    - nothing provides requested tzdata 2021e he74cb21_0
    - nothing provides requested pycparser 2.21 pyhd8ed1ab_0
    - nothing provides requested certifi 2021.10.8 py39h2804cbe_1
    - nothing provides requested pycosat 0.6.3 py39h5161555_1009
    - nothing provides requested conda-package-handling 1.7.3 py39h5161555_1
    - nothing provides requested cryptography 36.0.0 py39hfb8cd70_0
    - nothing provides requested pip 21.3.1 pyhd8ed1ab_0
    - nothing provides requested requests 2.26.0 pyhd8ed1ab_1
    - nothing provides requested conda 4.10.3 py39h2804cbe_4
    - package requests-2.26.0-pyhd8ed1ab_0 requires charset-normalizer >=2.0.0,<2.0.1, but none of the providers can be installed
  ```

The error about the installer is transient sometimes, so doing a re-run can fix it. Have any package pins changed? It looks like the issue now is trying to solve the environment

mikemhenry:
On closer inspection, it looks like it is crashing when the installer (miniforge) is setting up, so an update may have broken it. I probably won't get to it today, but I will see if setting it to an older versions fixes this issue, and then open up a bug report with them.

peastman:
Thanks!

I'll merge this PR, since it fixed the problem installing AMD OpenCL.  (And revealed that some tests segfault under it.  I guess I'll look at that next.)

Issue #3368, opened 2021-12-06
Computed parameters for CustomNonbondedForce


peastman:
I've been thinking about ways to optimize #3361, and I thought of a feature that could be useful in many cases.  In that issue, a large part of the energy expression is really computing per-particle parameters.  In principle they only need to be computed once for each particle, but because of how CustomNonbondedForce works, they instead end up getting recomputed for each interaction.  That makes it slower than it needs to be.

What if we allowed you to specify "computed parameters"?  You would provide an expression that would get evaluated for each particle.  It could depend on global and per-particle parameters, but not on other computed parameters.  The computed parameter could then be used in the energy expression.

As an example, suppose you're doing an alchemical simulation.  You have a global parameter `lambda` that is used to interpolate between two per-particle parameters `sigma1` and `sigma2`.  You could write

```python
force.addComputedParameter('sigma', 'lambda*sigma1 + (1-lambda)*sigma2')
```

You could then use `sigma` in the energy expression.  This would remove the interpolation from the inner loop.  It also could save memory in the interaction kernel by only needing to load one parameter (`sigma`) in place of two (`sigma1` and `sigma2`).

peastman:
Actually, I just realized this is basically the same as what CustomGBForce already does with `addComputedValue()`.  The only difference is that it wouldn't support computing parameters as sums over pairs.

zhang-ivy:
This seems like it could be very helpful. 

>Actually, I just realized this is basically the same as what CustomGBForce already does with addComputedValue(). The only difference is that it wouldn't support computing parameters as sums over pairs.

But `CustomGBForce` is different enough from `CustomNonbondedForce` (i.e. doesn't have a `setLongRangeCorrection()` attribute), meaning it would be worth implementing this feature, right?

peastman:
Correct.

zhang-ivy:
@peastman : I didn't have a chance to ask you about this during our call today, but will you have time to implement this in the next week? I was hoping to have this feature available before I try to reduce the complexity of the custom expression, but if you don't expect it to done soon, I won't wait for it.

peastman:
This is a major feature, so it will take a bit of time.  Also, I'm on vacation the next two weeks.

jchodera:
Just a small addition: It would be helpful to be able to define derived global parameters too, since we will otherwise generally be computing a lot of derived global parameters over and over again in each pair computation.

peastman:
Can you give an example?

[...]

Issue #3351, opened 2021-11-24
Vectorize Lepton expressions


peastman:
With the OpenCL and CUDA platforms, there's little difference in speed between standard forces and equivalent custom forces.  On the CPU platform, there can be a much bigger difference.  That's especially true for CustomNonbondedForce.

The main reason for this is that Lepton expressions don't get vectorized.  NonbondedForce uses the vector unit to evaluate 4 or 8 interactions at once, but CustomNonbondedForce needs to evaluate them one at a time.  Implementing vectorization of expressions would make CustomNonbondedForce a lot faster on the CPU platform.

This also could help with #3229, since vectorization would allow the numeric integrals to be computed much more quickly.

lukeczapla:
@peastman Are the Lepton expressions those strings with formulas that everyone's using?  It's a bit like an old scripting language but with very powerful math syntax for derivatives and integration and all.  Here is what I am trying to write:

saveGlobalVariable(1, 5000);   // it would mean to save value once every timestep and copy to main RAM every 5000 steps
                                                    // if your platform is a GPU or "interesting" OpenCL device
So now would that be a plugin or also just a new feature too?  I know I need one for each platform and the Reference would be easy, is the only difference between Reference and CPU is that CPU can take advantage of say, 72 cores on a fancy HPC server?   Global variables are always type "real" (i.e., single/double) or could they be integers too?

>  With the OpenCL and CUDA platforms, there's little difference in speed between standard forces and equivalent custom forces. On the CPU platform, there can be a much bigger difference. 

This is true, I wrote my C++ code to use the standard forces (SwitchingDistance + Cutoffs) to approximate the force-switching used by some CHARMM CustomNonbondedForce from Im's lab, and it was barely any different in ns/day between the custom and built in variants with everything else (PME spacings and all factors).   23 ns/d vs. 21.5 ns/d - now looking at the difference between the Python and C++ one to see if anything comes out, meaning running both on the same basic GPU.

> The main reason for this is that Lepton expressions don't get vectorized

So if we write a ComputePerDof variable for example, you want to arrange the data linearly or linearly packed for things like SIMD GPU routines?  Even with 3D data (bins in space) I had a problem like 15 years ago and could do something like linearly packing it and getting the most out of, e.g., the 2-4 GB of CPU RAM at the time.  The GPUs now have even more than that.

[...]

PR #3379, opened 2021-12-14
Enable analytical long-range correction integrals to be specified for CustomNonbondedForce

jchodera:
This addresses #3301 by enabling users to optionally specify an analytical form of the long-range correction to be used instead of numerical computation. This has the potential for greatly reducing the cost of updating global parameters, as in nonequilibrium switching simulations. By default, the standard numerical integration scheme is used, and it is easy to toggle the two scheme on and off should one want to verify the expression is correct.

As an example, for this `CustomNonbondedForce` implementing a Lennard-Jones 12-6 potential with cutoff:
```C++
    CustomNonbondedForce* customNonbonded = new CustomNonbondedForce("4*eps*((sigma/r)^12-(sigma/r)^6); sigma=0.5*(sigma1+sigma2); eps=sqrt(eps1*eps2)");
    customNonbonded->addPerParticleParameter("sigma");
    customNonbonded->addPerParticleParameter("eps");
```
the long-range correction integral can be specified by adding:
```C++
    customNonbonded->setUseLongRangeCorrectionIntegral(true);
    customNonbonded->setLongRangeCorrectionIntegral("(16*pi/3) * eps * sigma^3 * ((1/3)*(sigma/r)^9 - (sigma/r)^3); sigma=0.5*(sigma1+sigma2); eps=sqrt(eps1*eps2); pi = 3.1415926535;");
```
This is tested in `TestCustomNonbondedForce.h` with `testLongRangeCorrectionAnalytical()`.

This doesn't have to be our final solution here, but we can at least test and benchmark the performance.

@zhang-ivy : Can you upload a ZIP of the test case we were exploring today and I can benchmark this?


jchodera:
Hm, I'm not sure how `examples/Makefile` got swept up in this PR, and it seems like my editor has converted some tabs to spaces in some of the files. Sorry about this---I'll figure out what happened there.

zhang-ivy:
@jchodera : Here's a zipped directory of the script and system/state xmls we were using today. I've also included a REST hybrid system/state xml as well, but we were experimenting with the vanilla hybrid system today.

[20211214_experiments_with_john.zip](https://github.com/openmm/openmm/files/7714518/20211214_experiments_with_john.zip)

peastman:
I really don't like this feature.  It requires a lot of work from the user, and gives them lots of opportunities to make mistakes.  And if they do make a mistake, they get wrong results, but in a subtle way so they likely will never discover all their results are wrong and their simulations are invalid.

It also isn't a generally useful feature.  It's designed for one particular user (I won't name names!), and is only relevant to such a specific combination of requirements, it's quite likely that no other user in the world will ever have a good reason to use it.  It's only relevant because you're 1) using CustomNonbondedForce 2) with the long range correction enabled, while 3) modifying a global parameter on every time step and 4) also querying the energy on every step.  And 5) this affects performance for your system (it won't for a lot of systems), and 6) you care about the tiny change in the correction when you change the parameter (for a lot of applications, it's irrelevant).

For your application, I think it's much better to use a linear interpolation method like we agreed on at our meeting, while also investigating ways to make the existing computation faster (such as #3368 and #3351, or possibly doing the integration on the GPU).

jchodera:
> I really don't like this feature. It requires a lot of work from the user, and gives them lots of opportunities to make mistakes. And if they do make a mistake, they get wrong results, but in a subtle way so they likely will never discover all their results are wrong and their simulations are invalid.

Let's take this one by one:
* It is optional
* It gives additional flexibility in terms of the expression that will be used for long-range corrections
* It's easy to test if it gives the correct results with a simple toggle of a single flag
* It greatly reduces the number of function evaluations if used

> It also isn't a generally useful feature. It's designed for one particular user (I won't name names!),

It's designed primarily for one particular _class of applications_---nonequilibrium methods---where introducing time-dependent changes in global parameters that impact long-range corrections would otherwise be too slow. That's not the same as it being designed for a particular user. We have wrestled with this in all our nonequilibrium calculations, but have finally concluded we can't afford to omit this term anymore.

> For your application, I think it's much better to use a linear interpolation method like we agreed on at our meeting, while also investigating ways to make the existing computation faster (such as #3368 and #3351, or possibly doing the integration on the GPU).

If you can come up with a similarly simple API extension that accomplishes this, we're more than happy to use it!

Important considerations:
* It has to be easily defined in the `System` object via the API, so that it can be serialized
* It can't require complex manual calculations (more complex than doing a simple analytical integral, as here)
* It can't require instantiating a Context to compute something that then gets stuffed back into the System

For example, if there was something as simple as 
```python
nbforce.setLongRangeInterpolation('lambda_lrc', 0.0, { 'lambda_sterics_old' : 1.0, 'lambda_sterics_new' : 0.0, 'lambda_sterics_core' : 0.0 })
nbforce.setLongRangeInterpolation('lambda_lrc', 1.0, { 'lambda_sterics_old' : 0.0, 'lambda_sterics_new' : 1.0, 'lambda_sterics_core' : 1.0 })
```
where we can simply tell `CustomNonbondedForce` to linearly interpolate the correction using the global parameter `lmbda_lrc` from 0 to 1 by computing the LRC at the two specified combinations of other global parameters, that would be perfect. This is super clunky, however---if you can think of something simpler, we're happy to use this!

Of course, another possibility is to move the integration to the GPU to speed things up further without the need to interpolate at all. I still think this may be possible...

jchodera:
Alternatively, could lepton be extended to do symbolic 1D integration? It's not as simple as the chain rule, unfortunately...

peastman:
> It is optional

Doesn't matter.  It pollutes the API.  If you add a feature people shouldn't use, you can be absolutely certain that some of them will end up using it anyway!

When you're maintaining a package for the long term, one of the most important things to know is what features to say no to.  You can always argue, "It's just one minor function," or, "It's optional," or, "It might be useful for someone."  So you keep letting them in, and after ten years you discover your package has turned into a mess.

> It gives additional flexibility in terms of the expression that will be used for long-range corrections

No it doesn't.  There is exactly one and only one correct expression (the one that is the actual integral of the energy).  Any other would simply be wrong and would give wrong results.

> It's designed primarily for one particular class of applications---nonequilibrium methods

No, it's designed for one particular user.  Most nonequilibrium methods won't have the very unusual combination of factors that make this an issue for you.

> Alternatively, could lepton be extended to do symbolic 1D integration?

Ha!  Analytic differentiation is simple and mechanical.  Analytic integration is an art.  Packages like Maxima or Mathematica have huge amounts of code to implement many different techniques for analytic integration, and even then they still can't do it reliably.

jchodera:
I do understand the desire to not pollute the API. That's why it would be
best to just speed up the exact result where this would not be needed. :)

The long range correction does not necessarily have an unambiguous meaning.
The method you implement is only an approximation to LJPME, for
example---there could be other more accurate choices for the integral (like
GB is!) or there could be fast but not too inaccurate approximations.

I wouldn't knock nonequilibrium simulations. They're an important use case
that are rapidly coming to dominate alchemical free energy calculations.
The flexibility afforded by handling sterics with CustomNonbondedForce is
not a restricted use case.

I'm super curious what you think about the other possibilities:
* Numerical integration on the GPU
* Extending the API cleanly for the case where we want to interpolate
linearly, but given the conditions on not substantially increasing the
complexity of any code that uses this feature

peastman:
> I wouldn't knock nonequilibrium simulations.

Neither would I!  I didn't say they aren't important.  I said most of them don't have the very specific combination of factors that make this an issue for you.

> * Numerical integration on the GPU

It's definitely a possibility.  The advantage is that the GPU can probably do it a lot faster than the CPU.  The disadvantage is that it will always add cost, whereas the CPU can run in parallel with the GPU, making the computation completely free if it's fast enough.

> * Extending the API cleanly for the case where we want to interpolate
> linearly, but given the conditions on not substantially increasing the
> complexity of any code that uses this feature

Anything that involves a new API will add some complexity, but it shouldn't be too much.  More important, it shouldn't increase the complexity of other features.  Your CustomNonbondedForce is really complicated because it does a bunch of things at once, and each extra thing you make it do increases the difficulty of the others.  This feature shouldn't be like that.

Of course, the very best option is if the computation just gets faster with no user code changes required.

Speaking of code complexity, #3368 should help with the existing code.  It will let you break up the huge, complicated expression into a series of smaller, more modular ones.

jchodera:
For reference, here is the timing for this feature on a full-size system (185391 particles) from @zhang-ivy with her new alchemical protocol:

No LRC: 0.007777 s
Numerical integration, 1 thread: 0.0459807 s (6.0x slower)
Numerical integration, 8 threads: 0.02133307 s (2.7x slower)
Analytical integration, 1 thread: 0.015357 s (2.0x slower)
Analytical integration, 8 threads: 0.007249 (not measurably slower)

In other words, it accomplishes our goals and only adds the need for two additional lines of code.

I'll attach a ZIP of the test input files shortly.

zhang-ivy:
@peastman @jchodera: If the main problem with this feature/PR is that it gives users opportunities to make mistakes, what if we modified this feature to allow users the option to analytically integrate the LRC, but only using a predefined expression (one that John/Peter agree on together)?

For example, the feature might introduce something like:
```python
setIntegrateLongRangeCorrectionAnalytical(true)
```

which would by default by false, but when true, would allow the user to analytically integrate the long range correction (without choosing the integral).

peastman:
I don't understand.  How would it know what the integral of your energy expression is?

jchodera:
That's already implemented [here](https://github.com/openmm/openmm/pull/3379/files#diff-fb6a991524733b2a581720db8f8003af117faeeebfb91662da01623c208a5becR311-R314)!

@zhang-ivy will post an example she ran into where the current numerical integration scheme actually fails at some intermediate `\lambda`. We're not sure if this is a failure of the numerical integration scheme or if the expression no longer has a finite integral in this case---hopefully you can give us some pointers on how to introduce code to dump out the exact combination of parameters that result in numerical integration failure. 

In either case (numerical integration failures or intermediate \lambda results in an expression with infinite integral), it does suggest that we need more control over exactly what expression is integrated for intermediate \lambda (or what expression to use for the integral).

zhang-ivy:
>I don't understand. How would it know what the integral of your energy expression is?

I must be understanding the long range correction incorrectly. I thought that the LRC integral was independent of the custom expression, but I must be wrong. 

If we are worried about users making mistakes, isn't the current implementation of the `CustomNonbondedForce` prone to users making mistakes as well, given that the user-supplied custom expression could lead to incorrect results? Allowing the user to also specify the analytical integral seems to have the same level of risk.

>@zhang-ivy will post an example she ran into where the current numerical integration scheme actually fails

I ran nonequilibrium switching with the rest factory we've been talking about. I used the full RBD:ACE2 system and turned the LRC on. Towards the end of the forward annealing, I am seeing this error:
```
  what():  CustomNonbondedForce: Long range correction did not converge.  Does the energy go to 0 faster than 1/r^2?
``` 
I double checked to make sure there are no instances of sigma = 0 in the hybrid system. 

When I run nonequilibrium switching with the rest factory with the LRC off, I am seeing this error (also towards the end of forward annealing):

```
  File "/lila/home/zhangi/choderalab/perses_benchmark/perses_protein_mutations/code/31_rest_over_protocol/run_neq_rest_o
ver_protocol_save_solvent_last_frame_LRC_off.py", line 161, in <module>
    integrator.step(2500)
  File "/home/zhangi/miniconda3/envs/perses-rest-over-protocol2/lib/python3.9/site-packages/openmm/openmm.py", line 1092
5, in step
    return _openmm.CustomIntegrator_step(self, steps)
openmm.OpenMMException: Particle coordinate is nan
```

This seems to indicate that there is might be a problem with the custom expressions I wrote, right? I'm not sure what the best way to debug this is (I'm not seeing any nans when I run vanilla MD). 

Attached is the script I ran and an xml of the hybrid system. 
[rest-system-lrc-nan-issues.zip](https://github.com/openmm/openmm/files/7817943/rest-system-lrc-nan-issues.zip)

jchodera:
> I must be understanding the long range correction incorrectly. I thought that the LRC integral was independent of the custom expression, but I must be wrong.

For `CustomNonbondedForce`, the long-range correction is computed by numerically integrating `r` over the product of the surface area of a sphere (`4*pi*r^2`) and the potential you specify, over `r \in [r_cutoff, \infty)`. The potential has to go to zero as `r \rightarrow \infty` faster than `1 / r^2` for this to converge, which _should_ be the case for all potentials we care about (even alchemical ones).

> If we are worried about users making mistakes, isn't the current implementation of the CustomNonbondedForce prone to users making mistakes as well, given that the user-supplied custom expression could lead to incorrect results? Allowing the user to also specify the analytical integral seems to have the same level of risk.

Absolutely! And my implementation has the analytical integral off by default, but an option that can be specified and toggled easily.

What we are finding now is that it's also possible for the potential function to have well-defined integrals at the lambda endpoints we care about, but may have a numerically difficult integral at intermediate lambda values, necessitating we specify a different functional form or integration scheme at intermediate lambdas (or, as in this PR, just give the integral we want to evaluate instead).

I ran nonequilibrium switching with the rest factory we've been talking about. I used the full RBD:ACE2 system and turned the LRC on. Towards the end of forward annealing, I am seeing this error:
```
  what():  CustomNonbondedForce: Long range correction did not converge.  Does the energy go to 0 faster than 1/r^2?
```
This is where we need some help from @peastman about how to instrument OpenMM to identify exactly which parameters are causing this integration failure. We'll then be able to check if somehow we're modifying the integral so it's no longer finite or if it's just a numerical integration failure.

peastman:
> If we are worried about users making mistakes, isn't the current implementation of the CustomNonbondedForce prone to users making mistakes as well, given that the user-supplied custom expression could lead to incorrect results? Allowing the user to also specify the analytical integral seems to have the same level of risk.

There is no right expression for the energy.  You can use whatever function you think makes sense for your application.  A particular function might not give the effect you were hoping for, but that's another matter.  Given your energy function, however, there is one and only one correct integral of it.  Anything else is unambiguously wrong.

> What we are finding now is that it's also possible for the potential function to have well-defined integrals at the lambda endpoints we care about, but may have a numerically difficult integral at intermediate lambda values

In that situation, I think the best solution is to accurately compute the coefficient at the endpoints (changing the integral expression would not do that), then linearly interpolate it between those endpoints.

jchodera:
> In that situation, I think the best solution is to accurately compute the coefficient at the endpoints (changing the integral expression would not do that), then linearly interpolate it between those endpoints.

We'd still be happy with this if we can come up with an API that satisfies the requirements enumerated above!
https://github.com/openmm/openmm/pull/3379#issuecomment-995189923

jchodera:
We've addressed this through performance improvements, though we have not fully eliminated the performance gap---we're down to ~10x slowdown. See [#3229.](https://github.com/openmm/openmm/issues/3229#issuecomment-1121105100)

peastman:
Hopefully a smaller speed difference than that at this point.  Those tests were run before #3606, which further improved it.

PR #3383, opened 2021-12-16
Optimize sorting of non-uniformly distributed data

### LINKED ISSUE #3357: Optimizing sorting when building neighbor list
This is a followup to the discussion that started in #3347.  I'm moving it here so it won't get lost in the discussion for a merged PR.

When running the amber20-stmv benchmark parallelized across four A100s, @dmclark17 found `sortBuckets()` was taking a significant amount of time.  In contrast, when I run the same benchmark on a single Titan V that kernel takes very little time.  We want to figure out why that's happening and what we can do about it.

He observed that some blocks had very large numbers of elements, forcing it to use the slower code path that doesn't use shared memory.  It chooses the number of buckets so that the average number of elements in each bucket will be no more than half the maximum it can fit in shared memory.

https://github.com/openmm/openmm/blob/0c82c2647de98da5c6dab7bf7a7b8b19705aadc0/platforms/cuda/src/CudaSort.cpp#L71-L72

That ensures most buckets can be sorted with the fast path.  If the data is uniformly distributed, all of them will.  But if the distribution is very nonuniform, there could be some that require the slow path.

One solution is to increase the number of buckets further.  That would impact speed in several ways.

- It would reduce the number of buckets requiring the slow path.
- It would make each bucket faster to sort, since the bitonic merge sort scales as O(n*(log(n))^2).
- It would add some overhead in `computeRange()` and `computeBucketPositions()`.
- It could lead to some buckets having very few elements, which would mean some threads being left idle in `sortBuckets()`.

Another possible optimization is to try to skip the sort on steps when the neighbor list doesn't need to be rebuilt.  Here is how the code currently works.

1. `findBlockBounds()` clears the flag for whether a rebuild is needed.
2. We sort the blocks by the size of their bounding boxes.
3. `sortBoxData()` checks all atoms to see how far they have moved and sets a flag if any has moved far enough to require rebuilding.

Those operations don't have to be done in those particular kernels.  They were just convenient places to put them.  In principle we could do the check in `findBlockBounds()` instead, which would allow us to skip the sort if it isn't needed.  That would require the flag to get cleared at some earlier point, and also require the result to be sent back to the host.

peastman:
Fixes #3357.  This improves performance with very large systems where it can't use the short list version of the kernel.  It's especially important for multi-GPU simulations.

I benchmarked it on 1 to 4 A100s.  Here are the speeds for cellulose (about 400,000 atoms) with CUDA.

| GPUs | Original | Optimized |
| --- | --- | --- |
1 | 105.3 | 111.5
2 | 164.4 | 173.5
3 | 193.1 | 206.9
4 | 210.7 | 222.2

And for STMV (about 1 million atoms).

| GPUs | Original | Optimized |
| --- | --- | --- |
1 | 32.2 | 33.3
2 | 51.1 | 53.1
3 | 61.1 | 65.6
4 | 71.1 | 73.6

Here is cellulose with OpenCL.

| GPUs | Original | Optimized |
| --- | --- | --- |
1 | 69.2 | 70.1 | 
2 | 83.8 | 85.6 | 
3 | 92.8 | 95.2 | 
4 | 90.3 | 90.9 |

And STMV with OpenCL.

| GPUs | Original | Optimized |
| --- | --- | --- |
1 | 18.2 | 18.4
2 | 24.5 | 24.5
3 | 27.0 | 27.4
4 | 29.3 | 29.4

cc @dmclark17

### DIFF
--- a/platforms/cuda/include/CudaSort.h
+++ b/platforms/cuda/include/CudaSort.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2010-2018 Stanford University and the Authors.      *
+ * Portions copyright (c) 2010-2021 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -77,8 +77,12 @@ public:
      *                   on the heap with the "new" operator.  This object takes over ownership of it,
      *                   and deletes it when the CudaSort is deleted.
      * @param length     the length of the arrays this object will be used to sort
+     * @param uniform    whether the input data is expected to follow a uniform or nonuniform
+     *                   distribution.  This argument is used only as a hint.  It allows parts
+     *                   of the algorithm to be tuned for faster performance on the expected
+     *                   distribution.
      */
-    CudaSort(CudaContext& context, SortTrait* trait, unsigned int length);
+    CudaSort(CudaContext& context, SortTrait* trait, unsigned int length, bool uniform=true);
     ~CudaSort();
     /**
      * Sort an array.
@@ -94,7 +98,7 @@ private:
     CudaArray buckets;
     CUfunction shortListKernel, shortList2Kernel, computeRangeKernel, assignElementsKernel, computeBucketPositionsKernel, copyToBucketsKernel, sortBucketsKernel;
     unsigned int dataLength, rangeKernelSize, positionsKernelSize, sortKernelSize;
-    bool isShortList;
+    bool isShortList, uniform;
 };
 
 /**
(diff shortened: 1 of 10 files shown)


dmclark17:
Looks great! I am seeing slightly higher benchmark numbers across the board on a DGX A100—for STMV with 4 A100 GPUs, I am seeing 77ns/day with the optimization and 73ns/day without the change. With the hacky bounding box size transformation, I was hitting 82ns/day on 4 A100, so there may still be some cases were the global memory code path is being taken. I tried tuning some of the sorting parameters and found that doubling the thread block size and number of buckets for the non-uniform case improved performance to 78.95ns/day on 4 A100s.

peastman:
I made some further improvements to better distribute the elements between buckets.  Running STMV on a Titan V, the time for `sortBuckets()` is now down to about 45 us.

dmclark17:
Here is the performance I am seeing with the new version. I will try to generate some profiles as well!

System | GPUs (A100) | Initial pull request (ns/day) | New version (ns/day)
-- | -- | -- | --
STMV | 1 | 34.45 | 35.62
STMV | 2 | 55.37 | 57.52
STMV | 4 | 77.00 | 80.95
STMV | 8 | 85.72 | 89.18
cellulose | 1 | 115.11 | 117.97
cellulose | 2 | 180.02 | 182.46
cellulose | 4 | 232.08 | 236.99
cellulose | 8 | 242.42 | 249.37

peastman:
Merging this now that 7.7 is out.

Issue #3382, opened 2021-12-15
Add option to benchmark.py to run all available benchmarks and store in computer-readable format


jchodera:
Currently, `benchmark.py` can only run one benchmark at a time, and only outputs human-readable text.

It would be convenient to have a `--run-all` option that would run all available benchmarks (or possibly those specified combinatorially by a YAML file?) to make it easy to generate and capture (in a computer-readable manner) comprehensive benchmarks on a new system.

The output could, for example, be a YAML file that makes it easy for both humans to read and to manipulate by converting it into a Python `dict`.

peastman:
If you omit the `--test` argument it runs all the tests except the Amber ones.

jchodera:
But I still have to combinatorially regress over all other arguments?
And there's no machine-readable output, right?

peastman:
Correct.

jchodera:
Would it be OK if I added two command-line arguments?
1. A `--run-all` argument that combinatorially runs all combinations of (platform, test, ensemble, precision)
2. A `--outfile` argument that causes the output to also be captured in a machine- and human-readable format (like YAML)

If we wanted to be more flexible about output formats, we could use [pydantic](https://pydantic-docs.helpmanual.io/), which can support multiple output schemas (e.g. JSON, YAML, etc), but it would introduce another dependency.

I've got access to some hardware for benchmarking, and it would be valuable to be able to capture machine-readable output for our benchmarking purposes.

jchodera:
Alternatively, we may want to pull the benchmark code and data files into the `openmm.app` namespace so that benchmarking can be more easily automated programmatically, or run without needing to find the `$HOME/path/to/anaconda/env/{yourenvname}/share/openmm/examples/` directory, which is a significant challenge for most users. For example, it could be something like
```python
python -m openmm.benchmarks.run
```

jchodera:
I've made some improvements:
* Accept an `--all` argument that combinatorially regresses over all valid combinations of (test, platform, precision, heavy_hydrogens, ensemble)
* Added an `--outfile <filename>` argument that allows the benchmark to be written to a machine-readable JSON or YAML file. We could have this create or append if you want to be able to chain together arbitrary benchmarks.
* Added a `--style <style>` argument to select between a legacy-like output (key: value) and a more readable table:
```
  Test   Platform   Precision   Heavy Hydrogens   dt (fs)   Ensemble   ns/day    
 ─────────────────────────────────────────────────────────────────────────────── 
[...]

PR #3385, opened 2021-12-18
Automate benchmarking

jchodera:
This PR makes a number of improvements to the `examples/benchmark.py` script without changing the underlying benchmarks, following #3382.

* The `--all` option was added, enabling all (testsystems, constraints, ensembles, platforms, precisions) to be run.
* The `--outfile <filename>` option was added, enabling machine-readable files to be written to `.json` or `.yaml` files. (YAML requires pyyaml)
* The `--style <style>` option was added to select the output style. The default is `legacy`, which uses the old `key: value` terminal output. If `rich` is specified, rich will be used to generate a form\
atted output table.
* A supported precision must now be correctly specified for each platform specified for benchmarking or else no benchmark will be run.
* Caching is used to reduce the overhead of setting up systems for repeated benchmarks.

The JSON/YAML output format includes information about the system, such as
```JSON
{
    “benchmarks”: [
        {
            “test”: “gbsa”,
            “constraints”: “HBonds”,
            “hydrogen_mass”: “1.5”,
            “ensemble”: “NVE”,
            “platform”: “CUDA”,
            “precision”: “single”,
            “timestep_in_fs”: 4.0,
            “steps”: 6191,
            “elapsed_time”: 0.937712,
            “ns_per_day”: 2281.7342638251403
        },
        {
            “test”: “gbsa”,
            “constraints”: “HBonds”,
            “hydrogen_mass”: “1.5”,
            “ensemble”: “NVE”,
            “platform”: “CUDA”,
            “precision”: “mixed”,
            “timestep_in_fs”: 4.0,
            “steps”: 5337,
            “elapsed_time”: 0.924172,
            “ns_per_day”: 1995.8051098713229
        }
    ],
    "system": {
        "hostname": "csctmp-r750xa-5.hpc.local",
        "timestamp": "2021-12-18T04:52:03.299571",
        "openmm_version": "7.7.0.dev-92c6dbb",
        "cpuinfo": " Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz",
        "cpuarch": "x86_64",
        "system": "Linux",
        "nvidia_driver": "495.29.05",
        "gpu": "NVIDIA A10"
    }
}
```

Eventually, we should probably refactor the benchmark systems, runners, and automation into `openmm.benchmarks` so they can more easily be used in testing or other tooling, and also easily be called from the CLI.


jchodera:
Whoops---I intended to commit only changes to `benchmark.py`. I'll try again.

Issue #132, opened 2013-09-08
Constraints involving massless particles


cwehmeyer:
Dear all,

I would like to constrain an entire molecule during an MD simulation to a fix position in space. However, if that molecule is rigid, the default way of setting the atoms' masses to zero yields an exception (see the attached example).

Is it possible to implement a check such that all particles within a constraint group may have zero mass?

Thank you all in advance!

Christoph

Example:

```
import simtk.openmm.app as app
import simtk.openmm as openmm
import simtk.unit as unit

temperature = 300.0 * unit.kelvin
timestep = 2.0 * unit.femtosecond
collision = 1.0 / unit.picosecond
tolerance = 1.0e-8

pdb = app.PDBFile( 'tip3p.pdb' ) # taken from ...python2.7/site-packages/simtk/openmm/app/data/
ff = app.ForceField( 'tip3p.xml' )
system = ff.createSystem( pdb.getTopology(), nonbondedMethod=app.PME, nonbondedCutoff=9*unit.angstrom )
integrator = openmm.LangevinIntegrator( temperature, collision, timestep )
integrator.setConstraintTolerance( tolerance )

### set the masses of the first water molecule to zero
system.setParticleMass( 0, 0.0 )
system.setParticleMass( 1, 0.0 )
system.setParticleMass( 2, 0.0 )

context = openmm.Context( system, integrator )
context.setPositions( pos )

### enforce the constraints
context.applyConstraints( tolerance )
```

Output:

```
Traceback (most recent call last):
  File "constraint-test.py", line 25, in <module>
    context = openmm.Context( system, integrator )
  File "/home/mi/schiffm/local/opt/python/openmm-git/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 3505, in __init__
    this = _openmm.new_Context(*args)
Exception: A constraint cannot involve a massless particle
```

stefdoerr:
I am facing the same issue so I assume there was no update on this. Trying to constrain protein atoms with mass 0 but since I also use `constraints=HBonds` it complains. I ended up using mass 1e10 in the end but it would be nice if the default `mass=0` worked.

stefdoerr:
Actually no. I just noticed that mass `1e10` doesn't work for minimization. The atoms still move.

jchodera:
Minimization should ignore atom masses. 

[...]

Issue #102, opened 2013-08-13
GromacsTopFile.createSystem() fails for a rather large example


leeping:
Hi there,

The GromacsTopFile.createSystem() method fails for a rather large topology that I was using for an OpenMM example.  I verified that grompp does successfully build the .tpr file here.  

https://dl.dropboxusercontent.com/u/5381783/gmxfail.tar.bz2

```
In [2]: top = GromacsTopFile('topol.top')

In [3]: top.createSystem()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-d13415455b1a> in <module>()
----> 1 top.createSystem()

/home/leeping/local/lib/python2.7/site-packages/simtk/openmm/app/gromacstopfile.pyc in createSystem(self, nonbondedMethod, nonbondedCutoff, constraints, rigidWater, implicitSolvent, soluteDielectric, solventDielectric, ewaldErrorTolerance, removeCMMotion)
    501                         params = self._bondTypes[types[::-1]][3:5]
    502                     else:
--> 503                         raise ValueError('No parameters specified for bond: '+fields[0]+', '+fields[1])
    504                     # Decide whether to use a constraint or a bond.
    505                     useConstraint = False

ValueError: No parameters specified for bond: 1, 2
```

Issue #100, opened 2013-08-10
Organize Issues into Milestones


jchodera:
I think it would be a great idea to organize the various issues listed here into Milestones for upcoming releases, so that we have a more up-to-date sense of which features and fixes are on track for which releases.

PR #3386, opened 2021-12-18
Automate benchmarking and production of computer-readable output

jchodera:
This PR makes a number of improvements to the `examples/benchmark.py` script without changing the underlying benchmarks, addressing #3382.
* By default, all available benchmarks are run (including the AMBER ones), but this can be overridden:
* By specifying one or more of `--test <tests>`, `--platform <platforms>`, `--precision <precisions>` --bond-constraints <constraints>`, `--ensemble <ensembles>`, with a single argument or a comma-separated list of arguments (e.g. `CUDA,OpenCL`), the set of benchmarks to be run can be restricted to a desired subset
* The `--heavy` option really changes constraints from `HBonds` to `AllBonds` (in addition to modifying the timestep), so this option has been renamed to `--bond-constraints`
* The `--outfile <filename>` option was added, enabling machine-readable files to be written to `.json` or `.yaml` files. (YAML requires the optional `pyyaml` package)
* The `--style <style>` option was added to select the output style. The default is `legacy`, which uses the old `key: value` terminal output. If `rich` is specified, rich will be used to generate a formatted output table (requires the optional `rich` package).
* Caching is used to reduce the overhead of setting up systems for repeated benchmarks.

To generate a JSON benchmark output with a `rich` table printed to the terminal:
```
$ python benchmark.py --outfile benchmark.json --style rich --seconds 1
```
You will see output like
```
timestamp: 2021-12-20T20:07:58.933640
openmm_version: 7.7.0.dev-92c6dbb
cpuinfo: Intel(R) Xeon(R) Platinum 8360Y CPU @ 2.40GHz
cpuarch: x86_64
system: Linux
nvidia_driver: 495.29.05
gpu: NVIDIA A100 80GB PCIe
                                                                                                                   
  Test                       Precision   Constraints   H mass (amu)   dt (fs)   Platform    Ensemble       ns/day  
 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
  gbsa                       single      HBonds        1.5            4.0       NVE         CUDA           1954.7  
  gbsa                       mixed       HBonds        1.5            4.0       NVE         CUDA           1939.7  
  gbsa                       double      HBonds        1.5            4.0       NVE         CUDA            654.8  
  gbsa                       single      HBonds        1.5            4.0       NVE         OpenCL         1276.6  
  gbsa                       mixed       HBonds        1.5            4.0       NVE         OpenCL         1222.0  
  gbsa                       double      HBonds        1.5            4.0       NVE         OpenCL          621.1  
  gbsa                       single      HBonds        1.5            4.0       NVT         CUDA           1954.2  
  gbsa                       mixed       HBonds        1.5            4.0       NVT         CUDA           1929.0  
  gbsa                       double      HBonds        1.5            4.0       NVT         CUDA            650.9  
  gbsa                       single      HBonds        1.5            4.0       NVT         OpenCL         1229.0  
  gbsa                       mixed       HBonds        1.5            4.0       NVT         OpenCL         1216.8  
  gbsa                       double      HBonds        1.5            4.0       NVT         OpenCL          610.8  
```

The JSON/YAML output format includes information about the system, such as
```JSON
{
    "benchmarks": [
        {
            "test": "gbsa",
            "constraints": "HBonds",
            "hydrogen_mass": "1.5",
            "ensemble": "NVT",
            "platform": "CUDA",
            "precision": "single",
            "timestep_in_fs": 4.0,
            "steps": 12673,
            "elapsed_time": 1.998786,
            "ns_per_day": 2191.2244732552654
        },
        {
            "test": "gbsa",
            "constraints": "HBonds",
            "hydrogen_mass": "1.5",
            "ensemble": "NVT",
            "platform": "OpenCL",
            "precision": "single",
            "timestep_in_fs": 4.0,
            "steps": 7458,
            "elapsed_time": 1.99769,
            "ns_per_day": 1290.2326186745693
        },
        {
            "test": "gbsa",
            "constraints": "AllBonds",
            "hydrogen_mass": "4",
            "ensemble": "NVT",
            "platform": "CUDA",
            "precision": "single",
            "timestep_in_fs": 5.0,
            "steps": 7547,
            "elapsed_time": 2.000243,
            "ns_per_day": 1629.9539605937878
        }
    ],
    "system": {
        "hostname": "csctmp-r750xa-5.hpc.local",
        "timestamp": "2021-12-20T06:48:42.653108",
        "openmm_version": "7.7.0.dev-92c6dbb",
        "cpuinfo": "Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz",
        "cpuarch": "x86_64",
        "system": "Linux",
        "nvidia_driver": "495.29.05",
        "gpu": "NVIDIA A10"
    }
}
```


jchodera:
@peastman : Should we be worried about these test failures on AMD OpenCL?
```
Errors while running CTest
    Start 100: TestOpenCLNonbondedForceSingle
    Start 101: TestOpenCLNonbondedForceMixed
1/4 Test #100: TestOpenCLNonbondedForceSingle .....***Exception: SegFault 38.21 sec

    Start 102: TestOpenCLNonbondedForceDouble
2/4 Test #101: TestOpenCLNonbondedForceMixed ......***Exception: SegFault 39.96 sec

    Start 132: TestOpenCLVerletIntegratorDouble
3/4 Test #132: TestOpenCLVerletIntegratorDouble ...   Passed   15.46 sec
4/4 Test #102: TestOpenCLNonbondedForceDouble .....***Exception: SegFault 42.13 sec


25% tests passed, 3 tests failed out of 4

Total Test time (real) =  80.35 sec

The following tests FAILED:
	100 - TestOpenCLNonbondedForceSingle (SEGFAULT)
	101 - TestOpenCLNonbondedForceMixed (SEGFAULT)
	102 - TestOpenCLNonbondedForceDouble (SEGFAULT)
```

jchodera:
In a future update, perhaps we could stand up a benchmark server that could collect (and serve up plots of) machine-readable submissions of benchmarks

jchodera:
Here is a ZIP archive of NVT benchmarks for CUDA and OpenCL on NVIDIA A10, A40, and A100 GPUs:
[openmm-7.7.0rc1-benchmarks.zip](https://github.com/openmm/openmm/files/7747709/openmm-7.7.0rc1-benchmarks.zip)

jchodera:
At the suggestion of https://github.com/openmm/openmm/issues/3382#issuecomment-998349960, I am now capturing properties from the `Platform` in use by the `Context`, like this:
```JSON
{
    "benchmarks": [
        {
            "test": "amber20-dhfr",
            "cutoff": 0.9,
            "constraints": "HBonds",
            "hydrogen_mass": "1.5",
            "ensemble": "NVE",
            "timestep_in_fs": 4.0,
            "precision": "single",
            "platform": "OpenCL",
            "platform_properties": {
                "DeviceIndex": "1",
                "DeviceName": "Intel(R) Iris(TM) Plus Graphics 655",
                "OpenCLPlatformIndex": "0",
                "OpenCLPlatformName": "Apple",
                "Precision": "single",
                "UseCpuPme": "false",
                "DisablePmeStream": "false"
            },
            "steps": 58,
            "elapsed_time": 1.040505,
            "ns_per_day": 19.264491761212103
        }
    ],
    "system": {
        "hostname": "lski2838",
        "timestamp": "2021-12-21T00:20:28.649980",
        "openmm_version": "7.6.0.dev-ad113a0",
        "cpuinfo": "Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz",
        "cpuarch": "i386",
        "system": "Darwin"
    }
}
```

ex-rzr:
> @peastman : Should we be worried about these test failures on AMD OpenCL?
> 
> ```
> Errors while running CTest
> Test project /home/runner/work/openmm/openmm/build
>     Start 100: TestOpenCLNonbondedForceSingle
>     Start 101: TestOpenCLNonbondedForceMixed
> 1/4 Test #100: TestOpenCLNonbondedForceSingle .....***Exception: SegFault 38.21 sec
> 
>     Start 102: TestOpenCLNonbondedForceDouble
> 2/4 Test #101: TestOpenCLNonbondedForceMixed ......***Exception: SegFault 39.96 sec
> 
>     Start 132: TestOpenCLVerletIntegratorDouble
> 3/4 Test #132: TestOpenCLVerletIntegratorDouble ...   Passed   15.46 sec
> 4/4 Test #102: TestOpenCLNonbondedForceDouble .....***Exception: SegFault 42.13 sec
> 
> 
> 25% tests passed, 3 tests failed out of 4
> 
> Total Test time (real) =  80.35 sec
> 
> The following tests FAILED:
> 	100 - TestOpenCLNonbondedForceSingle (SEGFAULT)
> 	101 - TestOpenCLNonbondedForceMixed (SEGFAULT)
> 	102 - TestOpenCLNonbondedForceDouble (SEGFAULT)
> ```

I know what causes this. I'll open an issue in a couple of hours and describe what I found.

ex-rzr:
@jchodera BTW, there is an issue with amber20-factorix: https://github.com/openmm/openmm/issues/3391
I've noticed that you encountered it too as your json results don't have results for it.
I guess it would be great to have a special value (0?) or a flag for failed tests. What do you think?

There is a build-in way to pass multiple value for an argument (`nargs='*'`), see:
https://github.com/openmm/openmm/blob/331654d67d2fa157135fc30cbc3ec9236819a688/examples/benchmark.py#L210-L211
For such arguments you need to pass a space-separated values without =, like `--test gbsa amber20-dhfr`, the result is a list.
It also checks passed values for correctness (i.e. compares it with `choices`).

I'm not sure though that it's better that your way with a comma-separated list, just FYI.

peastman:
I don't know what causes the segfaults with AMD CPU OpenCL.  They happen on CI, but I've never been able to reproduce them when running locally.  We're also using a fairly old version of AMD OpenCL (the old APP SDK).  I've been meaning to try updating it to use the ROCm toolkit and see if that fixes them.  I just haven't gotten around to it.

I think I like the comma separated list better than the alternate syntax.  At least, it seems closer to how I would expect it to work.

peastman:
Could we switch some of the defaults back to running single tests rather than a matrix of tests?  Default ensemble should be NVT instead of all, default precision should be single instead of all, and default constraints should be hbonds instead of all.  You're optimizing the arguments for running a whole lot of benchmarks at once, but that's something we do infrequently.  I run this script constantly, almost on a daily basis, when doing optimizing or profiling.  Having to specify three extra arguments to run a single test will make it a lot less convenient for that.

examples/benchmark.py
```diff
@@ -174,40 +350,140 @@ def runOneTest(testName, options):
...
+parser.add_argument('--pme-cutoff', default=0.9, dest='pme_cutoff', type=float, help='direct space cutoff for PME in nm [default: 0.9]')
 parser.add_argument('--seconds', default=60, dest='seconds', type=float, help='target simulation length in seconds [default: 60]')
-parser.add_argument('--polarization', default='mutual', dest='polarization', choices=('direct', 'extrapolated', 'mutual'), help='the polarization method for AMOEBA: direct, extrapolated, or mutual [default: mutual]')
+parser.add_argument('--polarization', default='mutual', dest='polarization', choices=POLARIZATION_MODES, help='the polarization method for AMOEBA: {POLARIZATION_MODES} [default: mutual]')
 parser.add_argument('--mutual-epsilon', default=1e-5, dest='epsilon', type=float, help='mutual induced epsilon for AMOEBA [default: 1e-5]')
-parser.add_argument('--heavy-hydrogens', action='store_true', default=False, dest='heavy', help='repartition mass to allow a larger time step')
+parser.add_argument('--bond-constraints', default=None, dest='bond_constraints', help='hbonds: constrain bonds to hydrogen, use 1.5*amu H mass; allbonds: constrain all bonds, use 4*amu H mass [default: all]')
```
peastman, line 382: The revised description no longer says that it uses a larger time step.  Probably also good to point out this is ignored for AMOEBA.

examples/benchmark.py
```diff
@@ -31,140 +124,223 @@ def downloadAmberSuite():
...
+                cutoff = pme_cutoff
             elif testName == 'apoa1ljpme':
                 method = app.LJPME
-                cutoff = options.cutoff
+                cutoff = pme_cutoff
             else:
+                # Reaction field uses hard-coded cutoff
                 method = app.CutoffPeriodic
-                cutoff = 1*unit.nanometers
+                cutoff = 1*unit.nanometers # JDC: Shouldn't this be larger for reaction field?
```
peastman, line 205: 1 nm is a common value to use with reaction field.

examples/benchmark.py
```diff
@@ -31,140 +124,223 @@ def downloadAmberSuite():
...
         else:
             ff = app.ForceField('amber99sb.xml', 'amber99_obc.xml')
             pdb = app.PDBFile('5dfr_minimized.pdb')
-            method = app.CutoffNonPeriodic
+            method = app.CutoffNonPeriodic # JDC: Shouldn't this be app.NoCutoff?
```
peastman, line 219: 2 nm cutoff is a common value to use with GB.  No cutoff is too slow to be usable for anything except very small systems.

examples/benchmark.py
jchodera: Any idea where this number comes from? Is there at least some precedent we can cite?

examples/benchmark.py
jchodera: Any idea where this number comes from? Is there at least some precedent we can cite?

jchodera:
> I guess it would be great to have a special value (0?) or a flag for failed tests. What do you think?

That's a great idea---I can capture those in the JSON as well, along with the exception/error message.

jchodera:
> There is a build-in way to pass multiple value for an argument (nargs='*'), see:

Thanks! That's a better approach---I'll switch to that.

jchodera:
> @jchodera BTW, there is an issue with amber20-factorix: #3391

Thanks for opening this issue---I'm trying to track down what the correct behavior here should be.

jchodera:
> Could we switch some of the defaults back to running single tests rather than a matrix of tests? Default ensemble should be NVT instead of all, default precision should be single instead of all, and default constraints should be hbonds instead of all. You're optimizing the arguments for running a whole lot of benchmarks at once, but that's something we do infrequently. I run this script constantly, almost on a daily basis, when doing optimizing or profiling. Having to specify three extra arguments to run a single test will make it a lot less convenient for that.

Happy to do this if this makes sense. We'd like to support three main use cases, I imagine:
1. @peastman running a specific set of profiling (corresponding to most representative use cases?) daily to test code improvements---is this the old defaults?
2. A new user or developer producing a complete profiling of new hardware (perhaps to automatically submit to a future benchmark server)
3. Profiling a specific test system / combination to test new code/kernels

What would you recommend as far as simple arguments for supporting these three use cases?

jchodera:
As an alternative, we could provide a way to specify a YAML file that would contain one or more "benchmarking profiles", where you could simply specify
```
python benchmark.py --profile legacy
```
and it would pluck the combination to run from `benchmarking-profiles.yaml` by default (which could also be user-configured). That way, you can set up a bunch of profiles you use frequently and easily specify which one to use.

peastman:
> @peastman running a specific set of profiling (corresponding to most representative use cases?) daily to test code improvements---is this the old defaults?

That isn't something I do much.  The really important one is

> Profiling a specific test system / combination to test new code/kernels

That's the thing that I may do many times in a single day.  It isn't a specific test or set of tests.  It's whatever is relevant for the code I'm working on right at that moment.  Often interspersed with adding debugging code to collect more information during a particular run.

jchodera:
OK, so you're asking me to _just_ change the defaults for ensemble, precision, and constraints so you don't need to specify `--ensemble NVT --precision single --constraints hbonds` each time you want to run tests?

jchodera:
@peastman : I can quickly finish up this PR so we can use this version (and capture JSON output) for benchmarking the 7.7 release.

peastman:
> so you're asking me to just change the defaults for ensemble, precision, and constraints

Exactly!

jchodera:
Thanks for the clarification! Will do!

examples/benchmark.py
```diff
@@ -403,63 +412,73 @@ def runOneTest(testName, options):
...
+
+    from rich import box
+    table = Table(box=box.SIMPLE)
+    table.add_column("[red]Test", width=24)
+    table.add_column("[blue]Precision")
+    table.add_column("[blue_violet]Constraints")
+    table.add_column("[blue_violet]H mass (amu)")
+    table.add_column("[blue_violet]dt (fs)")
+    table.add_column("[blue]Platform", width=9)
+    table.add_column("[orange1]Ensemble")
```
tictooc, line 571: Rich output for "Platform" and "Ensemble" are opposite of the table headers

jchodera: Great catch---thanks!

jchodera:
@peastman : I've made the requested changes. This should be ready for inclusion in the beta.

Suggest we recommend folks try
```
$ conda install rich
$ python benchmark.py --style rich --ensemble=NVT,NPT --precision=single,mixed,double --bond-constraints=hbonds --seconds 5 --outfile benchmark.json
```
and send us the `benchmark.json` files.
Output looks like this (though note I haven't updated my branch, so these numbers don't reflect the latest performance data):
```
$ python benchmark.py --style rich --ensemble=NVT,NPT --precision=single,mixed,double --bond-constraints=hbonds --seconds 5 --outfile benchmark-test.json
hostname: lt01
timestamp: 2022-07-04T05:39:25.950560
openmm_version: 7.7.0.dev-0644f05
cpuinfo: Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz
cpuarch: x86_64
system: Linux
nvidia_driver: 515.43.04
gpu: NVIDIA GeForce GTX 1080 Ti
                                                                                                                   
  Test                       Precision   Constraints   H mass (amu)   dt (fs)   Ensemble   Platform        ns/day  
 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────── 
  gbsa                       single      HBonds        1.5            4.0       NVT        CUDA            1803.3  
  gbsa                       mixed       HBonds        1.5            4.0       NVT        CUDA            1596.7  
  gbsa                       double      HBonds        1.5            4.0       NVT        CUDA              63.5  
  gbsa                       single      HBonds        1.5            4.0       NVT        OpenCL           732.9  
  gbsa                       mixed       HBonds        1.5            4.0       NVT        OpenCL           706.3  
  gbsa                       double      HBonds        1.5            4.0       NVT        OpenCL            59.9  
  gbsa                       mixed       HBonds        1.5            4.0       NVT        CPU                2.8  
  gbsa                       double      HBonds        1.5            4.0       NVT        Reference          0.6  
  rf                         single      HBonds        1.5            4.0       NVT        CUDA            1111.9  
  rf                         mixed       HBonds        1.5            4.0       NVT        CUDA             945.8  
  rf                         double      HBonds        1.5            4.0       NVT        CUDA              76.1  
  rf                         single      HBonds        1.5            4.0       NVT        OpenCL           829.3  
  rf                         mixed       HBonds        1.5            4.0       NVT        OpenCL           735.7  
  rf                         double      HBonds        1.5            4.0       NVT        OpenCL            55.6  
  rf                         mixed       HBonds        1.5            4.0       NVT        CPU                5.8  
  rf                         double      HBonds        1.5            4.0       NVT        Reference          0.3  
  rf                         single      HBonds        1.5            4.0       NPT        CUDA            1027.2  
...
```
If they don't want to install `rich`, simply omitting `--style rich` will use the default `simple` output format.

In the meantime, we can work on a way to automatically capture the JSON emitted to a webpage to automate collection of performance stats.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
+        outfile.write('</config>\n')
+
+    import yaml
+    with open(os.path.join(directory, 'metadata.yaml'), 'wt') as outfile:
+        outfile.write(yaml.dump(metadata))
+
+def runOneTest(testName, options):
+    """Perform a single benchmarking simulation."""
+
+    # Return if we do not support the requested precision
```
peastman, line 286: I'd prefer to remove this check.  It makes it impossible to benchmark the Reference or CPU platform without manually specifying precision, which will be annoying in practice.  Let's just say that the precision option is ignored for these platforms.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
+
+    explicit = (testName not in ('gbsa', 'amoebagk'))
+    amoeba = (testName in ('amoebagk', 'amoebapme'))
+    apoa1 = testName.startswith('apoa1')
+    amber = (testName.startswith('amber'))
+    hydrogenMass = None
+
+    if amoeba:
+        test_result['epsilon'] = options.epsilon
+    elif (testName in ['pme', 'ljpme']) or testName.startswith('amber'):
```
peastman, line 304: That should be `['pme', 'apoa1pme', 'apoa1ljpme']`.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
+        test_result['constraints'] = 'None'
+        test_result['hydrogen_mass'] = '1'
+    elif options.bond_constraints == 'hbonds':
+        test_result['constraints'] = 'HBonds'
+        test_result['hydrogen_mass'] = '1.5'
+    elif options.bond_constraints == 'allbonds':
+        test_result['constraints'] = 'AllBonds'
+        test_result['hydrogen_mass'] = '4'
+    else:
+        raise ValueError(f'Unknown bond_constraints: {bond_constraints}')
```
peastman, line 317: This duplicates logic that's already present in `retrieveTestSystem()`.  It would be better if we could unify them.  Otherwise there's a high risk that someone will modify one and not realize they have to modify the other as well.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
             else:
                 integ = mm.LangevinMiddleIntegrator(temperature, friction, dt)
-        positions = pdb.positions
-        system = ff.createSystem(pdb.topology, nonbondedMethod=method, nonbondedCutoff=cutoff, constraints=constraints, hydrogenMass=hydrogenMass)
-    if options.ensemble == 'NPT':
-        system.addForce(mm.MonteCarloBarostat(1*unit.bar, temperature, 100))
-    print('Step Size: %g fs' % dt.value_in_unit(unit.femtoseconds))
+        elif options.bond_constraints == 'allbonds':
+            dt = 0.005*unit.picoseconds
+            hydrogenMass = 4*unit.amu
```
peastman, line 349: This is never used.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
     context.setPositions(positions)
     if amber:
-        if inpcrd.boxVectors is not None:
-            context.setPeriodicBoxVectors(*inpcrd.boxVectors)
         mm.LocalEnergyMinimizer.minimize(context, 100*unit.kilojoules_per_mole/unit.nanometer)
     context.setVelocitiesToTemperature(temperature)
 
+    # Take one step (to enforce constraints and trigger kernel compulation) and store the initial state
+    integ.step(1)
+    state = context.getState(getPositions=True, getVelocities=True, getEnergy=True, getForces=True, getParameters=True)
```
peastman, line 393: This is unnecessary.  `timeIntegration()` already takes care of it.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
+
+        metadata = {
+            'name' : testName,
+            'description' : f'OpenMM Benchmark Suite : {testName}',
+            'precision' : options.precision,
+            'num_atoms' : context.getSystem().getNumParticles(),
+            'nonbonded_method' : f'{nonbonded_method}',
+            'timestep' : integ.getStepSize() / unit.picoseconds,
+            'integrator' : f'{integ.__class__.__name__}',
+            }
```
peastman, line 443: Does all of this really belong here?  Serializing the system and integrator is fine, but this is quite a lot of F@H specific code being put into OpenMM.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
 parser.add_argument('--seconds', default=60, dest='seconds', type=float, help='target simulation length in seconds [default: 60]')
-parser.add_argument('--polarization', default='mutual', dest='polarization', choices=('direct', 'extrapolated', 'mutual'), help='the polarization method for AMOEBA: direct, extrapolated, or mutual [default: mutual]')
+parser.add_argument('--polarization', default='mutual', dest='polarization', choices=POLARIZATION_MODES, help=f'the polarization method for AMOEBA: {POLARIZATION_MODES} [default: mutual]')
 parser.add_argument('--mutual-epsilon', default=1e-5, dest='epsilon', type=float, help='mutual induced epsilon for AMOEBA [default: 1e-5]')
-parser.add_argument('--heavy-hydrogens', action='store_true', default=False, dest='heavy', help='repartition mass to allow a larger time step')
+parser.add_argument('--bond-constraints', default=None, dest='bond_constraints', help=f'hbonds: constrain bonds to hydrogen, use 1.5*amu H mass; allbonds: constrain all bonds, use 4*amu H mass, and use larger timestep (if not AMOEBA): {BOND_CONSTRAINTS} [default: hbonds]')
```
peastman, line 473: To make it slightly clearer: "This option is ignored for AMOEBA."

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
+    platform = mm.Platform.getPlatformByName(options.platform)
     integ.setConstraintTolerance(1e-5)
     if len(properties) > 0:
         context = mm.Context(system, integ, platform, properties)
     else:
         context = mm.Context(system, integ, platform)
+
+    # Store information about the Platform used by the Context
+    platform = context.getPlatform()
+    test_result['precision'] = options.precision # requested precision
```
peastman, line 381: Only store precision for CUDA and OpenCL.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
+    TESTS = requested_tests
+
+if args.precision is not None:
+    # Use specified subset of precisions
+    requested_precisions = args.precision.split(',')
+    if not set(requested_precisions).issubset(PRECISIONS):
+        parser.error(f'Available precisions: {PRECISIONS}')
+    PRECISIONS = requested_precisions
+else:
+    PRECISIONS = ['single'] # Peter's preferred default
```
peastman, line 534: In this and the others below, why not use the `default` option to `add_argument()`?  Then this block could be simplified to

```python
precisions = args.precision.split(',')
if not set(precisions).issubset(PRECISIONS):
    parser.error(f'Available precisions: {PRECISIONS}')
```

Overwriting `PRECISIONS` is also confusing.  People will assume a name in all capitals is an immutable constant.  Putting the user's selection into a different variable called `precisions` is clearer.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
...
-            print('Test failed: %s' % ex.message)
+        except OpenMMException as e:
+            if args.verbose:
+                print(e)
+            pass
+elif args.style == 'rich':
+    from rich.live import Live
+    from rich.table import Table
+
+    from rich import box
```
peastman, line 570: I'm not familiar with `rich`.  We don't document anywhere that it's a dependency.  And adding a new dependency just for this seems excessive.

peastman (commented):
Thanks!  I made a first pass through it.

examples/benchmark.py
peastman: Even better, we can check `'Precision' in platform.getPropertyNames()` to see whether the selected platform supports precision.  That way we don't need to hardcode a list of platforms.

examples/benchmark.py
```diff
@@ -6,6 +6,99 @@
...
+    # Do nothing if filename is None
+    if filename is None:
+        return
+
+    import os
+    all_results = { 'benchmarks' : list() }
+    if system_info is not None:
+        all_results['system'] = system_info
+
+    if filename.endswith('.yaml'):
```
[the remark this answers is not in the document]
jchodera: I couldn't find any examples of writing JSON with the `yaml` package, so I don't know how to implement this now.

examples/benchmark.py
```diff
@@ -31,183 +124,469 @@ def downloadAmberSuite():
         tarfh.extractall(path=dirname)
     return dirname
 
-def runOneTest(testName, options):
-    """Perform a single benchmarking simulation."""
+import functools
+@functools.cache
```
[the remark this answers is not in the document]
jchodera: I've changed this to `lru_cache`, which is present in earlier versions.

examples/benchmark.py
jchodera: Fixed, but it would be better if there was a simpler, more modular way to register tests and their permitted values for these combinations of options.

examples/benchmark.py
jchodera: We're aiming to benchmark each release of OpenMM on all available GPUs, which requires serializing out all the benchmarks to a new Folding@home project each release. This is the easiest way to do it without maintaining separate benchmarking scripts, which runs the risk of having to keep both in perfect sync to be a valid comparison.

This is just an example script anyway, so there's no harm in keeping it here if it serves an active purpose each release and within the development cycles.

jchodera:
@peastman: I've made the requested changes.

Should we change the default 60 seconds (which is pretty long and inconvenient for most users) to something more reasonable, like 5 or 10 seconds?

examples/benchmark.py
jchodera: [`rich`](https://github.com/Textualize/rich) is an amazingly popular (1.2M downloads) flexible terminal formatting library.

It's not a dependency---it's optional and not enabled by default. The user needs to specify `--style rich` to enable this.

examples/benchmark.py
jchodera: It's necessary to ensure subsequent serialization for Folding@home benchmark WU generation will match the same initial conditions for local benchmarking.

examples/benchmark.py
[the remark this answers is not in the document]
peastman: Just use the `json` package for either filename extension.  YAML is a strict superset of JSON.  Every JSON file is also a YAML file.

jchodera:
@peastman: Would there be utility in refactoring this into a module so that benchmarks could be run via
```python
python -m openmm.benchmark.cli ...
```
or to have an entry point created to enable a CLI entry point in the path:
```bash
$ openmm-benchmark ...
```
instead of requiring users to find where the examples directory is installed?

examples/benchmark.py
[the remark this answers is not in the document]
jchodera: The purpose of using YAML is that the preferred serialization is human readable, whereas JSON is not easily human-readable. Generating JSON formatted output for a YAML file would defeat the purpose of having YAML as a human-readable alternative in the first place.

examples/benchmark.py
peastman: `timeIntegration()` already takes some initial steps to ensure everything is warmed up.  Since those steps are nondeterministic (assuming constant temperature), the starting state at the end of them will be different either way.  But that also makes no difference for benchmarking.  The speed difference for very slightly different starting states is at the noise level.

examples/benchmark.py
peastman: An optional dependency is still a dependency.  It's something that needs to be tested, and that will break our code when they make changes to future versions.

Let's use a simple tab or comma delimited output.  It may not look quite as pretty on the screen, but that's not a serious concern for benchmarking.  And it will be very easy to copy and paste into a spreadsheet.

peastman:
> Should we change the default 60 seconds (which is pretty long and inconvenient for most users) to something more reasonable, like 5 or 10 seconds?

10 seconds is way too short to get reliable results on a GPU.  If I'm running quick benchmarks for testing I'll usually specify 30 seconds, which gets the fluctuations down to a few percent.  For official benchmarks, 60 seconds is better.

CPU benchmarks are a lot less noisy.  10 seconds is usually enough to get fairly consistent results with them.  But using different defaults for different platforms is probably just unnecessary complexity.

examples/benchmark.py
```diff
@@ -389,9 +389,10 @@ def runOneTest(testName, options):
         mm.LocalEnergyMinimizer.minimize(context, 100*unit.kilojoules_per_mole/unit.nanometer)
     context.setVelocitiesToTemperature(temperature)
 
-    # Take one step (to enforce constraints and trigger kernel compulation) and store the initial state
-    integ.step(1)
-    state = context.getState(getPositions=True, getVelocities=True, getEnergy=True, getForces=True, getParameters=True)
+    if options.serialize:
+        # Take a step to enforce constraints on positions and velocities if needing to serialize for Folding@home
+        integ.step(1)
+        state = context.getState(getPositions=True, getVelocities=True, getEnergy=True, getForces=True, getParameters=True)
```
peastman, line 395: If constraints are what you're concerned about, call `context.applyConstraints()`.  Taking a step and retrieving a state aren't necessary.

We just called `setVelocitiesToTemperature()` in the previous line, which already applies constraints to velocities.

examples/benchmark.py
jchodera: Removing this check would mean that benchmarks that run over multiple combinations of platform and precision will erroneously report that the Reference and CPU platforms were run in multiple precision modes (single, mixed, double) where the precision is actually ignored.

examples/benchmark.py
peastman: You forgot to remove this line when you added the check below.

examples/benchmark.py
jchodera: After applying position constraints, the velocity constraints will no longer be accurate, will they?

examples/benchmark.py
peastman: You probably don't want to use a single run of the script for all options on all platforms.  Different platforms require different settings.  For example, the CPU platform generally should use a shorter PME cutoff than the GPU platforms.

And this check will be a serious annoyance for me.  I run this script constantly, often many times in a single day.  Having to add an unnecessary argument every time I do will be a pain.  It also will be very confusing to other users.  If you try to benchmark the Reference or CPU platform and don't explicitly specify a precision, it will simply exit without running any benchmarks and not explain why.

examples/benchmark.py
jchodera: > An optional dependency is still a dependency. 

That makes absolutely no sense.

> It's something that needs to be tested, and that will break our code when they make changes to future versions.

We are doing zero continuous integration testing right now, so this is irrelevant.

> Let's use a simple tab or comma delimited output. It may not look quite as pretty on the screen, but that's not a serious concern for benchmarking. And it will be very easy to copy and paste into a spreadsheet.

Happy for you to implement this if you want, but this library is orders of magnitude more popular than OpenMM, so I don't agree there is a serious issue with this being a problematic optional dependency.

examples/benchmark.py
jchodera: > You probably don't want to use a single run of the script for all options on all platforms. Different platforms require different settings. For example, the CPU platform generally should use a shorter PME cutoff than the GPU platforms.

I'd love to move to a strategy where we autotune any platform-dependent optimizable parameters, but failing that, is there a way we could encode all these optimal combinations in some modular way?

examples/benchmark.py
peastman: > We are doing zero continuous integration testing right now, so this is irrelevant.

All code needs to be tested, whether it's done through CI or manually.  Do you think the benchmarking script has never been tested?  And if it's not being tested through CI, that's all the more recent not to introduce dependencies that could cause it to spontaneously break at unknown times.

> Happy for you to implement this if you want

Ok, I'll make the changes and push them to your branch.

examples/benchmark.py
peastman: Autotuning is great as a future feature, but you have made a change that will make my life more difficult.  I probably run this script 10 times for every time anyone in the world other than me runs it.  The normal use case for this script is to run a single benchmark.  That will continue to be by far the most common way it gets run.  Please remove this check.

peastman: I suggest we remove the ability to test multiple platforms on a single run of the script.  It's clear that creates more problems than it's worth.  There are too many differences between platforms to make that a sensible thing to do in most cases/

examples/benchmark.py
jchodera: > All code needs to be tested, whether it's done through CI or manually. Do you think the benchmarking script has never been tested? And if it's not being tested through CI, that's all the more recent not to introduce dependencies that could cause it to spontaneously break at unknown times.

No, I'm just saying that an equivalent level of testing is being performed now as was performed before. Nothing has degraded by introducing this optional library---I have tested it with and without.

> We are doing zero continuous integration testing right now, so this is irrelevant.

All code needs to be tested, whether it's done through CI or manually. Do you think the benchmarking script has never been tested? And if it's not being tested through CI, that's all the more recent not to introduce dependencies that could cause it to spontaneously break at unknown times.

> Ok, I'll make the changes and push them to your branch.

Feel free to. I've run out of time to implement things you clearly want. The only thing I care about in this PR is the ability for others to run the benchmarking script and generate a human- and machine-readable artifact they can share with us. It would be even better if we could automate the deposition of that artifact somewhere central, but I don't have a good idea about how to easily do this yet.

peastman:
Argh.  I just pushed some changes to the script, and now it's listing every change made to OpenMM in the last year as a change in this PR.  Even though all of them are already present in the repository, and so shouldn't be seen as changes.  I'm not sure what's causing that or how to fix it.

Anyway, here are the changes I made to the script.

- I removed the dependency on `rich`.
- I removed the option to benchmark multiple platforms at once.  It causes problems and usually isn't a good idea anyway.
- It no longer requires you to specify the precision on the command line.  For Reference and CPU it automatically fills in the correct precision.
- I fixed some bugs.

peastman:
@jchodera can you look over the changes to benchmark.py and see if they look ok?

If I can't figure out why git is confused, I might need to close this PR and open a new one with just the updates to the script.

jchodera:
Apologies for the delay---I was out of commission with COVID for several weeks.

This is definitely an improvement over the previous benchmarking scripts!

Suggestions for final changes before merging:
* **Help text:** We need to add some more text to the help that is printed when you use `python benchmark.py` or `python benchmark.py --help` to give one or more examples of how we expect folks to run it. 
* **What do we want folks to run and send back?** Since there is no longer a way to specify multiple platforms, we're back to folks manually having to run a bunch of commands to run the benchmark, which means we're not going to get everything we want from folks. Are we sure there isn't a way to automate "run all valid combinations"? If not, what's the most important thing we want people to run and send back? Can we add this as an example to the help text too?

Thanks for finishing this up!

peastman:
I'm really sorry you were sick.  I hope you're doing better now?

I expect it to be very uncommon for someone to want to send us full benchmark suites for multiple platforms on a single computer.  If it has an NVIDIA GPU, they'll benchmark CUDA.  If it's a different kind of GPU, they'll benchmark OpenCL.  There might occasionally be cases where someone would benchmark CPU.  There's no reason for anyone ever to benchmark Reference.  So I don't expect that to be a problem in practice.  And if they really do want to benchmark multiple platforms, they'll just run the script once for each platform.

peastman:
This is superseded by #3887.

PR #3394, opened 2021-12-27
Add Fortran type translation for `long long`

yishutu:
There are some methods with `long long` types, which were introduced in #3248. However, Fortran cannot recognize such type, making compiling errors with Fortran code.

I have added the translation for `long long` to `integer*8` in the translation list. Making the Fortran wrappers compatible.


peastman:
Thanks, looks good!

Issue #3409, opened 2022-01-12
Is it possible to break down a force into individual interactions?


Gareth-elliott:
Hi OpenMM Developers,

I have been converting some parameters from tinker to openmm and (especially after the fix in #3376 ) this has been mostly successful. I did have an issue today with the conversion of one particular molecule (ethanol). All the terms are fine, except the stretch-bend term:

TINKER:
 Energy Component Breakdown :           Kcal/mole        Interactions
 Bond Stretching                           0.9611                8
 Angle Bending                             0.2219               13
 **Stretch-Bend                              0.0186                9**
 Torsional Angle                           0.0011               12
 Van der Waals                             0.8595               15
 Atomic Multipoles                        -0.9873               15
 Polarization                             -0.2529               15

OPENMM:
AmoebaBond = 0.9610670410837722
AmoebaAngle = 0.22193213789688226
AmoebaInPlaneAngle = 0.0
**AmoebaStretchBend = -0.00495009064193708**
PeriodicTorsionForce = 0.0010703517696935574
AmoebaVdwForce = 0.8595058606240733
AmoebaMultipoleForce = -1.2402446877061468
CMMotionRemover = 0.0

My question is if it is possible to decompose the stretch-bend force into the individual interactions to make sure that there are the same number and that everything is assigned correctly (or if there is a better way to check). Something like the output from tinker ANALYZE:

 Individual Stretch-Bend Interactions :

 Type                  Atom Names                  dSB 1     dSB 2      Energy

 StrBend         4-H        1-C        5-C        0.0067    0.0043      0.0023
 StrBend         1-C        5-C        6-H       -0.0097   -0.0114     -0.0056
 StrBend         7-H        5-C        8-O       -0.0250    0.0085      0.0076
 StrBend         5-C        8-O        9-H        0.0121   -0.0233      0.0099
 StrBend         2-H        1-C        5-C        0.0059    0.0037      0.0020
 StrBend         3-H        1-C        5-C        0.0059    0.0037      0.0020
 StrBend         1-C        5-C        8-O       -0.0039    0.0015     -0.0016
 StrBend         1-C        5-C        7-H       -0.0097   -0.0114     -0.0056
 StrBend         6-H        5-C        8-O       -0.0250    0.0085      0.0076

Thanks,
Gareth

peastman:
You can print out all stretch-bend terms and their parameters by doing something like this:

```python
for force in system.getForces():
  if force.getName() == 'AmoebaStretchBend':
[...]

PR #3420, opened 2022-01-21
Fix for non-equal force constants in stretch-bend

Gareth-elliott:
Fix for #3409 suggested by @peastman.

Feel free to edit etc (its possible I may have got some formatting incorrect), and let me know if you want me to add anything? (maybe a test?)

G


peastman:
Thank you!

PR #3421, opened 2022-01-22
TargetArch.cmake supports ARM Macs

peastman:
This is necessary to allow compiling on ARM Macs.  At least, it *should* be necessary, and for me it *is* necessary.  So I'm really confused about how our existing ARM Mac builds manage to work.  Can anyone explain what's going on?

cc @mikemhenry


mikemhenry:
I'm really not sure, I checked with `file` (or whatever osx/bsd uses) to ensure the binary was arm, but I did set arch manually as part of the CI, so maybe that is how it worked? :shrug:

PR #3424, opened 2022-01-25
Fix GIL for OpenMM::Context::setVelocitiesToTemperature

raimis:
Release the GIL when calling `OpenMM::Context::setVelocitiesToTemperature`, for the same reason as https://github.com/openmm/openmm/pull/3061

Fix https://github.com/openmm/openmm-torch/issues/61.

### DIFF
--- a/wrappers/python/src/swig_doxygen/swig_lib/python/exceptions.i
+++ b/wrappers/python/src/swig_doxygen/swig_lib/python/exceptions.i
@@ -51,3 +51,17 @@
     }
     PyEval_RestoreThread(_savePythonThreadState);
 }
+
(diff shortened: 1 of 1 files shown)


raimis:
@peastman could you review?

peastman:
Looks good, thanks.

jchodera:
@peastman can we tag this issue for potential inclusion in a bugfix release
to enable the openmm-ml release? Maybe we just want to use a new milestone
or label for that purpose now, and we can discuss whether a bugfix release
is needed when everything else is ready?

peastman:
Sure, that would be fine.

PR #3463, opened 2022-02-15
Fix GIL for OpenMM::CustomCVForce::getCollectiveVariableValues

raimis:
Release the GIL when calling `OpenMM::CustomCVForce::getCollectiveVariableValues`, for the same reason as https://github.com/openmm/openmm/pull/3061 and https://github.com/openmm/openmm/pull/3424

Fix https://github.com/openmm/openmm-ml/issues/22.


jchodera:
Are there other areas where we should evaluate whether we need to also release the GIL?

raimis:
@peastman 
> Looks good. You have this marked as a draft. Are you planning further changes?

No, just forgot to mark it ready.

peastman:
Done!

Issue #3446, opened 2022-02-04
ZeroDivisionError:float division by zero


HrishiDhondge:
I am getting this error while using pdbfixer, but when I trace back the error; this actually arises from the pdbstructure.py file.
To replicate this error: 
- 1) Download any NMR structure from PDB (2RS2)
- 2) Separate the chains using VMD as follows 
```
vmd -dispdev text 2RS2.pdb
set sel1 [atomselect top "chain A and protein"]
$sel1 writepdb protA.pdb
exit
```
- 3) Now use pdbfixer on the protA.pdb with all default parameters

This error arises due to the fact that VMD assigns the structure wrong/faulty unit cell parameters. For structures other than X-ray crystallography, it should be as follows (according to PDB file format documentation)
`CRYST1    1.000    1.000    1.000  90.00  90.00  90.00 P 1           1 `

But the separated chain from VMD has 
`CRYST1    0.000    0.000    0.000  90.00  90.00  90.00 P 1           1 `

and the main cause of the problem here. 
So I propose to add a little modification in a few lines from the pdbstructure.py script -
The original lines are as follows:
```
elif command == "CRYST1": 
    a_length = float(pdb_line[6:15])*0.1 
    b_length = float(pdb_line[15:24])*0.1 
    c_length = float(pdb_line[24:33])*0.1  
``` 

I propose one of the two ways:    
```
elif command == "CRYST1": 
    a_length = float(pdb_line[6:15]) *0.1
    b_length = float(pdb_line[15:24])*0.1
    c_length = float(pdb_line[24:33]) *0.1
    if (a_length ==b_length) and (b_length == c_length) and (c_length == 0):
        a_length, b_length, c_length = 1.0, 1.0, 1.0
``` 
The above method sets the default unit cell parameters for other methods, in case the parameters are set wrong in the PDB file.

```
elif command == "CRYST1": 
    a_length = float(pdb_line[6:15]) *0.1
    b_length = float(pdb_line[15:24])*0.1
    c_length = float(pdb_line[24:33]) *0.1
    if (a_length ==b_length) and (b_length == c_length) and (c_length == 0):
        continue
```
The above method simply ignores the unit cell parameters and proceed with the other computations.

Which one of these two would be better to incorporate in order to avoid such errors?

peastman:
It's a good question.  I think I would go with this

```python
if not (a_length == b_length == c_length == 0):
[...]

PR #3469, opened 2022-02-18
Improved handling of CRYST1 records #3446

HrishiDhondge:


wrappers/python/openmm/app/internal/pdbstructure.py
```diff
@@ -196,7 +196,8 @@ def _load(self, input_stream):
                 alpha = float(pdb_line[33:40])*math.pi/180.0
                 beta = float(pdb_line[40:47])*math.pi/180.0
                 gamma = float(pdb_line[47:54])*math.pi/180.0
-                self._periodic_box_vectors = computePeriodicBoxVectors(a_length, b_length, c_length, alpha, beta, gamma)
+                if not (a_length == b_length == c_length == 0):
```
swails, line 199: What about if 1 or 2 dimensions are zero?  Simulations won't work well in that case, either.

swails: Thanks for the correction. Edited my suggestion.

HrishiDhondge: Yes, this suggestion seems more precise.

peastman:
This fixes #3446.

peastman:
Thanks!

PR #3473, opened 2022-02-19
Add library path with os.add_dll_directory()

peastman:
I'm hopeful this will fix the problems with Python 3.10 on Windows.


peastman:
Maybe it fixed the problem, but I can't tell because now it doesn't even get that far.  Some time between #3463 (three days ago) and #3464 (two days ago), the Windows builds start failing to even run CMake correctly.

```
CMake Error at CMakeLists.txt:21 (PROJECT):
  The CMAKE_C_COMPILER:

    cl.exe

  is not a full path and was not found in the PATH.
```

I don't know why.  Builds are spontaneously breaking faster than I can fix them.

PR #3480, opened 2022-02-22
Added option for box shape when building solvent boxes

### LINKED ISSUE #3124: Add truncated octahedral solvation option for Modeller.addSolvent()
It turns out that it's quite difficult to automate solvation within a truncated octahedron.

Could we add an option to `Modeller.addSolvent()` that would allow the user to select between standard box shapes?

cc : @glass-w, who can provide more information on use cases.

peastman:
Fixes #3124.


peastman:
Does anyone have comments, or shall I merge this?

peastman:
The PowerPC build with conda-forge compilers is failing with segfaults and nans.  It's not related to this PR, which only has Python changes.  The same errors are showing up in #3481.  It's probably due to a change in a compiler version or something like that.

PR #3481, opened 2022-02-23
OpenCL platform has option to choose throttle level

### LINKED ISSUE #2955: lowering CPU usage of OpenMM GPU to free a CPU core
Current CPU usage of OpenMM on GPU is quite high in Folding@Home:  I do see 100% of one core on Intel GPUs with OpenCL and Nvidia GPUs with Cuda on both Linux and Windows.

Because of this, in Folding@Home, one CPU thread is reserved to each GPU slot, so I'm opening this issue so we can explore ways to lower that CPU usage and free cores back to CPU folding (or just save some watts).

The first cause of CPU usage is GPU events synchronisation:
For CUDA, it seems there is already a contextProperties["UseBlockingSync"] that could be set to "true" (already default in openmm) instead of false in F@H openmm: https://github.com/openmm/openmm/issues/1541

For OpenCL, since OpenCL 2.2, the _cl_khr_throttle_hints_ extension is defined and allows to specify the throttling that's needed.
It allows to create a command queue with _CL_QUEUE_THROTTLE_KHR_ set to _CL_QUEUE_THROTTLE_LOW_KHR_ (4). That tells the runtime it may run in a way that's optimized, with lowest energy consumption.
On Intel GPUs, that hint is supported and translates to blocking sync, hence waiting on most OpenCL events takes no CPU.

I've been able to test on Intel GPUs using https://github.com/intel/opencl-intercept-layer to force that setting (DefaultQueueThrottleHint=4) on Windows and Linux and it brings CPU usage down from 100% to 5-18% of a single thread.

The cost in OpenMM GPU performance varies: with F@H 13429 work-unit it's **none** in Windows in _high perf_ mode, -12% in _balanced_ mode. On Linux it's worse though, -35%.

Since there is a varying impact on performance, maybe it should be up to OpenMM user / F@H client to make a choice - could an OpenMM option be exposed to use low throttle queue as well as blocking sync, up to F@H custom slots settings?

Then F@H testers will be able to more easily determine the impact on various platforms and the strategy at F@H client level could be improved afterward. Linking this setting to the "Folding Power" one could maybe provide a good default option to most users.

peastman:
Hopefully this fixes #2955.  I don't have an Intel GPU I can test it on.  @ph0b and @bb30994 are either of you able to test it?  When creating a Context, set the `"OpenCLThrottle"` property to either `"low"`, `"medium"`, or `"high"`.  The default is medium.

### DIFF
--- a/platforms/opencl/src/OpenCLContext.cpp
+++ b/platforms/opencl/src/OpenCLContext.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2009-2020 Stanford University and the Authors.      *
+ * Portions copyright (c) 2009-2022 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -58,6 +58,12 @@ using namespace std;
 #ifndef CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV
   #define CL_DEVICE_COMPUTE_CAPABILITY_MINOR_NV 0x4001
 #endif
+#ifndef CL_QUEUE_THROTTLE_KHR
+    #define CL_QUEUE_THROTTLE_KHR 0x1097
+    #define CL_QUEUE_THROTTLE_HIGH_KHR (1<<0)
+    #define CL_QUEUE_THROTTLE_MED_KHR (1<<1)
+    #define CL_QUEUE_THROTTLE_LOW_KHR (1<<2)
+#endif
 
 const int OpenCLContext::ThreadBlockSize = 64;
 const int OpenCLContext::TileSize = 32;
@@ -77,7 +83,8 @@ static bool isSupported(cl::Platform platform) {
             vendor.find("Intel") == 0);
 }
 
-OpenCLContext::OpenCLContext(const System& system, int platformIndex, int deviceIndex, const string& precision, OpenCLPlatform::PlatformData& platformData, OpenCLContext* originalContext) :
+OpenCLContext::OpenCLContext(const System& system, int platformIndex, int deviceIndex, const string& precision, const string& throttle,
+        OpenCLPlatform::PlatformData& platformData, OpenCLContext* originalContext) :
         ComputeContext(system), platformData(platformData), numForceBuffers(0), hasAssignedPosqCharges(false),
         integration(NULL), expression(NULL), bonded(NULL), nonbonded(NULL) {
     if (precision == "single") {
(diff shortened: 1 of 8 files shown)


peastman:
Is anyone able to test this?  I don't have any appropriate hardware.

yafshar:
> Is anyone able to test this? I don't have any appropriate hardware.

It fails to run with below error `exception: Error initializing context: clCreateContext (-64)`  
The exception is happening at the `clCreateContext` in `opencl.hpp:2872`  and `cl_khr_throttle_hints` is enabled.

peastman:
What hardware and OpenCL implementation are you using?  What version of OpenCL does it support?

yafshar:
> What hardware and OpenCL implementation are you using? What version of OpenCL does it support?

I have tried on `Intel Iris Plus Graphics 650` and `Intel Iris Graphics 550` with Intel OpenCL implementation. It is OpenCL 3.0.

peastman:
Error -64 is `CL_INVALID_PROPERTY`.  That suggests it doesn't accept the `CL_QUEUE_THROTTLE_KHR` property.  But it only gets included if OpenCL reports that it supports the `cl_khr_throttle_hints` extension:

```c++
if (platforms[bestPlatform].getInfo<CL_PLATFORM_EXTENSIONS>().find("cl_khr_throttle_hints") != string::npos) {
    cprops.push_back(CL_QUEUE_THROTTLE_KHR);
    if (throttle == "low")
        cprops.push_back(CL_QUEUE_THROTTLE_LOW_KHR);
    else if (throttle == "medium")
        cprops.push_back(CL_QUEUE_THROTTLE_MED_KHR);
    else if (throttle == "high")
        cprops.push_back(CL_QUEUE_THROTTLE_HIGH_KHR);
    else
        throw OpenMMException("Illegal value for OpenCLThrottle");
}
```

Can you verify that Intel OpenCL 1) really does claim to support that extension, but 2) refuses to accept the property for it?

platforms/opencl/src/OpenCLContext.cpp
```diff
@@ -286,9 +293,21 @@ OpenCLContext::OpenCLContext(const System& system, int platformIndex, int device
             compilationDefines["SYNC_WARPS"] = "barrier(CLK_LOCAL_MEM_FENCE)";
         vector<cl::Device> contextDevices;
         contextDevices.push_back(device);
-        cl_context_properties cprops[] = {CL_CONTEXT_PLATFORM, (cl_context_properties) platforms[bestPlatform](), 0};
+        vector<cl_context_properties> cprops = {CL_CONTEXT_PLATFORM, (cl_context_properties) platforms[bestPlatform]()};
+        if (platforms[bestPlatform].getInfo<CL_PLATFORM_EXTENSIONS>().find("cl_khr_throttle_hints") != string::npos) {
+            cprops.push_back(CL_QUEUE_THROTTLE_KHR);
```
yafshar, line 298: @peastman Here, I think you are mixing the queue properties with context properties!
I suggest st like below,
```cpp
        if (platforms[bestPlatform].getInfo<CL_PLATFORM_EXTENSIONS>().find("cl_khr_throttle_hints") != string::npos) {
            vector<cl_queue_properties> qprops = {CL_QUEUE_THROTTLE_KHR};
            if (throttle == "low")
                qprops.push_back(CL_QUEUE_THROTTLE_LOW_KHR);
            else if (throttle == "medium")
                qprops.push_back(CL_QUEUE_THROTTLE_MED_KHR);
            else if (throttle == "high")
                qprops.push_back(CL_QUEUE_THROTTLE_HIGH_KHR);
            else
                throw OpenMMException("Illegal value for OpenCLThrottle");
            qprops.push_back(0);
        }
```
and then you need `clCreateCommandQueueWithProperties`, but I think you should update the `CommandQueue` interface in the `opencl.hpp` file, since currently it seems to accept only `QueueProperties`.

peastman:
Good point, thanks!  I'll fix it.

peastman:
It looks like it's impossible to specify this property without using `clCreateCommandQueueWithProperties()`, which was added in OpenCL 2.0?  We still only require 1.2.  Is there any way to specify it without breaking compatibility with older versions?

yafshar:
>It looks like it's impossible to specify this property without using clCreateCommandQueueWithProperties(), which was added in OpenCL 2.0? We still only require 1.2. Is there any way to specify it without breaking compatibility with older versions?

I also tried a couple of things, but unfortunately, I think there is no way to support it for OpenCL 1.2. If your program uses an OpenCL 1.2 and tries to call `clCreateCommandQueueWithProperties`, it will crash

peastman:
Thanks.  I'm closing this since we still need to support 1.2.

PR #3485, opened 2022-02-24
Use importlib_metadata instead of pkg_resources

### LINKED ISSUE #3397: Use importlib_metadata instead of pkg_resources to ensure compatibility with condacolab
We currently use `pkg_resources` to look for entry points where OpenMM force fields are installed, but this [does not work with tools like condacolab](https://github.com/conda-incubator/condacolab/issues/20).

The recommended scheme is now apparently [to use `importlib_metadata`](https://github.com/conda-incubator/condacolab/issues/20#issuecomment-1003342938).

We should update our usage (and perhaps our conda dependencies) to be more broadly compatible.

peastman:
Fixes #3397.

cc @jchodera @jaimergp

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -57,11 +57,11 @@ def _getDataDirectories():
     if _dataDirectories is None:
         _dataDirectories = [os.path.join(os.path.dirname(__file__), 'data')]
         try:
-            from pkg_resources import iter_entry_points
-            for entry in iter_entry_points(group='openmm.forcefielddir'):
+            from importlib_metadata import entry_points
+            for entry in entry_points().select(group='openmm.forcefielddir'):
                 _dataDirectories.append(entry.load()())
         except:
-            pass # pkg_resources is not installed
+            pass # importlib_metadata is not installed
     return _dataDirectories
 
 def _convertParameterToNumber(param):
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -56,12 +56,24 @@ def _getDataDirectories():
     global _dataDirectories
     if _dataDirectories is None:
         _dataDirectories = [os.path.join(os.path.dirname(__file__), 'data')]
+        eps = None
         try:
-            from importlib_metadata import entry_points
-            for entry in entry_points().select(group='openmm.forcefielddir'):
-                _dataDirectories.append(entry.load()())
+            from importlib.metadata import entry_points
+            eps = entry_points()
         except:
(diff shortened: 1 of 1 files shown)


wrappers/python/openmm/app/forcefield.py
```diff
@@ -57,11 +57,11 @@ def _getDataDirectories():
...
         _dataDirectories = [os.path.join(os.path.dirname(__file__), 'data')]
         try:
-            from pkg_resources import iter_entry_points
-            for entry in iter_entry_points(group='openmm.forcefielddir'):
+            from importlib_metadata import entry_points
+            for entry in entry_points().select(group='openmm.forcefielddir'):
                 _dataDirectories.append(entry.load()())
         except:
-            pass # pkg_resources is not installed
+            pass # importlib_metadata is not installed
```
jaimergp, line 64: I think from Python 3.8 this is a native module, so you can do `import importlib.metadata`.

The full construct would look more like:

```python
try:
    from importlib import metadata as _metadata
except ModuleNotFoundError:
    import importlib_metadata as _metadata

_metadata.entry_points().select(...)
```

peastman: It appears the API of the native module has changed with time.  `select()` doesn't exist prior to Python 3.10.  Instead `entry_points()` returns a dict.  `importlib_metadata` also supports the dict interface, but prints a deprecation warning to the console.  Supporting all the different combinations will be a bit of a mess.

peastman: Does this look right to you?

```python
eps = None
try:
    from importlib.metadata import entry_points
    eps = entry_points()
except:
    try:
        from importlib_metadata import entry_points
        eps = entry_points()
    except:
        pass # importlib_metadata is not installed
if eps is not None:
    if hasattr(eps, 'select'):
        entries = eps.select(group='openmm.forcefielddir')
    else:
        entries = eps['openmm.forcefielddir']
    for entry in entries:
          _dataDirectories.append(entry.load()())
```

peastman: Or we could just assume ` importlib_metadata` will be installed.  I think it always will be on miniconda?

jaimergp: I don't think we can assume it will always be installed (but we can add it to the `run` dependencies). That said, the code proposed above does look solid!

peastman: I tried the above code, but several of the builds are failing.  I'm going to revert it and just assume `importlib_metadata` is installed.  It generally seems to be anyway, and we can add it as a dependency if we want to make absolutely sure.

PR #3489, opened 2022-02-28
Fix possible bug in `AmberPrmTopFile.createSystem()`

### LINKED ISSUE #3488: Possible issue in `AmberPrmTopFile.createSystem` logic
I'm on the trail of a case in which `GBSAOBCForce.setUseSwitchingDistance()` is called inside of `AmberPrmTopFile.createSystem`, which causes a crash because that force does not have that attribute. I could be totally wrong here, but by my eye the [indentation here](https://github.com/openmm/openmm/blob/7.7.0/wrappers/python/openmm/app/amberprmtopfile.py#L294-L310) doesn't match what I'd expect. The `if switchingDistance ...` logic on L300 seems like it should be inside of the `if isinstance(force, mm.NonbondedForce):` check on L295 so that it's only called when `force` is an object that  supports a switching distance. Otherwise, that block seems to be called on whatever is the last force in `sys.getForces()`, which may typically be a non-bonded force but as far as I'm aware that's not guaranteed. These lines haven't been modified since they were added (https://github.com/openmm/openmm/pull/1272) and I don't see other issues on the tracker, so I guess it normally works.

The most minimal reproduction I can come up with is something like this:

```
In [1]: from openmm import app, unit

In [2]: sys = app.AmberPrmtopFile("gbsa.tmp.prmtop").createSystem(
   ...:     nonbondedMethod=app.CutoffNonPeriodic,
   ...:     implicitSolvent=app.OBC2,
   ...:     switchDistance=0.7 * unit.nanometer,
   ...: )
/Users/mwt/miniconda3/envs/openff-interchange-env/lib/python3.8/site-packages/openmm/app/internal/amber_file_parser.py:1102: UserWarning: Non-optimal GB parameters detected for GB model OBC2
  warnings.warn(
AttributeError                            Traceback (most recent call last)
Input In [2], in <module>
----> 1 sys = app.AmberPrmtopFile("gbsa.tmp.prmtop").createSystem(
      2     nonbondedMethod=app.CutoffNonPeriodic,
      3     implicitSolvent=app.OBC2,
      4     switchDistance=0.7 * unit.nanometer,
      5 )

File ~/miniconda3/envs/openff-interchange-env/lib/python3.8/site-packages/openmm/app/amberprmtopfile.py:310, in AmberPrmtopFile.createSystem(self, nonbondedMethod, nonbondedCutoff, constraints, rigidWater, implicitSolvent, implicitSolventSaltConc, implicitSolventKappa, temperature, soluteDielectric, solventDielectric, removeCMMotion, hydrogenMass, ewaldErrorTolerance, switchDistance, gbsaModel)
    307     if _strip_optunit(switchDistance, u.nanometer) < 0:
    308         # Detects negatives for both Quantity and float
    309         raise ValueError('switchDistance must be non-negative!')
--> 310     force.setUseSwitchingFunction(True)
    311     force.setSwitchingDistance(switchDistance)
    313 return sys

AttributeError: 'GBSAOBCForce' object has no attribute 'setUseSwitchingFunction'
```

Does this seem sensible?

I opened a PR that fixes this for my local configuration: #3499

This can probably be reproduced with some vanilla files but this is the one I used: [gbsa.tmp.prmtop.txt](https://github.com/openmm/openmm/files/8157442/gbsa.tmp.prmtop.txt)

mattwthompson:
May resolve #3488 if my diagnosis is correct

The repro in the linked issue becomes:
```python3
In [1]: from openmm import app, unit

In [2]: sys = app.AmberPrmtopFile("gbsa.tmp.prmtop").createSystem(
   ...:     nonbondedMethod=app.CutoffNonPeriodic,
   ...:     implicitSolvent=app.OBC2,
   ...:     switchDistance=0.7 * unit.nanometer,
   ...: )
/Users/mwt/miniconda3/envs/openff-interchange-env/lib/python3.8/site-packages/openmm/app/internal/amber_file_parser.py:1102: UserWarning: Non-optimal GB parameters detected for GB model OBC2
  warnings.warn(

In [3]: import openmm

In [4]: for force in sys.getForces():
   ...:     if type(force) == openmm.NonbondedForce:
   ...:         print(force.getSwitchingDistance())
0.7 nm

### DIFF
--- a/wrappers/python/openmm/app/amberprmtopfile.py
+++ b/wrappers/python/openmm/app/amberprmtopfile.py
@@ -288,7 +288,7 @@ class AmberPrmtopFile(object):
                     sys.setParticleMass(atom2.index, hydrogenMass)
                     sys.setParticleMass(atom1.index, sys.getParticleMass(atom1.index)-transferMass)
         for force in sys.getForces():
-            if isinstance(force, mm.NonbondedForce):
+            if isinstance(force, (mm.NonbondedForce, mm.CustomNonbondedForce)):
                 force.setEwaldErrorTolerance(ewaldErrorTolerance)
                 if switchDistance and nonbondedMethod is not ff.NoCutoff:
                     # make sure it's legal
--- a/wrappers/python/openmm/app/amberprmtopfile.py
+++ b/wrappers/python/openmm/app/amberprmtopfile.py
@@ -288,8 +288,9 @@ class AmberPrmtopFile(object):
                     sys.setParticleMass(atom2.index, hydrogenMass)
                     sys.setParticleMass(atom1.index, sys.getParticleMass(atom1.index)-transferMass)
         for force in sys.getForces():
-            if isinstance(force, (mm.NonbondedForce, mm.CustomNonbondedForce)):
+            if isinstance(force, mm.NonbondedForce):
                 force.setEwaldErrorTolerance(ewaldErrorTolerance)
+            if isinstance(force, (mm.NonbondedForce, mm.CustomNonbondedForce)):
                 if switchDistance and nonbondedMethod is not ff.NoCutoff:
                     # make sure it's legal
                     if (_strip_optunit(switchDistance, u.nanometer) >=
--- a/wrappers/python/openmm/app/amberprmtopfile.py
+++ b/wrappers/python/openmm/app/amberprmtopfile.py
@@ -290,19 +290,19 @@ class AmberPrmtopFile(object):
         for force in sys.getForces():
             if isinstance(force, mm.NonbondedForce):
                 force.setEwaldErrorTolerance(ewaldErrorTolerance)
+                if switchDistance and nonbondedMethod is not ff.NoCutoff:
+                    # make sure it's legal
+                    if (_strip_optunit(switchDistance, u.nanometer) >=
+                            _strip_optunit(nonbondedCutoff, u.nanometer)):
+                        raise ValueError('switchDistance is too large compared '
+                                         'to the cutoff!')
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by peastman** [comment]
That's definitely a bug.  I suggest a slightly different fix.  If the file uses nbfix terms, LJ will be implemented with a CustomNonbondedForce instead of a NonbondedForce.  So I would put that block inside a new loop:

```python
for force in sys.getForces():
    if isinstance(force, mm.NonbondedForce) or isinstance(force, mm.CustomNonbondedForce):
        ...
```

mattwthompson:
Seems good to me -  I've updated with that behavior.

wrappers/python/openmm/app/amberprmtopfile.py
```diff
@@ -288,21 +288,21 @@ def createSystem(self, nonbondedMethod=ff.NoCutoff, nonbondedCutoff=1.0*u.nanome
                     sys.setParticleMass(atom2.index, hydrogenMass)
                     sys.setParticleMass(atom1.index, sys.getParticleMass(atom1.index)-transferMass)
         for force in sys.getForces():
-            if isinstance(force, mm.NonbondedForce):
+            if isinstance(force, (mm.NonbondedForce, mm.CustomNonbondedForce)):
                 force.setEwaldErrorTolerance(ewaldErrorTolerance)
```
peastman, line 292: CustomNonbondedForce doesn't have a `setEwaldErrorTolerance()` method.  That's why we need separate checks for that (which should only be called on NonbondedForce) and for the switching function (which should be called on either of them).

peastman:
Looks good.  Thanks!

mattwthompson:
Thanks Peter!

Issue #2913, opened 2020-11-07
Does CustomRMSDForce prevent atoms within the specified atom group from being periodically imaged?


jchodera:
When a `CustomRMSDForce` is defined for a set of atoms, is this atom set prevented from being placed in different periodic images by `MonteCarloBarostat`?

jchodera:
For example, I have a simple `CustomCVForce` to restrain the RMSD between a protein and a ligand that are not otherwise covalently connected:
```python
        from simtk import unit, openmm
        kB = unit.AVOGADRO_CONSTANT_NA * unit.BOLTZMANN_CONSTANT_kB 
        temperature = 300 * unit.kelvin
        kT = kB * temperature     
        sigma = 1.0 * unit.angstrom
        buffer = 2.0 * unit.angstrom
        custom_cv_force = openmm.CustomCVForce('step(RMSD-buffer)*(K_RMSD/2)*(RMSD-buffer)^2')
        custom_cv_force.addGlobalParameter('K_RMSD', kT / sigma**2)
        custom_cv_force.addGlobalParameter('buffer', buffer)
        rmsd_force = openmm.RMSDForce(self.hybrid_positions, rmsd_atom_indices)
        custom_cv_force.addCollectiveVariable('RMSD', rmsd_force)
        system.addForce(custom_cv_force)
```
My guess is that the barostat is imaging the ligand into different boxes from the protein, since I see this if I report the CV value:
```
iteration    27 RMSD 0.854 A
iteration    28 RMSD 0.853 A
iteration    29 RMSD 0.775 A
iteration    30 RMSD 0.760 A
iteration    31 RMSD 0.768 A
iteration    32 RMSD 0.758 A
iteration    33 RMSD 46.048 A
iteration    34 RMSD 46.021 A
iteration    35 RMSD 0.833 A
```
Would adding a zero-energy `CustomBondForce` between two of these atoms prevent this?

jchodera:
It appears that this is the case: Adding a zero-energy `CustomBondForce` seems to solve this!
```
iteration    49 RMSD 0.715 A
iteration    50 RMSD 0.667 A
iteration    51 RMSD 0.860 A
iteration    52 RMSD 0.653 A
iteration    53 RMSD 0.642 A
iteration    54 RMSD 0.651 A
iteration    55 RMSD 0.638 A
iteration    56 RMSD 0.639 A
iteration    57 RMSD 0.812 A
iteration    58 RMSD 0.701 A
[...]

PR #3490, opened 2022-03-01
Barostat does not reset cell offsets

peastman:
See https://github.com/openmm/openmm/issues/2898#issuecomment-1055664566.


jchodera:
@peastman : Does this have any impact on #2913?

peastman:
No effect.  The RMSDForce doesn't look at the cell offsets.

PR #3505, opened 2022-03-07
Fix carboxylate atom radii in GB-Neck2 implicit solvent

jgreener64:
Currently no atoms get marked as carboxylate because this line always returns `False`.

This is missed by the force test as alanine dipeptide has no carboxylate. The `GBn2_NoCutoff.xml` file used at https://github.com/openmm/openmm/blob/master/wrappers/python/tests/TestCharmmFiles.py#L304 will change - let me know if you want me to update this.


peastman:
Thanks for catching this!  Yes, fixing the test would be great.

jgreener64:
Sure, have done so.

peastman:
Thanks!

PR #3510, opened 2022-03-10
Workaround for error in finding broken URLs

peastman:
The docs build has been failing because linkinator incorrectly reports https://www.pnas.org/content/105/17/6290 as being broken.  I [reported the issue](https://github.com/JustinBeckwith/linkinator/issues/386), but so far haven't received any reply.  For now, I'm telling it to skip checking any URLs at www.pnas.org.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -650,7 +650,7 @@ jobs:
         shell: bash -l {0}
         env:
           # Pipe-seperated list of domains to skip
-          SKIP_DOMAIN: support.amd.com
+          SKIP_DOMAIN: support.amd.com|www.pnas.org
         run: |
           set +e
           # Linkinator accepts regex for its --skip argument, so we


jchodera:
What if we just replace the pnas.org link with the doi.org link?
https://doi.org/10.1073/pnas.0711686105

peastman:
That doesn't work either:

```
$ npx linkinator https://doi.org/10.1073/pnas.0711686105
🏊‍♂️ crawling https://doi.org/10.1073/pnas.0711686105
[503] https://doi.org/10.1073/pnas.0711686105
[200] https://www.cloudflare.com/5xx-error-landing/

  [503] https://doi.org/10.1073/pnas.0711686105
https://doi.org/10.1073/pnas.0711686105
  [200] https://www.cloudflare.com/5xx-error-landing/
ERROR: Detected 1 broken links. Scanned 2 links in 0.639 seconds.
```

PR #3512, opened 2022-03-11
computeDrudeTemperature

Olllom:
Adds a convenience function to compute the instantaneous Drude temperature in the `DrudeNoseHooverIntegrator` and `DrudeLangevinIntegrator`.

### DIFF
--- a/plugins/drude/openmmapi/src/DrudeHelpers.cpp
+++ b/plugins/drude/openmmapi/src/DrudeHelpers.cpp
@@ -127,12 +127,12 @@ vector<Vec3> assignDrudeVelocities(const System &system, double temperature, dou
     return velocities;
 }
 
-double computeSystemTemperatureFromVelocities(const System& system, const vector<Vec3>& velocities) {
+pair<double, double> computeSystemTemperatureFromVelocities(const System& system, const vector<Vec3>& velocities) {
     vector<int> normalParticles;
     vector<pair<int, int> > pairParticles;
     findParticlesAndPairs(system, normalParticles, pairParticles);
-    double energy = 0.0;
-    int dof = 0;
+    double energy = 0.0, drudeEnergy = 0;
+    int dof = 0, drudeDof = 0;
     
     // Kinetic energy and degrees of freedom from normal particles.
     
@@ -151,10 +151,14 @@ double computeSystemTemperatureFromVelocities(const System& system, const vector
         int p2 = pair.second;
         double mass1 = system.getParticleMass(p1);
         double mass2 = system.getParticleMass(p2);
+	double reducedMass = (mass1*mass2)/(mass1+mass2);
         if (mass1 != 0 || mass2 != 0) {
             Vec3 momentum = mass1*velocities[p1] + mass2*velocities[p2];
             energy += momentum.dot(momentum)/(mass1+mass2);
-            dof += 3;
+	    Vec3 drudeVelocity = velocities[p1] - velocities[p2];
+            drudeEnergy += reducedMass*drudeVelocity.dot(drudeVelocity);
+	    dof += 3;
+	    drudeDof += 3;
         }
     }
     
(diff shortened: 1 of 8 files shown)


plugins/drude/openmmapi/src/DrudeHelpers.cpp
```diff
@@ -127,12 +127,12 @@ vector<Vec3> assignDrudeVelocities(const System &system, double temperature, dou
     return velocities;
 }
 
-double computeSystemTemperatureFromVelocities(const System& system, const vector<Vec3>& velocities) {
+pair<double, double> computeSystemTemperatureFromVelocities(const System& system, const vector<Vec3>& velocities) {
```
peastman, line 130: The function name is no longer accurate.  Maybe rename it to `computeTemperaturesFromVelocities()`, and add a comment above describing what it computes?

plugins/drude/openmmapi/src/DrudeLangevinIntegrator.cpp
```diff
@@ -30,6 +30,7 @@
  * -------------------------------------------------------------------------- */
 
 #include "openmm/DrudeLangevinIntegrator.h"
+#include "SimTKOpenMMRealType.h"
```
peastman, line 33: What is this for?  I don't think you added any references to things defined in this file.

plugins/drude/openmmapi/src/DrudeNoseHooverIntegrator.cpp
```diff
@@ -30,6 +30,7 @@
  * -------------------------------------------------------------------------- */
 
 #include "openmm/DrudeNoseHooverIntegrator.h"
+#include "SimTKOpenMMRealType.h"
```
peastman, line 33: Same here (unless I missed something).

peastman:
Aside from the minor comments above, this looks good.

plugins/drude/openmmapi/src/DrudeHelpers.cpp
Olllom: done

plugins/drude/openmmapi/src/DrudeLangevinIntegrator.cpp
Olllom: good catch, thanks

plugins/drude/openmmapi/src/DrudeNoseHooverIntegrator.cpp
Olllom: done

peastman:
There's an inconsistency in the implementation.  We need to decide how it should work.

When you call `getState()` to query the kinetic energy, it shifts the velocities to account for the integrator's half step offset.  That's necessary so the kinetic and potential energy correspond to the same point in time.  Otherwise, you couldn't add them and get a conserved total energy.

StateDataReporter computes the temperature based on the shifted kinetic energy.  To be consistent, I therefore made `computeSystemTemperature()` also apply the same offset.  It computes the on-step rather than half-step temperature.

For DrudeLangevinIntegrator, you made `computeDrudeTemperature()` also apply the offset.  But for DrudeNoseHooverIntegrator, you implemented it by calling `computeMaskedKineticEnergy()`, which does not apply an offset.  So the two classes compute slightly different things.

1. Should `computeDrudeTemperature()` compute on-step or half-step temperature?  It should be consistent across both classes.
2. Should NoseHooverIntegrator be applying the offset when it computes kinetic energy in other places?

Olllom:
The Nose-Hoover implementation is a Velocity Verlet so that all velocities, kinetic energies, and temperatures should be reported full-step. This includes `computeSystemTemperature` and `computeDrudeTemperature`. 

Actually, the second-to-last commit in this PR https://github.com/openmm/openmm/pull/3512/commits/698fa7433a8e4041ecdd4d30d9b42c61252a686c had parallel implementations of `computeDrudeTemperature` for Nose-Hoover and Langevin. This should work in principle but for some reason failed the ReferenceDrudeNoseHoover tests [here](https://github.com/openmm/openmm/runs/5514903885?check_suite_focus=true).

From how I understand the code, both the current version and https://github.com/openmm/openmm/pull/3512/commits/698fa7433a8e4041ecdd4d30d9b42c61252a686c should return the full-step temperature but their return values were slightly different (only on the Reference platform). I did not find a bug.

Issue #3361, opened 2021-12-02
Slowdown in `getState()` in REST hybrid system vs vanilla hybrid system?


zhang-ivy:
When running hamiltonian repex (for relative free energy calculations) using openmmtools, we first propagate all replicas (run MD) and then compute the energy of each thermodynamic state at each replica. I've written a new factory for generating REST capable hybrid systems (see details at bottom) and am seeing a slowdown (compared to the original/vanilla hybrid systems) in the `getState()` [call](https://github.com/choderalab/openmmtools/blob/master/openmmtools/states.py#L972) when we compute the energy at each thermodynamic state. Each call (after the first) takes ~0.0006 sec (vanilla hybrid system) vs ~0.06 seconds (REST hybrid system).

I've done some initial experiments to determine where the slowdown is coming from (all independent experiments):
1. If I delete the custom nonbonded forces + standard nonbonded force: rest system getState() calls speed up from 0.06 to ~0.0002 sec
2. If I turn off the long range correction in the `CustomNonbondedForce` for sterics,  rest system getState() calls speed up from 0.06 to ~0.001 sec

As you can see in experiment 2, even with the long range correction off, there is still a slowdown (0.001 sec vs 0.0006 sec) incurred by the REST hybrid system compared to the vanilla system. 

@peastman : Any suggestions for how to determine where the rest of the slowdown is coming from (see below for more details on the differences in implementations)?  I've attached scripts and and system/state xmls in case you have time to help diagnose this (note you'll need to install perses from the `rest-over-protocol` branch).

-- Details on REST hybrid system -- 
Handles nonbondeds with:
- a `CustomNonbondedForce` to handle softcored direct space PME electrostatics interactions (long range correction off),
- a `CustomNonbondedForce` to handle softcored steric interactions (long range correction on),
- a `CustomBondForce` to handle direct space PME electrostatics and steric exceptions, and
- a `NonbondedForce` to handle reciprocal space PME electrostatics (long range correction off).

This system differs from the vanilla hybrid system in the following ways:
- The vanilla hybrid system handles nonbondeds with: 
    - a `NonbondedForce` that handles electrostatics and sterics (for non-alchemical interactions)
    - a `CustomNonbondedForce` that handles sterics for alchemical interactions only (softcored)
- REST hybrid system has a lot more global and per-particle parameters
[rest-system-slowdown-files.zip](https://github.com/openmm/openmm/files/7644988/rest-system-slowdown-files.zip)

[...]

PR #3520, opened 2022-03-17
Optimized computing powers in CompiledExpression

peastman:
This is another optimization to JIT compilation of custom expressions on the CPU.  Simulating DHFR with CHARMM36 on the CPU platform, it increases the speed from 7.4 ns/day to 10 ns/day.  There may also be some benefit for #3361.

### DIFF
--- a/devtools/ci/gh-actions/conda-envs/build-ubuntu-latest.yml
+++ b/devtools/ci/gh-actions/conda-envs/build-ubuntu-latest.yml
@@ -7,6 +7,7 @@ dependencies:
 - cmake
 - make
 - ccache
+- sysroot_linux-64 2.17
 # host
 - python
 - cython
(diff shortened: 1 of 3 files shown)


swails:
Is this real from the conda-forge compilers?

```
/home/runner/work/openmm/openmm/libraries/asmjit/asmjit/core/cpuinfo.cpp:18:12: fatal error: sys/auxv.h: No such file or directory
   18 |   #include <sys/auxv.h>
      |            ^~~~~~~~~~~~
compilation terminated.
```

peastman:
I think it's real.  I don't know why they're missing that file.  It's [part of glibc](https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sys/auxv.h;h=b5ab30ab777ec6d6aed94628db3d97534ad2e564;hb=HEAD).  I found [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1443400) from four years ago reporting that it was missing on some BSDs, but it should always be present on Linux.

I think we need to raise the issue with the conda-forge maintainers.  I'm not sure what the best place to report it is.  On the feedstock for the compilers package?

peastman:
The fix was to update the package containing glibc.  That build is passing now.

PR #3521, opened 2022-03-18
DCDFile reports CHARMM version 21 instead of 24

### LINKED ISSUE #3507: truncated octahedron from LEaP
Hi,

I have been preparing my systems in the LEaP module of AmberTools and have noticed that simulations setup with a truncated octahedron (solvateoct) solvent box are being treated as rectangular boxes in my OpenMM simulations. This creates imaging issues during processing as well as likely applying some unwanted forces on the system during equilibration. 

Is there a way to have OpenMM work with the input files I've generated with AmberTools?

Thanks for your help!

peastman:
Fixes #3507.  This should produce DCD files that are read correctly by cpptraj.  @drroe does this look correct to you?


drroe:
> @drroe does this look correct to you?

This is essentially what cpptraj does now so I think it should work.

peastman:
Thanks!

PR #3538, opened 2022-03-30
Added a few Lepton tests

peastman:
These tests were supposed to have been included in #3520 but accidentally got left out.


peastman:
The CUDA builds are failing with network errors trying to install CUDA.  That isn't related to this PR and hopefully it's just an intermittent problem.

Issue #3544, opened 2022-04-01
Drop support for processors without AVX


peastman:
The first CPUs with AVX were released in 2011.  It has been standard on all x86 processors from both Intel and AMD for many years.  What do you think about dropping support for processors that don't include it?

This came up in the context of #3351.  Supporting vectorization of Lepton expressions both with and without AVX support would add a lot of complexity.  Alternatively I could make older processors still work, but they wouldn't get JIT compilation of expressions anymore, which would make custom forces on the CPU platform slower.

jchodera:
We could do a quick Twitter poll to see if there is any need to support x86
without AVX.

Any suggestions on the wording and options?

peastman:
You could just ask whether they're still using old x86 processors without AVX support.  Options could include "Yes", "Yes, but not for anything where CPU performance is critical", and "No".

jchodera:
Just posted the twitter poll here:
https://twitter.com/openmm_toolkit/status/1510685264481763328

I had to tailor the options a bit since each option must be 25 characters or less.

peastman:
I see the result are up.  40% of respondents said they're still using an older CPU, which is a lot higher than I would have expected.  Clearly we need to maintain compatibility a while longer.  A much smaller fraction (18.5%) are using the CPU platform on it, and presumably some even smaller fraction of those are using custom forces.  Also, the text you used ("use CPU") doesn't really capture the distinction I was looking for ("anything where CPU performance is critical").

As an intermediate step, what about continuing to support older CPUs, but JIT compilation of custom expressions will require AVX?  They'll still work without JIT compilation, just not as quickly.

swails:
In my recalled experience, even single point calculations on the CPU platform were uselessly slow when using custom nonbonded potentials (like the custom GB forces which constitute the defacto implementation of all the GB models in Amber except one) on standard protein sized systems.

swails:
65 votes are also only a small number, and probably overrepresent the old processor support requests given they would seem to have the most to lose by not responding.

So on the one hand JIT probably matters the most to the old systems that will be unusably slow without it, but the workarounds are simple:

- Use an older version of OpenMM 
- Get a newer cpu (plenty of compatible refurb options at affordable costs)
- Run on a budget CPU
- Use the OpenCL platform on the CPU (I think?)

I'd drop JIT support for pre-AVX procs, personally.

jchodera:
Here are the poll results, for reference:
[...]

PR #3552, opened 2022-04-06
Vectorize custom expressions on CPU

peastman:
This is a major update to JIT compilation of custom expressions on the CPU.  The biggest change is a new CompiledVectorExpression class that vectorizes computation of expressions with AVX (on x86) or NEON (on ARM).  This should allow speedups to custom forces through vectorization.  Nothing in OpenMM uses this class yet.  That will come in future PRs.

There also are other improvements to JIT compilation that should produce speedups even without vectorization.  It now produces optimized code for `min()`, `max()`, `abs()`, `floor()`, `ceil()`, and `select()`.  Expressions that use those functions should now be faster than before.

As discussed in #3544, JIT compilation now requires x86 processors to support AVX.  Older processors will still work, but custom forces will be slower on the CPU platform.


peastman:
It looks like the Windows test failure is probably caused by a bug in AsmJit.  See https://github.com/asmjit/asmjit/issues/363.  Hopefully there will be a fix soon.

PR #3558, opened 2022-04-08
Generate more informative error message if residue template not found

### LINKED ISSUE #3557: When residue template cannot be found, print more informative error message
Right now, when `ForceField` fails to match a residue template, it only emits the residue name and index in the topology.
If the topology contains multiple chains or residue id numbering does not start at 0, this information is not as helpful as it might be.
```
ValueError: No template found for residue 1168 (ASN). The set of atoms matches ASN, but the bonds are different.
```
Instead, an error message like
```
ValueError: No template found for chain G, resid 213, resname ASN (residue index 1168 in topology). 
The set of atoms matches ASN, but the bonds are different.
```
would be much more helpful.

An example of this appears in issue #3549.

jchodera:
Fixes #3557.

Instead of generating error messages that index the residue within the entire topology when a residue template is not found:
```
ValueError: No template found for residue 1168 (ASN). The set of atoms matches ASN, but the bonds are different.
```
we now generate a more useful error message:
```
ValueError: No template found for chainid <G>, resid <213>, resname <ASN> (residue index in topology 1168). 
The set of atoms matches ASN, but the bonds are different.
```

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -486,14 +486,14 @@ class ForceField(object):
     def registerScript(self, script):
         """Register a new script to be executed after building the System."""
         self._scripts.append(script)
-    
+
     def registerTemplateMatcher(self, matcher):
         """Register an object that can override the default logic for matching templates to residues.
 
         A template matcher is a callable object that can be invoked as::
-        
+
             template = f(forcefield, residue, bondedToAtom, ignoreExternalBonds, ignoreExtraParticles)
-        
+
         where ``forcefield`` is the ForceField invoking it, ``residue`` is an openmm.app.Residue object,
         ``bondedToAtom[i]`` is the set of atoms bonded to atom index i, and ``ignoreExternalBonds`` and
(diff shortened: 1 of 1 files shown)


peastman:
Lots of test failures.  You don't need to restructure the loop just to get the chain.  Given a residue you can access it as `res.chain`.

jchodera:
Sorry about that---don't quite have an openmm dev environment set up on my osx-arm laptop yet!

jchodera:
Looks like tests pass except this AMD OpenCL test:
```
The following tests FAILED:
	 80 - TestOpenCLLangevinIntegratorMixed (Subprocess aborted)
Errors while running CTest
    Start 80: TestOpenCLLangevinIntegratorMixed
1/1 Test #80: TestOpenCLLangevinIntegratorMixed ...Subprocess aborted***Exception:   7.99 sec
corrupted double-linked list


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   8.00 sec

The following tests FAILED:
	 80 - TestOpenCLLangevinIntegratorMixed (Subprocess aborted)
Errors while running CTest
Rerunning failing tests...
```

peastman:
Looks good.

PR #3574, opened 2022-04-14
Fixed inconsistency in computing kinetic energy

### LINKED ISSUE #3570: Kinetic energy computed differently for CUDA/OpenCL vs CPU/Reference platforms for LangevinMiddleIntegrator
It looks like `LangevinMiddleIntegrator` computes the kinetic energy differently between CUDA/OpenCL and CPU/Reference platforms:
```
$ python test-energy.py
CUDA             : kinetic energy : 416673.43781221827 kJ/mol
OpenCL           : kinetic energy : 416673.43781221827 kJ/mol
CPU              : kinetic energy : 415709.5186723646 kJ/mol
Reference        : kinetic energy : 415709.5186723646 kJ/mol
```
See attached ZIP file for script and input files to reproduce:
[kinetic-energy-discrepancy.zip](https://github.com/openmm/openmm/files/8486018/kinetic-energy-discrepancy.zip)

If the integrator is modified to use `VerletIntegrator`, the kinetic energies are all consistent:
```
$ python test-energy.py
CUDA             : kinetic energy : 405598.4631463683 kJ/mol
OpenCL           : kinetic energy : 405598.4631342366 kJ/mol
CPU              : kinetic energy : 405598.46493097965 kJ/mol
Reference        : kinetic energy : 405598.4704745067 kJ/mol
```

peastman:
Fixes #3570.  Many integrators apply a half step offset to the velocities when computing kinetic energy.  After updating the velocities, it needs to apply constraints to them to make sure the updated velocities still respect the constraints.

LangevinMiddleIntegrator does not apply an offset, but on the CPU and Reference platforms it was still applying constraints.  In the example from that issue, the velocities being loaded into the Context did not respect the constraints.  This led to different kinetic energies being reported by different platforms.

I changed Reference and CPU to behave the same way OpenCL and CUDA do: if no offset is applied, do not apply constraints.  That means if you call `setVelocities()` or `setState()`, the kinetic energy will exactly match the velocities you set, even if they include a component along a constrained direction.

### DIFF
--- a/platforms/cpu/src/CpuKernels.cpp
+++ b/platforms/cpu/src/CpuKernels.cpp
@@ -113,7 +113,7 @@ static double computeShiftedKineticEnergy(ContextImpl& context, vector<double>& 
     int numParticles = context.getSystem().getNumParticles();
     
     // Compute the shifted velocities.
-    
+
     vector<Vec3> shiftedVel(numParticles);
     for (int i = 0; i < numParticles; ++i) {
         if (masses[i] > 0)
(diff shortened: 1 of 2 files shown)


jchodera:
> LangevinMiddleIntegrator does not apply an offset

Is this because you found the non-offset estimator was a better predictor of temperature? There are ramifications in computing things like heat and work, but presumably we are not supporting those applications, instead suggesting folks use CustomIntegrator. Just wanted to make sure this was documented somewhere.

peastman:
Correct.  It was a tradeoff.  Applying the offset would be consistent with LangevinIntegrator, and would mean that kinetic+potential would equal the total energy at a particular point in time.  Not applying it gives a much better representation of the thermal distribution.  Since the main use of kinetic energy in constant temperature simulations is to monitor the temperature, we decided the latter was more important.

PR #3594, opened 2022-05-04
Skip more URLs that linkinator reports as broken

peastman:
The docs build recently started failing again.  For some reason linkinator now reports more URLs as broken, even though it didn't before and they still work.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -650,7 +650,7 @@ jobs:
         shell: bash -l {0}
         env:
           # Pipe-seperated list of domains to skip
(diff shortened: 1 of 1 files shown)


anand-bhat:
Would using the  `--retry-errors` option in linkinator be of any use here? It retries links with 50x errors thrice by default, which would hopefully address any network flakiness or concurrency/ rate limits.

peastman:
It isn't flakiness.  There are certain URLs that it consistently reports as being broken even though they're correct.  I [opened an issue](https://github.com/JustinBeckwith/linkinator/issues/386) about it, but never got a reply.

PR #3614, opened 2022-05-24
Added temperature adjustment to MC barostat for simulated tempering

### LINKED ISSUE #3612: Simulated tempering in the isothermal-isobaric ensemble
Hi OpenMM-devs,

I am attempting to run simulated tempering in the isothermal-isobaric ensemble, however upon inspection of the code it appears to me it only works in the canonical ensemble or non-MC volume regulated simulations?

I was wondering, could one run with a Monte Carlo barostat by addition of the following code?
```python
# Select the initial temperature.

self.currentTemperature = 0
self.simulation.integrator.setTemperature(self.temperatures[self.currentTemperature])
for force in self.simulation.system.getForces():
    if force.getName() == 'MonteCarloBarostat':
        force.setDefaultTemperature(self.temperatures[self.currentTemperature])
```
and
```python
# Select this temperature.
self._hasMadeTransition = True
self.currentTemperature = j
self.simulation.integrator.setTemperature(self.temperatures[j])
for force in self.simulation.system.getForces():
    if force.getName() == 'MonteCarloBarostat':
        force.setDefaultTemperature(self.temperatures[self.currentTemperature])
```
Looking forward hearing from you and hoping to get even more hands-on with OpenMM in the future.

SHervoe-Hansen:
The title more or less explains it all:
This pull request allows simulated tempering to be run in isobaric ensembles using Monte Carlo barostats by adjusting the temperature of the barostat.

fixes #3612

### DIFF
--- a/wrappers/python/openmm/app/simulatedtempering.py
+++ b/wrappers/python/openmm/app/simulatedtempering.py
@@ -160,6 +160,9 @@ class SimulatedTempering(object):
         
         self.currentTemperature = 0
         self.simulation.integrator.setTemperature(self.temperatures[self.currentTemperature])
+        for param in self.simulation.context.getParameters():
+            if 'MonteCarloTemperature' in param:
+                self.simulation.context.setParameter(param, self.temperatures[self.currentTemperature])
         
         # Add a reporter to the simulation which will handle the updates and reports.
         
@@ -230,6 +233,9 @@ class SimulatedTempering(object):
                     self._hasMadeTransition = True
                     self.currentTemperature = j
                     self.simulation.integrator.setTemperature(self.temperatures[j])
+                    for param in self.simulation.context.getParameters():
+                        if 'MonteCarloTemperature' in param:
+                            self.simulation.context.setParameter(param, self.temperatures[self.currentTemperature])
                 if self._updateWeights:
                     # Update the weight factors.
                     
--- a/wrappers/python/tests/TestSimulatedTempering.py
+++ b/wrappers/python/tests/TestSimulatedTempering.py
@@ -54,31 +54,37 @@ class TestSimulatedTempering(unittest.TestCase):
             expectedEnergy = (0.5*MOLAR_GAS_CONSTANT_R*t).value_in_unit(kilojoules_per_mole)
             self.assertAlmostEqual(expectedEnergy, meanEnergy, delta=expectedEnergy*0.3)
 
-    def testHarmonicOscillatorNPT(self):
-        system = System()
-        system.addParticle(1.0)
-        system.addParticle(1.0)
-        force = HarmonicBondForce()
-        force.addBond(0, 1, 1.0, 1000.0)
-        system.addForce(force)
-        mcBarostat = MonteCarloBarostat(1*bar, 300*kelvin, 2)
+
+    def testAlanineDipeptideExplicit(self):
+        prmtop = AmberPrmtopFile('systems/alanine-dipeptide-explicit.prmtop')
+        inpcrd = AmberInpcrdFile('systems/alanine-dipeptide-explicit.inpcrd')
+        system = prmtop.createSystem(nonbondedMethod=PME, nonbondedCutoff=1*nanometer, constraints=HBonds)
+        mcbarostat = MonteCarloBarostat(1*bar, 100*kelvin, 2)
         system.addForce(mcBarostat)
-        integrator = LangevinIntegrator(300*kelvin, 10/picosecond, 0.001*picosecond)
-        topology = Topology()
-        chain = topology.addChain()
-        residue = topology.addResidue('H2', chain)
-        topology.addAtom('H1', element.hydrogen, residue)
-        topology.addAtom('H2', element.hydrogen, residue)
+        integrator = LangevinIntegrator(100*kelvin, 1/picosecond, 0.001*picosecond)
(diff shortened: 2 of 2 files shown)


peastman:
Looks good!  How about adding a test case to TestSimulatedTempering.py?  It can be very simple.  Create a minimal system that includes a MonteCarloBarostat, run a small number of steps, and then verify that `integrator.getTemperature()`, `st.temperatures[st.currentTemperature]`, and `context.getParameter('MonteCarloTemperature')` are all equal to each other.

SHervoe-Hansen:
> verify that `integrator.getTemperature()`, `st.temperatures[st.currentTemperature]`, and `context.getParameter('MonteCarloTemperature')` are all equal to each other.

I think that is a great idea, and I have added a simple example. However, I am not sure if the unit test could be made a bit more complete by comparing the results obtained to an analytical probability density of some sort?  Sadly going from _NVT_ to _NPT_ for a_N_  independent harmonic oscillators system becomes a bit nontrivial for me to solve.

SHervoe-Hansen:
Alternatively, would it be better maybe with a small explicit solvent from the existing systems?

peastman:
That looks fine.  This test is just to validate the code you added for setting temperatures.  The other test case already checks distributions.  Running 40,000 steps is way more than required for this purpose.

SHervoe-Hansen:
So I am having some issues adding periodic boundaries to the system.
```python
from openmm import *
from openmm.app import *
from openmm.unit import *

system = System()
system.addParticle(1.0)
system.addParticle(1.0)
force = HarmonicBondForce()
force.addBond(0, 1, 1.0, 1000.0)
system.addForce(force)
mcBarostat = MonteCarloBarostat(1*bar, 300*kelvin, 2)
system.addForce(mcBarostat)
integrator = LangevinIntegrator(300*kelvin, 10/picosecond, 0.001*picosecond)
topology = Topology()
chain = topology.addChain()
residue = topology.addResidue('H2', chain)
topology.addAtom('H1', element.hydrogen, residue)
topology.addAtom('H2', element.hydrogen, residue)
system.usesPeriodicBoundaryConditions()
```

`system.usesPeriodicBoundaryConditions()` returns False. I don't really understand why.

Finally, upon creating the simulation object I get the error:
`OpenMMException: A barostat cannot be used with a non-periodic system`

Could advise me how to enable periodic boundaries before creatiing the simulation object?

SHervoe-Hansen:
If I can use one of the existing systems the following unit test could instead be conducted:
```python
import unittest
from openmm import *
from openmm.app import *
from openmm.unit import *

class TestSimulatedTempering(unittest.TestCase):
    def testAlanineDipeptideExplicit(self):
        prmtop = AmberPrmtopFile('systems/alanine-dipeptide-explicit.prmtop')
        inpcrd = AmberInpcrdFile('systems/alanine-dipeptide-explicit.inpcrd')
        system = prmtop.createSystem(nonbondedMethod=PME, nonbondedCutoff=1*nanometer, constraints=HBonds)
        mcbarostat = MonteCarloBarostat(1*bar, 200*kelvin, 2)
        system.addForce(mcBarostat)
        integrator = LangevinIntegrator(200*kelvin, 1/picosecond, 0.001*picosecond)
        simulation = Simulation(topology, system, integrator, Platform.getPlatformByName('Reference'))
        simulation.context.setPositions(inpcrd.positions)
        simulation.context.setPeriodicBoxVectors(*inpcrd.boxVectors)

        st = SimulatedTempering(simulation, numTemperatures=10, minTemperature=200*kelvin, maxTemperature=400*kelvin, tempChangeInterval=4, reportInterval=10000)
        self.assertEqual(10, len(st.temperatures))
        self.assertEqual(200*kelvin, st.temperatures[0])
        self.assertEqual(400*kelvin, st.temperatures[-1])

        # We let the simulation run and assert at every step T(mcBarostat) == T(integrator) == T(tempering)
        sampled_temperatures = []
        for i in range(100):
            st.step(2)
            self.assertEqual(0, 1)
            self.assertEqual(st.temperatures[st.currentTemperature], integrator.getTemperature())
            self.assertEqual(st.temperatures[st.currentTemperature], simulation.context.getParameter('MonteCarloTemperature')*kelvin)
            sampled_temperatures.append(st.currentTemperature)
        self.assertEqual(set(range(10)) ^ set(sampled_temperatures), set())
```

With a wall time of 40.8 s on my machine.

The last assert is obviously stochastic, but serves to check that the temperature has in fact changed during the simulated tempering and the integrator and barostat temperature as well. However, one could remove this check and just set the initial temperature of `integrator` and `mcbarostat ` to a different value than 200 Kelvin I guess?

Looking forward to hearing your response.

peastman:
A system is considered periodic if any force in the system makes use of periodic boundary conditions.  If you call `force.setUsesPeriodicBoundaryConditions(True)`, it will consider the system to be periodic.

> self.assertEqual(set(range(10)) ^ set(sampled_temperatures), set())

I don't think we can assume that.  It needs time to learn the weight factors.  In the first 200 steps it likely won't have visited all the temperatures yet.

> However, one could remove this check and just set the initial temperature of integrator and mcbarostat to a different value than 200 Kelvin I guess?

That seems reasonable.  Set them initially to a value that isn't in `st.temperatures`, say 100K.  That way we'll know it has successfully changed the temperature.

peastman:
Could you switch the test back to the harmonic oscillator?  A tiny system works just as well for this test, and it will be much faster.

Issue #3633, opened 2022-06-08
MetaDynamics ZeroDivisionError for bias that is exactly equal to 1


CoryKornowicz:
https://github.com/openmm/openmm/blob/dbb4827e6569c0c1b2eb42ad9bd40e17b400a40a/wrappers/python/openmm/app/metadynamics.py#L108

This line dictates the bias must be greater than or equal to one, but when this line is exectuted 

https://github.com/openmm/openmm/blob/dbb4827e6569c0c1b2eb42ad9bd40e17b400a40a/wrappers/python/openmm/app/metadynamics.py#L127

The self._deltaT is set to 0 and that causes 
```
height = self.height*np.exp(-energy/(unit.MOLAR_GAS_CONSTANT_R*self._deltaT))
```
to through a ZeroDivisionError. 

Thoughts on setting line 108 to 
```
 if biasFactor <= 1.0: 
```
to avoid the issue of 0 division?

peastman:
Good catch!  You're exactly right.  A PR would be welcome!

peastman:
Fixed by #3636.

PR #3636, opened 2022-06-09
ZeroDivisionError Fix to Metadynamics Bias

CoryKornowicz:
Proposed Solution to Issue #3633


wrappers/python/openmm/app/metadynamics.py
```diff
@@ -105,7 +105,7 @@ def __init__(self, system, variables, temperature, biasFactor, height, frequency
             temperature = temperature*unit.kelvin
         if not unit.is_quantity(height):
             height = height*unit.kilojoules_per_mole
-        if biasFactor < 1.0:
+        if biasFactor <= 1.0:
             raise ValueError('biasFactor must be >= 1')
```
peastman, line 109: Can you also correct the error message?  "must be > 1" rather than >=.

PR #3637, opened 2022-06-10
Prevent Windows from defining macros that interfere with other code

mikemhenry:


peastman:
It's best to put it at the very top of the file, before the first `#include`.  In order to work, the symbol has to be defined before anything includes the header with the unwanted macros.

mikemhenry:
Looks like the failing test on windows fails occasionally, but I don't have the permissions to re-run it.

peastman:
It's just a stochastic failure.  You can ignore it.

peastman:
Thanks!

PR #3649, opened 2022-06-20
Reduced the cost of updating tabulated functions

peastman:
See https://github.com/openmm/openmm/pull/3307#issuecomment-1156478973.  Updating a discrete function should now be very fast.  I think there's still room to improve the speed of updating continuous functions.  I'll see what I can do about that in a future PR.

@tristanic can you try this out and see how it works for you?

### DIFF
--- a/openmmapi/include/openmm/TabulatedFunction.h
+++ b/openmmapi/include/openmm/TabulatedFunction.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2014 Stanford University and the Authors.           *
+ * Portions copyright (c) 2014-2022 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -57,6 +57,8 @@ namespace OpenMM {
 
 class OPENMM_EXPORT TabulatedFunction {
 public:
+    TabulatedFunction() : updateCount(0) {
+    }
     virtual ~TabulatedFunction() {
     }
     /**
@@ -67,12 +69,18 @@ public:
      * Get the periodicity status of the tabulated function.
      */
     bool getPeriodic() const;
+    /**
+     * Get the value of a counter that is updated every time setFunctionParameters()
+     * is called.  This provides a fast way to detect when a function has changed.
+     */
+    int getUpdateCount() const;
     virtual bool operator==(const TabulatedFunction& other) const = 0;
     virtual bool operator!=(const TabulatedFunction& other) const {
         return !(*this == other);
     }
 protected:
     bool periodic;
+    int updateCount;
 };
 
 /**
--- a/openmmapi/src/TabulatedFunction.cpp
+++ b/openmmapi/src/TabulatedFunction.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2014-2021 Stanford University and the Authors.      *
+ * Portions copyright (c) 2014-2022 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -39,6 +39,10 @@ bool TabulatedFunction::getPeriodic() const {
     return periodic;
 }
 
+int TabulatedFunction::getUpdateCount() const {
+    return updateCount;
+}
+
 Continuous1DFunction::Continuous1DFunction(const vector<double>& values, double min, double max, bool periodic) {
     this->periodic = periodic;
     setFunctionParameters(values, min, max);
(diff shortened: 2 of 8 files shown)


tristanic:
I'm at home getting over a cold today. Tried to build on my home WIndows PC (using Visual Studio 2015, following the instructions at http://docs.openmm.org/latest/userguide/library/02_compiling.html#id9), but ran into what looks like a completely unrelated error (actually, lots of the same, scattered throughout the `asmjit` library). Not sure what's going on there (I think the compiler might be refusing to inline the comparison operators despite their declaration as `__forceinline`)... will try again once I'm back at my Linux machine.

```
19>c:\users\tristan\gits\openmm-tabulated\libraries\asmjit\asmjit\core/archtraits.h(238): error C2593: 'operator <=' is ambiguous
19>  c:\users\tristan\gits\openmm-tabulated\libraries\asmjit\asmjit\core\../core/operand.h(216): note: could be 'bool asmjit::_abi_1_9::operator <=(asmjit::_abi_1_9::RegGroup,asmjit::_abi_1_9::RegGroup) noexcept' [found using argument-dependent lookup]
19>  c:\users\tristan\gits\openmm-tabulated\libraries\asmjit\asmjit\core\../core/operand.h(162): note: or       'bool asmjit::_abi_1_9::operator <=(asmjit::_abi_1_9::RegType,asmjit::_abi_1_9::RegType) noexcept' [found using argument-dependent lookup]
19>  c:\users\tristan\gits\openmm-tabulated\libraries\asmjit\asmjit\core\../core/type.h(155): note: or       'bool asmjit::_abi_1_9::operator <=(asmjit::_abi_1_9::TypeId,asmjit::_abi_1_9::TypeId) noexcept' [found using argument-dependent lookup]
19>  c:\users\tristan\gits\openmm-tabulated\libraries\asmjit\asmjit\core/archtraits.h(238): note: while trying to match the argument list '(asmjit::_abi_1_9::RegType, asmjit::_abi_1_9::RegType)'
```

peastman:
Can you try a newer compiler?  The CI builds use VS 2022.  The [build instructions for asmjit](https://asmjit.com/doc/group__asmjit__build.html) say they officially support VS 2017 and later, though 2015 has been reported to work.  But maybe it doesn't really and we need to update the minimum version in the documentation.

tristanic:
OK, yep - builds fine in VS 2022. Will try to give it a test drive later this evening.

tristanic:
Just gave it a try. Very nice! 

```
len(params[3])
Out[26]: 896896

%prun f.updateParametersInContext(c)
          5 function calls in 0.008 seconds

   Ordered by: internal time

        1    0.006    0.006    0.006    0.006 {built-in method openmm._openmm.CustomCompoundBondForce_updateParametersInContext}
        1    0.002    0.002    0.008    0.008 openmm.py:6475(updateParametersInContext)
        1    0.000    0.000    0.008    0.008 {built-in method builtins.exec}
        1    0.000    0.000    0.008    0.008 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

As you expected, the `Continuous3DFunction` still takes a lot longer (1.7 seconds for the same case) - but I can switch to permanently using the `Discrete3DFunction` MDFF implementation with very little downside. Simulation speed is functionally identical (actually seems ever-so-slightly faster!), and in any case beyond about 800-900 time steps/second the graphical rendering becomes rate-limiting anyway.

peastman:
Great!  I'll merge it.

I've been investigating ways to speed up continuous functions.  The bottleneck for them is fitting the splines.  In the case of 3D functions, most of the time is spent in these nested loops:

https://github.com/openmm/openmm/blob/d7da750ad3d2c8064ffb28417402f44e1cdb8bd7/openmmapi/src/SplineFitter.cpp#L583-L623

If we parallelize the outer loop, that should make it a few times faster.

tristanic:
If you're going to be revisiting that code anyway, it might be worth also looking into ways to sensibly change how the coefficients are laid out in GPU memory. Currently they all go into a single array - the reason I did the `Discrete3DFunction` MDFF implementation in the first place is that the `Continuous3DFunction` hits the maximum allowed array size for large (but not _ridiculously_ large) maps. One array for each coefficient, perhaps?

tristanic:
Hmm… it’s been a long while since I last looked at that, but I distinctly remember the error message being about the maximum array size rather than out-of-memory. But the gist of it is that with 64 coefficients per voxel, a 2^31 array allows for ~32M voxels - roughly a 30x30x30 nm map at 3 Å resolution with a dmin/3 grid spacing (about the size of a ribosome).

On 21 Jun 2022, at 22:14, Peter Eastman ***@***.***> wrote:

﻿

What limit on array size do you encounter? It ought to be ok with arrays of at least 2^31 elements, maybe more. Since each element is a float, that would be 8 GB just to store that one array.

peastman:
It's possible this has improved since you last tested it.  When allocating an array, we used to represent the size as an `int`.  In #3428 we changed it to a `size_t`.  That should remove limits on allocating large arrays.

In this case, though, you'll still be limited because when we look up elements in the array within the kernel, the index is represented as an `int`.  That's a limit on the number of elements rather than bytes, though, so it's four times higher.

peastman:
Yes, fitting 3D splines (and 2D ones to a lesser extent) could probably be done pretty efficiently on a GPU.  A potential issue is that we'd prefer to do it in double precision so we don't lose a lot of accuracy.  Low end GPUs often don't support double precision, and even high end ones targeted at consumers tend to have terrible double precision performance.  I'm not sure how much precision we'd lose by doing the fit in single precision, but it could be significant.

PR #3654, opened 2022-06-22
OPC and OPC3 water

aizvorski:


aizvorski:
@peastman I'm working on a unit test right now; I was thinking the density of an equilibrated box would be a pretty good test.  I don't have a reference for the equilibrium geometry for a pair of molecules.

Is there a specific way to compare energy amber <-> openmm?

Also, how were the equilibrated water boxes eg openmm/app/data/tip3p.pdb made?  Would OPC and OPC3 need their own?

peastman:
I suggest keeping it to just an energy comparison for a minimal system.  Simulating a water box long enough to converge the density will take a while, and we want unit tests to be as fast as possible (preferably much less than a second for each one).  And the result will be stochastic, so it might occasionally fail.  Computing the energy of two molecules is fast and deterministic.

> Is there a specific way to compare energy amber <-> openmm?

You can just compute the energy with sander and hardcode the result in the test case (with a comment saying where the value came from).

aizvorski:
@peastman I added an OPC water box and set up a unit test, the energies do match pretty closely.

First, using ambertools-22.0-py38h6177452_1 installed from conda-forge, in tleap

```
source leaprc.water.opc
loadoff opcbox.off
saveamberparm OPCBOX opcbox.prmtop opcbox.inpcrd
savepdb OPCBOX opcbox.pdb
```

Then, get energy in sander (not super familiar with running sander so please let me know if there is a better way to just get energy without minimizing or running dynamics):

mdin
```
Test
 &cntrl
  imin=1,
  maxcyc=0,
  cut=7.0,
 /
```

sander command and results
```
sander -i mdin -p opcbox.prmtop -c opcbox.inpcrd
...
   NSTEP       ENERGY          RMS            GMAX         NAME    NUMBER
      1      -2.6476E+03     2.0113E+01     9.0801E+01     O         609

 BOND    =        0.0587  ANGLE   =        0.0000  DIHED      =        0.0000
 VDWAALS =      362.2542  EEL     =    -3009.9362  HBOND      =        0.0000
 1-4 VDW =        0.0000  1-4 EEL =        0.0000  RESTRAINT  =        0.0000
```

Energy calculated in openmm using opcbox.pdb and opc.xml = `-2647.2222697324237 kcal/mol`, after applyConstraints(1e-12) `-2647.441600693312 kcal/mol`

aizvorski:
@peastman I serialized a system created from the opcbox.prmtop and opcbox.inpcrd, and kept modifying the opc.xml file until a system serialized from opcbox.pdb and opc.xml matched as closely as possible.  However - that requires values which are slightly different from the ones described in ambertools

For example:
```
ambertools-22.0-py38h6177452_1/dat/leap/lib/opcbox.off

 "O" "OW" 0 1 131072 1 8 0.0
 "H1" "HW" 0 1 131072 2 1 0.679142
 "H2" "HW" 0 1 131072 3 1 0.679142
 "EPW" "EP" 0 1 131072 4 0 -1.358284

opcbox.xml (created from opcbox.prmtop)

                                <Particle eps=".8903586017365637" q="0" sig=".316655208182147"/>
                                <Particle eps="0" q=".679142001832919" sig=".08908987181403394"/>
                                <Particle eps="0" q=".679142001832919" sig=".08908987181403394"/>
                                <Particle eps="0" q="-1.3582839981780566" sig=".08908987181403394"/>
```

Is a charge difference of 1.8e-09 something to worry about?  Where does this come from?  Which value should we use?

(I think there may be similar very slight differences in geometry - bond lengths are exact, but angles formed by the OW-HW and HW-HW triangle are not exactly the same as the HW-OW-HW angle of 103.6 degrees)

jchodera:
OpenMM uses a different set of physical constants than Amber's `sander`: OpenMM uses [CODATA 2018](https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/platforms/reference/include/SimTKOpenMMRealType.h#L63-L84), while Amber `sander` uses some historical choices [the square root of `1/(4*pi*epsilon0)` in kcal/mol/A used by `sander` is given [here](https://github.com/ParmEd/ParmEd/blob/1be8ca0f2f5b232db79384cb02454a9c53f37749/parmed/constants.py#L14). 

As a result, I'd expect to see some differences on the order of 3e-5 in energies and forces.

aizvorski:
> OpenMM uses a different set of physical constants than Amber's `sander`: 

@jchodera Thanks!  Which do you think is better: 

(a) have opc.xml use the same numerical values as leap/lib/opcbox.off and leap/parm/frcmod.opc, but have a system created from tleap -> prmtop -> openmm be different from a system created from pdb -> openmm, or 
(b) have opc.xml use different numbers, but have a system created tleap -> prmtop -> openmm be identical to pdb -> openmm?

I'm still not sure where the difference in numerical values comes from now, even though the unit definitions are different.  Do we essentially convert units when reading prmtop files now?  Ie, does reading a prmtop cause 1 prmtop-e charge to be converted to 1.0001 opemm-e charge or something similar?

jchodera:
> Which do you think is better:

@peastman may have a more succinct statement of our philosophy on converting parameters developed with other packages, but I believe out approach to date has been to convert the parameters directly, even if the difference in physical constants causes a difference in potential energies between packages. More modern approaches for converting parameterized systems like [openff-interchange](https://github.com/openforcefield/openff-interchange) from @mattwthompson may have a different philosophy here.

> does reading a prmtop cause 1 prmtop-e charge to be converted to 1.000001 opemm-e charges or something similar?

It's important to remember that Amber `prmtop` files store the _square root_ of the atomic charge (in units of `|e|^{1/2}`). If you go from `off` -> `prmtop` -> `xml` -> OpenMM, it will compute the square root, print it to a `prmtop` file truncating the decimal places (fortran`%FORMAT(5E16.8)`), read it again, and then square it. This will certainly introduce differences in the exact numerical values.

mattwthompson:
What currently happens in in OpenMM's infrastructure when Amber infrastructure reports a value using its unit definitions? I would be floored if this issue hasn't been run into years ago; as far as I can tell this isn't the first set of Amber parameters that are being ported into OpenMM code.

Let me add that it's a difficult problem and I wouldn't at all blame people for sweeping under the rug in the past.

peastman:
Different values for fundamental constants is a problem through the whole community.  I believe CHARMM uses yet another set of values which are different from either OpenMM or Amber.  Still, when people convert force fields they generally use the exact parameter values.  This means you shouldn't expect energies from different codes to match to more than about five significant digits.  In practice that doesn't generally matter, since other uncertainties are much larger.

aizvorski:
@peastman Ready for review! 

- [x] Both OPC and OPC3 are done
- [x] Exact parameter values for bond lengths and charges (from frcmod.opc, opcbox.off and same for opc3)
- [x] Derivation for sigma, epsilon, angle and virtual site weights is shown in comments in the xml files
- [x] Added opcbox.pdb and opc3box.pdb
- [x] Added unit tests for energies, the values match those calculated with sander to 0.02% or better
- [x] Updated docs

wrappers/python/tests/TestOpcWater.py
```diff
@@ -0,0 +1,79 @@
...
+import unittest
+import openmm as mm
+from openmm import app
+from openmm import unit
+
+
+class TestOpcWater(unittest.TestCase):
+
+    def test_OpcEnergy(self):
+        pdb = app.PDBFile('../openmm/app/data/opcbox.pdb')
```
peastman, line 11: Data files used in test cases should go in `wrappers/python/tests/systems`.  Inside the test case, you can refer to it as `systems/opcbox.pdb`.

wrappers/python/tests/TestOpcWater.py
```diff
@@ -0,0 +1,79 @@
+import sys
+import unittest
+import openmm as mm
+from openmm import app
+from openmm import unit
+
+
+class TestOpcWater(unittest.TestCase):
```
peastman, line 8: Rather than creating a separate class for testing these two water models, let's put the tests in TestForceField.py.  It already contains tests for specific force field XML files.

wrappers/python/tests/TestOpcWater.py
```diff
@@ -0,0 +1,79 @@
...
+        system = forcefield.createSystem(
+            topology,
+            nonbondedMethod=app.PME,
+            nonbondedCutoff=0.7*unit.nanometer,
+            constraints=app.HBonds,
+            rigidWater=True,
+        )
+
+        box_vectors = pdb.topology.getPeriodicBoxVectors()
+        system.setDefaultPeriodicBoxVectors(*box_vectors)
```
peastman, line 24: You don't need to do this.  The default box vectors automatically get set from the topology.

peastman:
Aside from the minor comments above about the test case, this looks great.

wrappers/python/tests/TestOpcWater.py
aizvorski: @peastman These are equilibrated boxes that may be useful when adding water in modeller, not just as a test case; analogous to openmm/app/data/tip3p.pdb and tip4pew.pdb etc.  I wasn't sure how to use stuff outside of tests/ in a test, but it seems like a relative path would work?

wrappers/python/tests/TestOpcWater.py
aizvorski: @peastman Okay, I moved them over

wrappers/python/tests/TestOpcWater.py
aizvorski: removed (and tested, it works)

wrappers/python/tests/TestOpcWater.py
peastman: We can always move them in the future if we decide to use them for something else.  Since the test case is the only thing they're currently used for, let's keep them with the test.

aizvorski: @peastman Okay, done

PR #3696, opened 2022-07-12
Avoid clashes when adding solvent to non-rectangular boxes

### LINKED ISSUE #3692: Water clashes with protein for octahedral boxes not centred in unit cell
Hi, 

If I add an octahedral solvent box to the attached protein ([input_protein.pdb.txt](https://github.com/openmm/openmm/files/9082148/input_protein.pdb.txt)), I end up with a large number of water molecules clashing with my protein. We are using the solvated structure in FEP calculations and this is causing us to obtain poor results using our normal simulation protocol. However, if I first move the protein so that it will be centred in the unit cell, I do not get any water clashes. I also do not see this problem with orthogonal boxes. Please could you investigate? 

The following is the code I used to generate the water box (using the latest OpenMM from conda-forge) and some code that deletes waters that are too close. I find 2159 waters are too close to the protein when I run this script. 

```
import math
import openmm
import openmm.app as app
from openmm.app import element as elem
import openmm.unit as units

pdb_file = app.PDBFile('input_protein.pdb')
force_field = app.ForceField('amber/protein.ff14SB.xml', 'amber/tip3p_standard.xml')
modeller = app.Modeller(pdb_file.topology, pdb_file.positions)
padding_nm = 1.0*units.nanometers
max_solute_size = max(max((pos[i] for pos in modeller.positions)) - min((pos[i] for pos in modeller.positions)) for i in range(3))
octahedral_vectors = openmm.Vec3(1,0,0), openmm.Vec3(1/3,2*math.sqrt(2)/3,0), openmm.Vec3(-1/3,math.sqrt(2)/3,math.sqrt(6)/3)
box_vectors = [(max_solute_size + 2.0*padding_nm)*vec for vec in octahedral_vectors]
modeller.addSolvent(force_field, model='tip3p', boxVectors=box_vectors)

# Delete waters that are too close to the solute
count = 0
vdw_radius_per_sigma = 0.56
water_o_sigma = 0.32
water_radius = water_o_sigma*vdw_radius_per_sigma
temp_system = force_field.createSystem(modeller.topology)
nonbonded = None
for i in range(temp_system.getNumForces()):
    if isinstance(temp_system.getForce(i), openmm.NonbondedForce):
        nonbonded = temp_system.getForce(i)
cutoff = [water_radius]*temp_system.getNumParticles()
for i in range(temp_system.getNumParticles()):
    params = nonbonded.getParticleParameters(i)
    if params[2] != 0*units.kilojoules_per_mole:
        cutoff[i] += params[1].value_in_unit(units.nanometer)*vdw_radius_per_sigma
if nonbonded is not None:
    waters_to_delete = []
    for i, res in enumerate(modeller.topology.residues()):
        if res.name == 'HOH':
            water_o_atom_pos = None
            for water_atom in res.atoms():
                if water_atom.element and water_atom.element.atomic_number == 8:
                    water_o_atom_pos = modeller.positions[water_atom.index]
            if water_o_atom_pos is None:
                continue
            for j, r in enumerate(modeller.topology.residues()):
                if r.name != 'HOH':
                    for at in r.atoms():
                        other_atom_pos = modeller.positions[at.index]
                        distance = math.sqrt((water_o_atom_pos.x - other_atom_pos.x)**2 + (water_o_atom_pos.y - other_atom_pos.y)**2 + (water_o_atom_pos.z - other_atom_pos.z)**2)
                        if distance < cutoff[at.index]:
                            count += 1
                            print(f'deleting water {count}')
                            waters_to_delete.append(res)
                            break
                    if len(waters_to_delete) and waters_to_delete[-1] == res:
                        break
    modeller.delete(waters_to_delete)
```

Thanks, 

Rosie

peastman:
Fixes #3692.  @rosemary-cresset can you try this out and see if it fixes the problem for you?

### DIFF
--- a/wrappers/python/openmm/app/modeller.py
+++ b/wrappers/python/openmm/app/modeller.py
@@ -602,7 +602,7 @@ class Modeller(object):
             filteredWaters = []
             cells.cells = {}
             for i in range(len(lowerSkinPositions)):
-                cell = tuple((int(floor(lowerSkinPositions[i][j]/cells.cellSize[j]))%cells.numCells[j] for j in range(3)))
+                cell = cells.cellForPosition(lowerSkinPositions[i])
                 if cell in cells.cells:
                     cells.cells[cell].append(i)
                 else:
@@ -1587,6 +1587,8 @@ class _CellList(object):
         self.cells = {}
         self.numCells = tuple((max(1, int(floor(vectors[i][i]/maxCutoff))) for i in range(3)))
         self.cellSize = tuple((vectors[i][i]/self.numCells[i] for i in range(3)))
+        self.vectors = vectors
+        self.periodic = periodic
         invBox = Vec3(1.0/vectors[0][0], 1.0/vectors[1][1], 1.0/vectors[2][2])
         for i in range(len(self.positions)):
             pos = self.positions[i]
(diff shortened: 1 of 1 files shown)


rosemary-cresset:
This seems to fix the problem - thanks very much for the quick fix.

peastman:
Great, thanks for creating such a good test case!

zhang-ivy:
@rosemary-cresset : Just curious, when you re-ran [your code](https://github.com/openmm/openmm/issues/3692#issue-1300384167) with this PR, did you see that `waters_to_delete` went from 2159 to 0? 

I ran your code on one of my solvated PDBs without and with this PR, and I'm seeing that `waters_to_delete` goes from 640 to 300, so I think there are still clashy waters?

peastman:
When I ran it the number went down to about 17.  But I think those were just from the script using a slightly different standard than what Modeller uses.  For example, several of them involved Na ions.  The way it adds ions is to first tile a pre-equilibrated water box, then select a random set of waters to replace with ions.  Since Na has a slightly larger radius than a water oxygen, it sees them as clashes.  But really they're fine and energy minimization has no trouble dealing with them.

rosemary-cresset:
Hi @zhang-ivy. Thanks for testing. I got roughly the behaviour that @peastman is describing and sensible results from subsequent simulations.

zhang-ivy:
Ah gotcha, thanks. I was trying to evaluate whether the same problem was happening for my system and whether this PR solves the problem, but I am seeing that `waters_to_delete` is still 300 after using this PR, and I'm not sure what to make of that. I wonder if @rosemary-cresset 's code is more stringent than necessary for my purposes, but I don't fully understand every aspect of it (e.g. I don't know why  how `vdw_radius_per_sigma` was chosen to be 0.56), so I think I'm just going to proceed with using a cubic box for my system, as @peastman mentions [here](https://github.com/openmm/openmm/issues/3692#issuecomment-1181126926) that there shouldn't be a problem with how clashes are handled for the cubic box. 

Thanks for the issue and PR!

zhang-ivy:
My simulations (replica exchange free energy calcs) run fine -- no nans. I have been seeing water sampling problems in my simulations, so I was just worried that starting with waters in potentially clashy positions would make the sampling problems unnecessarily more difficult.

PR #3700, opened 2022-07-13
Kernel source headers included in installation

craabreu:
Installs header files `[Common|Cuda|OpenCL]KernelSources.h`, generated during compilation, in the corresponding include directories.


peastman:
Thanks!

PR #3706, opened 2022-07-19
Fix #3705: Force reordering of atoms after setPositions() for efficiency

### LINKED ISSUE #3705: Atom reordering does not correctly occur on CUDA/OpenCL platforms after setPositions() is called
We have discovered that calling `setPositions()` calls [`reorderAtoms()`](https://github.com/openmm/openmm/blob/109f6b2535da4e0c0dd88007d6ca06b4add2ce81/platforms/cuda/src/CudaKernels.cpp#L221) to optimally reorder atoms for efficiency, but this does not actually cause atoms to be reordered because of a check for [`stepsSinceReorder < 250`](https://github.com/openmm/openmm/blob/3d62421b62494d3d3143feedf0d580c74501f92f/platforms/common/src/ComputeContext.cpp#L435). 

This seems like unintended behavior, and that adding `cu.setStepsSinceReorder(99999)` right before [cu.reorderAtoms()](https://github.com/openmm/openmm/blob/109f6b2535da4e0c0dd88007d6ca06b4add2ce81/platforms/cuda/src/CudaKernels.cpp#L221) would ensure the atoms are correctly reordered after `setPositions()` is called.

cc https://github.com/choderalab/perses/issues/613#issuecomment-1189514739

I can open a PR to fix this.

jchodera:
This fixes #3705, as verified in https://github.com/choderalab/perses/issues/613#issuecomment-1189514739 
image

**Edit:** I'm not sure why the diff is showing blank lines with empty spaces are being replaced by blank lines here.
I'm not sure why the source files had spaces on blank lines in the first place---my editor (Atom) must have replaced those with true blank lines.

I'm not quite sure how to reconfigure my editor to leave those spaces on blank lines untouched.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -128,12 +128,12 @@ void CudaUpdateStateDataKernel::getPositions(ContextImpl& context, vector<Vec3>&
         float4* posq = (float4*) cu.getPinnedBuffer();
         cu.getPosq().download(posq);
     }
-    
+
     // Filling in the output array is done in parallel for speed.
-    
+
     cu.getPlatformData().threads.execute([&] (ThreadPool& threads, int threadIndex) {
         // Compute the position of each particle to return to the user.  This is done in parallel for speed.
-        
+
         const vector<int>& order = cu.getAtomIndex();
         int numParticles = cu.getNumAtoms();
         Vec3 boxVectors[3];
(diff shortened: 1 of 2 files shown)


jchodera:
Closing this, since @peastman indicates his desire to take a different heuristic approach in https://github.com/openmm/openmm/issues/3705#issuecomment-1189693787 to avoid degrading performance in cases where `setPositions()` is called repeatedly or only with energy/force computations (rather than dynamics steps) in between.

PR #3711, opened 2022-07-26
CustomCVForce avoids breaking up molecules

peastman:
This might or might not be the fix for https://github.com/openmm/openmm-ml/issues/36.  Each Force object reports what atoms it creates bonds between, so all the atoms in a molecule are translated together when applying periodic boundary conditions.  CustomCVForce did not propagate the bond specifications from the Forces it contained.

This only applied to output when calling `getState()` with `enforcePeriodicBox=true`.  There's a different mechanism used to identify molecules when internally translating positions to preserve accuracy.  That mechanism was handled correctly, so it never broke up molecules during normal integration.  It only happened when returning output from `getState()`.

cc @dominicrufa

### DIFF
--- a/openmmapi/include/openmm/internal/CustomCVForceImpl.h
+++ b/openmmapi/include/openmm/internal/CustomCVForceImpl.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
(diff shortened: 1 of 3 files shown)


peastman:
I'll go ahead and merge this so it will get into tonight's dev build and be easier to test.

jchodera:
Thanks, @peastman!

@dominicrufa: The nightly builds had already started before this was merged, so I've started a new `omnia-dev` build run so nightlies with this branch merged will be available in the morning.

PR #3737, opened 2022-08-12
Improved support for devices without 64 bit atomics

peastman:
This uses a trick suggested by @dscerutti.  Low end embedded GPUs often don't support 64 bit atomic operations.  As long as you don't care about the return value, it's possible to reproduce a 64 bit atomic add with a sequence of 32 bit ones.  That's slower than a native 64 bit atomic, but in many cases it's still faster than the workarounds we were using before.

This has two big advantages.  First, it allows major code simplifications.  There are a lot of places where we previously had two different code paths for devices with and without 64 bit atomics.  This change allows those duplicate code paths to be eliminated.

Second, there were some features we just weren't able to support on GPUs without 64 bit atomics.  With this change, all forces can be used on all GPUs, even embedded ones.


jchodera:
Wow!  That's pretty fantastic: 
image

peastman:
My favorite kind of feature is one that eliminates a lot of code. :)

philipturner:
I've been trying to imagine how to perform 64-bit atomics on Apple M1 (based on a mobile architecture, A14 bionic). The Metal Shading Language specification technically supports 64-bit atomics, but only for minimum and maximum operations for `unsigned long`. Metal 3.0 introduced 32-bit floating-point atomics, supporting addition, subtraction, compare-exchange, the works. 

A while back with Metal 2.3, I used compare-exchange [to perform 32-bit floating point atomics](https://github.com/philipturner/ARHeadsetKit/blob/174385004728e8c9d047904f0497d9d7793aa764/Sources/ARHeadsetKit/Other/Metal%20Utilities/Private/ColorUtilities.metal#L65-L106). I was accumulating sparse 3D data about color in an AR app. I don't think it's possible to perform 64-bit operations this way for emulated FP64. How problematic will that be, if 64-bit atomics only work for integers?

peastman:
We only require integers.  We do force accumulation in 64 bit fixed point.  In fact, addition is the only atomic operation we use.

Issue #3752, opened 2022-08-23
Add support for other water models for Modeller.addSolvent


kntkb:
I would like to request a feature enhancement for `Modeller.addSolvent`. According to the [document](http://docs.openmm.org/latest/api-python/generated/openmm.app.modeller.Modeller.html), the supported water models are currently tip3p, spce, tip4pew, and tip5p. Can we add support for other water models such as OPC models?

peastman:
All three point water models are pretty similar.  If you tell it to build a TIP3P water box, then build a System with OPC and call `minimizeEnergy()`, I think it ought to work fine.  Can you try that?

kntkb:
@peastman  Thank you! I was able to solve the problem!

PR #3755, opened 2022-08-24
update modeller.addSolvent docstring

kntkb:
Regarding the discussion [Add support for other water models for Modeller.addSolvent #3752](https://github.com/openmm/openmm/issues/3752), I thought it would be helpful to update the `Modeller.addSovlent` docstring.

### DIFF
--- a/wrappers/python/openmm/app/modeller.py
+++ b/wrappers/python/openmm/app/modeller.py
@@ -404,14 +404,19 @@ class Modeller(object):
         between periodic copies to be achieved with a smaller box.  The most compact option is a rhombic dodecahedron,
         for which the box volume is 70.7% the volume of a cubic box with the same amount of padding.
 
+        There exist many different water models, many of which are very similar to each other. This method creates 
+        preequilibrated water boxes for a limited set of water models. In most cases, they work equally well for other models 
+        that involve the same number of particles. For example, to simulate a box of TIP3P-FB water, use this method to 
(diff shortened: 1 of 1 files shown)


peastman:
I think we can make the description a little clearer.  How about instead adding this paragraph just above the Parameters section?

> There exist many different water models, many of which are very similar to each other.  This method creates preequilibrated water boxes for a limited set of water models.  In most cases, they work equally well for other models that involve the some number of particles.  For example, to simulate a box of TIP3P-FB water, use this method to create a box of TIP3P water, construct a System using TIP3P-FB parameters, and perform a local energy minimization to correct for the small differences between the models.  Likewise, a box of TIP4P-Ew water can be used for most four site water models.

kntkb:
> I think we can make the description a little clearer. How about instead adding this paragraph just above the Parameters section?
> 
> > There exist many different water models, many of which are very similar to each other.  This method creates preequilibrated water boxes for a limited set of water models.  In most cases, they work equally well for other models that involve the some number of particles.  For example, to simulate a box of TIP3P-FB water, use this method to create a box of TIP3P water, construct a System using TIP3P-FB parameters, and perform a local energy minimization to correct for the small differences between the models.  Likewise, a box of TIP4P-Ew water can be used for most four site water models.

Yes, I think this is better. We just need to fix the typo. `involve the some number of particles` -> `involve the same number of particles`

peastman:
Thanks for catching that.  Can you update the PR?

peastman:
Thanks!

Issue #3034, opened 2021-02-25
Metadynamics simulations thread Lock CPU on first step


gavinbas:
Hi All,

We've had good success running metadynamics simulations starting from parmed objects, but have noticed that some percentage of the runs (<10%) of the simulations get stalled on the first step. We were able isolate the incident and include files to reproduce the error across platforms. The simulation reaches the first step, GPU utilization bursts and then falls flat while CPU utilization goes to 100% and hangs. The initial bias file is written. We're running OpenMM Version : 7.4.2.dev-dc9d188

[segwalker_stall.zip](https://github.com/openmm/openmm/files/6044920/segwalker_stall.zip)

There are two ways to reproduce the issue from attached files, if you have ommtk installed (pip install ommtk) you can run the simpler reproduce_ommtk.py but I've also reproduced the bare minimum from openMM calls in reproduce_openmm.py. Below is a gdb stack trace output of where the job hangs.
              
              Thread 1 "python" received signal SIGINT, Interrupt.
              0x00007fffc6bb4e22 in JAMA::Eigenvalue<double>::hqr2() ()
                 from /opt/conda/lib/python3.7/site-packages/simtk/openmm/../../../../libOpenMM.so
              (gdb) bt
              #0  0x00007fffc6bb4e22 in JAMA::Eigenvalue<double>::hqr2() ()
                 from /opt/conda/lib/python3.7/site-packages/simtk/openmm/../../../../libOpenMM.so
              #1  0x00007fffc6bb1356 in JAMA::Eigenvalue<double>::Eigenvalue(TNT::Array2D<double> const&) () from /opt/conda/lib/python3.7/site-packages/simtk/openmm/../../../../libOpenMM.so
              #2  0x00007fffc561210d in double OpenMM::CudaCalcRMSDForceKernel::executeImpl<float>(OpenMM::ContextImpl&) () from /opt/conda/lib/plugins/libOpenMMCUDA.so
              #3  0x00007fffc6aa3f54 in OpenMM::ContextImpl::calcForcesAndEnergy(bool, bool, int) ()
                 from /opt/conda/lib/python3.7/site-packages/simtk/openmm/../../../../libOpenMM.so
              #4  0x00007fffc55f67bf in OpenMM::CudaCalcCustomCVForceKernel::execute(OpenMM::ContextImpl&, OpenMM::ContextImpl&, bool, bool) () from /opt/conda/lib/plugins/libOpenMMCUDA.so
[...]

PR #3758, opened 2022-08-30
Detect NaN to avoid infinite loop in JAMA::Eigenvalue

dwilliams-nobias:
JAMA::Eigenvalue can run in an infinite loop if there are upstream NaN in atom positions. Throwing an exception is probably preferable. This checks for NaN directly before the JAMA call.

See: #3034


peastman:
Thanks!

Issue #3663, opened 2022-06-26
Add all Amber ion params


aizvorski:
Amber has quite a lot of different kinds of ion parameters in `dat/leap/parm/frcmod.ions*`.  There are sets for each water model, which can be either the Li/Merz "lm" or Joung/Cheatham "jc" sets, and also "hfe", "iod" or standard, and also 12-6 or 12-6-4  (Amber 22 manual, section "3.6. Ions").  To some extent these can be mixed and matched with water models, for example OPC can be used with either the lm 12-6 opc standard set or the jc tip4pew set (manual , section 3.5.1).

OpenMM so far takes a different approach where a water model xml file seems to have builtin ions (no mix and match), and also they are tied to the *protein* forcefield, not the water model.  So for example tip3p.xml doesn't have ions, but amber14/tip3p.xml has one specific set of ions.

I'd like to convert all Amber ion frcmod files into OpenMM xml files with the same names and the same contents (to avoid any naming confusion).  So for example ambertools/dat/leap/parm/frcmod.ionsjc_tip4pew -> openmm/app/data/ionsjc_tip4pew.xml  This is needed to get usable ions for OPC and OPC3, as well as (hopefully) not unduly restrict which combinations are available to use.

1. Does this sound okay?  Before taking this on (which is likely to be quite a lot of work), are there any potential issue with this plan, or something else we should do instead?
2. Could someone show me how openmm/wrappers/python/openmm/app/data/amber14/tip3p.xml was generated?  (it seems auto-generated, apologies if there was a readme that already explains this).  I'd like to be able to regenerate those before doing more conversions.

peastman:
Force field conversions are done at https://github.com/openmm/openmmforcefields.  It converts more files than what we ultimately bundle with OpenMM.  You can find the full set of Amber files at https://github.com/openmm/openmmforcefields/tree/master/amber/ffxml.  It still groups the ions and water parameters together though.  It's all configured through a [single file](https://github.com/openmm/openmmforcefields/blob/master/amber/solvents.yaml) that specifies what output files to generate and what input parameter files to include in each one.

@jchodera and @zhang-ivy might have thoughts on your suggestion of splitting them out.

rkingsbury:
@aizvorski thanks for posting; I am also seeking this exact same capability and just ran across your post. Let me know if I can help. It sounds like, in principle, we can edit `solvents.yaml` in `openmmforcefields` so as to generate the combinations of water+ion parameters we need?

[...]

PR #3773, opened 2022-09-11
Add all Amber ion params

aizvorski:
This converts one-to-one all Amber ion frcmod files (from AmberTools 22) into OpenMM xml files with the same names and the same contents.  Implements what is proposed in #3663

Run the conversion:
```
python ./devtools/forcefield-scripts/processAmberIons.py ~/anaconda3/pkgs/ambertools-22.0-py38h6177452_1
```
(or path to wherever ambertools is installed)

There is a script that makes comparing the generated xml files with older xml files a bit easier:
```
python ./devtools/forcefield-scripts/diffIons.py wrappers/python/openmm/app/data/amber14/tip4pew.xml wrappers/python/openmm/app/data/ions/ionsjc_tip4pew.xml  
python ./devtools/forcefield-scripts/diffIons.py wrappers/python/openmm/app/data/amber14/tip4pew.xml wrappers/python/openmm/app/data/ions/ions234lm_126_tip4pew.xml  
```

As far as I can tell, aside from some changes in the names of residues used for heavy ions (where Amber itself is not perfectly consistent), this matches the old sets in any case in which the two overlap.

However, it also adds sets for OPC, OPC3, and a proper set for TIP3PFB (where it seems the old xml files use TIP3P parameters), as well as adding the iod and hfe sets, ions1lm_* sets, etc.

Usage:
```
forcefield = app.ForceField(
    'amber14-all.xml', 
    'opc3.xml', 
    'ions/ionslm_126_opc3.xml',
)

forcefield = app.ForceField(
    'amber14-all.xml', 
    'tip3pfb.xml', 
    'ions/ionslm_126_fb3.xml',
)
```

cc @zhang-ivy

### DIFF
--- a/devtools/forcefield-scripts/diffIons.py
+++ b/devtools/forcefield-scripts/diffIons.py
@@ -1,0 +1,26 @@
+import sys
+import re
+import subprocess
+
+file1, file2 = sys.argv[1], sys.argv[2]
+
+with open(file1) as fh:
+    text1 = fh.read()
+
+with open(file2) as fh:
+    text2 = fh.read()
+
+m = re.search(r'Type class="([^\-]*)', text1)
+if m:
+    prefix1 = m.group(1)
+m = re.search(r'Type class="([^\-]*)', text2)
+if m:
+    prefix2 = m.group(1)
+
+text2a = text2.replace(prefix2, prefix1)
+text2b = re.sub(r'<!--.*?-->', '', text2a)
+
+with open('new.xml', 'w') as fh:
+    fh.write(text2b)
+
+subprocess.call(['diff', '-uw', file1, 'new.xml'])
--- a/devtools/forcefield-scripts/processAmberIons.py
+++ b/devtools/forcefield-scripts/processAmberIons.py
@@ -39,7 +39,9 @@ def amber_frcmod_ions_to_openmm_xml(
             
     ionname_to_residuename['Na+'] = 'NA' # this is a duplicate in atomic_ions.lib
     ionname_to_residuename['Cl-'] = 'CL' # this is a duplicate in atomic_ions.lib
-    
+    ionname_to_residuename['Cu2+'] = 'CU' # consistent with previous residue names
+    ionname_to_residuename['Cu+'] = 'CU1'
+
     with open(frcmod_file) as fh:
         lines = fh.readlines()
         lines = [x.strip() for x in lines]
@@ -109,7 +111,9 @@ def amber_frcmod_ions_to_openmm_xml(
     output_lines.append('  </AtomTypes>')
 
     output_lines.append('  <Residues>')
-    for ionname, mass, extra in mass_data:
+    # NOTE this is a hack to sort residues in a similar way to the older converted xml files, for easier diff
+    # NOTE it would be better to keep them in the order in which ions are listed
+    for ionname in sorted([x[0] for x in mass_data], key=str.casefold):
         if not ionname in ionname_to_residuename:
             print(f'Warning: ion name {ionname} not found in list of atomic ions')
             continue
@@ -153,7 +157,8 @@ os.makedirs(output_path, exist_ok=True)
 
 for frcmod_file in frcmod_files:
     if '1264' in frcmod_file:
-        print(f'Skipped {frcmod_file} because it has 12-6-4 parameters')
+        print(f'Skipped {frcmod_file} because it has 12-6-4 parameters, only 12-6 are supported')
+        continue
     setname = os.path.basename(frcmod_file)
     setname = setname.replace('frcmod.', '')
     xml_file = f'{output_path}/{setname}.xml'
--- a/devtools/forcefield-scripts/processAmberIons.py
+++ b/devtools/forcefield-scripts/processAmberIons.py
@@ -1,0 +1,164 @@
+import re
+import os
+import sys
+import glob
+import datetime
+import subprocess
+
+# from openmm import unit
+
+kilocalorie_to_kilojoule =  4.184 # 1.0*unit.kilocalorie/unit.kilojoule
+angstrom_to_nanometer = 0.1 # 1.0*unit.angstrom/unit.nanometer
+
+
+def md5sum(file):
+    result = subprocess.check_output(['md5sum', file]).decode('utf-8').split()[0]
+    return result
+
+
+def amber_frcmod_ions_to_openmm_xml(
+    frcmod_file, 
+    atomic_ions_file, 
+    xml_file, 
+    setname=None, 
+    include_extra_comments=True,
+):
+    with open(atomic_ions_file) as fh:
+        lines = fh.readlines()
+
+    ionname_to_residuename = {}
+    ionname_to_charge = {}
+    for i, line in enumerate(lines):
+        if 'unit.atoms table' in line:
+            name, type_, typex, resx, flags, seq, elmnt, chg = lines[i+1].split()
+            name = name.replace('"', '')
+            type_ = type_.replace('"', '')
+            # print(name, type_, typex, resx, flags, seq, elmnt, chg)
+            ionname_to_residuename[type_] = name
+            ionname_to_charge[type_] = float(chg)
+            
+    ionname_to_residuename['Na+'] = 'NA' # this is a duplicate in atomic_ions.lib
+    ionname_to_residuename['Cl-'] = 'CL' # this is a duplicate in atomic_ions.lib
+    
+    with open(frcmod_file) as fh:
+        lines = fh.readlines()
+        lines = [x.strip() for x in lines]
+        
+    mass_data = []
+    nonbon_data = []
+    comments_data = []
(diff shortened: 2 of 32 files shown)


aizvorski:
One particular note for TIP3P-FB (and TIP4P-FB)

This would be loaded in Amber/AmberTools (version 22) by loading dat/leap/cmd/leaprc.water.fb3 in tleap.  This pulls in  dat/leap/parm/frcmod.tip3pfb and dat/leap/parm/frcmod.ionslm_126_fb3

In OpenMM's current xml files:
amber14/tip3pfb.xml converted  from parm/frcmod.ionsjc_tip3p, "Monovalent ion parameters for Ewald and TIP3P water from Joung & Cheatham JPCB (2008)"
```    <Atom epsilon="0.8103692536" sigma="0.3037964628858557" type="tip3pfb_standard-K+"/>```

In this PR:
ions/ionslm_126_fb3.xml from parm/frcmod.ionslm_126_fb3 "Li/Merz ion parameters of atomic ions for the TIP3P-FB water model (12-6 set)", "HFE set for the TIP3P-FB water model from Sengupta et al., JCIM, 2021, 61, 869"
```    <Atom epsilon="0.59238168208" sigma="0.3037964628858557" type="ionslm_126_fb3-K+"/>```

In a nutshell, before this PR, OpenMM's TIP3P-FB uses TIP3P ions, and TIP4P-FB uses TIP4PEW ions.  Newer sets are available, and default in recent Amber versions - and may have very different numerical values.  

I think we may strive for maximum consistency with Amber here - it is probably easiest to do that if we do the conversions one-to-one (don't merge multiple Amber files into one xml file) and keep the same file names.

**Comment by aizvorski** [comment]
@peastman Any feedback on this PR?  What is needed before merging it?

**Comment by peastman** [comment]
Sorry for not replying sooner.  I was waiting to see what other people with more experience on force field conversion would say.

I think we probably want to start by putting this in https://github.com/openmm/openmmforcefields.  It aims to provide a complete conversion of all Amber files, generally with names matching the original files.  That's what you have here.  Then we look through them and select which ones to copy over to the main OpenMM package, often renaming them in the process.  So if other people agree, I think we want to start by putting your scripts and converted files into openmmforcefields.

**Comment by peastman** [comment]
I notice one difference from the current files.  You converted just the ions, not the water.  The current files combine them in a single file that you include to get both water and a set of ions designed to work with that water model.  For example, your `ions234lm_126_tip4pew.xml` file lists these sources:

```
parm/frcmod.ions234lm_126_tip4pew
lib/atomic_ions.lib
```

Compare that to the existing `tip4pew.xml` file which has these sources:

```
parm/frcmod.ionsjc_tip4pew
parm/frcmod.ions234lm_126_tip4pew
lib/atomic_ions.lib
tip4pew.xml
```

aizvorski:
@peastman Thanks for getting back to me!  I knew about openmmforcefields, and I first started by trying to coax it to do what this PR does, but it didn't work.

> It aims to provide a complete conversion of all Amber files, generally with names matching the original files.

Maybe that's what it aims to do, but when it comes to ions it doesn't really do that!  It's not a complete conversion, and the names don't match the original files.  Examples: original parm/frcmod.ionslm_126_opc3 -> not converted; original parm/frcmod.ionsjc_tip3p -> converted to amber/ffxml/tip3pfb_standard.xml, name doesn't match.

This creates confusion, it's probably one of the reasons why openmm's data/amber14/tip3pfb.xml for example has different ions from what Amber recommends.

> You converted just the ions, not the water. The current files combine them in a single file that you include to get both water and a set of ions designed to work with that water model

Yes, that's intentional - I tried to explain the case for this in #3663.  There isn't always a single correct set of ions for all use cases, and also some sets are recommended with more than one water model, if they are combined that would have duplication of the actual params.

I think it would be best to have some aliases (essentially xml files that only include several others, the way amber14-all.xml does) but otherwise neither duplicate nor combine water and ions.

aizvorski:
@peastman About whether to put this here or in openmmforcefields - if we do openmmforcefields, there would be two remaining issues which are essentially openmm bugs that would have to be fixed somehow:

1. No usable set of ions for OPC and OPC3, and
2. Wrong ions by default for TIP3P-FB and TIP4P-FB

Fixing those would require pulling in several of these xml files anyway.  Why *not* pull them all in?

peastman:
> Maybe that's what it aims to do, but when it comes to ions it doesn't really do that! It's not a complete conversion, and the names don't match the original files.

Right, but that's its goal.  Which is why it makes sense to start by fixing it there.  All the modern Amber files in OpenMM start out in openmmforcefields and then get copied over.

aizvorski:
@peastman Moved over to https://github.com/openmm/openmmforcefields/pull/239

Could you let me know who would be the main folks who need to look at that?

I know it is potentially a big change to switch the existing water+ions models to this, so I want to be sensitive to that (and do some very thorough before and after tests) but just adding more xml files should be low risk.

peastman:
Great, thanks!

> Could you let me know who would be the main folks who need to look at that?

@jchodera @mattwthompson and @mikemhenry are the primary maintainers of openmmforcefields.

aizvorski:
@peastman The port of this to openmmforcefields just landed in https://github.com/openmm/openmmforcefields/pull/242

Could we at least cherry pick opc_standard.xml, opc3_standard.xml and the included ions/ionslm_126_opc.xml and ions/ionslm_126_opc3.xml from that?  That would give us fully working OPC in time for openmm 8.0.  Please let me know and I can do a new PR just for that.

jchodera:
The OPC water model is officially recommended in the [Amber21 reference manual](https://ambermd.org/doc12/Amber21.pdf) over TIP-3P, so I would recommend we include it:
image

We're unfortunately still limited in not being able to support ff19SB until someone helps fix up the CMAP Amber -> OpenMM conversion support in [ParmEd](https://parmed.github.io/ParmEd/html/index.html).

jchodera:
I'll note the Amber21 manual also notes that OPC provides improvements over TIP3P for other force fields we _do_ distribute:
image

peastman:
> Could we at least cherry pick opc_standard.xml, opc3_standard.xml

I don't see any files with those names.  Which ones did you mean?

All the current Amber files merge the water model and the ions into a single file.  Can we make merged versions for these too?  I don't think we should change the current arrangement without careful consideration, and we certainly shouldn't have some water models work one way and others work the other.

aizvorski:
@peastman I meant from openmmforcefields not from this PR.  https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/ffxml/amber/opc_standard.xml and https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/ffxml/amber/opc3_standard.xml (plus the files they include)

> All the current Amber files merge the water model and the ions into a single file. Can we make merged versions for these too? 

I argued for why merged files are a non-optimal design decision in https://github.com/openmm/openmm/issues/3663  Why have the ability to include xml files if not to avoid duplication?

Rather than make merged files (which continues with the same problem) or split all files (which is a decently large, and if not done carefully, possibly breaking change), I would recommend simply adding new files using the new style, and leaving existing files as-is, for now.  That seems like the most conservative approach.  Risk avoidance > absolute consistency :)  

It seems most users would not see any difference in how this works in practice.  What user-facing problem are you concerned about?

aizvorski:
@peastman I made a PR here #3819

peastman:
> I would recommend simply adding new files using the new style, and leaving existing files as-is, for now.

I don't think that's a viable solution.  It would lead to confusion and incompatibility.  We made the decision to use merged files, and while we can certainly revisit that decision, it needs to be done carefully and with proper discussion.  Having some water models work one way and others work the other way would be very confusing.  And if we add split files for the new water models now, then discuss it in the future and decide we want to stick with merged ones, then fixing them will break compatibility.

aizvorski:
@peastman Could you describe what form you'd like this PR to have?  I'm a bit unsure of the details.  Do you want to just have one opc.xml file which is merged with ionslm_126_opc.xml? (analogous for opc3)  Or one file for just water and another one merged?  I can make either version pretty quick, but not sure what we're aiming for.

peastman:
Basically the same as the existing water files in `data/amber14`.  Take a look at https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/data/amber14/tip3pfb.xml for example.  There's one file that contains both the water model and ion parameters designed to go with it.

While adding the new water models, we also should add them to [the documentation](http://docs.openmm.org/latest/userguide/application/02_running_sims.html#amber14).  The source for that table is in `docs-source/usersguide/application/02_running_sims.rst`.  References go in `docs-source/usersguide/references.bib`.

aizvorski:
@peastman Okay, I changed #3819 to have merged files as you described, could you have a look?

aizvorski:
Closing this rather old pull request: all Amber ion params got into openmmforcefields in https://github.com/openmm/openmmforcefields/pull/239 (please use them from there if needed); OPC/OPC3 water models with params with ions are here in #3819

PR #3779, opened 2022-09-15
Made `Atom` objects be hashed by values.

ywu-roivant:
When an `Atom` object is copied, the copy should give the same hash. This is needed when `Atom` objects are used as dictionary keys, e.g., at here: https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/pdbfile.py#L439

### DIFF
--- a/wrappers/python/openmm/app/topology.py
+++ b/wrappers/python/openmm/app/topology.py
@@ -463,6 +463,14 @@ class Atom(object):
         ## A user defined identifier for this Atom
         self.id = id
 
+    def __hash__(self):
+        return hash(str(self))
+
+    def __eq__(self, other):
+        if isinstance(other, Atom):
+            return str(other) == str(self)
+        return False
+
     def __repr__(self):
         return "<Atom %d (%s) of chain %d residue %d (%s)>" % (self.index, self.name, self.residue.chain.index, self.residue.index, self.residue.name)
 


peastman:
Can you describe what problem this is intended to solve?  The current implementation treats every Atom as a unique object that is part of a unique Topology.  This changes it so that if two Atoms have the same string representation they are treated as the same atom, even if they come from different Topologies.  What's the reason for that?

ywu-roivant:
Atom objects are allowed to be copied. I don't know exactly where copying happens. What I see is those iterated by `topology.bond()` are not the same iterated by `topology.chains().atoms()`; and then [this function](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/pdbfile.py#L402) will fail at [here](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/pdbfile.py#L439). There may be problem in my code constructing the Topology object.
This change proposes to treat two atoms as **equal** atoms when their values are equal, it doesn't treat the two atoms as the **same** atom. Two atoms can still be two unique atoms even when their values are equal. But it does alter the hashing behavior, from id-based hashing to value-based hashing.

peastman:
> What I see is those iterated by topology.bond() are not the same iterated by topology.chains().atoms()

That means there's something wrong with your Topology.  I just created a Topology by loading a PDB file and ran the following test.

```python
atoms1 = set(topology.atoms())
atoms2 = set()
for chain in topology.chains():
    atoms2 = atoms2.union(set(chain.atoms()))
for atom1, atom2 in topology.bonds():
    if atom1 not in atoms1 or atom2 not in atoms1 or atom1 not in atoms2 or atom2 not in atoms2:
        print('error')
```

It didn't report any errors.

How did you create your Topology?

ywu-roivant:
Yah. It could be my code doing something wrong with Topology. As long as it's something easy to fix at my side, I am fine  with the current ID-based hashing. At the surface, my code that constructs the Topology does nothing fancy but calling addAtom and addBond. I need a bit time to look deeper.

PR #3791, opened 2022-09-24
Fix #3790: Collision rate for MTS BAOAB Langevin integrator now correctly accounts for number of substeps

### LINKED ISSUE #3790: MTSLangevinIntegrator friction bug
Hi all, I think I found a bug in the [MTS Langevin Integrator](https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/wrappers/python/openmm/mtsintegrator.py) code.

Currently the routine defines constants `a` and `b` as

https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/wrappers/python/openmm/mtsintegrator.py#L170-L171

and then applies an Ornstein-Uhlenbeck update using these constants _at every substep_

https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/wrappers/python/openmm/mtsintegrator.py#L187-L192

The problem being that if I have $N$ substeps then I effectively am using $N\times$ the friction in simulation, which can affect diffusion coefficients and the like.

An easy fix would be to redefine the `a` and `b` constants to divide the per-substep OU update by the number of total substeps, something like

    total_substeps = math.prod( [group_size for (group_id, group_size) in groups] )
    self.addGlobalVariable("a", math.exp(-friction*dt / total_substeps))
    self.addGlobalVariable("b", math.sqrt(1-math.exp(-2*friction*dt / total_substeps)))

That way the total damping should be correct. An alternative would be to give the user control of how the damping is allocated between the groups.

Cheers - CM

jchodera:
Previously, `openmm.MTSLangevinIntegrator` would use an effective collision rate that is too large when there were multiple steps, since every substep used the outer timestep instead of the inner substep timestep to apply the friction.

This PR corrects this, using a fix suggested by Charlie Matthews (@c-matthews) in #3790.

### DIFF
--- a/wrappers/python/tests/TestIntegrators.py
+++ b/wrappers/python/tests/TestIntegrators.py
@@ -156,13 +156,42 @@ class TestIntegrators(unittest.TestCase):
         temperature = averageEnergy*2/(dofs*MOLAR_GAS_CONSTANT_R)
         self.assertTrue(290*kelvin < temperature < 310*kelvin)
 
+    def testMTSLangevinIntegratorFriction(self):
+        """Test the MTSLangevinIntegrator on a force-free particle to ensure friction is properly accounted for (issue #3790)"""
+        # Create a System with a single particle and no forces
+        system = System()
+        system.addParticle(12.0*amu)
+        platform = Platform.GetPlatformByName('Reference')
+        initial_positions = Vec3(0,0,0)
+        initial_velocities = Vec3(1,0,0)
+        nsteps = 125 # number of steps to take
+
+        def get_final_velocities(nsubsteps):
+            """Get the final velocity vector after a fixed number of steps for the specified number of substeps"""
+            integrator = MTSLangevinIntegrator(0*kelvin, 1/picosecond, 4*femtoseconds, [(0,ngroups)])
+            context = Context(system, integrator, platform)
+            context.setPositions(initial_positions)
+            context.setVelocities(initial_velocity)
+            integrator.step(nsteps)
+            final_velocities = context.getState(getVelocities=True).getVelocities()
+            del context, integrator
+            return final_velocities
+
+        # Compare sub-stepped MTS with single-step MTS
+        reference_velocities = get_final_velocities(1)
+        for nsubsteps in range(2,6):
+            mts_velocities = get_final_velocities(nsubsteps)
+            self.assertAlmostEqual(reference_velocities.x, mts_velocities.x)
+            self.assertAlmostEqual(reference_velocities.y, mts_velocities.y)
+            self.assertAlmostEqual(reference_velocities.z, mts_velocities.z)
+
     def testNoseHooverIntegrator(self):
         """Test partial thermostating in the NoseHooverIntegrator (only API)"""
         pdb = PDBFile('systems/alanine-dipeptide-explicit.pdb')
         ff = ForceField('amber99sbildn.xml', 'tip3p.xml')
         system = ff.createSystem(pdb.topology, nonbondedMethod=PME)
 
-        integrator = NoseHooverIntegrator(1.0*femtosecond) 
+        integrator = NoseHooverIntegrator(1.0*femtosecond)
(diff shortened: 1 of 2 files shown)


peastman:
The tests are failing with

>  AttributeError: module 'math' has no attribute 'prod'

The number of substeps per step is just `groups[-1][1]`.  The second element of each tuple is the number of substeps per complete step for a group, not the number per substep of the previous group.  So we just want the maximum, and since we've already sorted them, that's `groups[-1]`.

It might be useful to add a test case for this in TestIntegrators.py.  Something like this:

- Create System with a single particle and no forces applied to it.
- Create a MTSLangevinIntegrator with a few sets of nested groups (which don't really apply any forces) and a temperature of zero (so velocity will decay toward zero at a known rate).
- Give the particle an initial nonzero velocity, take some steps, and see if the velocity is decaying at the expected rate.

Or if you want something a bit fancier, you could do something similar to [this test for LangevinMiddleIntegrator](https://github.com/openmm/openmm/blob/ffc3ee78863e3778558ba7b0bcfde55f8c3d8897/tests/TestLangevinMiddleIntegrator.h#L49-L90), which simulates a damped harmonic oscillator and checks against the analytical solution.

wrappers/python/tests/TestIntegrators.py
```diff
@@ -156,13 +156,42 @@ def testMTSLangevinIntegrator(self):
...
+            del context, integrator
+            return final_velocities
+
+        # Compare sub-stepped MTS with single-step MTS
+        reference_velocities = get_final_velocities(1)
+        for nsubsteps in range(2,6):
+            mts_velocities = get_final_velocities(nsubsteps)
+            self.assertAlmostEqual(reference_velocities.x, mts_velocities.x)
+            self.assertAlmostEqual(reference_velocities.y, mts_velocities.y)
+            self.assertAlmostEqual(reference_velocities.z, mts_velocities.z)
```
peastman, line 186: It's better to check against the analytical result.  Otherwise you're just checking that the velocity doesn't depend on `nsubsteps`, which is a useful check but doesn't prove it's correct.

```python
self.assertAlmostEqual(math.exp(-context.getTime()*friction, mts_velocities.x))
self.assertAlmostEqual(0, mts_velocities.y)
self.assertAlmostEqual(0, mts_velocities.z)
```

jchodera: Done!

jchodera:
Thanks! And apologies for "debugging by CI". I still don't have an OpenMM dev environment set up on my M1 laptop.

peastman:
I see we never got around to finishing this up.  Is it ready to merge?

I notice that in a different test case, you changed

```python
integrator.step(500)
```

to

```python
integrator.step(5000)
```

What was the reason for that?  It adds about 10 seconds to running the test suite, which makes it almost as expensive as all the other integrator tests put together.

peastman:
@jchodera we still need to finish this up.  Can you check the issue mentioned above?

jchodera:
Sorry about this---I've tried to revert this.

jchodera:
@peastman : It looks like the answer to your question is "the test fails if you don't take a sufficient number of steps":
```
        # See if the temperature is correct.
        totalEnergy = 0*kilojoules_per_mole
        steps = 50
        for i in range(steps):
            integrator.step(10)
            totalEnergy += context.getState(getEnergy=True).getKineticEnergy()
        averageEnergy = totalEnergy/steps
        dofs = 3*system.getNumParticles() - system.getNumConstraints() - 3
        temperature = averageEnergy*2/(dofs*MOLAR_GAS_CONSTANT_R)
>       self.assertTrue(290*kelvin < temperature < 310*kelvin)
E       AssertionError: False is not true
```

jchodera:
@peastman : Why don't you take this over? I'm time limited for the next week.

jchodera:
Seems like that worked!

Issue #3611, opened 2022-05-20
Handling mixed periodic/non-periodic CVs in Metadynamics


sukritsingh:
In setting up a metadynamics simulation with a system and state file, I was hoping to be able to setup a simulation using a dihedral and a pair of distances as my collective variables. However, when I set it up using the following: 

```
# unpack the equilirated system
bz2_file = f'metadynamics/system.xml.bz2'
with bz2.open(bz2_file, 'rb') as infile:
    system = XmlSerializer.deserialize(infile.read().decode())

# define atom indices for measuring distances
d1_atom1_ind=2464
d1_atom2_ind=974
d2_atom2_ind=641
#define atoms for measuring dihedrals 
phe_chi = [2454, 2455, 2458, 2459]

chi = CustomTorsionForce('theta')
chi.addTorsion(phe_chi[0], phe_chi[1], phe_chi[2], phe_chi[3])
chi_bias = BiasVariable(chi, -np.pi, np.pi, 0.35, True, 181)

d1 = CustomBondForce('r')
d1.addBond(d1_atom1_ind, d1_atom2_ind)
d1_bias = BiasVariable(d1, 0.3, 2.3, 0.05, periodic=False)

d2=CustomBondForce('r')
d2.addBond(d1_atom1_ind, d2_atom2_ind)
d2_bias = BiasVariable(d2, 1.2, 5.0, 0.05, periodic=False)

meta = Metadynamics(system, [d1_bias, d2_bias, chi_bias], 300, biasFactor=5, height=1, frequency=1, saveFrequency=10, biasDir="metadynamics/")
```
 I get the the following `ValueError`:
```
ValueError: Metadynamics cannot handle mixed periodic/non-periodic variables
```

Is there a reason OpenMM can't handle mixed periodic/non-periodic CVs? In previous metadynamics simulations (used PLUMED though) I've never experienced issues with mixed CVs like these, so my baseline assumption is that it's reasonable to use a distance and a dihedral as pair of CVs. 

Is there a workaround? I noticed I can get `meta` to not throw errors if I do something like: 
`chi_bias = BiasVariable(chi, -np.pi, np.pi, 0.35, False, 181)`
where `Periodic=False` is the argument passed. However, for dihedrals this would be incorrect so I'm hesitant to do this as a workaround.

Files + command to get to this error are attached here:  
[meta-period-example.tar.gz](https://github.com/openmm/openmm/files/8743329/meta-period-example.tar.gz)

[...]

PR #3813, opened 2022-10-21
Parallelize fitting 3D splines

peastman:
This is a first step toward addressing the performance problems discussed in #3611.  With a good CPU, fitting large 3D tabulated functions should be several times faster.

In principle this could be done for 2D functions too, but those are so much faster that there likely wouldn't be a lot of benefit.

@sukritsingh can you try this out?


sukritsingh:
Happy to test it out! One question @peastman: What's the best/preferred way to install a local copy of this branch to run and test? I've been installing and creating fresh environments of OpenMM using `conda` commands, so I'm not too sure what the way to get this branch is. Should I just clone a copy of your fork and checkout the branch to run it? And then I'd compile/install from source?

Sorry if this is a super basic question - just want to make sure I do it the appropriate/correct way. Thanks!

peastman:
Here's the commands github suggests:

```
git checkout -b peastman-spline master
git pull git@github.com:peastman/openmm.git spline
```

Then build from source.

peastman:
@sukritsingh have you had a chance to test this out?  We're quickly running out of time to get this into 8.0.

sukritsingh:
So sorry for the delay! I'm working on this today. I had a few fires and bugs to deal with for FAH and other stuff that needed my attention this week. I'll use the same scripts/systems as before so hopefully the bottleneck is just installation/setup.

sukritsingh:
I cloned the openmm repo, ran the two github commands you said above, then ran the following commands: 
```
cd openmm
mkdir build 
cd build
ccmake ..
```
changing just the install prefix to a local folder and then running `make`. However, the following error popped up: 
```
In file included from /home/singhs15/src/openmm/platforms/reference/src/SimTKReference/ReferencePME.cpp:43:0:
/home/singhs15/src/openmm/libraries/pocketfft/include/pocketfft_hdronly.h: In function ‘void* pocketfft::detail::aligned_alloc(std::size_t, std::size_t)’:
/home/singhs15/src/openmm/libraries/pocketfft/include/pocketfft_hdronly.h:165:35: error: ‘max_align_t’ was not declared in this scope
   align = std::max(align, alignof(max_align_t));
                                   ^
make[2]: *** [CMakeFiles/OpenMM.dir/build.make:1994: CMakeFiles/OpenMM.dir/platforms/reference/src/SimTKReference/ReferencePME.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2149: CMakeFiles/OpenMM.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```
Any insight as to what might be causing the issues?

This is on our HPC system where I used the 2080 Ti used to run the previous tests. I'm building it now on a local desktop with an A4000 as an alternative.

peastman:
What compiler are you using?  `max_align_t` is part of the standard library since C++11.

sukritsingh:
The gcc loaded by default in the HPC system is `7.3.0`: 
```
 gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/admin/opt/common/CentOS_7/gcc/gcc-7.3.0/bin/../libexec/gcc/x86_64-pc-linux-gnu/7.3.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: ../configure --prefix=/opt/common/CentOS_7/gcc/gcc-7.3.0 --enable-languages=c,c++ --disable-multilib : (reconfigured) ../configure --prefix=/opt/common/CentOS_7/gcc/gcc-7.3.0 --enable-languages=c,c++ --disable-multilib
Thread model: posix
gcc version 7.3.0 (GCC) 
```

Should it be a later version? I think the cluster also has gcc 10.2.0

sukritsingh:
Update: I successfully compiled this branch on a desktop with an A4000 GPU - running benchmarking tests now so should have an update soon!

sukritsingh:
Posted below are the results of the head to head comparisons, all run for 100,000 steps on an A4000 GPU this time. Same system and state as #3611 

The 'Old spline fitting' was generated using the previous environment in #3611. The "New parallel splines" data was generated using this branch with an OpenMM 8.0 Beta. I've generated the chart with two different versions of the x-axis (one linear, the other as regularly spaced out values) to for easy visualization/comparison. As before, the shaded region represents a benchmark of the equilibrium simulation with no metadynamics with 5% variance in performance. As before, I set `save_frequency` to a constant value (of 50,000). I also made sure to remove any mention of `meta._syncWithDisk()`. 

Overall: Parallelizing the splines works is a great improvement! It now achieves performance comparable to native equilibrium MD with a `frequency=5000`, while the dropoff in performance for the old code happened much sooner.

Chart with linear x-axis:
dev-meta-speed-nsday-linearX

Chart with same data with regularized spacing for easier visualization: 
dev-meta-speed-nsday-regularX

peastman:
Thanks!

PR #3814, opened 2022-10-22
Minor spelling nit

philipturner:
I just encountered this while reading through the contributing guidelines, and thought it might be important to fix since it's public documentation. I have a bad habit of pointing out typos, so for less important cases like typos in source code comments I probably won't make a PR.

If this is discouraged because it triggers integration tests, please let me know so I can avoid repeating something similar in the future.


peastman:
Congratulations on your *successful* contribution. :)

PR #3819, opened 2022-10-25
Add OPC and OPC3 ion sets

aizvorski:
Make merged xml files from `opc*xml` and `ions/ionslm_126_opc*.xml` from https://github.com/openmm/openmmforcefields/commit/b12c2e871b60275a10b3f8f56dfc2f6d5447591f


peastman:
When I create a force field with

```python
forcefield = ForceField('amber14-all.xml', 'amber14/opc3.xml')
```

it fails with the exception

```
ValueError: Residue template U with the same override level 0 already exists.
```

That's because the new file defines a residue

```xml
    <Residue name="U">
      <Atom charge="4.0" name="U" type="ionslm_126_opc3-U4+"/>
    </Residue>
```

which conflicts with the standard uracil residue.  Comparing to `tip3p.xml`, in that case the corresponding residue has the name `U4+`:

https://github.com/openmm/openmm/blob/7e1ebf1cfde67f47af158206bd11b859627447fe/wrappers/python/openmm/app/data/amber14/tip3p.xml#L228-L230

Does anyone know why we're getting a different name now?  Did something change in the openmmforcefields script?

aizvorski:
@peastman This should be fixed now. 

This was kind of an overlap of several issues:

* the conversion code that produced these is completely written from scratch since it was intended to convert everything (and it was hard to get the old code to do that)
* the naming in the Amber file lib/atomic_ions.lib which defines the names of the residues and atoms for all ions is pretty inconsistent (is the residue all upper case or mixed case? is the charge included in the name? etc).  it also has some duplicates, and some odd cases which were probably handled specially before
* the new ion sets add some ions that didn't have parameters before (Ag+, Cu+, Tl+); in atomic_ions.lib these are called AG CU and TL (as opposed to the divalent/trivalent Ag, Cu and Tl).  This last one kinda raises a question, does openmm consider formal charge on atoms when matching residues?)

The current state after a couple of renames is that every ion which exists in both tip3p.xml and the new xml files should have the same name.

peastman:
> This last one kinda raises a question, does openmm consider formal charge on atoms when matching residues?

That's a good question.  No, it doesn't.  Templates are matched only on elements and bonds.  In fact, Topology doesn't even store formal charges (though we've discussed adding it), and most of the file formats we create Topologies from don't report it.

The result is that if you use these files and try to parametrize an Ag, Cu, or Ti ion, you'll get an exception because it won't know which version to use:

https://github.com/openmm/openmm/blob/e2a5b1686f60ba9bc53c145b00126d0b3b2ad48b/wrappers/python/openmm/app/forcefield.py#L1013-L1017

I don't know whether you consider that a problem or not.  These aren't ions that people use very often, at least in biomolecular simulations.

aizvorski:
@peastman Hmm - would this also happen for FE vs FE2?  That is rather more common, and also it is in the existing tip3p.xml etc

I like storing formal charges and requiring charges to match.  PDB format can definitely include charges too, the PDBs written from a lot of prep tools add this (at the same time they add hydrogens etc).  Does the current OpenMM PDB reader/writer ignore them?

peastman:
> would this also happen for FE vs FE2?

Correct.

>  Does the current OpenMM PDB reader/writer ignore them?

Also correct.

aizvorski:
@peastman Okay, thanks!

I think this is okay as is for now, since the same issues exist in the current xml files (and are nontrivial to fix without changes to template matching).  What do you think - okay to merge?

peastman:
I tried to run a simulation using the OPC3 file.  It immediately blows up.  I believe there's an error in this line:

```xml
<Bond class1="OW" class2="HW" length="0.978882" k="502416.0"/>
```

The length is measured in nm.  That would be a *very* large water molecule!

It looks like this was copied directly from the earlier `opc3.xml` file, which also has the same error.  In TestForceField.py there's a pair of test cases for OPC and OPC3, but they don't catch this error because they only check rigid water (so bonded terms are omitted), and they don't try to run dynamics or apply constraints, either of which would have revealed the problem.

How about modifying those tests as follows:

1. Make them check the new files as well as the old ones.
2. After computing the energy, call `context.applyConstraints(1e-5)` and check that the energy doesn't change significantly.

aizvorski:
@peastman Fixed!  

OPC was already okay, there was already a constraints test for OPC just not OPC3 (the bugs show up exactly in the parts that don't have tests :)

peastman:
Looks good, thanks!

> the bugs show up exactly in the parts that don't have tests :)

Funny how that happens.  Almost as if the tests were doing something useful. :)

PR #3842, opened 2022-11-11
Initialize pinnedBuffer at OpenCLContext creation

craabreu:
This initialization is needed to prevent segfault during object destruction in certain circumstances (e.g., when a `Force` is changed before calling `Context.reinitialize()` and this change causes the corresponding `ForceImpl` to throw an exception).


peastman:
Thanks!

PR #3860, opened 2022-11-29
Attempt at fixing failing pypy build

peastman:


peastman:
For some reason the pypy build recently started failing with an error during installation:

```
Solving environment: ...working... failed
  Warning: 
  ResolvePackageNotFound: 
    - python3.7.*=*pypy
```

I don't know why it started failing.  The only way I eventually found to make it work again was to use a different environment file for that build that specifies `pypy` in place of `python`.  I'm not sure if that's really the best solution though.  If anyone has insight it would be appreciated.

I also switched it to use mambaforge, which at one point I thought might be needed.  I don't think it really is, but I left it in because it seems to save about a minute and a half in the "prepare base dependencies" stage.

@mikemhenry if you have time to take a look at my changes, that would be great.

PR #3868, opened 2022-12-12
update hosted env to use newer openmm-torch version

mikemhenry:


mikemhenry:
Once the build is green here https://github.com/conda-forge/openmm-torch-feedstock/commits/rc

We should be good to merge this in and update the hosted env, I want to wait till it hits the CDN so that people won't get a "version can't be found" error

peastman:
We need to add the pytorch 1.13 builds first.

mikemhenry:
PR for that here: https://github.com/conda-forge/openmm-torch-feedstock/pull/31

peastman:
I tried to create an environment with the new file, but it fails with this error:

```
Encountered problems while solving:
  - package openmm-ml-1.0beta-pyh79ba5db_0 requires openmm-torch 1.0beta, but none of the providers can be installed
```

peastman:
Possibly we should rebuild the openmm-torch package with the same version as before (1.0beta), just updating the build number.  There weren't any code changes in it, so there's no need for an updated version number.  That will maintain compatibility with the other beta packages.

mikemhenry:
> Possibly we should rebuild the openmm-torch package with the same version as before (1.0beta), just updating the build number. There weren't any code changes in it, so there's no need for an updated version number. That will maintain compatibility with the other beta packages.

I agree, honestly, that is what I thought I was doing when I changed the version number. I will fix that with https://github.com/conda-forge/openmm-torch-feedstock/pull/31

devtools/hosted-envs/openmm-8-rc1-linux.yaml
```diff
@@ -0,0 +1,14 @@
...
+channels:
+  - conda-forge/label/openmm_rc
+  - conda-forge/label/openmm-torch_rc
+  - conda-forge
+
+dependencies:
+  - openmm==8.0.0rc1
+  - openmm-ml==1.0rc1
+  - openmm-torch==1.0rc1
+  - nnpops==0.2
```
peastman, line 12: This should be 0.3.

mikemhenry:
@peastman good catch! Fixed + uploaded new envs to my mmh channel

PR #3887, opened 2023-01-05
Overhaul benchmarking script to add lots of new features

peastman:
These are the changes from #3386.  That branch somehow got messed up and I'm not sure how to fix it, so it's easiest to create a new PR.

In addition to the changes there, I added more help text with usage examples.

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -3,8 +3,98 @@ import openmm.app as app
 import openmm as mm
 import openmm.unit as unit
 from datetime import datetime
-import os
-from argparse import ArgumentParser
+import argparse
+
+def cpuinfo():
+    """Return CPU info"""
+    import os, platform, subprocess, re
+    if platform.system() == "Windows":
+        return platform.processor()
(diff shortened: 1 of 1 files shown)


jchodera:
This looks good!

I'd just add one more example that shows folks what we would like them to run for benchmarking to send back to us, e.g.
```
To run a series of benchmarks and generate a YAML file to send back to the OpenMM developers to report on performance on your hardware, please run (e.g. for NVIDIA GPUs):

python benchmark.py --platform=CUDA --precision=single --style=table --outfile single.yaml
python benchmark.py --platform=CUDA --precision=mixed --style=table --outfile mixed.yaml
```
Otherwise looks great!

jchodera:
For tracking benchmarks on a variety of hardware.
Mixed precision is of interest too. We don't use single precision.

Eventually, we can add an option to submit the YAML to a webserver, which would automate the capture of benchmarks, but it's useful to capture the detailed data provided in the YAML at the moment, since the table doesn't contain as much detail.

PR #3902, opened 2023-01-12
Optimized reducing energy

peastman:
This reduces the overhead of evaluating the energy on the CUDA and OpenCL platforms.  If you're doing a lot of energy evaluations for a very small system or inexpensive force, this can significantly improve performance.  In most other cases the effective is negligible.

### DIFF
--- a/platforms/cuda/src/CudaContext.cpp
+++ b/platforms/cuda/src/CudaContext.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2009-2021 Stanford University and the Authors.      *
+ * Portions copyright (c) 2009-2023 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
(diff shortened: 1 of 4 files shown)


philipturner:
@bdenhollander I tested this in the Metal plugin, and it harms performance across the board for FP32. I'll try again once I integrate mixed precision, which is more computationally intensive to sum. The only exception seems to be `amber20-stmv`, a supermassive system.

| Benchmark | Before | After (32 cores) | After (4 cores) |
| --------- | ------ | ---------------- | --------------- |
| gbsa | 633.6 | 635.3 | 634.4 |
| pme | 266.6 | 266.3 | 257.3 |
| apoa1rf | 151.6 | 151.4 | 151.4 |
| amber20-dhfr | 280.7 | 280.3 | 278.4 |
| amber20-stmv | 6.898 | 6.910 | 6.918 |

philipturner:
@peastman I achieved on order-of-magnitude speedup by unrolling the first loop. This allows for latency hiding and fusion of memory access instructions. The speedup is only detectable in [ui.perfetto.dev](https://ui.perfetto.dev) (not visible from ns/day) because there’s a driver latency bottleneck on the CPU.

https://github.com/philipturner/openmm-metal/commit/8ccf366c3d23ab938261646f6cd95a6720ba9613#diff-4dd67ae2feb39d13a3e32f5389fa24dbf7fc8bdc3f86adca0ee605f97db4928a

On that note, there also is a marginal speedup from using multiple threadgroups, looking like Amdahl’s Law scaling. It harms performance for larger systems because multiple GPU kernels are running concurrently. Parallelization simply consumes more power.

| Atoms | OpenMM OpenCL | Metal TG=1 | Metal TG=2 | Metal TG=4 | Metal TG=8 |
| ----- | ------------- | ----- | ----- | ----- | ----- |
| 96    | ~65 µs        | 9 µs  | 7 µs  | 6 µs  | 6 µs  |
| 306   | ~65 µs        | 9 µs  | 7 µs  | -     | -     |
| 774   | ~65 µs        | 9 µs  | 7 µs  | -     | -     |
| 2661  | ~65 µs        | 9 µs  | 7 µs  | -     | -     |
| 4158  | ~65 µs        | 9 µs  | 7 µs  | 6 µs  | 6 µs  |

Finally, I know how Apple force-unrolls loops in Metal Shading Language. Haven't tested it with OpenCL.

```cpp
#pragma clang loop unroll(full)
```

PR #3911, opened 2023-01-16
Double the performance on M1 and AMD Mac GPUs

### LINKED ISSUE #2525: Slow OpenCL performance on MacBook Pro (Vega 20 GPU)
A colleague using ISOLDE on his 2018 MacBook Pro reports glacially slow MD performance. Currently, ISOLDE just relies on OpenMM (currently 7.4.0.dev-b71e92d) to pick the fastest available GPU, but I suspect this might be misfiring here. From his description the speed is about the same as what I get on my MacBook Air, so the most likely explanation is that it's choosing to run on the onboard Intel GPU rather than the Radeon...

Unfortunately I don't have easy access to a high-end MacBook Pro to test on (but will ask around). Any advice? My first thought is to do a custom build with the constructor for `OpenCLContext` reporting its decision-making process to stdout, and ask him to run it.

### LINKED ISSUE #3847: Low performance on M1 GPU
I ran the benchmark for ApoA1 and compared then to the RTX 2080, which has equivalent performance (10 TFLOPS) and bandwidth (400 GB/s) to the M1 Max. There is an order of magnitude difference in GPU performance:

```
M1 Max:
(base) philipturner@M1-Max-MacBook-Pro Desktop % python3 benchmark.py --platform CPU --test apoa1rf
Platform: CPU

Test: apoa1rf
Ensemble: NVT
Step Size: 4 fs
Integrated 1841 steps in 62.0415 seconds
10.2552 ns/day
(base) philipturner@M1-Max-MacBook-Pro Desktop % python3 benchmark.py --platform OpenCL --test apoa1rf
Platform: OpenCL
Precision: single

Test: apoa1rf
Ensemble: NVT
Step Size: 4 fs
Integrated 7457 steps in 61.8533 seconds
41.6654 ns/day
(base) philipturner@M1-Max-MacBook-Pro Desktop % python3 benchmark.py --platform OpenCL --test apoa1pme
Platform: OpenCL
Precision: single

Test: apoa1pme
Ensemble: NVT
Step Size: 4 fs
Integrated 5796 steps in 63.1291 seconds
31.7302 ns/day
(base) philipturner@M1-Max-MacBook-Pro Desktop % python3 benchmark.py --platform OpenCL --test apoa1ljpme
Platform: OpenCL
Precision: single

Test: apoa1ljpme
Ensemble: NVT
Step Size: 4 fs
Integrated 4614 steps in 62.6888 seconds
25.4367 ns/day
```

From https://openmm.org/benchmarks:

```
Nvidia RTX 2080:
RF: 283.927 ns/day
PME: 194.643 ns/day
LJPME: 158.666

CPU:
RF: 8.823
PME: 7.26
LJPME: 6.27
```

One hypothesis is that the Apple GPU has low `__local` memory performance. It's at least 2 times slower than other vendors. But that does not fully explain the 6x difference in performance. I've faced similar problems in GROMACS with the M1 Max being an order of magnitude slower than an Nvidia GPU with equivalent TFLOPS, and only 2x faster than CPU.

Another hypothesis is that we face the problems from this issue: https://github.com/openmm/openmm/issues/2525. An AMD GPU has much slower performance on macOS than expected. Is there a way to narrow down the performance regression and find what's causing it?

### LINKED ISSUE #3899: RDNA2 Performance Tuning
I've been profiling, tuning, and benchmarking OpenMM on an [AMD Radeon RX 6600](https://www.techpowerup.com/gpu-specs/radeon-rx-6600.c3696) with the goal of improving FAH on AMD. It has been a challenge to improve performance overall without dropping gbsa and amoebagk performance by 8%.

My changes are listed in this commit: https://github.com/openmm/openmm/commit/67fc75b0e68a36bee09296f418efd3c50d8e5339 I don't have other GPUs to test for regressions so if any of these are useful on other devices, it would be great to get them merged.

| Benchmark  | Atoms | OpenMM 7.7.0 | Tuned 7.7.0 | Change |
|------------|------:|-------------:|------------:|-------:|
| gbsa       | 2489  | 378.719      | 368.491     | -2.70% |
| rf         | 23558 | 388.071      | 403.127     | 3.88%  |
| pme        | 23558 | 265.914      | 278.566     | 4.76%  |
| apoa1rf    | 92224 | 131.435      | 145.921     | 11.02% |
| apoa1pme   | 92224 | 80.181       | 86.1428     | 7.44%  |
| apoa1ljpme | 92224 | 61.3489      | 66.472      | 8.35%  |
| amoebagk   | 2489  | 1.33752      | 1.34371     | 0.46%  |
| amoebapme  | 23558 | 4.24589      | 4.7204      | 11.18% |
| | | | Average | 5.55% |

| Test       | Precision | Constraints | H mass (amu) | dt (fs) | Ensemble | Platform | OpenMM 8.0 n/s | Tuned 8.0 n/s | Change |
|------------|-----------|-------------|-------------:|--------:|----------|----------|---------------:|--------------:|-------:|
| gbsa       | mixed     | HBonds      | 1.5          | 4       | NVT      | OpenCL   | 377.407        | 376.896       | -0.14% |
| rf         | mixed     | HBonds      | 1.5          | 4       | NVT      | OpenCL   | 386.766        | 399.861       | 3.27%  |
| pme        | mixed     | HBonds      | 1.5          | 4       | NVT      | OpenCL   | 265.2          | 278.561       | 4.80%  |
| apoa1rf    | mixed     | HBonds      | 1.5          | 4       | NVT      | OpenCL   | 124.19         | 131.482       | 5.55%  |
| apoa1pme   | mixed     | HBonds      | 1.5          | 4       | NVT      | OpenCL   | 79.0307        | 84.8213       | 6.83%  |
| apoa1ljpme | mixed     | HBonds      | 1.5          | 4       | NVT      | OpenCL   | 60.5439        | 65.2943       | 7.28%  |
| amoebagk   | mixed     | None        | 1            | 2       | NVT      | OpenCL   | 1.38164        | 1.39667       | 1.08%  |
| amoebapme  | mixed     | None        | 1            | 2       | NVT      | OpenCL   | 4.23471        | 4.74045       | 10.67% |
| | | | | | | | | Average | 4.92% |

philipturner:
Apple silicon has a ~3x speedup on apoa1rf. The M1 Max went from 40 ns/day to 113 ns/day. Other benchmarks sped up less significantly. There's still a ways to go before surpassing the lower-spec RX 6600 (131 ns/day) or approaching the RTX 2080 (284 ns/day). For example, we could use VkFFT during PME. The theoretical limit should be another 1.4x speedup, ~160 ns/day @ 45 W.

I attempted to port https://github.com/openmm/openmm/pull/2989 and it harmed performance. Most likely, because it fetched an extra 4th element from threadgroup memory each iteration. M1 has higher-latency and half-throughput threadgroup memory. This kernel isn't bandwidth-bound, but the low occupancy (6 simds/scheduler) likely can't hide the high latency (~48 cycles). I can retry with SIMD shuffles in a future Metal backend.

| M1 Max (Single Precision) | Before | After | Speedup |
| ------ | ------ | ----- | ------- |
| gbsa | 264 ns/day | 380 ns/day | +44% |
| rf | 153 ns/day | 304 ns/day | +99% |
| pme | 120 ns/day | 191 ns/day | +59% |
| apoa1rf | 38-42 ns/day | 112-113 ns/day | +169-195% |
| apoa1pme | 33 ns/day | 63 ns/day | +91% |
| apoa1ljpme | 28 ns/day | 47 ns/day | +68% |
| amoebagk | 0.80 ns/day | 1.19 ns/day | +49% |
| amoebapme | 2.72 ns/day | 3.90 ns/day | +43% |

Modest Windows RDNA speedup below (credit @bdenhollander). His tuning compounds on the fact that we're recognizing Mac AMD GPUs as having the correct SIMD width. Most speedup comes from correctly identifying this number (32 or 64, not 1). That change alone improved each benchmark 38%, 90%, 150%, 76%, 57%, 26%, and 31% respectively on M1. Hopefully someone else can verify a ~2x speedup on GCN or Mac RDNA.

| RX 6600 (Mixed Precision) | Before | After | Speedup |
| ------- | ------ | ----- | ------- |
| gbsa | 377 ns/day | 377 ns/day | +0% |
| rf | 387 ns/day | 400 ns/day | +3% |
| pme | 265 ns/day | 279 ns/day | +5% |
| apoa1rf | 124.2 ns/day | 131.5 ns/day | +6% |
| apoa1pme | 79.0 ns/day | 84.8 ns/day | +7% |
| apoa1ljpme | 60.5 ns/day | 65.3 ns/day | +8% |
| amoebagk | 1.38 ns/day | 1.40 ns/day | +1% |
| amoebapme | 4.23 ns/day | 4.74 ns/day | +12% |

Most of the other M1 speedup came from staying at 768 threads/core. Previous parameters were either an odd multiple (4x256 = 1024 for forces) or vastly undersubscribed\* (6x64 = 384). As discovered in [metal-benchmarks](https://github.com/philipturner/metal-benchmarks#alu-bottlenecks), the M1 was designed to reach peak ALU utilization at 24 simds/core. This magic number also allocates the most registers possible with the assembly ISA, freeing us from the need to balance register pressure.

> \* Setting 16x256 = 4096 threads/core (forceThreadBlocks) should <i>over</i>subscribe AMD's 2560-thread limit (actually ~500-1000 from register pressure). Is it possible to go lower?

As final evidence of a speedup, the M1 Max jumped from 13 W to 31 W during apoa1rf. Slightly more power-efficient (2.9 ns/day/W -> 3.6 ns/day/W) and climate-friendly, especially for a large-scale project like Folding@Home ;) We can also distribute more work to M1 GPUs and slightly less to power-hungry Nvidia ones.

Resolves #2525, resolves #3847, resolves #3899


philipturner:
Refs: #2489

cc: @tristanic @theschles

peastman:
Very nice!

Can we break this up into a few PRs?  Let's start with setting `simdWidth` on Apple GPUs.  That's a clear improvement that lets it use more efficient algorithms, and it doesn't affect any non-Apple GPUs.  We need to be careful with the other changes.  If you're just working on one GPU, it's easy to tweak settings and make things slightly faster.  But the changes often make it slower on other GPUs, even ones with the same manufacturer and architecture.  Those changes will need thorough testing across a variety of GPUs.

philipturner:
> Let's start with setting `simdWidth` on Apple GPUs.

Could that PR also modify the threadblocks/core on M1 (6 -> 12)? On other processors, it's assigned in the same code section. `forceThreadBlocks` could happen in a separate PR, along with a second attempt to accelerate findInteractingBlocks.

peastman:
Let's keep them separate.  In my testing, the benefit of changing the number of thread blocks is barely above the noise level.  The changes in speed with number of thread blocks are also erratic, which suggests we need to analyze them to figure out what's going on.  For example, the GBSA test gets slower when you increase the number of blocks from 6 to 8, gets still slower when you increase it from 8 to 10, but then gets faster again when you increase it from 10 to 12.  That means we're probably seeing competing effects on the speeds of different kernels.

philipturner:
> For example, the GBSA test gets slower when you increase the number of blocks from 6 to 8, gets still slower when you increase it from 8 to 10, but then gets faster again when you increase it from 10 to 12.

I only ran GBSA tests for one trial (constant ForceBlocks=4, varying general ThreadBlocks), and it never exceeded 1% variation. How much is your M1 Pro varying?

The M1 chip physically runs 768-3072 threads, so the only thing more threads do is boost ALU utilization. With 384 threads, we're actually assigning 768 threads. But making the other half do no work. It's like GCN, but assigning less than one wavefront per quarter-CU.

However if you allocate over 5 KB of threadgroup memory to each block, we would have a problem. It's impossible for 768 threads to coexist; we're oversubscribing.

peastman:
Here are the speeds I measured for GBSA.

|Blocks|ns/day|
|---|---|
|6|318|
|8|316|
|10|307|
|12|316|

philipturner:
The culprit is 10 threadgroups/core. It's generally bad to assign odd numbers of simds. I never tested 7, 9, 10, 11, 14, or 20. Jumping by anything other than multiples of 2 will create uneven workloads between the 4 schedulers.

The scheduler's probably optimized certain prime numbers of simds. ~4-6 simds saturates IMUL32, 8 saturates FFMA32, and 12 and 16 are double those. The next mode is 24. There's little motive for the prime factors 5, 7, 9, etc. You have 20 simds (10 threadgroups), sandwiched between two modes.

Try n=2,3,4,5,6,8,10,12,14,16,20,24 (1 trial) @ 15 seconds. Run 5 trials for 6, 12, 24 @ 15 seconds, then 3 trials @ 30 seconds to ensure the difference is noise. Total 10 minutes of benchmarking. I'll do the same; we should see similar results.

philipturner:
One more revision - can you also run the extended 5+3 trials on 3 threadgroups/core?

philipturner:
GBSA, 2489 atoms, 32 cores, up to 24576-98304 threads

| Threadgroups/Core | Simds/Core | ns/day (15 seconds) |
| ----------------- | ---------- | ------------------- |
| 1 | 2 | 319.07 |
|2 | 4 | 369.53
|3 | 6 | 391.58
|4 | 8 | 364.69
|5 | 10 | 363.48
|6 | 12 | 366.90
|8 | 16 | 365.88
|10 | 20 | 363.57
|12 | 24 | 363.02
|14 | 28 | 365.32
|16 | 32 | 365.52
|20 | 40 | 364.88
|24 | 48 | 364.89

| Extra Trials | 15 s | 15 s | 15 s | 15 s | 30 s | 30 s | 30 s |
| ------------ | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
2 | 374.78 | 368.46 | 367.62 | 367.80 | 368.42 | 367.53 | 367.93
3\* | 377.33 | 378.42 | 382.61 | 377.80 | 377.18 | 377.45 | 377.78
6 | 365.16 | 367.10 | 365.14 | 364.97 | 364.78 | 364.86 | 363.37
10 | 366.27 | 365.62 | 364.41 | 363.33 | 364.32 | 364.15 | 363.88
12 | 363.90 | 364.53 | 364.94 | 364.29 | 363.16 | 365.27 | 363.63
24 | 368.52 | 364.34 | 363.44 | 362.48 | 362.80 | 365.55 | 364.17

\*Multiple theories:

1) 192 simds doesn't divide into 128. The GPU has a 2 schedulers @ dual-dispatch mode (a previous theory), switching between that and a 4 schedulers @ single-dispatch mode. I read somewhere that Volta does this.

2) You're allocating way too much threadgroup memory/threadgroup somewhere. 16-20 KB?

3) Or something much simpler - I had 6144 threads but 2500 atoms. Having less threads will drastically increase the arithmetic intensity, outweighing the harm to theoretical ALU %. Can we make another GBSA with more atoms?

The last would indicate competing interests between kernels. One is O(n) and the other is O(n log(n)). I assume you'd never create something O(n^2).

peastman:
There's more to it than that.  You're assuming the only effect of this number is on the GPU's thread scheduler.  It also affects other things.  For example, more threads mean we have to create larger accumulation buffers, which take longer to reduce.  It also means each thread computes a smaller number of interactions, which adds overhead and impacts cache efficiency.  More thread blocks can also add overhead to launching the kernel.  And so on.

A good rule of thumb is that you want enough threads to cover latencies but no more.  Based on your results, it looks like 3 is about optimal.

philipturner:
> Based on your results, it looks like 3 is about optimal.

That number does _not_ cover the IADD32/FFMA/BITWISE ALU latencies. It might be specific to this simulations's small size/number of GPU cores. It's also not optimal on other benchmarks. Could we keep it at 6 until we run more different benchmarks?

| apoa1rf | 3 threadblocks | 6 threadblocks | 12 threadblocks |
| ------- | -------------- | -------------- | --------------- |
| ns/day | 104.0 | 103.5 | 107.0 (fastest) |

peastman:
> Could we keep it at 6 until we run more different benchmarks?

Yes, that's what I recommend.  Here are the speeds I get for a variety of benchmarks with 3, 6, or 12 thread blocks.

|Benchmark|3|6|12|
|---|---|---|---|
|gbsa|321|318|316|
|rf|220|220|221|
|pme|120|126|129|
|apoa1rf|59.7|58.1|59.2|
|apoa1pme|30.6|32.6|34.1|
|apoa1ljpme|22.8|24.4|25.6|
|amoebagk|0.62|0.53|0.53|
|amoebapme|2.03|2.03|2.13|

The overall pattern seems to be that GB/GK is faster with fewer thread blocks, PME is faster with more thread blocks, and RF mostly doesn't change.

philipturner:
Perhaps that's because the FFT in PME requires more simds/core to hide memory latencies. I don't mean to pester you, but are there plans to replace fft/fftR2C with VkFFT CUDA and OpenCL? It can be just a fast-path for devices where you test it works. Furthermore, embedding it into `master` reduces the complexity of the HIP and Metal plugins.

peastman:
I'd never heard of VkFFT before.  For CUDA we use cufft.

ex-rzr:
@philipturner 

OpenMM-HIP uses VkFFT, here is a reference: https://github.com/amd/openmm-hip/blob/master/platforms/hip/src/HipFFTImplVkFFT.cpp

philipturner:
> I'd never heard of VkFFT before.

VkFFT is a lot faster and not hard to integrate.

peastman:
Sounds like a good thing to investigate.  How about opening a new issue proposing it?

bdenhollander:
> *Setting 16x256 = 4096 threads/core (forceThreadBlocks) should oversubscribe AMD's 2560-thread limit (actually ~500-1000 from register pressure). Is it possible to go lower?

I'm not familiar with a 2560-thread limit. I see significantly more threads listed in Radeon Profiler on my 14 CU RX 6600.

gbsa - default 4x multiplier
image

gbsa - 8x multiplier
image
Utilization goes up but doesn't hit 100% because of 14/16 Occupancy due to LDS usage. The long tail off where the GPU ends up unevenly loaded results in longer kernel execution time and a performance decrease.

gbsa - 16x Multiplier
image
Utilization starts high and then it looks more like 4x multiplier with a short tail. Performance is better than 8x but worse than 4x.

apoa1rf - default 4x multiplier
image

apoa1rf - 8x multiplier
image
Utilization goes to 100% and the decreasing tail is about 1/2 of the execution time. Performance improves by about 15% compared to 4x.

apoal1rf - 16x multiplier
image
Utilization goes to 100% for about 1/3 of time, nearly 100% for another 1/3, and then trails off more sharply. Performance improves by about 20% compared to 4x.

philipturner:
> I'm not familiar with a 2560-thread limit.

I meant 2560 threads per CU. When I speak threads, SIMDs, IPC, etc, it's always per-GPU core. That means CU on both GCN and RDNA, not WGP (an attempt to portray building blocks as twice as powerful as they actually are). Whatever contains ~300 KB of register file and ~70 KB of shared memory, that's a GPU core.

bdenhollander:
> I meant 2560 threads per CU. When I speak threads, SIMDs, IPC, etc, it's always per-GPU core. That means CU on both GCN and RDNA, not WGP (an attempt to portray building blocks as twice as powerful as they actually are). Whatever contains ~300 KB of register file and ~70 KB of shared memory, that's a GPU core.

By that metric, my GPU would be considered 28 CUs so max threads would be 28x2560 = 71680. It would take a 20x multiplier to produce 2240 wavefronts and 71680 threads.

philipturner:
> By that metric, my GPU would be considered 28 CUs so max threads would be 28x2560 = 71680. It would take a 20x multiplier to produce 2240 wavefronts and 71680 threads.

20 x 28 x 256 = 143360

bdenhollander:
> > By that metric, my GPU would be considered 28 CUs so max threads would be 28x2560 = 71680. It would take a 20x multiplier to produce 2240 wavefronts and 71680 threads.
> 
> 20 x 28 x 256 = 143360

RDNA reports the [# of WGPs](https://stackoverflow.com/questions/63975762/opencl-reports-half-the-expected-compute-units) when queried for `CL_DEVICE_MAX_COMPUTE_UNITS`.

For this line of code it would be 20 x 14 = 280 and then later on x 256 = 71680.
```C++
numForceThreadBlocks = 20*context.getDevice().getInfo<CL_DEVICE_MAX_COMPUTE_UNITS>();
```

philipturner:
> RDNA reports the [# of WGPs](https://stackoverflow.com/questions/63975762/opencl-reports-half-the-expected-compute-units) when queried for CL_DEVICE_MAX_COMPUTE_UNITS.

Then with 16, you'd actually have 8 workgroups per CU which is 2048/2560 threads. That's not oversubscribing. Unless you are assigning >128 bytes of register memory per thread (32 floats). On the Apple GPU, that's the minimum you can allocate (3072 threads/core) while the maximum is 512 B (768 threads/core).

Regarding threadgroup memory, you shouldn't allocate over 8 KB per 256-threadgroup. As long as no kernels exceed that, you aren't oversubscribing.

bdenhollander:
gbsa - 20x mulitplier
image
Utilization starts high and then stays above 50% with a short tail. Performance on `computeNonbonded` is better than 16x but worse than 4x. `computeBornSum` and `computeGBSAForce1` improve compared to 4x so overall ns/day is better than 4x.

apoa1rf - 20x multiplier
image
Performance is quite similar to 16x. The area where chart is at 100% is when the initial 896 wavefronts are in progress and then it gets bumpier as those finish and new ones start.

CoryKornowicz:
All of the mixed and double precision tests fail on the latest m1-tuning build. I understand the metal kernel cannot handle anything larger than float32 (unless you craft a unique solution), but will the older implementations still stick around to enable that option?

philipturner:
> All of the mixed and double precision

We're not removing any older implementations here. M1 GPU lacks hardware FP64, AMD GPU has native FP64.

peastman:
When you run `ccmake`, set `OPENMM_BUILD_OPENCL_DOUBLE_PRECISION_TESTS` to `OFF`.  That will tell it not to build and run any tests that require double precision to be supported.

philipturner:
I'm closing the PR because it has been superseded. I made a lot of progress since then! This PR increased `apoa1rf` by +70 ns/day (42 -> 112), further work added another +40 ns/day (112 -> 151.7).

PR #3914, opened 2023-01-17
Fix SIMD width and force thread blocks

philipturner:
Cherry picked optimizations from https://github.com/openmm/openmm/pull/3911 which are strongly confirmed by evidence.

#### Fix SIMD width

Set the proper SIMD width, which boosts apoa1rf performance from 38 ns/day to 103 ns/day (+171%).

#### Fix force thread blocks

Changing force thread blocks only increases performance in every benchmark. There are twin maxima at 768 and 1536 threads/core, except GBSA (only 768 threads/core). This boosts apoa1rf to 109 ns/day (+6%).

[OpenMM ForceThreadBlocks Tests.xlsx](https://github.com/openmm/openmm/files/10431103/OpenMM.ForceThreadBlocks.Tests.xlsx)

#### Not implemented

apoa1rf could reach 113 ns/day (+4%) if we change generic thread blocks from 384 to 768 threads/core. That doesn't harm GBSA, but GBSA's maximum is at 192 threads/core. We defer dealing with that disparity to another PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 __pycache__
 build
 build?
+
+# macOS
+.DS_Store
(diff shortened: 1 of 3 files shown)


peastman:
As discussed in #3911, let's remove all the changes relating to the number of thread blocks.  Make this a clean PR that just sets `simdWidth` and nothing else.  The other changes aren't directly related, and they need investigation to understand the effect.

philipturner:
Re-opening a new PR, as I'd like to keep the commits and cited evidence intact.

PR #3917, opened 2023-01-17
[macOS GPU Support] Fix SIMD width

philipturner:
See #3914

### DIFF
--- a/platforms/opencl/src/OpenCLContext.cpp
+++ b/platforms/opencl/src/OpenCLContext.cpp
@@ -211,7 +211,11 @@ OpenCLContext::OpenCLContext(const System& system, int platformIndex, int device
             throw OpenMMException("This device does not support double precision");
         string vendor = device.getInfo<CL_DEVICE_VENDOR>();
         int numThreadBlocksPerComputeUnit = 6;
-        if (vendor.size() >= 6 && vendor.substr(0, 6) == "NVIDIA") {
+      
+        if (vendor.size() >= 5 && vendor.substr(0, 5) == "Apple") {
+            simdWidth = 32;
+        }
+        else if (vendor.size() >= 6 && vendor.substr(0, 6) == "NVIDIA") {
             compilationDefines["WARPS_ARE_ATOMIC"] = "";
             simdWidth = 32;
             if (device.getInfo<CL_DEVICE_EXTENSIONS>().find("cl_nv_device_attribute_query") != string::npos) {


peastman:
With this change I'm getting a test failure in `TestOpenCLAmoebaVdwForce`.  However when I run that test without this change, I still occasionally get the failure.  The difference is just that it happens every time with this change.  I suspect the root cause is something unrelated.  I'll investigate.

platforms/opencl/src/OpenCLContext.cpp
```diff
@@ -211,7 +211,11 @@ OpenCLContext::OpenCLContext(const System& system, int platformIndex, int device
             throw OpenMMException("This device does not support double precision");
         string vendor = device.getInfo<CL_DEVICE_VENDOR>();
         int numThreadBlocksPerComputeUnit = 6;
-        if (vendor.size() >= 6 && vendor.substr(0, 6) == "NVIDIA") {
+      
+        if (vendor.size() >= 5 && vendor.substr(0, 5) == "Apple") {
+            simdWidth = 32;
```
bdenhollander, line 216: Should this do a CL_DEVICE_TYPE check or is SIMD width for Apple CPUs and GPUs both 32? Does Apple expose its CPUs as OpenCL devices?

peastman: Apple does expose the CPU as an OpenCL device, but the support is so primitive that it's basically useless.  For that reason we just skip that device:

https://github.com/openmm/openmm/blob/e1c4e1f8ee3e574f582b107412c7f9b595610012/platforms/opencl/src/OpenCLContext.cpp#L136-L137

But you're right, we should add a check for the device type here too, just in case Apple ever makes it useful.

philipturner: No Apple silicon does not expose the CPU as an OpenCL device. x86 Macs do because they have an Intel CPU.
```
NULL platform behavior
  clGetPlatformInfo(NULL, CL_PLATFORM_NAME, ...)  Apple
  clGetDeviceIDs(NULL, CL_DEVICE_TYPE_ALL, ...)   Success [P0]
  clCreateContext(NULL, ...) [default]            Success [P0]
  clCreateContextFromType(NULL, CL_DEVICE_TYPE_DEFAULT)  Success (1)
    Platform Name                                 Apple
    Device Name                                   Apple M1 Max
================================================================================
  clCreateContextFromType(NULL, CL_DEVICE_TYPE_CPU)  No devices found in platform <-- Here
================================================================================
  clCreateContextFromType(NULL, CL_DEVICE_TYPE_GPU)  Success (1)
    Platform Name                                 Apple
    Device Name                                   Apple M1 Max
  clCreateContextFromType(NULL, CL_DEVICE_TYPE_ACCELERATOR)  No devices found in platform
  clCreateContextFromType(NULL, CL_DEVICE_TYPE_CUSTOM)  Invalid device type for platform
  clCreateContextFromType(NULL, CL_DEVICE_TYPE_ALL)  Success (1)
    Platform Name                                 Apple
    Device Name                                   Apple M1 Max

```

peastman: Ok, that's fine then.

peastman:
Thanks!

Issue #3916, opened 2023-01-17
.DS_Store files automatically saved to the Git source tree


philipturner:
We can edit the `.gitignore` to prevent this, as well as other common intermediate files, from appearing. I frequently have to revert commits and manually remove .DS_Store files.

```
__pycache__
build
build?

# macOS
.DS_Store
```

peastman:
That would be reasonable.

peastman:
This was fixed by #3918.

PR #3918, opened 2023-01-17
[macOS Support] Add .DS_Store

philipturner:
Refs: https://github.com/openmm/openmm/issues/3916

Specific files for other operating systems can come later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 __pycache__
 build
 build?
+
+# macOS
+.DS_Store


peastman:
Looks good.  OpenMM 8.0 is in release candidate, so we're currently only merging critical bug fixes.  I'll merge this as soon as 8.0 is out.

philipturner:
> only merging critical bug fixes.

I assume that includes setting correct SIMD width on Mac? I'd be disappointed otherwise. Setting correct force threads could wait.

peastman:
Critical bug fixes means things that corrupt data, crash, or produce incorrect results.  Optimizations need to wait for 8.1.  As long as no critical bugs are found, the current release candidate will become the release in about a week.

peastman:
Thanks!

PR #3928, opened 2023-01-23
fix: py Interface omit errors in xml attributes.

### LINKED ISSUE #3897: (bug&suggestion) py Interface parsing xml files omit some errors to incorrect xml attributes
**Dear maintainers,**

This is not a critical problem, more like a suggestion.

## where it is
I noticed that in the forcefield file wrapping of the Python interface, specifically in the `AmoebaMultipoleGenerator` parsing method, there is the following code:

https://github.com/openmm/openmm/blob/4c2f3323fb7ece01e36aa16f0471b9543926e881/wrappers/python/openmm/app/forcefield.py#L4970-L4974

As comment says(Line 4931), these lines is using for set the value to 0 if the element is missing in one of kz, kx, ky. But I'm confused about the current using of `try ... except ...` which catches all exceptions. 

## problems
In some special cases (e.g. some xml data filled in manually), such as when `xml` provides `kz='-1.0'` or typo like `kz='l'`, although human might understand the meaning and know that such a definition is wrong, the python will ignore it and without throwing any errors, and most important, it set a 0.

## suggestions
It is therefore sincerely recommended that developers consider using :
`kIndices.append(int(atom.attrib.get(kString,0)))`, and raise any other exceptions to reminder users.

This may not be a critical problem, but I strongly recommend that communities check other similar `try ... except ...` statements in Python wrappers to ensure that users can know possible problems in the xml.

saltball:
As described in #3897 , this pr aims to avoid the potential problem of over-catching errors to omit errors when parsing the xml files of the Amoeba forcefield.

I'm unsure if I need to add unit tests related to the incorrect xml node attributes. Because there are so many possibilities for errors, I suspect that we probably don't need unit tests to ensure that error raises properly.

This will close #3897 .

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -3495,14 +3495,11 @@ class AmoebaAngleGenerator(object):
                 angleList = []
                 angleList.append(float(angle.attrib['angle1']))
 
-                try:
+                if 'angle2' in angle.attrib:
                     angleList.append(float(angle.attrib['angle2']))
-                    try:
-                        angleList.append(float(angle.attrib['angle3']))
-                    except:
-                        pass
-                except:
-                    pass
+                if 'angle3' in angle.attrib:
+                    angleList.append(float(angle.attrib['angle3']))
+
                 generator.angle.append(angleList)
                 generator.k.append(float(angle.attrib['k']))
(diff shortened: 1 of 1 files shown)


peastman:
This looks good.  I'll merge it once 8.0 is released, hopefully around the end of this week.

peastman:
Thanks!

Issue #3925, opened 2023-01-21
Add support for writing atom subsets to reporters?


jchodera:
It would be useful to enable the [OpenMM `Simulation` reporters](http://docs.openmm.org/latest/api-python/app.html#reporting-output) that output PDB, PDBx, or DCD trajectories to write only atom subsets. Often, in explicit solvent simulations, only the positions of a subset of atoms (such as the solute) are of interest.

peastman:
What should the API for this look like?  Presumably you would specify the subset when you create the reporter.  A list of particle indices?  Or Atom objects?

jchodera:
I propose we adopt the [same API as MDTraj reporters that support this](https://www.mdtraj.org/1.9.8.dev0/api/generated/mdtraj.reporters.DCDReporter.html#mdtraj.reporters.DCDReporter), and include an optional `atomSubset` keyword argument that takes an array-like object specifying the zero-indexed atom index subset to write, defaulting to `None`.

peastman:
This was implemented in #3931.

jgreener64:
Are there any plans to allow this for `DCDReporter`? I might be able to have a go at implementing it.

peastman:
Sure, go ahead!

sef43:
Worth noting that the MDTraj DCD reporter does have this feature: https://www.mdtraj.org/1.9.8.dev0/api/generated/mdtraj.reporters.DCDReporter.html#mdtraj.reporters.DCDReporter

jgreener64:
> Worth noting that the MDTraj DCD reporter does have this feature

Thanks for pointing that out. It doesn't seem to have the ability to append though, which would be useful in my case (https://github.com/mdtraj/mdtraj/issues/441).

PR #3931, opened 2023-01-25
Add support for writing atom subsets to PDBReporter

sef43:
This is a first attempt at a way to output atom subsets with PDBReporter addressing issue #3925
It creates a subset topology and position list with just the specified atoms, similar to [MDTraj](https://www.mdtraj.org/1.9.8.dev0/api/generated/mdtraj.reporters.DCDReporter.html#mdtraj.reporters.DCDReporter).
It then passes the new topology and positions to PDBFile.

### DIFF
--- a/wrappers/python/openmm/app/pdbreporter.py
+++ b/wrappers/python/openmm/app/pdbreporter.py
@@ -33,7 +33,7 @@ __author__ = "Peter Eastman"
 __version__ = "1.0"
 
 from openmm.app import PDBFile, PDBxFile, Topology
-from openmm.unit import angstroms, Quantity
+from openmm.unit import angstroms
 
 class PDBReporter(object):
     """PDBReporter outputs a series of frames from a Simulation to a PDB file.
@@ -99,21 +99,6 @@ class PDBReporter(object):
         """
         if self._atomSubset is not None:
             if self._subsetTopology is None:
-                # check atomSubset is valid 
-                # TODO: move to constructor
-                if len(self._atomSubset)==0:
-                    raise ValueError('atomSubset cannot be an empty list')
-                if not all(a==int(a) for a in self._atomSubset):
-                    raise ValueError('all of the indices in atomSubset must be integers')
-                if len(set(self._atomSubset)) != len(self._atomSubset):
-                    raise ValueError('atomSubset must contain unique indices')
-                if sorted(self._atomSubset) != self._atomSubset:
-                    raise ValueError('atomSubset must be sorted in ascending order')
-                if self._atomSubset[0] < 0:
-                    raise ValueError('The smallest allowed value in atomSubset is zero')
-                if self._atomSubset[-1] >= simulation.topology.getNumAtoms():
-                    raise ValueError('The maximum allowed value in atomSubset must be less than the total number of particles')
-                
                 self._createSubsetTopology(simulation.topology)
 
             topology = self._subsetTopology
@@ -143,6 +128,21 @@ class PDBReporter(object):
         topology : Topology
             The Topology to create a subset from
         """
+        # check atomSubset is valid 
+        if len(self._atomSubset)==0:
+            raise ValueError('atomSubset cannot be an empty list')
+        if not all(a==int(a) for a in self._atomSubset):
+            raise ValueError('all of the indices in atomSubset must be integers')
+        if len(set(self._atomSubset)) != len(self._atomSubset):
+            raise ValueError('atomSubset must contain unique indices')
+        if sorted(self._atomSubset) != self._atomSubset:
+            raise ValueError('atomSubset must be sorted in ascending order')
+        if self._atomSubset[0] < 0:
+            raise ValueError('The smallest allowed value in atomSubset is zero')
+        if self._atomSubset[-1] >= topology.getNumAtoms():
+            raise ValueError('The maximum allowed value in atomSubset must be less than the total number of particles')
+        
+
         self._subsetTopology = Topology()
         
         # convert to set for fast look up
@@ -158,9 +158,10 @@ class PDBReporter(object):
                         if posIndex in atomSubsetSet:
                             atom = self._subsetTopology.addAtom(atom.name, atom.element, r, atom.id)
                         posIndex += 1
+
+        self._subsetTopology.setPeriodicBoxVectors(topology.getPeriodicBoxVectors())
+
         
-
-
     def __del__(self):
         if self._topology is not None:
             PDBFile.writeFooter(self._topology, self._out)
@@ -182,10 +183,25 @@ class PDBxReporter(PDBReporter):
         state : State
             The current state of the simulation
         """
+
+        if self._atomSubset is not None:
+            if self._subsetTopology is None:
+                self._createSubsetTopology(simulation.topology)
+
+            topology = self._subsetTopology
+
+            #PDBFile will convert to angstroms so do it here first instead
+            positions = state.getPositions().value_in_unit(angstroms) 
+            positions = [positions[i] for i in self._atomSubset]
+
+        else:
+            topology = simulation.topology
+            positions = state.getPositions()
+
         if self._nextModel == 0:
-            PDBxFile.writeHeader(simulation.topology, self._out)
+            PDBxFile.writeHeader(topology, self._out)
             self._nextModel += 1
-        PDBxFile.writeModel(simulation.topology, state.getPositions(), self._out, self._nextModel)
+        PDBxFile.writeModel(topology, positions, self._out, self._nextModel)
         self._nextModel += 1
         if hasattr(self._out, 'flush') and callable(self._out.flush):
             self._out.flush()
--- a/wrappers/python/openmm/app/pdbreporter.py
+++ b/wrappers/python/openmm/app/pdbreporter.py
@@ -130,16 +130,22 @@ class PDBReporter(object):
         """
         # check atomSubset is valid 
         if len(self._atomSubset) == 0:
+            self._out.close()
             raise ValueError('atomSubset cannot be an empty list')
         if not all(a == int(a) for a in self._atomSubset):
+            self._out.close()
             raise ValueError('all of the indices in atomSubset must be integers')
         if len(set(self._atomSubset)) != len(self._atomSubset):
+            self._out.close()
             raise ValueError('atomSubset must contain unique indices')
         if sorted(self._atomSubset) != self._atomSubset:
+            self._out.close()
             raise ValueError('atomSubset must be sorted in ascending order')
         if self._atomSubset[0] < 0:
+            self._out.close()
             raise ValueError('The smallest allowed value in atomSubset is zero')
         if self._atomSubset[-1] >= topology.getNumAtoms():
+            self._out.close()
             raise ValueError('The maximum allowed value in atomSubset must be less than the total number of particles')
         
         self._subsetTopology = Topology()
--- a/wrappers/python/openmm/app/pdbreporter.py
+++ b/wrappers/python/openmm/app/pdbreporter.py
@@ -32,7 +32,8 @@ from __future__ import absolute_import
 __author__ = "Peter Eastman"
 __version__ = "1.0"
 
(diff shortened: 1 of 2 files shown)


peastman:
Thanks!  This looks like a good start.  We can make it more efficient in a few ways.

There's no need to build a new Topology every time `report()` is called.  You can create it only once and then reuse it.  In the constructor set

```python
self._subsetTopology = None
```

Then in `report` you can have

```python
if self._subsetTopology is None:
    self._createSubsetTopology(simulation.topology)
```

Notice that I've also simplified the API a little.  There's no need to pass the subset to `_createSubsetTopology()` or get a return value from it.  If you make it an instance method, it already has access to `self._atomSubset`, and it can store the result into the correct field.

I'm not sure it's worth having a method to subset the positions at all.  You add about 20 lines to the file to avoid repeating a single line.  Which at the moment isn't even repeated, though PDBxReporter will need to do the same thing.

Removing units from each element of a list individually is very expensive.  It's much faster to remove them from the whole list at once.  We can also save time by changing units only once.  `state.getPositions()` returns positions in nanometers.  You strip off the units, then create a new quantity which again has units of nanometers.  It gets passed to PDBFile, which has to convert them to angstroms.  This will be faster:

```python
positions = positions.value_in_unit(angstroms)
positions = [positions[i] for i in self._atomSubset]
```

Since the positions passed to PDBFile are not Quantities, it will just write them out without attempting any unit conversion.

peastman:
I noticed another problem: if you pass a list for `atomSubset` that is not sorted, you'll get wrong results.  The topology subset will have the atoms in the original order, while the position subset will have them in the order of `atomSubset`.  We probably should require them to be in order, and throw an exception if they aren't.  Alternatively we could sort the list, but it's probably better to throw an exception.  An out of order list is likely to indicate a user error.

All the checks on the list of atoms can be done in the constructor.  They don't need to be repeated every time `report()` is called.

This line will make `_subsetTopology()` very slow:

```python
if posIndex in atomSubset:
```

Determining whether a value appears in a list requires a full scan through the list.  Since it gets repeated for every atom, the cost will be O(n^2) in the number of atoms.  If you convert it to a set first, that will make it much faster.

sef43:
Thanks for the feedback @peastman, I have done most of the changes. I have not yet moved the checks on `atomSubset` to the constructor, this is because I don't see how to check the max number of atoms in the simulation here? I have kept the checks all in the same place at the moment.

peastman:
True, we don't know the number of atoms yet there.  How about putting the checks in `_createSubsetTopology()`?  We have the information we need there, and they'll only be run once.

sef43:
Do we need to deal with creating a subset of the bond information? PDBReporter prints out the non standard bonds as connect records in the footer, does it make sense to still do this if a subset of atoms are output?

peastman:
Yes, we need to subset the bonds correctly.  Otherwise you'll get a PDB file with invalid CONECT records.

wrappers/python/openmm/app/pdbreporter.py
```diff
@@ -91,15 +97,81 @@ def report(self, simulation, state):
...
+
+        # store a map from posIndex to Atom object for when we add the bonds
+        indexToAtom = {}
+
+        posIndex = 0
+        for chain in topology.chains():
+            c = self._subsetTopology.addChain(chain.id)
+            residues = list(chain.residues())
+            for res in residues:
+                r = self._subsetTopology.addResidue(res.name,c,res.id,res.insertionCode)
```
peastman, line 159: Can you make the formatting consistent?  Spaces after the commas.  There are also a few places where you don't put spaces around `=` in assignments.

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,204 @@
...
+        modeller.addSolvent(ff, padding=1.0*unit.nanometer)
+
+        system = ff.createSystem(modeller.topology)
+        
+        with tempfile.TemporaryDirectory() as tempdir:
+            filename = os.path.join(tempdir, 'temptraj.pdb')
+            simulation = app.Simulation(modeller.topology, system, mm.LangevinMiddleIntegrator(1.0*unit.kelvin, 1.0/unit.picosecond, 0.0000001*unit.picoseconds))
+            simulation.context.setPositions(modeller.positions)
+
+            simulation.minimizeEnergy(maxIterations=100)
```
peastman, line 81: Why the energy minimization?

sef43:
Made the changes, I have removed the minimization from the tests, this was left over from my manual testing. I just chose a PDB file in the test dataset that had connect records

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # just the alanine-dipeptide atoms
+            subset = list(range(0,22))
+
+            simulation.reporters.append(app.PDBReporter(filename, 10, atomSubset=subset))
+            simulation.step(10)
+
+            # check if the output out trajectory contains the expected number of atoms
+            checkpdb = app.PDBFile(filename)
+            self.assertEqual(len(checkpdb.positions), len(subset))
```
peastman, line 30: Can we do a more thorough check?  This verifies how many atoms it wrote, but not that they're the correct ones.  Can we check that the atom names, elements, and positions match what we expect?

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,197 @@
...
+                simulation.context.setPositions(self.pdb.positions)
+
+                simulation.reporters.append(app.PDBReporter(filename, 10, atomSubset=subset))
+                
+                self.assertRaises(ValueError, lambda: simulation.step(10))
+    
+    def testBondSubset(self):
+        """ Test that CONNECT records are output correctly when using atomSubset"""
+
+        # use a testcase that has CONNECT records in the input pdb file
```
peastman, line 65: It's spelled CONECT.  Because PDB files only have six columns for the record name.  Modern technology!

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # just the alanine-dipeptide atoms
+            subset = list(range(0,22))
+
+            simulation.reporters.append(app.PDBxReporter(filename, 10, atomSubset=subset))
+            simulation.step(10)
+
+            # check if the output out trajectory contains the expected number of atoms
+            checkpdb = app.PDBxFile(filename)
+            self.assertEqual(len(checkpdb.positions), len(subset))
```
peastman, line 126: Same here.

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,197 @@
...
+
+                simulation.reporters.append(app.PDBxReporter(filename, 10, atomSubset=subset))
+                
+                self.assertRaises(ValueError, lambda: simulation.step(10))
+
+
+    def testBondSubset(self):
+        """ Test that CONNECT records are output correctly when using atomSubset"""
+
+        # use a testcase that has CONNECT records in the input pdb file
```
peastman, line 161: In PDBx/mmCIF files, the record name for bonds is `struct_conn`.

peastman (commented):
Looks good!  I made a few final suggestions, mainly to improve a couple of the test cases.  Then it should be ready to merge.

peastman:
On the Pypy build one of the new test cases failed:

> FAILED TestPdbReporter.py::TestPDBReporter::testBondSubset - AssertionError: 130 != 20

sef43:
> On the Pypy build one of the new test cases failed:
> 
> > FAILED TestPdbReporter.py::TestPDBReporter::testBondSubset - AssertionError: 130 != 20

I think this is failing because when using  Pypy the `__del__` method of PDBReporter:
```
    def __del__(self):
        if self._topology is not None:
            PDBFile.writeFooter(self._topology, self._out)
        self._out.close()
```
which contains the code to write the footer with the bond info does not get called when the reporter is deleted:
```
            # clear reporters to ensure PDBReporter calls writeFooter and file.close
            simulation.reporters.clear()
            
            # check if the output out trajectory contains the expected number of atoms
            checkpdb = app.PDBFile(filename)
```
So when the pdb file is opened again the bond information has not been written yet. With normal (cpython) python it appears the `__del__ `does get called in time. I am not sure if that is guaranteed behaviour. On my local PyPy build i can use gc.collect() to force the full deletion and calling of `__del__` before the file is opened again. I am also not sure if that is guaranteed behaviour.

peastman:
It seems this is caused by the [different garbage collection methods](https://doc.pypy.org/en/latest/cpython_differences.html).  CPython uses reference counting, so `__del__()` is called as soon as you delete the last reference to it.  Pypy does not use reference counting, so `__del__()` doesn't get run until the next time it does a garbage collection.  Given that, calling `gc.collect()` before trying to read the file is probably the best solution.  It's not very clean, but I don't have a better solution.

sef43:
@peastman I have now done all the changes, the tests are more thorough, gc.collect should enable the pypy build tests to pass. I am still unsure if the Windows tests will pass, there were some issues with the file opening/closing when writing the test trajectories, will know when the CI runs

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,307 @@
...
+
+            # check if the output out trajectory contains the expected number of atoms
+            checkpdb = app.PDBxFile(filename)
+            self.assertEqual(len(checkpdb.positions), len(subset))
+
+            # check the positions are correct
+            validPositions  = [simulation.context.getState(getPositions=True).getPositions()[i] for i in subset]
+            # round to 5 decimal places before comparing
+            for i in range(len(validPositions)):
+                validPositions[i] = [round(validPositions[i][j]._value, 5) for j in (0, 1, 2)]*unit.nanometer
```
peastman, line 192: Does this actually make a difference?  `assertVecAlmostEqual()` ought to work just as well without rounding first.

wrappers/python/tests/TestPdbReporter.py
```diff
@@ -0,0 +1,307 @@
...
+            validBonds = list(validpdb.topology.bonds())
+            testBonds = list(testpdb.topology.bonds())
+            
+            self.assertEqual(len(validBonds), len(testBonds))
+
+            for validBond, testBond in zip(validBonds, testBonds):
+                self.assertEqual(str(validBond), str(testBond))
+
+
+    def assertVecAlmostEqual(self, p1, p2, tol=1e-7):
```
peastman, line 295: If you make this a function rather than a method, you won't need to repeat it.

peastman:
It's getting test failures on Windows:

> E                   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp5axilklo\\temptraj6.pdb'

and

> FAILED TestPdbReporter.py::TestPDBReporter::testInvalidSubsets - NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp5axilklo\\temptraj6.pdb'

wrappers/python/tests/TestPdbReporter.py
sef43: If I don't round the position to 4 decimal places for comparing to PDB input and 5 decimal places for PDBx then the assertion fails unless I start reducing the tolerance in assertVecAlmostEqual() to 1e-4 for PDB and 1e-5 for PDBx.
e.g. (1.00001,1.00001,1.00001)nm will be written to PDB file as (10.000, 10.000, 10.000) angstroms, which will fail in assertVecAlmostEqual(), the diff is 5.77e-5. 

Given that we have set the precision we are writing the PDB and PDBx files with, `%8.3f` and %10.4f` Angstrom, I think it is clearer to just round the more precise values from getState() before comparing them

peastman: I see, makes sense.

sef43:
> That seems to have fixed the tests. Is this ready to merge?

Yes it is

peastman:
Thanks!  This should be a useful new feature.

PR #3934, opened 2023-01-27
Use VkFFT for OpenCL

### LINKED ISSUE #3915: VkFFT integration for optimizing PME
Spinoff of #3911 - [VkFFT](https://github.com/DTolm/VkFFT) is a very fast FFT library, designed to be portable and faster than cuFFT. It started on Vulkan, then expanded to CUDA, OpenCL, and all the major GPU APIs. A very recent addition was Metal.

It's heavily tested and I'd be surprised if it breaks old hardware. However, it's a good idea to only run it on tested hardware. We can first run it only on Apple GPUs - Apple7 and Apple8 are practically identical. Then slowly expand to other vendors and architectures within vendors.

VkFFT is a header-only library; most people just copy the 40,000-line header verbatim. A separate idea, is that since HIP and Metal plugins use VkFFT, why not eliminate the code duplication? Make it part of the common platform. Of course backends can still copy VkFFT internally to use the most recent commits.

peastman:
Fixes #3915.

VkFFT is faster than the FFT implementation we currently use on OpenCL.  I tested the apoa1pme benchmark on a couple of GPUs.  On RTX 4080 this produces about a 10% speedup.  On M1 Pro it produces about a 5% speedup.

If someone has an AMD GPU they can test on, that would be great.  I have one, but OpenCL on that machine is currently unable to recognize the GPU.  I'm not sure why or how soon I'll be able to get it working.

I also tried using it for CUDA in place of cuFFT, but that made it slightly slower.


bdenhollander:
Here are results for AMD Radeon RX 6600 on Windows 10 with Python 3.8.15.
```
python benchmark.py --platform OpenCL --precision single --seconds 30
```
| Test        | OpenMM 8.0 ns/day | VkFFT ns/day | Change |
|-------------|-----------:|-------------:|-------:|
| gbsa        | 406.407    | 405.929      | -      |
| rf          | 430.722    | 430.964      | -      |
| pme         | 254.532    | 294.766      | 15.81% |
| apoa1rf     | 117.976    | 117.486      | -      |
| apoa1pme    | 73.6291    | 79.3003      | 7.70%  |
| apoa1ljpme  | 57.4547    | 61.9032      | 7.74%  |
| amoebagk    | 1.39495    | 1.397        | -      |
| amoebapme   | 3.66365    | crash        | crash  |

Crash info logged to Event Viewer when attempting to run amoebapme.
```
Faulting application name: python.exe, version: 3.8.15150.1013, time stamp: 0x637c8bcb
Faulting module name: ntdll.dll, version: 10.0.19041.2130, time stamp: 0xb5ced1c6
Exception code: 0xc0000374
Fault offset: 0x00000000000ff6a9
Faulting process id: 0x894
Faulting application start time: 0x01d932b8347bd72a
Faulting application path: C:\Users\[redacted]\.conda\envs\openmm-8.0\python.exe
Faulting module path: C:\WINDOWS\SYSTEM32\ntdll.dll
Report Id: ab011106-c532-4ce9-9db3-12589d64dbcb
Faulting package full name: 
Faulting package-relative application ID: 
```

peastman:
Thanks!  It's a nice improvement except for the crash...

bdenhollander:
> Thanks! It's a nice improvement except for the crash...

Sorry for the false alarm. I neglected to recompile `OpenMMAmoebaOpenCL` so it was crashing when trying to call `getReducedNumThreadBlocks()` that I added in https://github.com/openmm/openmm/commit/19b10b3ef1930c94d75565238101b4cd00e2dcdd. The chart above has been updated with amoebapme results.

bdenhollander:
Additional AMD Radeon RX 6600 data from FAH projects. Fantastic to see 0 regressions!

| FAH Project | Particles | Open 8.0 ns/day | VkFFT ns/day | Change |
|------------:|----------:|----------------:|-------------:|-------:|
| 13415       | 4038      | 126.7772        | 131.9623     | 4.09%  |
| 13428       | 90729     | 9.6257          | 10.0998      | 4.93%  |
| 14908       | 215357    | 16.9580         | 17.7235      | 4.51%  |
| 14944       | 105522    | 40.8515         | 46.8800      | 14.76% |
| 16918       | 65523     | 47.6549         | 52.1742      | 9.48%  |
| 17308       | 158384    | 45.0229         | 47.5445      | 5.60%  |
| 17424       | 167028    | 44.6111         | 47.5445      | 6.58%  |
| 18717       | 77795     | 34.8628         | 37.4136      | 7.32%  |

egallicc:
For reference, here are the 8.0.0rc1 OpenCL benchmarks compared to those of 8.0.0rc1 with the latest [HIP plugin](https://github.com/StreamHPC/openmm-hip). These are with an RX 6750 XT AMD card on Linux:

Test       |     Precision |  Constraints |  Hmass(amu) |  dt(fs)  |  OpenCL(ns/day)  |   HIP(ns/day) |  Ratio
--         |    --         |      --      |    --       |    --    |      --          |     --        |  --
gbsa       |       single  |    HBonds    |    1.5      |      4   |           565    |          1293 | 229% 
rf         |       single  |    HBonds    |    1.5      |      4   |           152    |          1220 | 803% 
pme        |       single  |    HBonds    |    1.5      |      4   |           145    |          877	 | 605% 
apoa1rf    |       single  |    HBonds    |    1.5      |      4   |           52.6   |          380	 | 722% 
apoa1pme   |       single  |    HBonds    |    1.5      |      4   |           52.8   |          263	 | 498% 
apoa1ljpme |       single  |    HBonds    |    1.5      |      4   |           60.7   |          210	 | 346% 
amoebagk   |       single  |    None      |    1        |      2   |           1.46   |          26.2 | 1795%
amoebapme  |       single  |    None      |    1        |      2   |           5.72   |          10.5 | 184%

I think the latest HIP plugin uses VkFFT. I assume 8.0.0rc1 OpenCL does not. I have yet to attempt to benchmark the trunk version. In any case, as you can see, HIP is vastly faster than OpenCL on these cards. (Why?)

Are there plans to integrate HIP into the main codebase?

peastman:
Some of those differences are pretty extreme.  All the ones I've seen in the past show about a 2x speed difference.  I don't know what the reason is for it.  I've generally suspected that AMD's OpenCL implementation is just badly written and does a very bad job of compliing the kernels.

There are legal issues that complicate distributing HIP code.  The only way to get either the runtime libraries or the development kit is to download it from their website, which includes a step where you read and agree to their license.    This means, for example, that it's impossible to build HIP code on conda-forge, because there's no legal way to install the development kit while building.  We've spoken to some people at AMD about this and they were going to talk to their legal department about finding a solution.  I don't know whether anything is happening about it.

egallicc:
We ran the benchmark suite using the default settings. We'll try to run replicates to confirm those extreme speed differences are real on this hardware. They are indeed hard to believe.

I built the openmm-hip plugin in a docker image based on conda-forge's linux-anvil-cos7-cuda image with the additional Rocm drivers installed as so:
```
RUN yum -y install epel-release
RUN yum -y repolist
RUN yum -y install https://repo.radeon.com/amdgpu-install/21.50.2/rhel/7.9/amdgpu-install-21.50.2.50002-1.el7.noarch.rpm
RUN yum -y install rocm-device-libs hip-devel hip-runtime-amd rocfft-devel hipfft-devel
```
I did not see an agreement popping up but the docker build output went by quickly and I could have missed it. 

Apparently, StreamHPC distributes OpenMM with the HIP plugin through [their conda channel](https://github.com/StreamHPC/openmm-hip#readme). So not a big deal if their channel is kept up to date.

I am not sure if the StreamHPC version supports CUDA. (The version I built from sources supports both CUDA and HIP--it is strangely satisfying to run replica exchange across AMD/HIP and NVIDIA/CUDA GPUs on the same machine.)  In any case, if it does, we would be in a strange situation where the conda-forge version of OpenMM is less capable than the StreamHPC version.

bdenhollander:
If there was a drastic decrease in kernel launch time due to less driver overhead, I could see performance doubling. Here's a profiling graph for gbsa with a lot of white space between groups of kernels. Imagine compressing that space horizontally because the GPU is active more often and performance would improve substantially.
image

Have you run through the tests to confirm results are valid? AMD's OpenCL driver automatically added CmdBarriers when it detected data dependencies between kernels. If HIP skipped that then many more kernels would run simultaneously rather than sequentially. This could produce gains that may or may not be valid.
image
image

egallicc:
Good idea to check carefully the tests for the correctness of the HIP implementation. We run production on this hardware. We haven't had MD crashes and we did not notice discrepancies or inconsistencies in the results.

philipturner:
@egallicc your vastly-improves throughputs [line up perfectly](https://openmm.org/benchmarks) with the RTX 2080, when extrapolating compute power.

> 
image


I'm surprised about the speedup from running stuff simultaneously. If you're correct, perhaps we'd increase throughput by running several simulations simultaneously. That would be useful for an evolutionary programming workflow, provided simulations contain at least several thousand atoms. My computer's hardware can't serve over two command streams simultaneously. Perhaps there's room for more concurrency if the GPU stalls completely during certain waits.

egallicc:
Thank you. We repeated the benchmarks. They are largely confirmed. The speed boost is indeed extreme and probably specific to our hardware (RX 6750 XT). I don't have another card to try.

Yes, with OpenCL the card uses a lot less power than with HIP. 80 W on average compared to 200 W with HIP. This could be caused by OpenCL using a fraction of the compute units or somehow waiting longer in between kernel launches as you are observing.

We are also running the tests (`make test`) with the HIP platform. As of now (~50% done), they are all passing. I will confirm as soon as they have all been completed.

StreamHPC's HIP implementation appears legit.

Maybe let's move this discussion to a separate thread? It could lead to ideas to speed up OpenCL.

peastman:
In that graph, which kernel does each color correspond to?  What are the point in the step where the GPU is idle?

philipturner:
> Yes, with OpenCL the card uses a lot less power than with HIP. 80 W on average compared to 200 W with HIP. This could be caused by OpenCL using a fraction of the compute units or somehow waiting longer in between kernel launches as you are observing.

It's strange that power only increases 2.5x but some throughputs increase 7x. Your graphics card's TDP is 250 W. For reference, my M1 Max is ~50-60 W and OpenMM is using only 30 W (apoa1rf). I could have room for a 2x throughput boost!

bdenhollander:
> In that graph, which kernel does each color correspond to? What are the point in the step where the GPU is idle?

The first two images in my comment are from the same profiling session. The light green `computeNonbonded` in the second image is the kernel ending at 1250 μs. Start from there and work up/left.

The big gaps are: 
- `integrateLangevinMiddlePart3` -> `calcCenterOfMassMomentum`
- `findBlocksWithInteractions` -> `computeBornSum`
- `computeNonbonded` -> `reduceForces`

bdenhollander:
> It's strange that power only increases 2.5x but some throughputs increase 7x. Your graphics card's TDP is 250 W. For reference, my M1 Max is ~50-60 W and OpenMM is using only 30 W (apoa1rf). I could have room for a 2x throughput boost!

Boosting voltage and frequency under low load uses considerable power. Average frequency is probably 200-400 MHz lower on HIP but way more units are active.

peastman:
I opened #3937.  Let's continue this discussion there.

muziqaz:
> Some of those differences are pretty extreme. All the ones I've seen in the past show about a 2x speed difference. I don't know what the reason is for it. I've generally suspected that AMD's OpenCL implementation is just badly written and does a very bad job of compliing the kernels.
> 
> There are legal issues that complicate distributing HIP code. The only way to get either the runtime libraries or the development kit is to download it from their website, which includes a step where you read and agree to their license. This means, for example, that it's impossible to build HIP code on conda-forge, because there's no legal way to install the development kit while building. We've spoken to some people at AMD about this and they were going to talk to their legal department about finding a solution. I don't know whether anything is happening about it.

Last I heard discussions are still ongoing. Large corp and stuff, things don't move fast :(
I ran comparison on my 6900xt, while difference between hip and opencl are staggering, they are not as extreme as egallicc's.
At some point this weekend I will run 7900xtx on it. That should show some unbelievable speedups, since 7000 series is running in limp mode on opencl (7900xtx is barely faster than 6900xt in FAH).

Issue #3924, opened 2023-01-20
Optimizing findBlocksWithInteractions for M1


philipturner:
I translated `findBlocksWithInteractions` to Metal Shading Language and ran it through Metal Frame Capture ([source code](https://github.com/philipturner/openmm-benchmarks/blob/09d58d0e62c7d981ce9eb64a0d87beffffc029de/FindInteractingBlocks/kernels.metal)). I stripped down a lot of extra code, just left the core inner loops. Without modification, a certain setup took 25573 us with 35% FP32 utilization.

If we explicitly force-unroll the inner loop through macros, the time will reduce down to 21559 us with 45% FP32 utilization. Do not trust the compiler to actually unroll loops, `#pragma unroll` does not work with Apple's compiler. This is all we can optimize through OpenCL.

Using Metal, we can bring execution time down to 17337 us by replacing the prefix sum with one call to `simd_prefix_inclusive_sum`. That brings FP32 utilization up to 56%.

Inside the loops, switching from threadgroup memory accesses to SIMD broadcasts harms performance. However, if and only if we fix the prefix sum afterward, something strange happens. The matrix multiplication idea from https://github.com/openmm/openmm/pull/2989 improves performance. It takes 16169 us with 50% FP32 utilization. Finally, we can compress `posBuffer` into `half` and reach 15501 us. This probably speeds it up because you can read two `half`s from one bank, compared to one `float`. Or it's because of lower register bandwidth.

I tried compressing a `float` as a `BFloat16`, but that harmed performance (25199 us). Probably because more ALU time is spent unpacking. Overall, we might speed up `findBlocksWithInteractions` the following amount:

| API Used | Speedup |
| -------- | ------- |
| OpenCL | +19% |
| Metal w/out compression | +58% |
| Metal w/ compression | +73% |

peastman:
Thanks, that's really interesting.

> Do not trust the compiler to actually unroll loops

That's unfortunate. :(  It's such a standard optimization in all compilers, it's really surprising they aren't doing it.

> Using Metal, we can bring execution time down to 17337 us by replacing the prefix sum with one call to simd_prefix_inclusive_sum.

I don't think we can do that using OpenCL on Mac.  OpenCL 2.0 introduced primitives for prefix sum and similar operations, but Apple has never supported anything beyond 1.2.  On AMD we might be able to use it, though.

> Finally, we can compress posBuffer into half and reach 15501 us.

[...]

PR #3959, opened 2023-02-12
Optimize `findBlocksWithInteractions` for Apple silicon GPUs

philipturner:
The Apple Metal/AIR backend compiler does not auto-inline loops, even when attempting to force-inline it. This should speedup the kernel by under 19%, speeding up the actual performance by the order of 1% in some benchmarks.

Resolves https://github.com/openmm/openmm/issues/3924

### DIFF
--- a/platforms/opencl/src/OpenCLContext.cpp
+++ b/platforms/opencl/src/OpenCLContext.cpp
@@ -218,6 +218,7 @@ OpenCLContext::OpenCLContext(const System& system, int platformIndex, int device
       
         if (vendor.size() >= 5 && vendor.substr(0, 5) == "Apple") {
             simdWidth = 32;
+            compilationDefines["APPLE_GPU_FAMILY"] = "";
         }
         else if (vendor.size() >= 6 && vendor.substr(0, 6) == "NVIDIA") {
             compilationDefines["WARPS_ARE_ATOMIC"] = "";
--- a/platforms/opencl/src/kernels/findInteractingBlocks.cl
+++ b/platforms/opencl/src/kernels/findInteractingBlocks.cl
@@ -79,6 +79,22 @@ __kernel void sortBoxData(__global const real2* restrict sortedBlock, __global c
 
 #define BUFFER_SIZE 256
 
+#define FORCE_UNROLL_4(__expr, __start_index) \
+__expr(__start_index + 0); \
+__expr(__start_index + 1); \
+__expr(__start_index + 2); \
+__expr(__start_index + 3); \
+
+#define FORCE_UNROLL_32(__expr) \
+FORCE_UNROLL_4(__expr, 0); \
+FORCE_UNROLL_4(__expr, 4); \
+FORCE_UNROLL_4(__expr, 8); \
+FORCE_UNROLL_4(__expr, 12); \
+FORCE_UNROLL_4(__expr, 16); \
+FORCE_UNROLL_4(__expr, 20); \
+FORCE_UNROLL_4(__expr, 24); \
+FORCE_UNROLL_4(__expr, 28); \
+
 __kernel void findBlocksWithInteractions(real4 periodicBoxSize, real4 invPeriodicBoxSize, real4 periodicBoxVecX, real4 periodicBoxVecY, real4 periodicBoxVecZ,
         __global unsigned int* restrict interactionCount, __global int* restrict interactingTiles, __global unsigned int* restrict interactingAtoms,
         __global const real4* restrict posq, unsigned int maxTiles, unsigned int startBlockIndex, unsigned int numBlocks, __global real2* restrict sortedBlocks,
@@ -202,10 +218,20 @@ __kernel void findBlocksWithInteractions(real4 periodicBoxSize, real4 invPeriodi
                         }
                         else {
 #endif
+#ifdef APPLE_GPU_FAMILY
+                            #define __findBlocksWithInteractions_loop1(j) \
+                            { \
+                                real3 delta = pos2-posBuffer[warpStart+j]; \
+                                interacts |= (delta.x*delta.x+delta.y*delta.y+delta.z*delta.z < PADDED_CUTOFF_SQUARED); \
+                            } \
+                            
+                            FORCE_UNROLL_32(__findBlocksWithInteractions_loop1)
+#else
                             for (int j = 0; j < TILE_SIZE; j++) {
                                 real3 delta = pos2-posBuffer[warpStart+j];
                                 interacts |= (delta.x*delta.x+delta.y*delta.y+delta.z*delta.z < PADDED_CUTOFF_SQUARED);
                             }
+#endif
 #ifdef USE_PERIODIC
                         }
 #endif
@@ -216,14 +242,22 @@ __kernel void findBlocksWithInteractions(real4 periodicBoxSize, real4 invPeriodi
                     atomCountBuffer[get_local_id(0)].x = (interacts ? 1 : 0);
                     SYNC_WARPS;
                     int whichBuffer = 0;
-                    for (int offset = 1; offset < TILE_SIZE; offset *= 2) {
-                        if (whichBuffer == 0)
-                            atomCountBuffer[get_local_id(0)].y = (indexInWarp < offset ? atomCountBuffer[get_local_id(0)].x : atomCountBuffer[get_local_id(0)].x+atomCountBuffer[get_local_id(0)-offset].x);
-                        else
-                            atomCountBuffer[get_local_id(0)].x = (indexInWarp < offset ? atomCountBuffer[get_local_id(0)].y : atomCountBuffer[get_local_id(0)].y+atomCountBuffer[get_local_id(0)-offset].y);
-                        whichBuffer = 1-whichBuffer;
-                        SYNC_WARPS;
-                    }
+//                    for (int offset = 1; offset < TILE_SIZE; offset *= 2) {
+                    #define __findBlocksWithInteractions_loop2(offset) \
+                    { \
+                        if (whichBuffer == 0) \
+                            atomCountBuffer[get_local_id(0)].y = (indexInWarp < offset ? atomCountBuffer[get_local_id(0)].x : atomCountBuffer[get_local_id(0)].x+atomCountBuffer[get_local_id(0)-offset].x); \
+                        else \
+                            atomCountBuffer[get_local_id(0)].x = (indexInWarp < offset ? atomCountBuffer[get_local_id(0)].y : atomCountBuffer[get_local_id(0)].y+atomCountBuffer[get_local_id(0)-offset].y); \
+                        whichBuffer = 1-whichBuffer; \
+                        SYNC_WARPS; \
+                    } \
+                    
+                    __findBlocksWithInteractions_loop2(1);
+                    __findBlocksWithInteractions_loop2(2);
+                    __findBlocksWithInteractions_loop2(4);
+                    __findBlocksWithInteractions_loop2(8);
+                    __findBlocksWithInteractions_loop2(16);
                     
                     // Add any interacting atoms to the buffer.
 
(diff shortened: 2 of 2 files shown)


peastman:
Here are the speed differences I see on the benchmarks with this change.

||Before|After|Ratio
|---|---|---|---|
gbsa|408.387|406.496|0.995
rf|255.078|258.069|1.012
pme|147.489|148.896|1.010
apoa1rf|59.8386|61.0192|1.020
apoa1pme|35.2012|35.4637|1.007
apoa1ljpme|26.1992|26.4046|1.008
amoebagk|0.535711|0.538338|1.005
amoebapme|2.16664|2.14889|0.992
amber20-dhfr|155.635|156.501|1.006
amber20-cellulose|6.96229|7.02266|1.009
amber20-stmv|1.82035|1.84319|1.013

All differences are within the noise level.  I'm not sure it's worth adding complexity to the code for a negligible speedup.

philipturner:
Could you rerun `gbsa` and `amoebapme` for about 3 additional trials? There seems to be a statistically significant, universal +1% speedup like I predicted. The only way this would harm performance is by thrashing the instruction cache, if the inner 32-loop had 375 bytes of instructions (47 instructions).

peastman:
> We could also spot other loops in the code base that would benefit from the optimization. Then the speedup might reach 2% or 3%.

Let's see if we can do that.  I'm just hesitant to add complexity to an already complex kernel for a 1% speedup on a subset of benchmarks.  Especially since it only happens on macOS, and will likely become unnecessary in the future.  Loop unrolling is a standard optimization most compilers do.  It's surprising that Apple's doesn't, and it's likely they'll add it in the future.

philipturner:
> It's surprising that Apple's doesn't, and it's likely they'll add it in the future.

I've seen instances where they do; it just unrolls a small amount, and you can't control it. It might unroll 2-4 small iterations but not 32. Probably a heuristic to account for the M1's comparatively small L1I.

peastman:
And I'm guessing they don't support `#pragma unroll 32`?

philipturner:
That doesn't do anything. I remember trying it [here](https://github.com/philipturner/openmm-benchmarks/blob/main/FindInteractingBlocks/kernels.metal) and saw no change. The repo provides a very simple way to test how something affects this kernel's performance.

bdenhollander:
> We could also spot other loops in the code base that would benefit from the optimization. Then the speedup might reach 2% or 3%.

Try unrolling some of the `< TILE_SIZE` loops, which are typically 32 iterations. `computeGKForces` and `computeElectrostatics` would be targets for amoebagk and amoebapme respectively.

philipturner:
>> `computeGKForces` and `computeElectrostatics` would be targets for amoebagk and amoebapme respectively.

That would probably harm performance, as the loops are too large. Perhaps force-unrolling up to four iterations might allow more ILP, but we can't overflow the instruction cache.

philipturner:
Ultimately the purpose is to try to implement the optimizations from #3924. They can only be fully realized with Metal, creating a 3% speedup (no `half`). Limiting to OpenCL features makes this only 1%.

philipturner:
Converged to draft because https://github.com/openmm/openmm/issues/3979

philipturner:
I just reached 138 ns/day on apoa1rf - see https://github.com/philipturner/openmm-metal/commit/830c2e823a74ffd1ed41dda98947c570cde05b4e. The matrix multiply optimization harmed performance, likely because there was already so much `threadgroup` traffic. In my [benchmarking](https://github.com/philipturner/openmm-benchmarks/blob/main/FindInteractingBlocks/kernels.metal) repo, I had simplified the shader and removed a lot of traffic. I think we can see ever greater speedups by incorporating the full CUDA implementation for M1 and maybe RDNA, now that we can use SIMD ballots. The CUDA implementation is modestly different and uses a `saveSinglePairs` kernel. It also proactively skips certain internal loop iterations, which could boost ns/day beyond the theoretical maximum I predicted.

peastman:
Don't worry about the `saveSinglePairs()` optimization.  The effect is tiny, maybe a 1-2% speedup at best.  It also is fairly brittle and needs to be re-tuned with every generation.  Even on CUDA it's probably more trouble than it's worth.

The biggest optimization is that we can used a `__ballot_sync()` combined with a `__popc()` to do a prefix sum of the binary flags over a whole warp in just a few instructions.

philipturner:
I tried that and now half of the tests are failing. The `apoa1rf` benchmark is silently crashing. Do you have any idea what's happening?

https://github.com/philipturner/openmm-metal/commit/a0ebbd3acaf488395a10c6eec15d329ddb46ba65

philipturner:
Even without the ballot optimization, some impressive improvements. I still need to tune force thread blocks for AMOEBA as @bdenhollander suggested on the Mac AMD thread. Going by RTX 4080 for comparison, we can reach 141 ns/day on M1 Max `apoa1rf`. Going by RTX 2080 = 2x its own TFLOPS, 150 ns/day.

- Optimization 1: fixing SIMD width, thread blocks, and loop unrolling for `findBlocksWithInteractions`
- Optimization 2: fixing command queue flushing
- Optimization 3: switching to `simd_prefix_inclusive_sum` in `findBlocksWithInteractions`


ns/day | OpenMM 8.0.0 | Opt. 1 | Opt. 2 | Opt. 3 | Extrapolating 2080 |
-- | -- | -- | -- | -- | --
gbsa | 264 | 378.0 | 632.8 | 631.8 | 974.9
rf | 153 | 304.5 | 421.5 | 452.8 | 568.5
pme | 120 | 199.6 | 246.8 | 258.6 | 463.5 |
apoa1rf | 42 | 114.9 | 129.0 | 138.5 | 149.7 |
apoa1pme | 33 | 67.2 | 72.3 | 75.3 | 103.7 |
apoa1ljpme | 28 | 50.2 | 53.5 | 55.1 | 83.6 |
amoebagk | 0.8 | 1.184 | 1.187 | 1.185 | 8.512 |
amoebapme | 2.72 | 3.97 | 4.00 | 4.05 | 7.72 |
amber20-dhfr | TBD | 212.0 | 267.0 | 275.7 | TBD |
amber20-cellulose | TBD | 14.36 | 14.78 | 15.69 | 17.72
amber20-stmv | TBD | 3.83 | 3.88 | 4.29 | 5.23

Speed vs 8.0.0 | Opt. 1 | Opt. 2 | Opt. 3 | Extrapolating 2080
-- | -- | -- | -- | --
gbsa | +43% | +140% | +139% | +269%
rf | +99% | +175% | +196% | +271%
pme | +66% | +106% | +116% | +286%
apoa1rf | +174% | +207% | +230% | +256%
apoa1pme | +104% | +119% | +128% | +214%
apoa1ljpme | +79% | +91% | +97% | +199%
amoebagk | +48% | +48% | +48% | +964%
amoebapme | +46% | +47% | +49% | +184%

Peak Watts | Opt. 1 | Opt. 2 | Opt. 3
-- | -- | -- | --
gbsa | 13.2 | 22.1 | 21.6
rf | 20.0 | 27.7 | 28.0
pme | 16.8 | 21.2 | 21.0
apoa1rf | 30.8 | 35.3 | 35.8
apoa1pme | 23.6 | 25.6 | 25.3
apoa1ljpme | 23.5 | 25.0 | 24.8
amoebagk | 15.4 | 15.3 | 15.2
amoebapme | 17.3 | 17.3 | 17.5
amber20-dhfr | 16.9 | 21.1 | 21.5
amber20-cellulose | 27.5 | 28.1 | 28.6
amber20-stmv | 27.1 | 27.4 | 28.5

Optimization 3 does not change watts, and improves everything except AMOEBA, GBSA, GK.

bdenhollander:
@philipturner There are loops in `gridInterpolateForce` and `gridSpreadCharge` that AMD unrolls. Search for `for (int ix = 0; ix < PME_ORDER; ix++)` and try unrolling all three levels or the innermost two depending on how many registers are available. This may improve performance on pme benchmarks.

`PME_ORDER` is set to `PmeOrder`, which is defined as `static const int PmeOrder = 5`;

philipturner:
If I'm going to maintain a separate Metal backend after all, I think there's less need to integrate the entire header into OpenMM. @peastman could we just paste the declaration for `simd_prefix_inclusive_sum(float)` and `simd_broadcast(float, ushort)`, and not worry about the header?

peastman:
That seems reasonable.

bdenhollander:
`V_MAD_I32_I24` and `V_MAD_U64_U32` are mentioned in all variants of the RDNA ISAs: [RDNA](https://www.amd.com/system/files/TechDocs/rdna-shader-instruction-set-architecture.pdf), [RDNA2](https://www.amd.com/system/files/TechDocs/rdna2-shader-instruction-set-architecture.pdf), and [RDNA3](https://www.amd.com/system/files/TechDocs/rdna3-shader-instruction-set-architecture-feb-2023_0.pdf).

VOP3 instructions might be what you're looking for.

philipturner:
> VOP3 instructions might be what you're looking for.

Apple's CMPSEL instruction takes 4 input operands, a massive 80 bits to encode. For example, it is used for the `max` and `select` operations. The hardware probably also uses it for control flow, fusing the comparison and branch selection.

```
max(x,y): CMPSEL x > y ? x : y
select(x,y, a < b): CMPSEL a < b ? x : y
```

AMD only supports 3 inputs. Can it fuse a comparison and selection into a single cycle?

Screenshot 2023-03-23 at 9 12 08 PM

philipturner:
GPT-4 was right. The bug stems from CUDA `__ffs` having very different behavior than OpenCL `ctz`.

> The count trailing zeros and count leading zeros functions may not work as expected on OpenCL devices. These functions are also vendor-specific functions that may have different implementations or behaviors on different devices. For example, some devices may return undefined values for zero inputs or have different rounding modes for floating-point inputs. You may want to check the documentation of your device and the OpenCL specification to see if there are any differences or limitations for these functions. You may also want to test these functions with some simple inputs and outputs to see if they match your expectations.

philipturner:
Optimization 4: ballot optimization



ns/day | OpenMM 8.0.0 | Opt. 1 | Opt. 2 | Opt. 3 | Opt. 4 | Extrapolating 2080
-- | -- | -- | -- | -- | -- | --
gbsa | 264 | 378.0 | 632.8 | 631.8 | 635.0 | 974.9
rf | 153 | 304.5 | 421.5 | 452.8 | 476.3 | 568.5
pme | 120 | 199.6 | 246.8 | 258.6 | 267.2 | 463.5
apoa1rf | 42 | 114.9 | 129.0 | 138.5 | <b><ins>151.4</ins></b> | <b><ins>149.7</ins></b>
apoa1pme | 33 | 67.2 | 72.3 | 75.3 | 79.0 | 103.7
apoa1ljpme | 28 | 50.2 | 53.5 | 55.1 | 56.9 | 83.6
amoebagk | 0.8 | 1.184 | 1.187 | 1.185 | 1.189 | 8.512
amoebapme | 2.72 | 3.97 | 4.00 | 4.05 | 4.07 | 7.72
amber20-dhfr | 131.7 | 212.0 | 267.0 | 275.7 | 279.2 | TBD
amber20-cellulose | 5.92 | 14.36 | 14.78 | 15.69 | <b><ins>19.93</ins></b> | <b><ins>17.72</ins></b>
amber20-stmv | 1.40 | 3.83 | 3.88 | 4.29 | 6.89 | 5.23

bdenhollander:
> AMD only supports 3 inputs. Can it fuse a comparison and selection into a single cycle?

I can't find anything like that for RDNA. Pre-GCN, [TeraScale/VLIW5/VLIV4](https://www.x.org/docs/AMD/old/r600isa.pdf) might have been able to do something similar but only when comparing to 0.
image

philipturner:
That would be a 3-input instruction, so it seems AMD doesn't have anything similar.

philipturner:
> @philipturner There are loops in gridInterpolateForce and gridSpreadCharge that AMD unrolls. Search for for (int ix = 0; ix < PME_ORDER; ix++) and try unrolling all three levels or the innermost two depending on how many registers are available. This may improve performance on pme benchmarks.

I just unrolled the 1 innermost loop for each of these functions, but both optimizations failed.

PR #3960, opened 2023-02-12
[macOS GPU Support] Fix `MTLCommandBuffer` bottlenecks for Apple silicon GPUs

philipturner:
The Apple Metal/MoltenCL backend has strange behavior regarding OpenCL events, causing them to infect any nearby commands. We explicitly flush whenever touching an event, ensuring a smooth command stream.

Refs: #3937

### DIFF
--- a/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
+++ b/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
@@ -358,7 +358,7 @@ void OpenCLNonbondedUtilities::prepareInteractions(int forceGroups) {
 
     #if __APPLE__ && defined(__aarch64__)
     // Segment the command stream to avoid stalls later.
-    if (groupKernels[forceGroups].hasForces && useCutoff && numTiles > 0)
+    if (groupKernels[forceGroups].hasForces)
         context.getQueue().flush();
     #endif
 }
--- a/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
+++ b/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
@@ -354,7 +354,13 @@ void OpenCLNonbondedUtilities::prepareInteractions(int forceGroups) {
     context.executeKernel(kernels.findInteractingBlocksKernel, context.getNumAtoms(), interactingBlocksThreadBlockSize);
     forceRebuildNeighborList = false;
     lastCutoff = kernels.cutoffDistance;
-    context.getQueue().enqueueReadBuffer(interactionCount.getDeviceBuffer(), CL_FALSE, 0, sizeof(int), pinnedCountMemory, NULL, &downloadCountEvent); 
+    context.getQueue().enqueueReadBuffer(interactionCount.getDeviceBuffer(), CL_FALSE, 0, sizeof(int), pinnedCountMemory, NULL, &downloadCountEvent);
+
+    #if __APPLE__ && defined(__aarch64__)
+    // Segment the command stream to avoid stalls later.
+    if (groupKernels[forceGroups].hasForces && useCutoff && numTiles > 0)
+        context.getQueue().flush();
+    #endif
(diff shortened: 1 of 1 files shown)


peastman:
Is this the same as the optimization we've been discussing in https://github.com/openmm/openmm/issues/3937, or is this something different?

philipturner:
Yes, it is the same optimization.

platforms/opencl/src/OpenCLNonbondedUtilities.cpp
```diff
@@ -354,7 +354,13 @@ void OpenCLNonbondedUtilities::prepareInteractions(int forceGroups) {
     context.executeKernel(kernels.findInteractingBlocksKernel, context.getNumAtoms(), interactingBlocksThreadBlockSize);
     forceRebuildNeighborList = false;
     lastCutoff = kernels.cutoffDistance;
-    context.getQueue().enqueueReadBuffer(interactionCount.getDeviceBuffer(), CL_FALSE, 0, sizeof(int), pinnedCountMemory, NULL, &downloadCountEvent); 
+    context.getQueue().enqueueReadBuffer(interactionCount.getDeviceBuffer(), CL_FALSE, 0, sizeof(int), pinnedCountMemory, NULL, &downloadCountEvent);
+
+    #if __APPLE__ && defined(__aarch64__)
+    // Segment the command stream to avoid stalls later.
+    if (groupKernels[forceGroups].hasForces && useCutoff && numTiles > 0)
```
peastman, line 361: It already checks for some of these conditions above:

```c++
    if (!useCutoff)
        return;
    if (numTiles == 0)
        return;
```

Now I'm confused about why it doesn't also return if `hasForces` is false, and likewise why the call to `updateNeighborListSize()` in `computeInteractions()` isn't inside the `if (kernels.hasForces)` block.  Is there ever a reason it would need to rebuild the neighbor list if there are no nonbonded forces in the specified force groups?  I can think of any.  I think we probably want to make those changes, though we need to make sure all the tests pass in case there's some reason I've forgotten about.

philipturner: I think that would belong in a separate PR. If changing that causes a bug, I don't want this optimization to be conflated with it and unnecessarily reverted.

philipturner:
Should I also change the compilation macros to a runtime query of the `CLDevice` vendor?

platforms/opencl/src/OpenCLNonbondedUtilities.cpp
peastman: Ok, that's fine.

peastman:
I'm not sure what the best check is.  A different OpenCL implementation on macOS might have different behavior.  So maybe we should check both the platform vendor and the device vendor?  But we don't currently support any other implementations and we don't currently have plans to add any.  So the macro is probably ok for now.

philipturner:
The I'd say get it over with and merge it.

peastman:
Sounds good.  I'll merge it once the tests finish running.

peastman:
Here's the performance difference I get from this change on M1 Pro.

||Before|After|Ratio
|---|---|---|---|
gbsa|299.897|408.387|1.36
rf|210.735|255.078|1.21
pme|130.44|147.489|1.13
apoa1rf|56.8595|59.8386|1.05
apoa1pme|33.7838|35.2012|1.04
apoa1ljpme|25.529|26.1992|1.03
amoebagk|0.534747|0.535711|1.00
amoebapme|2.15571|2.16664|1.01
amber20-dhfr|136.644|155.635|1.14
amber20-cellulose|6.89188|6.96229|1.01
amber20-stmv|1.81794|1.82035|1.00

There's a significant speedup on the DHFR benchmarks, a small speedup on the ApoA1 benchmarks, and minimal difference on the more expensive ones (AMOEBA or larger molecules).  It's never slower.

PR #3961, opened 2023-02-13
Remove unused offset variables

bdenhollander:
- Appear to be copy and pasted from getPositions and were not removed


peastman:
Thanks!

PR #3968, opened 2023-02-17
Fix misleading documentation of `Topology.bonds`

padix-key:
`Topology.bonds()` states that iterates over a tuple of `Atom` objects. While this is technically correct, as the `Bond` class is a named tuple, this also hides the fact, that `Bond` objects also provide the `order` and `type` attributes. Therefore, I propose to change the sentence, so that it uses the capitalized *Bonds* indicating that they are actually objects.

### DIFF
--- a/wrappers/python/openmm/app/topology.py
+++ b/wrappers/python/openmm/app/topology.py
@@ -230,8 +230,8 @@ class Topology(object):
                     yield atom
 
     def bonds(self):
-        """Iterate over all Bonds in the Topology.
-        A Bond can be treated as tuple of two Atoms.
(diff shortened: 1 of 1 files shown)


peastman:
Thanks for catching this!  It used to be just an ordinary tuple before we created the Bond class.  Perhaps we could make the description even more explicit, something like this?

> Iterate over all bonds in the Topology.  Each one is represented by a Bond object, which is a named tuple of two atoms.

padix-key:
I like your proposed change and adapted the text accordingly.

peastman:
Looks good.  Thanks!

PR #3969, opened 2023-02-17
Fix ReStructuredText formatting in docstring

padix-key:
This PR should fix the docstring rendering of `Quantity`, which currently looks the following way: http://docs.openmm.org/latest/api-python/generated/openmm.unit.quantity.Quantity.html#openmm.unit.quantity.Quantity


peastman:
Thanks!

PR #3976, opened 2023-02-23
Add OpenCLPlatformIndex to benchmark.py

bdenhollander:
Fix specifying a device will fail to run when multiple OpenCL platforms are available. Specifying `--verbose` prints the following exception: `Specified DeviceIndex but not OpenCLPlatformIndex. When multiple platforms are available, a platform index is needed to specify a device.`


peastman:
Looks good.

Issue #3457, opened 2022-02-10
Improve load balancing on multiple GPUs


peastman:
This is a followup to #3333.  We made changes there that significantly improved multi-GPU performance.  A major limiting factor at this point seems to be load balancing.  Ideally we want all GPUs to finish their work at the same time, but in practice we usually get a lot of variation so some GPUs spend extra time sitting idle.  We should investigate whether we can improve that.

peastman:
In order to improve the load balancing, I need to accurately measure how long it takes for each GPU to complete its computations and data transfers.  Currently that's done by calling `cuCtxSynchronize()` and measuring the time on the host.  It works, but not all that well.  The synchronization makes it expensive, so I don't want to do it too frequently.  And it isn't a very accurate way of measuring when the GPU has finished its work.

An alternative is `cuEventElapsedTime()`.  In principle that should allow direct timing measurements from the GPU without involving the host.  But according to the [documentation](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EVENT.html#group__CUDA__EVENT_1gdfb1178807353bbcaa9e245da497cf97) there are two major problems with it.  First, the time resolution is only about 0.5 ms, which isn't very good for this purpose.  And second, if you use any stream other than the default one its result may be wrong.  I think that mostly rules it out.

Is there a better way of measuring time?  The CUDA profiler reports exact times when each operation begins and ends with sub-microsecond resolution.  That suggests this information must be available somewhere.  Is there a way to retrieve it?

cc @dmclark17

dmclark17:
The time resolution of `cuEventElapsedTime` should be around 0.5 microseconds. The non-NULL stream behavior does seem potentially problematic. I will look into other solutions!

peastman:
@dmclark17 did you ever figure out if there's a way to get accurate timings that works with multiple streams?

dmclark17:
Sorry, I lost track of this—I think we might be able to use CUDA events for this use-case. It looks like CUPTI could also be a possibility, but I am not very familiar with the API: [https://docs.nvidia.com/cupti/r_overview.html#r_overview](https://docs.nvidia.com/cupti/r_overview.html#r_overview)

[...]

PR #3984, opened 2023-03-01
Initialize tilesAfterReorder with zero

ex-rzr:
It may contain a garbage, and if it is large then updateNeighborListSize does not force reorder atoms after 25 steps, multi-GPU load balancing (for amber20-stmv particularly) uses an initial order splitting work between GPUs suboptimally.

I've noticed a strange behavior for multi-GPU STMV benchmarks on the HIP platform:

1. Amber benchmarks use LocalEnergyMinimizer which sets positions multiple times
2. tilesAfterReorder may have a garbage which can be very large
3. So this check https://github.com/openmm/openmm/blob/master/platforms/cuda/src/CudaNonbondedUtilities.cpp#L428 does not trigger reordering at step 25
4. Multi-GPU load balancing works only during first 200 steps, reordering will happen at step 250
5. So a distribution between GPU is based on blocks of unordered atoms (IIRC there are >600k "unordered" tiles vs 500k tiles after reodering)
6. As the result some runs works slower, others - faster (the difference is about 10% for STMV on 4 GPUs)

(@peastman unrelated to this issue, but is it possible to add a link to https://github.com/amd/openmm-hip to https://openmm.org/ecosystem ? Thanks!)

### DIFF
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -64,7 +64,7 @@ private:
 };
 
 CudaNonbondedUtilities::CudaNonbondedUtilities(CudaContext& context) : context(context), useCutoff(false), usePeriodic(false), anyExclusions(false), usePadding(true),
-        blockSorter(NULL), pinnedCountBuffer(NULL), forceRebuildNeighborList(true), lastCutoff(0.0), groupFlags(0), canUsePairList(true) {
+        blockSorter(NULL), pinnedCountBuffer(NULL), forceRebuildNeighborList(true), lastCutoff(0.0), groupFlags(0), canUsePairList(true), tilesAfterReorder(0) {
     // Decide how many thread blocks to use.
 
     string errorMessage = "Error initializing nonbonded utilities";
@@ -423,7 +423,7 @@ void CudaNonbondedUtilities::computeInteractions(int forceGroups, bool includeFo
 bool CudaNonbondedUtilities::updateNeighborListSize() {
     if (!useCutoff)
         return false;
-    if (context.getStepsSinceReorder() == 0)
+    if (context.getStepsSinceReorder() == 0 || tilesAfterReorder == 0)
         tilesAfterReorder = pinnedCountBuffer[0];
     else if (context.getStepsSinceReorder() > 25 && pinnedCountBuffer[0] > 1.1*tilesAfterReorder)
         context.forceReorder();
--- a/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
+++ b/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
@@ -56,7 +56,7 @@ private:
 };
 
 OpenCLNonbondedUtilities::OpenCLNonbondedUtilities(OpenCLContext& context) : context(context), useCutoff(false), usePeriodic(false), anyExclusions(false), usePadding(true),
-        blockSorter(NULL), pinnedCountBuffer(NULL), pinnedCountMemory(NULL), forceRebuildNeighborList(true), lastCutoff(0.0), groupFlags(0) {
+        blockSorter(NULL), pinnedCountBuffer(NULL), pinnedCountMemory(NULL), forceRebuildNeighborList(true), lastCutoff(0.0), groupFlags(0), tilesAfterReorder(0) {
     // Decide how many thread blocks and force buffers to use.
 
     deviceIsCpu = (context.getDevice().getInfo<CL_DEVICE_TYPE>() == CL_DEVICE_TYPE_CPU);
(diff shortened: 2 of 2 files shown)


peastman:
That check isn't supposed to be triggered.  It's only there for unusual situations.  In normal simulations it never gets activated.

Aside from that check, reordering happens every 250 updates to the positions.  That includes updates that happen during minimization: there's a call to `reorderAtoms()` at the end of `CudaUpdateStateDataKernel::setPositions()`.  That's a short enough interval that the efficiency of the neighbor list usually doesn't decrease very much in that time.  The check based on `tilesAfterReorder` is there to catch the unusual cases where it does, such as when a user is evaluating a series of states that have large differences from each other.

A consequence is that the atoms should never be unordered.  The very first time positions get set, `reorderAtoms()` will immediately get called and reordering will happen.

So I think there must be some other reason for the poor load balancing.  Take a look at #3457.  We were discussing ways to improve it there but never came up with a real solution.  Perhaps HIP has the tools we need?

> is it possible to add a link to https://github.com/amd/openmm-hip to https://openmm.org/ecosystem ? Thanks!)

Yes, I'll add it.  Thanks for point out it was missing!

ex-rzr:
Yes, I read thread about multi-GPU balancing and tilesAfterReorder.

> That includes updates that happen during minimization: there's a call to reorderAtoms() at the end of CudaUpdateStateDataKernel::setPositions()

> The very first time positions get set, reorderAtoms() will immediately get called and reordering will happen.

Yes, but right after the first reordering LocalEnergyMinimizer changes positions (so atoms are unordered again!), it does this (I don't remember exactly) about 80 times, the next reordering will happen later at step 250. Load balancing finishes at step 200, so during these 200 steps it's measuring performance of unordered atoms and modifying GPU fractions but they are not optimal when reordering gets finally triggered.

So either I'm missing something or the logic should be changed somehow (perhaps multi-GPU load balancing should trigger it at least once?).

Could you check what is `pinnedCountBuffer[0]` before load balancing finishes and after reordering at step 250? I see "there are >600k "unordered" tiles vs 500k tiles after reordering".

> So I think there must be some other reason for the poor load balancing. Take a look at https://github.com/openmm/openmm/issues/3457. We were discussing ways to improve it there but never came up with a real solution. Perhaps HIP has the tools we need?

Sadly, HIP tools are still catching up. On traces collected with rocprof I saw timelines quite similar to Nsight's results in that issue.

peastman:
> right after the first reordering LocalEnergyMinimizer changes positions (so atoms are unordered again!)

That doesn't alter the ordering.  `setPositions()` takes the current ordering into account when copying the positions over:

https://github.com/openmm/openmm/blob/81271f93abeca82af709e1b288b2fc4aa0c272d5/platforms/cuda/src/CudaKernels.cpp#L194-L200

As long as the new positions aren't too different from the old ones, the existing ordering shouldn't be too bad.

> I see "there are >600k "unordered" tiles vs 500k tiles after reordering".

That's only a slightly suboptimal ordering.  If they were truly unordered, you would probably have orders of magnitude more tiles.  In fact, it would probably run out of memory trying to allocate space for them!

ex-rzr:
Ok, thanks for the clarification. You are right, I missed that!

So if ordering is preserved, this difference in numbers of tiles after LocalEnergyMinimizer (529210) and after reordering (408392) is a result of position deltas added by LocalEnergyMinimizer? I guess I underestimated its influence and thought that ordering was ignored completely.

peastman:
I believe that's correct.

We should still merge this.  Initializing memory is a good idea regardless!  But what about making this change?  In `updateNeighborListSize()` where it decides when to record the number of tiles

```c++
    if (context.getStepsSinceReorder() == 0)
        tilesAfterReorder = pinnedCountBuffer[0];
```

change it to

```c++
    if (context.getStepsSinceReorder() == 0 || tilesAfterReorder == 0)
        tilesAfterReorder = pinnedCountBuffer[0];
```

That should match the behavior that was really intended.  The very first time `updateNeighborListSize()` is called, it will always just record the number of tiles.  It will only force a reorder if it sees the number has increased too much.

ex-rzr:
Good idea. Done.

And finally I think that I know what happens:

I used to run the **old** benchmark.py (as I have some modifications useful for profiling on HIP). This script calls `setPeriodicBoxVectors` after `setPositions`:
https://github.com/openmm/openmm/blob/6bdaa4953963575e0c70d155ec85a81b2f5afd39/examples/benchmark.py#L161-L164

The new script calls `setDefaultPeriodicBoxVectors` **before** `setPositions`:
https://github.com/openmm/openmm/blob/master/examples/benchmark.py#L197-L198

So in fact with the old script periodic box vectors are set twice:
```
setPeriodicBoxVectors
a: (22.754963 0.000000 0.000000) b: (0.000000 22.963463 0.000000) c: (0.000000 0.000000 23.096564)
setPositions
reorderAtomsImpl
setPeriodicBoxVectors
a: (22.117231 0.000000 0.000000) b: (0.000000 22.319888 0.000000) c: (0.000000 0.000000 22.449258)
```
and reordering happens after the first `setPeriodicBoxVectors` (implicit) and `setPositions`.
But when the old benchmark.py sets periodic box vectors explicitly, there are multiple atoms that must be unwrapped and wrapped again:
https://github.com/openmm/openmm/blob/master/platforms/cuda/src/CudaKernels.cpp#L335-L344

And a new reordering doesn't happen, making the old order non-optimal, that's when LocalEnergyMinimizer and load balancing start working. In fact LocalEnergyMinimizer does not change atom positions to increase the number of tiles significantly.

Sorry for this confusion.

peastman:
I'm not sure of a good solution to that, or even if there's anything that needs to be solved.  Normally it's not a big deal.  It runs with a slightly sub-optimal order for the first 250 steps, then reorders and is fine for the rest of the simulation.  But it does potentially affect load balancing.  Then again, there are much bigger uncertainties in load balancing so it may not make much difference.

Issue #3997, opened 2023-03-14
XTC reporters


giadefa:
We would like to be able to use XTC reporters. It should be quite easy to do and we can do the PR.

Using XTC reporters is important because the size is 1/3 of DCD. When we run tons of simulations we have space problems. Yes, we can convert it later but it does not solve the size problem, in particular for GPUGRID.

@peastman ok?

mattwthompson:
There are some previous attempts: https://github.com/openmm/openmm/issues/2755 https://github.com/openmm/openmm/issues/1186

You can already do this via MDTraj: https://github.com/mdtraj/mdtraj/blob/1.9.7/mdtraj/reporters/xtcreporter.py

peastman:
I'd be ok with adding a XTC reporter, though the one in MDTraj might be fine for you.

sef43:
Also this for XTC and other formats: https://github.com/sef43/openmm-mdanalysis-reporter

But I think that an internal XTC reporter for OpenMM is a good idea!

giadefa:
we are doing the PR

PR #4001, opened 2023-03-16
Adding an XTC reporter

RaulPPelaez:
Following #3997 this PR attempts to add an XTCReporter to OpenMM.

- [x] Move and adapt xtc writer/reader from moleculekit (explicit permission granted by the authors to do so)
- [x] Create XTCFile
   - [x] Write tests
- [x] Create XTCReporter
   - [x] Write tests

The xtc writer/reader consists of cpp sources and a cython file. I added everything necessary to setup.py as an ext_modules.


peastman:
This is looking good.  Do try to make the API consistent with other classes.  Take a look at PDBFile and DCDFile and wherever possible use identical names for methods and attributes.  Also for consistency, the class should be called XTCFile instead of XTCTrajectoryFile.

raimis:
`xdrfile` is available on `conda-forge` (https://anaconda.org/conda-forge/xdrfile). We should reuse that library instead of embedding it into OpenMM.

RaulPPelaez:
The conda-forge package links to a documentation in a 404 page. Could you point to some resource about it? As in, does this provide a python interface? how does one call it? etc.

giadefa:
I would go as we were doing without adding an extra dependency.

raimis:
I think it is just a C library. Documentation should be somewhere on GROMACS page.

RaulPPelaez:
In that case its basically some version of what I uploaded. Its just a couple of headers and cpp files with a permissive licence, I agree with Gianni its simpler this way.

peastman:
Our general policy is to avoid external dependencies whenever possible.  Using the external package would limit us to only the platforms, python versions, and CPU architectures that package chose to provide.  It would add complexity for anyone building from source.  It would create the possibility of a future update to that package breaking OpenMM.  It would potentially bring in still more dependencies, if that package has any of its own, and create opportunities for version conflicts.

Since it's just a few source files, embedding them here makes things easier and avoids problems.

RaulPPelaez:
This PR is ready for review, all tests are passing.

peastman:
The API is still inconsistent with the other file classes.  As far as possible, someone should be able to take their existing code that uses DCDFile, replace the class name, and have it continue to work.  Even when that isn't possible, the structure, method names, and arguments should be as close as possible.

I'm not sure whether we want to support reading XTC files, or if that just adds unnecessary complexity.  Other packages like MDTraj and MDAnalysis already do a good job of that.  OpenMM doesn't provide readers for other trajectory formats, and trajectory analysis is really out of scope for it.

wrappers/python/openmm/app/xtcreporter.py
```diff
@@ -0,0 +1,77 @@
...
+        box = state.getPeriodicBoxVectors().value_in_unit(nanometers)
+        if (
+            box[0][1] != 0
+            or box[0][2] != 0
+            or box[1][0] != 0
+            or box[1][2] != 0
+            or box[2][0] != 0
+            or box[2][1] != 0
+        ):
+            raise Exception("XTCReporter does not support triclinic boxes")
```
peastman, line 71: I don't think that's true.  From https://manual.gromacs.org/archive/5.0.4/online/xtc.html:

> float box[3][3] 
> The computational box which is stored as a set of three basis vectors, to allow for triclinic PBC. For a rectangular box the box edges are stored on the diagonal of the matrix.

giadefa:
My views:
- Just writing XTC files is fine
- I agree that for a user it should be as easy as substituting the names
DCDxxxxx to XTCxxxxx in the python code of one simulation.

wrappers/python/openmm/app/xtcreporter.py
RaulPPelaez: The XTC format supports triclinic, but the reader/writer I took from moleculekit does not support it. Should be an easy change, though.

RaulPPelaez:
I pushed a commit trying to more closely imitate the naming scheme of dcdfile.
I removed the reading functionality.
The only difference I was not able to overcome is the fact that DCDFile takes a file, but the XTC parser I have requires a file name instead.

sef43:
The key benefit of having the xtc reader is that the tests can check the contents of the xtc files are correct, without a reader the tests just check that no errors are thrown when an xtc file is written.

peastman:
Since xtclib includes functions for reading, the test can always call them directly even if XTCFile doesn't wrap them in a public API.

RaulPPelaez:
The test now uses the internal library to read back a file and ensure the contents are correct.

RaulPPelaez:
@peastman could you review again? In particular the TextXtcfile.py in case I am doing something weird.

peastman:
I was waiting for you to make the changes discussed above: remove the interface for reading, make the API match DCDFile, and support triclinic boxes.

RaulPPelaez:
Thanks, @peastman .
I removed the reading interface, although I left some freestanding functions to facilitate testing in:
```python
from openmm.app.internal.xtc_utils
```
I made the reporter compatible with triclinic boxes, and added tests for it.
I followed the DCDFile interface as closely as I could. AFAICT they are now interchangeable.
Please review.

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -0,0 +1,156 @@
...
+import shutil
+
+
+class XTCFile(object):
+
+    """Parses an XTC file and provides access to the coordinates, box
+    vectors, time and step of each  frame.
+    This class also provides a method  to write new frames into an XTC
+    file from a set of coordinates, box vectors, time and step.
+    """
```
peastman, line 28: This description is no longer accurate.

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -0,0 +1,156 @@
...
+            The time step used in the trajectory
+        firstStep : int=0
+            The index of the first step in the trajectory
+        interval : int=1
+            The frequency (measured in time steps) at which states are written
+            to the trajectory
+        append : bool=False
+            If True, open an existing XTC file to append to.  If False, create a new file.
+        """
+        self._filename = file.name
```
peastman, line 49: I don't think we can depend on this working correctly.  Depending on how the file object was created, `name` might be a relative path, an absolute path, an integer referencing an OS-level file descriptor, or even just a human readable label like `'<stdout>'`.  If it was created by the `tempfile` module, there might not even exist any path for it.

Since XTCReporter already takes a filename instead of an open file object, and that's how most people will access it, it would probably be fine to just make this take a filename as well.

peastman (commented):
I've only done a brief skim through the C++ code.  I'm a bit confused about the index file.  If I understand correctly, it creates a second invisible file every time you write a XTC file?

giadefa:
Raul, you need to remove the index file. This is something that we use in
MoleculeKit only.

g

wrappers/python/openmm/app/internal/xtc_utils/include/xtc.h
```diff
@@ -0,0 +1,78 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+Contributors: Stefan Doerr, Raul P. Pelaez
+*/
+#ifdef PLATFORM_Linux
+    #if defined(__i386__) || defined(__x86_64__)
+__asm__(".symver memcpy,memcpy@GLIBC_2.2.5");
+    #endif
```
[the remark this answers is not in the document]
RaulPPelaez: I tried to leave the xtc parser code as unmodified as possible.
I removed those macros in a new commit.

RaulPPelaez:
Thanks guys, I removed the index file functionality in the xtc library. You may review again @peastman

wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
```diff
@@ -0,0 +1,326 @@
...
+    if (!natoms) {
+        fprintf(stderr, "xtc_read(): natoms is 0\n");
+        return -1;
+    }
+    xd = xdrfile_open(filename, "r");
+    if (NULL == xd) {
+        fprintf(stderr, "xtc_read(): could not open file\n");
+        return -1;
+    }
+    p = (rvec*)malloc(sizeof(rvec) * natoms);
```
peastman, line 87: Using malloc to create local workspace introduces a high risk of memory errors.  In C++, you should almost never explicitly allocate memory except in a constructor.  The correct way to do this is

```c++
vector<rvec> p(natoms);
```

You can call `p.data()` to get a pointer to the data.  The memory is guaranteed to be released correctly no matter where or how the function exits.

wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
```diff
@@ -0,0 +1,326 @@
...
+        frames[*nframes].step = step;
+        for (int i = 0; i < 3; i++) {
+            for (int j = 0; j < 3; j++) {
+                frames[*nframes].box[3 * i + j] = box[i][j];
+            }
+        }
+        frames[*nframes].pos = (float*)malloc(sizeof(float) * 3 * *natoms);
+        if (!(frames[*nframes].pos)) {
+            fprintf(stderr, "xtc_read(): Allocation error in xtc.c (2). nframes=%d natoms=%d\n", *nframes, *natoms);
+            return NULL;
```
peastman, line 195: Here's a perfect example of what I was saying about memory allocation.  You just leaked the memory pointed to by `p`.  C++ has better ways of doing memory allocation that eliminate the possibility of errors like this.

wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
```diff
@@ -0,0 +1,326 @@
...
+        fprintf(stderr, "xtc_read(): XTC file is corrupt\n");
+    }
+}
+
+struct XTC_frame* xtc_read(char* filename, int* natoms, int* nframes, double* dt, int* dstep) {
+    struct XTC_frame* frames = NULL;
+    *natoms = 0;
+    *nframes = 0;
+    rvec* p = NULL;
+    XDRFILE* xd = NULL;
```
peastman, line 157: This is another C-ism that can be handled better in C++.  You declare all your variables at the top of the function and initialize them to meaningless values.  Later on you assign the real values to them.  This creates a risk of bugs: there's a range of lines where the variable is declared and the compiler will let you use it, but any attempt to do so will probably create a bug.  It's better not to declare it until you're ready to assign the real value:

```c++
XDRFILE* xd = xdrfile_open(filename, "r");
```

That way, any attempt to reference `xd` before its value has been assigned will produce a compiler error.

wrappers/python/openmm/app/internal/xtc_utils/xtc.pyx
```diff
@@ -0,0 +1,122 @@
...
+    """ You need to pass the string with filename.encode("UTF-8") to this function """
+    return xtclib.xtc_nframes(filename)
+
+def get_xtc_natoms(char* filename):
+    """ You need to pass the string with filename.encode("UTF-8") to this function """
+    return xtclib.xtc_natoms(filename)
+
+
+@cython.boundscheck(False) # turn off bounds-checking for entire function
+@cython.wraparound(False)  # turn off negative index wrapping for entire function
```
peastman, line 56: Why do you disable error checking?

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -0,0 +1,160 @@
...
+            # update the file to say the trajectory consisted of a smaller number of larger steps (so the
+            # total trajectory length remains correct).
+            self._firstStep //= self._interval
+            self._dt *= self._interval
+            self._interval = 1
+            with tempfile.NamedTemporaryFile() as temp:
+                nframes = self._getNumFrames()
+                for i in range(nframes):
+                    read_frame = np.array([i]).astype(np.int32)
+                    _coords, _box, _time, _step = read_xtc_frames(
```
peastman, line 120: You don't need to prefix local variables with `_`.  It's confusing.  The `_` prefix is used to indicate private fields or methods of a class, not local variables.

wrappers/python/openmm/app/xtcreporter.py
```diff
@@ -0,0 +1,75 @@
...
+
+    To use it, create a XTCReporter, then add it to the Simulation's list of reporters.
+    """
+
+    def __init__(self, fileName, reportInterval, append=False, enforcePeriodicBox=None):
+        """Create a XTCReporter.
+
+        Parameters
+        ----------
+        fileName : string
```
peastman, line 15: This should be called `file` to be consistent with DCDReporter.

wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
```diff
@@ -0,0 +1,326 @@
...
+Contributors: Stefan Doerr, Raul P. Pelaez
+*/
+#include "xtc.h"
+#include "xdrfile.h"
+#include "xdrfile_xtc.h"
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
peastman, line 33: It's better to use the C++ versions of these headers, for example `<cstdio>` instead of `<stdio.h>`.  They put all the functions in the correct namespace.

peastman (commented):
I made a first pass through the C++ code.  A general observation is that you're writing C code in C++.  The filename ends in .cpp and you tell cython to compile it as C++, but for all that it's basically C!  You use pointers instead of references, malloc instead of STL containers, etc.  In a lot of cases, there are cleaner and safer ways of doing these things in C++.

giadefa:
I believe that it's actually C code, just compiled with the C++ compiler.
The XTC lib comes in C.

g

peastman:
Your C code has memory leaks. :)  Also a potential buffer overflow in the code that picks the index file name.

My comments were referring to the wrapper around xtclib (xtc.cpp).  I haven't looked at the xtclib code itself.  Since that's the same code used in Gromacs, I assume it works correctly.

RaulPPelaez:
I tried to leave the sources in xtc_utils as untouched as possible, as I assumed they would be well tested from moleculekit. Given the many issues the wrapper presents in its C'ish state I decided to just rewrite it in actual C++.

The underlying xtclib is still C so some ugly casts are required, but other than that RAII should avoid any leaks. I also do not know how to make cython work with std::vector (at least without copying), so that connection is still done via raw pointers.

I also removed several unused functions.

peastman (commented):
Much better!  I made just a few more minor comments.

We're getting some build failures.  The Mac build fails with

```
openmm/app/internal/xtc_utils/src/xtc.cpp:147:5: error: unknown type name 'constexpr'
    constexpr float prec = 1000;
    ^
```

This probably just needs a minor change to setup.py to tell it to use C++11.  Or just make it `const` instead of `constexpr`.  You're just setting it to a number, not anything that needs evaluating.

On Windows, all the tests for XTCFile are crashing:

```
_______________________________ TestXtcFile.py ________________________________
[gw0] win32 -- Python 3.11.3 D:\conda-envs\build\python.exe
worker 'gw0' crashed while running 'TestXtcFile.py::TestXtcFile::testAppend'
_______________________________ TestXtcFile.py ________________________________
[gw2] win32 -- Python 3.11.3 D:\conda-envs\build\python.exe
worker 'gw2' crashed while running 'TestXtcFile.py::TestXtcFile::test_xtc_box_from_topology'
_______________________________ TestXtcFile.py ________________________________
[gw3] win32 -- Python 3.11.3 D:\conda-envs\build\python.exe
worker 'gw3' crashed while running 'TestXtcFile.py::TestXtcFile::test_xtc_triclinic'
_______________________________ TestXtcFile.py ________________________________
[gw4] win32 -- Python 3.11.3 D:\conda-envs\build\python.exe
worker 'gw4' crashed while running 'TestXtcFile.py::TestXtcFile::test_xtc_cubic'
_______________________________ TestXtcFile.py ________________________________
[gw5] win32 -- Python 3.11.3 D:\conda-envs\build\python.exe
worker 'gw5' crashed while running 'TestXtcFile.py::TestXtcFile::testLongTrajectory'
```

I'm not sure what's causing that.

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -0,0 +1,156 @@
...
+        Parameters
+        ----------
+        positions : list
+            The list of atomic positions to write
+        unitCellDimensions : Vec3=None
+            The dimensions of the crystallographic unit cell.
+        periodicBoxVectors : tuple of Vec3=None
+            The vectors defining the periodic box.
+        """
+        if len(list(self._topology.atoms())) != len(positions):
```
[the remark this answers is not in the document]
RaulPPelaez: I took this verbatim from dcdfile:
https://github.com/openmm/openmm/blob/c7efe3d48d005fa00124b4a13722f5df6b86b855/wrappers/python/openmm/app/dcdfile.py#L120

I fixed this there as well, lmk if you'd rather me revert it.

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -0,0 +1,156 @@
...
+        ):
+            # This will exceed the range of a 32 bit integer.  To avoid crashing or producing a corrupt file,
+            # update the file to say the trajectory consisted of a smaller number of larger steps (so the
+            # total trajectory length remains correct).
+            self._firstStep //= self._interval
+            self._dt *= self._interval
+            self._interval = 1
+            with tempfile.NamedTemporaryFile() as temp:
+                nframes = self._getNumFrames()
+                coords, box, time, step = read_xtc(self._filename.encode("utf-8"))
```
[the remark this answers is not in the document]
RaulPPelaez: AFAIK XTC stores a constant number of atoms, but the rest of the parameters (box, step, time, coords and box) are stored per step. This means that this functionality requires going over every step.
There is this header comment in the xdrfile.h, which furthermore does not expose any functions to modify an existing file beyond appending frames:
```c++
/*
...
* We do not support getting or setting positions in XDR files, since it can
 * break in horrible ways for large (64-bit) files, resulting in silent data
 * corruption. Note that it works great to open/read/write 64-bit files if
 * your system supports it; it is just the random access we cannot trust!
```
I agree loading everything into memory is not a good idea, though. I implemented a function to read and write frame by frame C++ side, without loading the whole file into memory.
Still this requires storing the trajectory twice while this rewriting happens,  but I believe this is what it is.

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -0,0 +1,156 @@
...
+                if is_quantity(unitCellDimensions):
+                    unitCellDimensions = unitCellDimensions.value_in_unit(nanometers)
+                boxVectors = (
+                    Vec3(unitCellDimensions[0], 0, 0),
+                    Vec3(0, unitCellDimensions[1], 0),
+                    Vec3(0, 0, unitCellDimensions[2]),
+                )
+            boxVectors = np.array(
+                [[vec.x, vec.y, vec.z] for vec in boxVectors], dtype=np.float32
+            )
```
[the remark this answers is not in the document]
RaulPPelaez: I tried this, but I get:
```python
TypeError: Quantity.__float__ returned non-float (type Quantity)
....
boxVectors = np.array(boxVectors, dtype=np.float32)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: setting an array element with a sequence.
```

wrappers/python/openmm/app/xtcfile.py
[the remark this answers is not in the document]
RaulPPelaez: I mean storing the trajectory twice in the HDD, not in RAM.

wrappers/python/openmm/app/xtcfile.py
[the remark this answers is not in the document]
peastman: Thanks!

wrappers/python/openmm/app/xtcfile.py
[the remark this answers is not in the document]
peastman: That's probably the best we can do then.  A trajectory that long could easily be up into the terabyte range.  Hopefully they'll have enough disk space to briefly store a duplicate copy.

peastman:
That fixed the compilation error on Mac, but one test crashed:

```
________________________________ TestXtcFile.py ________________________________
[gw1] darwin -- Python 3.11.3 /Users/runner/miniconda3/envs/build/bin/python
worker 'gw1' crashed while running 'TestXtcFile.py::TestXtcFile::testLongTrajectory'
```

Windows still has the same crashes as before.

wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
```diff
@@ -44,7 +45,7 @@ static size_t Zf(size_t yidx, size_t nframes) {
 // Helper struct to manage the XDRFILE pointer
 struct XDRFILE_RAII {
     XDRFILE* xd;
-    XDRFILE_RAII(const char* filename, const char* mode) : xd(xdrfile_open(filename, mode)) {
+    XDRFILE_RAII(std::string filename, std::string mode) : xd(xdrfile_open(filename.data(), mode.data())) {
```
peastman, line 48: To convert a C++ string to a `char*`, call `.c_str()`.  The pointer returned by `.data()` isn't guaranteed to be null terminated.

RaulPPelaez: The evil person that wrote the function read_xtc_natoms made it so that it takes char*, refusing to accept filename.c_str().
It allows me to do this, though:
```c++
read_xtc_natoms(const_cast<char*>(filename.c_str()), &natoms)
```

wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    // Read the next frame from the XTC file and store it in this object
+    int readNextFrame(XDRFILE* xd) {
+        float in_prec;
+        auto* p_ptr = reinterpret_cast<rvec*>(positions.data());
+        int status = read_xtc(xd, natoms, &step, &time, box, p_ptr, &in_prec);
+        if (prec != in_prec) {
+            throw std::runtime_error("xtc_read(): precision mismatch\n");
+        }
```
peastman, line 89: I tracked down the test failure on Mac.  It's caused by these lines.  When you reach the end of the file, that's signaled by `read_xtc()` returning `exdrENDOFFILE`.  In that case, `in_prec` is uninitialized and doesn't match `prec`, leading it to throw an exception.

To fix it, change the check to `if (status == exdrOK && prec != in_prec)`.

RaulPPelaez: Amazing catch! would not have looked there for a long time -.- Lets see if it passes now.

peastman:
The problem is in the different handling of temporary files on Windows versus Linux.  From [the API docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) for `NamedTemporaryFile()`:

>  Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows).

Since the temporary file is already open, trying to open it again to write to it fails.  The fix is quite easy.  Just change

```python
with tempfile.NamedTemporaryFile() as temp:
    fname = temp.name
```

to

```python
with tempfile.TemporaryDirectory() as temp:
    fname = os.path.join(temp, 'traj.xtc')
```

The other issue, which I haven't investigated, is that the exception isn't getting handled correctly.  The C++ exception should have gotten translated into a Python exception with the message, "xtc file: Could not open file".  Instead it crashes with the much less useful message, "Fatal Python error: Aborted".

RaulPPelaez:
Oh well... Not used to Windows quirks. Sorry about that.
The exception translation does not arises in Linux as far as I can tell. Maybe in Windows the error is actually arising python side?
The XTC library requires the file to exist (it always opens the traj file with "append"), I added a line to create it if required. Hopefuly this fixes the windows ci.

wrappers/python/openmm/app/xtcfile.py
```diff
@@ -65,7 +65,8 @@ def __init__(self, fileName, topology, dt, firstStep=0, interval=1, append=False
         else:
             if os.path.isfile(self._filename) and os.path.getsize(self._filename) > 0:
                 raise FileExistsError(f"The file '{self._filename}' already exists.")
-
+            #XTC requires the file to exist before writing
+            open(self._filename, "a").close()
```
peastman, line 69: This isn't necessary.  `fopen()` with append mode already creates the file if it doesn't exist.

RaulPPelaez:
Lets see now!

RaulPPelaez:
I missed an instance of NamedTemporaryFile in xtcfile.py and a Windows test failed. It should work now.
An AMD OpenCL test also fails, but I do not think thats on me -.-

peastman:
All tests are now passing, except one unrelated stochastic failure.  Does this mean it's ready to merge?

peastman:
The test failure meant that it never got as far as running the python tests on Windows.  I'm rerunning that build just to be sure.

peastman:
One Python test failed, but I don't think it's related.

RaulPPelaez (commented):
This is ready to merge on my part.
Although I find it worrying, I do not think the failing test is related to any changes I made in this PR. Not sure what to do about it -.-
I improved the documentation of some functions.
I did not add this to the docs, would you like me to add the information for this reporter here ( https://github.com/openmm/openmm/blob/master/docs-source/usersguide/application/02_running_sims.rst#writing-trajectories ) before merging?

peastman:
> would you like me to add the information for this reporter here

Yes, that would be great.  Thanks.

RaulPPelaez:
I made just small additions, if it is ok then lets merge!

peastman:
Looks good.  Thanks for the excellent new feature!

RaulPPelaez:
Thanks! @peastman, in case you did not noticed, the Windows ci failed again but this time it was this:
```shell
12/125 Test   #9: TestReferenceCustomCVForce ....................Exit code 0xc0000374
***Exception:   0.41 sec
```
It seems that random tests fail whenever the CI runs -.-
Looking around, this error code signals "heap corruption".
I see some instances online where this error is caused by:
 - Cython https://bugs.python.org/issue35750
 - A default option related to regional UTF support in Windows 10 https://stackoverflow.com/questions/61935980/process-finished-with-exit-code-1073740940-0xc0000374 https://youtrack.jetbrains.com/issue/PY-33045
 This second one is really freaky, tbh.

Issue #3941, opened 2023-02-02
Investigating Mac AMD GPU performance


philipturner:
Continuation of discussion from #2489. We are troubleshooting AMD GPU performance on macOS, Bootcamp, and the Metal Plugin.

cc: @theschles

philipturner:
First of all, I fear the old Mac OpenMM backend is using the iGPU instead of the AMD GPU. Second, it seems to report `i386` instead of `x86_64`. I'll investigate this soon.

philipturner:
Okay, my Metal backend currently has an unknown bug impacting x86. I'll address that as soon as possible. On my Intel Mac mini with OpenCL, `mixed` precision and/or forcing to use CPU OpenCL device (`--device 0`) failed. OpenMM doesn't permit the CPU OpenCL device on Intel Macs. My iGPU got similar performance to your statistics' OpenCL single. That means it's automatically choosing the Intel GPU. Mixed precision has higher performance; it selected the AMD GPU because that's the only device supporting `mixed`.

Turner Mac Mini | CPU | OpenCL (iGPU)
-- | -- | --
gbsa | 22.50 | 54.37
rf | 27.55 | 30.33
pme | 24.23 | 24.34
apoa1rf | 7.62 | 7.52
apoa1pme | 6.47 | 5.38
apoa1ljpme | 5.73 | 4.62
amoebagk | aborted | 0.12
amoebapme | aborted | 0.09
amber20-dhfr | 27.97 | 28.02
amber20-cellulose | 1.51 | 0.76
amber20-stmv | aborted | aborted

> OpenCL system power was always around 29-31 W. I stopped measuring it because other processes were conflating measurements. By itself, the CPU was ~3.2 W.

Schlesinger | CPU | OpenCL (single, iGPU) | OpenCL (mixed, dGPU)
-- | -- | -- | --
gbsa | 16.52 | 45.72 | 67.29
rf | 19.13 | 26.37 | 33.72
pme | 15.43 | 21.55 | 24.85
apoa1rf | 5.00 | 6.50 | 9.87
apoa1pme | 4.47 | 4.60 | 7.66
apoa1ljpme | 3.88 | 4.05 | 6.58
amoebagk | 0.01 | 0.11 | 0.38
amoebapme | 0 | 0.04 | 0.51
amber20-dhfr | 18.12 | 24.51 | 26.37
amber20-cellulose | 1.04 | 0.63 | 1.18
[...]

PR #4009, opened 2023-03-24
[macOS GPU Support] Recognize AMD GPUs on macOS

philipturner:
Properly recognize AMD GPUs on x86-based Macs. The [Metal plugin](https://github.com/philipturner/openmm-metal) treats RDNA differently from GCN to maximize performance. This contribution to the OpenMM main branch instead errs on the side of caution.

#### Future Steps

- Test whether the [OpenCL integration utilities optimization](https://github.com/openmm/openmm/blob/3dc2f12ff20efcb609dfa5508431065d2c98d028/platforms/opencl/src/OpenCLIntegrationUtilities.cpp#L33-L38) will work. We should implement this in another PR, to decrease the number of confounding variables in the current PR.
- Test whether the flushing optimization will work.

Refs: https://github.com/openmm/openmm/pull/3911, https://github.com/openmm/openmm/issues/3941

cc: @bdenhollander, @theschles

### DIFF
--- a/platforms/opencl/src/OpenCLContext.cpp
+++ b/platforms/opencl/src/OpenCLContext.cpp
@@ -76,6 +76,7 @@ static void CL_CALLBACK errorCallback(const char* errinfo, const void* private_i
 static bool isSupported(cl::Platform platform) {
     string vendor = platform.getInfo<CL_PLATFORM_VENDOR>();
     return (vendor.find("NVIDIA") == 0 ||
+            vendor.find("AMD") == 0 ||
             vendor.find("Advanced Micro Devices") == 0 ||
             vendor.find("Apple") == 0 ||
             vendor.find("Intel") == 0);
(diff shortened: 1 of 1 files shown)


platforms/opencl/src/OpenCLContext.cpp
```diff
@@ -149,6 +150,9 @@ OpenCLContext::OpenCLContext(const System& system, int platformIndex, int device
                 if (devices[i].getInfo<CL_DEVICE_TYPE>() != CL_DEVICE_TYPE_GPU) {
                     processingElementsPerComputeUnit = 1;
                 }
+                else if (vendor.size() >= 5 && vendor.substr(0, 5) == "Apple") {
+                  processingElementsPerComputeUnit = 4 * 32;
+                }
```
bdenhollander, line 155: `vendor` is undefined here. `platformVendor` corresponds to `CL_PLATFORM_VENDOR` and `vendor` would be `CL_DEVICE_VENDOR` to match variable names after Line 216.

philipturner: I made a new variable called `deviceVendor` to explicitly differentiate the two versions of "vendor". This differs from the subsequent control flow block.

peastman:
I'm closing this old PR since it's now obsolete.  It's been long enough since Apple made x86 Macs that it's not worth adding complexity for them.

PR #4028, opened 2023-04-07
Support GROMACS files that use GROMOS force fields

### LINKED ISSUE #3901: Support another GROMACS bond type
I'm trying to run a simulation using gromacs topology and coordinate files.  When I try to load the topology file I get the following error:

.....
  File "gromacstopfile.py", line 247, in _processLine
    self._processBondType(line)
  File "gromacstopfile.py", line 406, in _processBondType
    raise ValueError('Unsupported function type in [ bondtypes ] line: '+line)
ValueError: Unsupported function type in [ bondtypes ] line:  S      S       2    0.2040  5.0300e+06

Note, I tried all of the gromos force field versions available on gromacs 2018.  And all of them yield the same error above.  A gromcas topology using the OPLS force field worked fine.  The protein is lysozyme.  All of the amino acids are standard.

I did look at the gromacstopfile.py code and it does indeed require the third column to equal 1.   Furthermore, the function _processAngleType only accepts function type values of 1 or 5.  

In my case, I need the following:
bondtype functiontype 2
angletype functiontype 2

Would it be possible to add those two function types to the code?  

Thank you;
-Sergio

peastman:
Fixes #3901.  This adds a number of features needed to support GROMOS force fields: a new bond type, a new angle type, combining NBFIX terms with geometric combination rules, and fixes to building exclusions when `gen-pairs` is set to `no`.


wrappers/python/openmm/app/gromacstopfile.py
```diff
@@ -318,8 +356,8 @@ def _processBond(self, line):
         fields = line.split()
         if len(fields) < 3:
             raise ValueError('Too few fields in [ bonds ] line: '+line)
-        if fields[2] != '1':
-            raise ValueError('Unsupported function type in [ bonds ] line: '+line)
+        if fields[2] not in ('1', '2'):
+                raise ValueError('Unsupported function type in [ bonds ] line: '+line)
```
lhmzhou, line 360: nit: indented too far

wrappers/python/openmm/app/gromacstopfile.py
```diff
@@ -329,8 +367,8 @@ def _processAngle(self, line):
         fields = line.split()
         if len(fields) < 4:
             raise ValueError('Too few fields in [ angles ] line: '+line)
-        if fields[3] not in ('1', '5'):
-            raise ValueError('Unsupported function type in [ angles ] line: '+line)
+        if fields[3] not in ('1', '2', '5'):
+                raise ValueError('Unsupported function type in [ angles ] line: '+line)
```
lhmzhou, line 371: second nit: indented too far

wrappers/python/openmm/app/gromacstopfile.py
swails, line 360:
```suggestion
            raise ValueError('Unsupported function type in [ bonds ] line: '+line)
```

wrappers/python/openmm/app/gromacstopfile.py
swails, line 371:
```suggestion
            raise ValueError('Unsupported function type in [ angles ] line: '+line)
```

PR #4034, opened 2023-04-12
Prevent molecules from accidentally getting wrapped by barostat

### LINKED ISSUE #4032: MonteCarloBarostat reimages molecules
I'm using the following script to simulate a box of 60 molecules. Each molecule contains 50 particles. When `MonteCarloBarostat` is applied, some molecules are jumping between cells. This jumping does not happen if the barostat is commented out. It seems that the barostat is forcing the center of each molecule to stay in the main cell.

```python
import sys
import numpy as np
from openmm import openmm as mm, app

pdb = app.PDBFile('omm-top.pdb')
with open('omm-sys.xml')  as f:
    system = mm.XmlSerializer.deserializeSystem(f.read())
system.addForce(mm.MonteCarloBarostat(1.0, 300.0))

integrator = mm.LangevinMiddleIntegrator(300.0, 1.0, 0.01)
platform = mm.Platform.getPlatformByName('CUDA')
properties = {'CudaPrecision': 'mixed'}
sim = app.Simulation(pdb.topology, system, integrator, platform, properties)
sim.context.setPositions(pdb.positions)
sim.minimizeEnergy(tolerance=100.0)

sim.reporters.append(app.StateDataReporter(sys.stdout, 1000, step=True, temperature=True, density=True, separator='\t'))
sim.reporters.append(app.DCDReporter('dump.dcd', 1000, enforcePeriodicBox=False))
positions_prev = sim.context.getState(getPositions=True, enforcePeriodicBox=False).getPositions(asNumpy=True)._value
for i in range(10):
    sim.step(1000)
    state = sim.context.getState(getPositions=True, enforcePeriodicBox=False)
    positions = state.getPositions(asNumpy=True)._value
    box_length = state.getPeriodicBoxVectors()._value[0][0]
    if np.any(positions - positions_prev > box_length * 0.8):
        print('  Jump')
    positions_prev = positions
```

These are two adjacent snapshots taken from the DCD output.

image

I'm using OpenMM 8.0.0 installed from conda-forge. The PDB and XML files to reproduce the issue are attached.

[pdb-xml.zip](https://github.com/openmm/openmm/files/11208531/pdb-xml.zip)

peastman:
Fixes #4032.

### DIFF
--- a/platforms/common/include/openmm/common/CommonKernels.h
+++ b/platforms/common/include/openmm/common/CommonKernels.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2008-2022 Stanford University and the Authors.      *
+ * Portions copyright (c) 2008-2023 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -1576,6 +1576,7 @@ private:
     ComputeArray moleculeStartIndex;
     ComputeKernel kernel;
     std::vector<int> lastAtomOrder;
+    std::vector<mm_int4> lastPosCellOffsets;
 };
 
 } // namespace OpenMM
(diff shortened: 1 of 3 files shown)


peastman:
@swails in the last few days the Jenkins OpenCL build has started failing.  I can't tell from the log what's wrong.  As far as it runs, that build looks exactly like the CUDA build (which works).  But it only gets as far as the `docker inspect -f . "$JD_TO_RUN"` stage, then doesn't run any more stages and reports a failure.

Issue #4063, opened 2023-05-05
O(n^2) scaling at several million atoms


philipturner:
I've been comparison some different MD libraries and came across a concern about OpenMM performance. amber20-stmv is slower than amber20-cellulose by a factor larger than the ratio of atoms, even though it has less % atoms with complex carbon-carbon bonds. I hope this is just because solvent molecules are more expensive than non-solvent, and O(n^2) scaling is not kicking in. I looked back at the performance improvement for several optimizations to OpenMM, and the ones impacting findBlocksWithInteractions had disproportionally more % improvement for larger AMBER systems. This could indicate that findBlocksWithInteractions is taking a significantly larger share of the compute time.

vs 8.0.0 | Atoms | Opt. 1 | Opt. 2 | Opt. 3 | Opt. 4
-- | -- | -- | -- | -- | --
gbsa | <s>2489</s> 23558 | 43% | 140% | 139% | 141%
rf |  23558 | 99% | 175% | 196% | 211%
pme |  23558 | 66% | 106% | 116% | 123%
apoa1rf | 92228 | 174% | 207% | 230% | 260%
apoa1pme |  92228 | 104% | 119% | 128% | 139%
apoa1ljpme |  92228 | 79% | 91% | 97% | 103%
amoebagk | 23558 | 48% | 48% | 48% | 49%
amoebapme | 23558 | 46% | 47% | 49% | 50%
amber20-dhfr |  <b>23558</b> | 61% | 103% | 109% | <b>112%</b>
amber20-cellulose | <b>408609</b> | 143% | 150% | 165% | <b>237%</b>
amber20-stmv | <b>1067095</b> | 174% | 177% | 206% | <b>392%</b>

I am concerned because I'm speculating whether to simulate up to 10-30 million atoms (10-100 ps / 6 hours) and anticipate O(n^2) dominating by 30 M. I'm also hoping for 1-3 million atoms (1-3 ns / 6 hours) and 300K atoms (10 ns / 6 hours). Long-range forces should use RF (O(n)) instead of PME (O(nlogn)).

At 2 million atoms, the gap between OpenMM and GROMACS has closed, even though OpenMM is using 4 fs and GROMACS is using 2 fs. GROMACS uses an O(nlogn) pairwise list which is costlier for smaller systems, but scales to much larger ones. I don't really have an option to use GROMACS though because I need the customizability of OpenMM. Hopefully it is just because LJ force computations take most of the time @ 2 M, regardless of which package is used.

[...]

PR #4066, opened 2023-05-08
Store bounding box sizes in half precision

peastman:
I finally found a use for half precision, though it's just for storage, not computation.  This reduces memory bandwidth when building the neighbor list.  For very large systems it can produce a bit of a speedup.  For smaller systems the effect is negligible.  See https://github.com/openmm/openmm/issues/4063#issuecomment-1537514168.

cc @philipturner

### DIFF
--- a/platforms/cuda/src/kernels/findInteractingBlocks.cu
+++ b/platforms/cuda/src/kernels/findInteractingBlocks.cu
@@ -2,6 +2,40 @@
 #define BUFFER_SIZE 256
 
 /**
+ * To use half precision, we're supposed to include cuda_fp16.h.  Unfortunately,
+ * it isn't included in the search path automatically, and there's no reliable
+ * way to find where it's located on disk.  Instead we provide our own definitions
+ * for the few symbols we need.
+ */
+struct __align__(2) __half {
+    unsigned short x;
+};
+__device__ __half __float2half_ru(const float f) {
+    __half h;
+    asm("{cvt.rp.f16.f32 %0, %1;}" : "=h"(*reinterpret_cast<unsigned short *>(&h)) : "f"(f));
+    return h;
+}
+__device__ float __half2float(const __half h) {
+    float f;
+    asm("{cvt.f32.f16 %0, %1;}" : "=f"(f) : "h"(*reinterpret_cast<const unsigned short *>(&h)));
+    return f;
+}
+struct half3 {
+    __device__ half3(float3 f) {
+        // Round up so we'll err on the side of making the box a little too large.
+        // This ensures interactions will never be missed.
+        v[0] = __float2half_ru(f.x);
+        v[1] = __float2half_ru(f.y);
+        v[2] = __float2half_ru(f.z);
+    }
+    __device__ float3 toFloat3() const {
+        return make_float3(__half2float(v[0]), __half2float(v[1]), __half2float(v[2]));
+    }
+private:
+    __half v[3];
+};
+
+/**
  * Find a bounding box for the atoms in each block.
  */
 extern "C" __global__ void findBlockBounds(int numAtoms, real4 periodicBoxSize, real4 invPeriodicBoxSize, real4 periodicBoxVecX, real4 periodicBoxVecY, real4 periodicBoxVecZ,
@@ -53,12 +87,12 @@ extern "C" __global__ void findBlockBounds(int numAtoms, real4 periodicBoxSize, 
  */
 extern "C" __global__ void sortBoxData(const real2* __restrict__ sortedBlock, const real4* __restrict__ blockCenter,
         const real4* __restrict__ blockBoundingBox, real4* __restrict__ sortedBlockCenter,
-        real4* __restrict__ sortedBlockBoundingBox, const real4* __restrict__ posq, const real4* __restrict__ oldPositions,
+        half3* __restrict__ sortedBlockBoundingBox, const real4* __restrict__ posq, const real4* __restrict__ oldPositions,
         unsigned int* __restrict__ interactionCount, int* __restrict__ rebuildNeighborList, bool forceRebuild) {
     for (int i = threadIdx.x+blockIdx.x*blockDim.x; i < NUM_BLOCKS; i += blockDim.x*gridDim.x) {
         int index = (int) sortedBlock[i].y;
         sortedBlockCenter[i] = blockCenter[index];
-        sortedBlockBoundingBox[i] = blockBoundingBox[index];
+        sortedBlockBoundingBox[i] = half3(trimTo3(blockBoundingBox[index]));
     }
     
     // Also check whether any atom has moved enough so that we really need to rebuild the neighbor list.
@@ -181,7 +215,7 @@ extern "C" __global__ __launch_bounds__(GROUP_SIZE,3) void findBlocksWithInterac
         unsigned int* __restrict__ interactionCount, int* __restrict__ interactingTiles, unsigned int* __restrict__ interactingAtoms,
         int2* __restrict__ singlePairs, const real4* __restrict__ posq, unsigned int maxTiles, unsigned int maxSinglePairs,
         unsigned int startBlockIndex, unsigned int numBlocks, real2* __restrict__ sortedBlocks, const real4* __restrict__ sortedBlockCenter,
-        const real4* __restrict__ sortedBlockBoundingBox, const unsigned int* __restrict__ exclusionIndices, const unsigned int* __restrict__ exclusionRowIndices,
+        const half3* __restrict__ sortedBlockBoundingBox, const unsigned int* __restrict__ exclusionIndices, const unsigned int* __restrict__ exclusionRowIndices,
         real4* __restrict__ oldPositions, const int* __restrict__ rebuildNeighborList) {
 
     if (rebuildNeighborList[0] == 0)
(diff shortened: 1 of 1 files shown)


philipturner:
The Apple GPU excels in a lot of the control flow operations present in the $O(n^2)$ part. That means removing a memory bandwidth bottleneck should cause a greater % improvement than with NV.

With GPT-4, it might be possible for me to convert the OpenMM Metal plugin into actual Metal shaders. Can we isolate the GPU shader for `findBlocksWithInteractions` so I can run it through Metal in a Swift script? I need all but the absolutely necessary shader arguments to be elided.

peastman:
> Can we isolate the GPU shader for findBlocksWithInteractions so I can run it through Metal in a Swift script?

I'm not sure what you're asking.  Can you clarify?

The changes here are mostly trivial, just converting floats to half before saving them and converting back when loading.  The extra code at the top is just to work around some CUDA annoyances.

philipturner:
I think I found a workaround that works for OpenCL! This compiled in the Apple OpenCL driver. I tried it after experiencing the following error:

```
(base) philipturner@M1-Max-MacBook-Pro CompileCL % xcrun metal -c File.cl --driver-mode=openclc -o File.air
File.cl:7:17: error: loading directly from pointer to type '__global half' requires
      cl_khr_fp16. Use vector data load builtin functions instead
  if (*input1 > *input2) {
                ^
1 error generated.
(base) philipturner@M1-Max-MacBook-Pro CompileCL % vim File.cl 
(base) philipturner@M1-Max-MacBook-Pro CompileCL % xcrun metal -c File.cl --driver-mode=openclc -o File.air
```

float vload_half(size_t offset, const __global half *p) float vload_half(size_t offset, const __local half *p) float vload_half(size_t offset, const __constant half *p) float vload_half(size_t offset, const __private half *p)  								For OpenCL C 2.0, or OpenCL C 3.0 or newer with the __opencl_c_ generic_address_space feature:  								float vload_half(size_t offset, const half *p) | Read sizeof(half) bytes of data from the address computed as (p + offset). The data read is interpreted as a half value. The half value is converted to a float value and the float value is returned. The computed read address must be 16-bit aligned.
-- | --
floatn vload_halfn(size_t offset, const __global half *p) floatn vload_halfn(size_t offset, const __local half *p) floatn vload_halfn(size_t offset, const __constant half *p) floatn vload_halfn(size_t offset, const __private half *p)  								For OpenCL C 2.0, or OpenCL C 3.0 or newer with the __opencl_c_ generic_address_space feature:  								floatn vload_halfn(size_t offset, const half *p) | Read (n * sizeof(half)) bytes of data from the address computed as (p + (offset * n)). The data read is interpreted as a halfn value. The halfn value read is converted to a floatn value and the floatn value is returned. The computed read address must be 16-bit aligned.

OpenCL code:

```opencl
                                                                                 
__kernel void testKernel(__global float* input1,                                 
                         __global half* input2,                                  
                         __global int* output)                                   
{                                                                                
  if (*input1 > vload_half(0, input2)) {                                         
    output[0] = 1;                                                               
  }                                                                              
}  
```

philipturner:
Optimization 1: `real4` -> `real3` for positions and box sizes

Optimization 2: `real3` -> `half` for box sizes

Width (nm) | Atoms | Steps/4 | Original | Opt. 1 | Opt. 2 | Opt. 2 Time / Before
-- | -- | -- | -- | -- | -- | --
2 | 774 | 1000 | 889 | 893 | 889 | 1.004
3 | 2661 | 1000 | 668 | 668 | 669 | 0.999
4 | 6282 | 1000 | 506 | 505 | 505 | 1.000
5 | 12255 | 1000 | 328 | 328 | 327 | 1.003
7 | 33840 | 1000 | 220 | 220 | 220 | 1.000
10 | 98880 | 1000 | 88.9 | 88.9 | 88.8 | 1.001
15 | 335025 | 100 | 26.4 | 26.7 | 26.7 | 1.000
20 | 792450 | 50 | 11.1 | 11.0 | 11.3 | 0.982
25 | 1550160 | 25 | 5.01 | 5.03 | 5.18 | 0.971
30 | 2682600 | 15 |   | 2.38 | 2.51 | 0.948
35 | 4254045 | 10 | 1.20 | 1.14 | 1.30 | 0.923
40 | 6354990 | 10 | 0.692 |   | 0.763 | 0.907
45 | 9056475 | 5 | 0.369 |   | 0.443 | 0.833
apoa1rf | 92224 | 6800 | 151.7 | 151.6 | 151.7 | 1.000

https://github.com/philipturner/openmm-metal/commit/0b487005873c3bbb1c40520fe6cc4fd1b198f0b4

peastman:
Does that mean `vload_half()` and `vstore_half()` are available even without the extension?  I'm looking at the OpenCL 1.2 spec, and it's a little unclear.  There's a footnote in section 6.12.7 that says, "The half*n* type is only defined by the **cl_khr_fp16** extension described in section 9.5 of the OpenCL 1.2 Extension Specification."  Does that include the half type itself, or do only vectors of halfs require the extension?

Anyway, if it works on all platforms, great!

peastman:
I was able to implement half precision support in OpenCL.  It works, but it has no measurable effect on speed on either NVIDIA or Apple.

|Width|Atoms|Old|New|Ratio|
|---|---|---|---|---|
15|335025|10.03|9.97|0.99
20|792450|29.67|29.79|1.00
25|1550160|78.21|78.23|1.00

Here's my implementation.

```c
real3 loadHalf3(const __global half* h, size_t i) {
    return (real3) (vload_half(3*i, h), vload_half(3*i+1, h), vload_half(3*i+2, h));
}

void storeHalf3(real3 f, __global half* h, size_t i) {
    // Round up so we'll err on the side of making the box a little too large.
    // This ensures interactions will never be missed.
    vstore_half_rtp((float) f.x, 3*i, h);
    vstore_half_rtp((float) f.y, 3*i+1, h);
    vstore_half_rtp((float) f.z, 3*i+2, h);
}
```

Possibly the problem is that it has to load the three components as three separate 16 bit memory transactions, rather than a single 48 bit one.  Loading all three at once requires the fp16 extension.

philipturner:
> Loading all three at once requires the fp16 extension.

The OpenCL C specification does not state any such requirement for this function. The function does not use the `half3` type.

| Function | Description |
| ---- | ---- |
| floatn vloada_halfn(size_t offset, const __global half *p) floatn vloada_halfn(size_t offset, const __local half *p) floatn vloada_halfn(size_t offset, const __constant half *p) floatn vloada_halfn(size_t offset, const __private half *p) | For n = 2, 4, 8 and 16, read sizeof(halfn) bytes of data from the address computed as (p + (offset * n)). The data read is interpreted as a halfn value. The halfn value read is converted to a floatn value and the floatn value is returned. The computed address must be aligned to sizeof(halfn) bytes. For n = 3, vloada_half3 reads a half3 from the address computed as (p + (offset * 4)) and returns a float3. The computed address must be aligned to sizeof(half) * 4 bytes.

```opencl
__global half* restrict sortedBlockBoundingBox,
vstorea_half3_rtp(float3(blockBoundingBox[index]), i, sortedBlockBoundingBox);

__global const half* restrict sortedBlockBoundingBox,
real3 blockSizeY = real3(vloada_half3(block2, sortedBlockBoundingBox));
```

Works on Apple, which does not have `cl_khr_fp16`:

```
  printf() buffer size                            1048576 (1024KiB)
  Built-in kernels                                (n/a)
  Device Extensions                               cl_APPLE_SetMemObjectDestructor cl_APPLE_ContextLoggingFunctions cl_APPLE_clut cl_APPLE_query_kernel_names cl_APPLE_gl_sharing cl_khr_gl_event cl_khr_byte_addressable_store cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_3d_image_writes cl_khr_image2d_from_buffer cl_khr_depth_images 
```

peastman:
I guess it's ambiguous.  The footnote quoted above says the half*n* types are only available with the extension.  I interpreted that to include the functions with half*n* in their names, but maybe that isn't what they intended.  I tried rewriting it to use those functions.  It works, but still has no effect on performance.

philipturner:
Can I see your data? Look at the places in my table where the performance ratio was 0.95 or less.

peastman:
Here are the times for box width 35:

old: 433.5
new: 439.3

It's slightly slower, but probably within the noise level.

philipturner:
Perhaps it's because the in-tree backend doesn't have a lot of optimizations from the Metal plugin. For example, the bottleneck is currently from unnecessarily accessing threadgroup memory each iteration. The bottleneck is not from L2 or SLC bandwidth, but rather from activity that's encapsulated inside the GPU core. Making it more compute-bound (+3 instructions) will worsen this bottleneck.

Before:

```
4 consecutive 32-bit registers = READ 128 bytes 
  READ masked in 32-bit chunks by 1110
ignore the last register
```

After:
```
2 consecutive 32-bit registers = READ 64 bytes
  READ masked in 16-bit chunks by 1110
X = F32_ADD(LEFT=decompress 16-bit operand, RIGHT=0)
Y = F32_ADD(LEFT=decompress 16-bit operand, RIGHT=0)
Z = F32_ADD(LEFT=decompress 16-bit operand, RIGHT=0)
```

For the Metal plugin, the inner loop has so few instructions (no log2(32) TG reads + log2(32) TG writes + log2(32) LOOP ITERATE; just one SIMD reduction) that it's okay to put some extra load on the ALU.

philipturner:
I predict that at several million atoms, your M1 Pro w/ OpenCL will underperform my M1 Max w/ Metal by a factor greater than 4. Can you record the ns/day?

peastman:
That seems very likely.  It has 2x the compute units, and Metal is faster than OpenCL.

PR #4075, opened 2023-05-19
Add link to cookbook and tutorials in docs

sef43:
This PR adds a link to the cookbook and tutorials page from the side panel navigation links of the OpenMM documentation.

### DIFF
--- a/docs-source/api-c++/conf.py
+++ b/docs-source/api-c++/conf.py
@@ -61,6 +61,11 @@ html_theme_options = {
             "relative": True,
         },
         {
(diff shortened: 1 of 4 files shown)


peastman:
Would it be better to have two separate links, one of the cookbook and one for the tutorials?

sef43:
I think it's good just having the one link for both: 

- less clutter.
- They are both accessed from the same top level index page.
- The top level page explains the difference between the cookbook and the tutorials so users can properly decide which one they need.

PR #4081, opened 2023-05-23
Allow negative pressure in MonteCarloAnisotropicBarostat

### LINKED ISSUE #4076: MonteCarloAnisotropicBarostat doesn't allow negative pressure
`MonteCarloAnisotropicBarostat` does not allow negative pressures. The other MC Barostats do not seem to have this check. Is there a good reason for the Anisotropic one having this? I have a use case that needs negative pressure applied to the system and the box size to only change in one direction.

sef43:
Fixes #4076


peastman:
Looks good, thanks!.

PR #4104, opened 2023-06-08
addSolvent now only replaces newly added water molecules with ions

### LINKED ISSUE #4103: Exclude specific waters from ion replacement `Modeller.addSolvent`
In some systems, we have structurally significant waters (e.g. involved in binding site interactions), that we would like to keep during our simulations.

However, when calling `Modeller.addSolvent`, there is a chance that those waters could get picked up and replaced by ions in `_addIons`.

As far as I am aware, there is no way to avoid this beyond renaming your crystallographic waters to a residue name that isn't `HOH`. If this is indeed the case, I was wonder if it might be possible to consider an enhancement where we could either pass in a list of excluded residues, or maybe just tell `addSolvent` to not pick up any water residues that already existed ahead of calling it?

richardjgowers:
previously pre-existing (i.e. potentially crystallographic waters) were candidates for replacement with an ion, now only newly added waters can be replaced with an ion

concentration calculations (i.e. total water molecules) still include these pre-existing water molecules

possibly fixes #4103

Currently this is always on/changing behaviour, I could put this behaviour behind some `preserveExistingWater=True` kwarg if required

### DIFF
--- a/wrappers/python/openmm/app/modeller.py
+++ b/wrappers/python/openmm/app/modeller.py
@@ -619,9 +619,10 @@ class Modeller(object):
             addedWaters = filteredWaters
 
         # Add the water molecules.
-
+        newIds = set()
         for index, pos in addedWaters:
             newResidue = newTopology.addResidue(residue.name, newChain)
+            newIds.add(newResidue.id)
             residue = pdbResidues[index]
             oxygen = [atom for atom in residue.atoms() if atom.element == elem.oxygen][0]
             oPos = pdbPositions[oxygen.index]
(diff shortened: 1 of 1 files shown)


peastman:
I think this is the right way to fix it.  I suggest though that concentration should only be based on newly added molecules.  The concentration is supposed to refer to the bulk solvent.  Counter ions and crystallographic waters are different.  So whatever is already in the model we leave alone.  The ionic strength specified by the argument means to add solvent with that ionic strength, regardless of what is already there.

`addMembrane()` needs similar changes.  That case is a little more complicated because water gets added in two ways.  First, we tile copies of a box that includes both membrane and water.  Then we add extra water if needed to make the box big enough.  In that case, both types of water (from the membrane box and extra water) should be considered in calculating ionic strength and should be eligible for replacement with ions.

richardjgowers:
@peastman I think `addMembrane()` might be fine, it looks like it only loops over the last two chains to search for candidate waters: https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/modeller.py#L1565

Concentration is now only based off new water molecules.  I've put the picking up of candidate water molecules into the loop of adding them to the new topology, this saves looping over the system twice.

Just to double check, this is a change in behaviour.  I can put this behind an opt-in kwarg, or are we happy that existing waters are always left alone?

peastman:
Looks good.  Tests are running now.

> Just to double check, this is a change in behaviour.

Correct, but it's also how it used to work, and how it was always intended to work.  The previous change was unintended, and this puts it back.

peastman:
Thanks for the fix!

Issue #4040, opened 2023-04-21
Integrate Alchemical Transfer Method (ATM) plugin into OpenMM main?


jchodera:
I'd like to propose we consider bringing in @egallicc's excellent Alchemical Transfer Method (ATM) plugin into the main OpenMM repo.

ATM has now been extensively validated for computing absolute ([[1]](https://pubs.acs.org/doi/10.1021/acs.jctc.1c00266) [[2]](https://link.springer.com/article/10.1007/s10822-021-00437-y) [[3]](https://pubs.acs.org/doi/full/10.1021/acs.jctc.9b01119)) and relative ([[1]](https://pubs.acs.org/doi/full/10.1021/acs.jcim.3c00178) [[2]](https://pubs.acs.org/doi/full/10.1021/acs.jcim.1c01129)) binding affinities.

@egallicc has created an [OpenMM plugin](https://github.com/Gallicchio-Lab/openmm-atmmetaforce-plugin) that is currently distributed separately, but (1) ensuring full compatibility with OpenMM releases and conda packages is challenging, (2) the current [LGPL license](https://github.com/Gallicchio-Lab/openmm-atmmetaforce-plugin/blob/main/LICENSE) may limit use, and (3) there may be opportunities to further generalize the `ATMMetaForce` when bringing it in to OpenMM to further expand its domain of use.

Currently, the API supports creating an `ATMMetaForce` object that mixes the current potential `U(x)` and a translated potential `U(x + direction)` according to a fixed expression:
```python
atmforce = ATMMetaForce(lambda1, lambda2,  alpha * kilojoules_per_mole, u0/kilojoules_per_mole, w0coeff/kilojoules_per_mole, umsc/kilojoules_per_mole, ubcore/kilojoules_per_mole, acore, direction, atmvariableforcegroups )
```
The `atmvariableforcegroups` specifies the force groups represented by `U(x)`. 

The functional forms that the parameters `alpha`, `u0`, `w0coeff`, `umsc`, `ubcore`, `acore` appear in are fixed expressions (reproduced below).

Integration with OpenMM would solve (1), which would also allow us to deploy it easily on Folding@home.

With the permission of @egallicc, we could incorporate this with a license compatible with OpenMM (such at MIT) to solve (2).

Are there opportunities to further generalize the implementation for (3)? Some thoughts:
[...]

PR #4110, opened 2023-06-13
Draft integration of the Alchemical Transfer Method (ATM) plugin

egallicc:
WORK IN PROGRESS. DO NOT MERGE.

This is the first attempt at the integration of the [Alchemical Transfer plugin](https://github.com/Gallicchio-Lab/openmm-atmmetaforce-plugin) into OpenMM proper. See  #4040.

It implements a new Force called ATMForce. As of now, it has the same functionality as the plugin API.

To do/help needed:

- Implement addForce()/getForce() methods to add and get the Forces managed by the ATMForce. I attempted to do so using a vector of Forces but I could not retrieve them to pass them to the inner Context. I kept getting errors in ATMForceImpl.cpp related to const qualifiers. `error: passing 'const OpenMM::ATMForce' as 'this' argument discards qualifiers [-fpermissive]`

- Python swig wrappers. The plugin has functioning swig wrappers. However, I could not penetrate OpenMM's swig system to a sufficient extent to add wrappers to the ATMForce methods. Any help is appreciated.

- Tests. We need to come up with a simple testing system. In the meantime, I prepared an ATMHelloChloride example by hacking HelloSodiumChloride.cpp. It moves the first sodium atom away and it calculates the change in energy.

- The rest of the proposed API, including the perturbation potential energy expression.


egallicc:
Done:
- addForce()/getForce() methods 

To do/help appreciated:

- Python swig wrappers. 
- Tests.  ATMHelloChloride example for now.
- The rest of the proposed API, including the perturbation potential energy expression.

peastman:
I'm about to start work on this.  What's the best way of coordinating?  I could push changes to your branch, or create a new branch and send changes as a pull request to you, or just copy the code over and start a new pull request.

egallicc:
> I'm about to start work on this. What's the best way of coordinating? 

Thanks! Pushing to our branch would be great. I added you as a collaborator.

peastman:
What would make good test cases for this?  I'd like to do that before making any significant code changes, so I can be sure I'm not breaking anything.

In the mean time, I'll start with the Python API and general cleanup, formatting, documentation, etc.

peastman:
I tried to push some changes, but I got the error, "ERROR: Permission to Gallicchio-Lab/openmm.git denied to peastman."  Can you make sure I have the right permissions?

egallicc:
> Can you make sure I have the right permissions?

Done

egallicc:
> What would make good test cases for this?

Would the ATMHelloChloride example suffice for now?

peastman:
Thanks!  I have the Python API working.

> Would the ATMHelloChloride example suffice for now?

It doesn't test anything.  It just creates the system, writes out a single PDB frame, and then exits.  We need some unit tests that do calculations with ATMForce and verify that it produces correct results.

peastman:
[TestCustomCVForce.h](https://github.com/openmm/openmm/blob/master/tests/TestCustomCVForce.h) could be a good first one to look at, since this class has a lot of similarities to it.

peastman:
Here are a couple of ideas for unit tests.

Create a system with only two particles, one that gets translated and one that doesn't.  Add a HarmonicBondForce between them.  That case is simple enough that you can compute the expected results analytically.  Try a few different lambda values and for each one check that all of the following have the expected values.

- The forces and energy returned by `context.getState()`
- The value returned by `getPerturbationEnergy()`
- The derivative of the energy with respect to each global parameter.

Another useful test is to create a system with a large number of particles (maybe 1000), each translated differently.  Add a CustomExternalForce that is a nonlinear function of the particle position.  By checking the force on each particle, you can tell what position it was evaluated at.

peastman:
When you call `calcForcesAndEnergy()`, you pass a variable with a set of force groups:

```c++
  double State1Energy = innercontextimpl1.calcForcesAndEnergy(true, do_energy, variable_force_groups_mask);
```

However, the variable containing the mask is never initialized, so the result is nondeterministic.  Is this left over from the old method of specifying the forces?  Should the argument be removed?

peastman:
I created the structure for writing test cases for ATMForce.  Put them in TestATMForce.h and invoke them from `main()`.  A global variable called `platform` will be defined, which you should pass as the platform when creating contexts.  The tests will be built for Reference, OpenCL, and CUDA.  When running them with ctest, it will run the OpenCL and CUDA tests in all three precision modes.

egallicc:
> ```c++
>   double State1Energy = innercontextimpl1.calcForcesAndEnergy(true, do_energy, variable_force_groups_mask);
> ```
> 
>  Should the argument be removed?

Oops! Yes, `variable_force_groups_mask` should have been removed.

egallicc:
I added a first simple test with two particles tethered by a harmonic spring as suggested. More extensive tests to come.

tests/TestATMForce.h
```diff
@@ -0,0 +1,114 @@
...
+
+  State state = context.getState(State::Energy | State::Forces );
+  double epot = state.getPotentialEnergy();
+  double epert = atm->getPerturbationEnergy(context);
+
+  ASSERT_EQUAL_TOL(lmbd, context.getParameter(atm->Lambda1()), 1e-6);
+  ASSERT_EQUAL_TOL(lmbd, context.getParameter(atm->Lambda2()), 1e-6);
+  ASSERT_EQUAL_TOL(lmbd*0.5*displx*displx, epot, 1e-6);
+  ASSERT_EQUAL_TOL(0.5*displx*displx, epert, 1e-6);
+  ASSERT_EQUAL_VEC(Vec3(-lmbd*displx, 0.0, 0.0), state.getForces()[1], 1e-6);
```
peastman, line 95: I suggest putting these lines inside a loop that tests a few values of lambda.

```c++
for (double lmbd : {0.0, 0.5, 1.0}) {
    context.setParameter(ATMForce::Lambda1(), lmbd);
    context.setParameter(ATMForce::Lambda2(), lmbd);
    ...
}
```

That will verify that you have the right dependence on it.

peastman:
I added some more test cases.  Can you take a look and make sure they're correct?  Do you have any ideas for other tests?

Once we're confident that the test coverage is enough to catch any problems, I can start on the changes to allow customizing the energy expression.

I'm a little confused about one aspect.  Why do you create two different inner Systems, which appear to be identical, and then two different Contexts for them?  It looks to me like a single System and Context would work just as well.

egallicc:
> I added some more test cases. 

Thanks! I was away. I will look at them now and consider more tests. 

I see that you added a test for atom reordering. I have been doubtful about the implementation of the ReorderListeners in ATM. In applications, molecules being displaced are always restrained in some way and do not get reordered. Nevertheless, it is an aspect that probably needs a professional review.

> Once we're confident that the test coverage is enough to catch any problems, I can start on the changes to allow customizing the energy expression.

Yes, thank you, that would be the next step. We decided that the existing expression should be used if one is not provided by the user. Do you have a plan/suggestion on how to achieve this? 

> I'm a little confused about one aspect. Why do you create two different inner Systems, which appear to be identical, and then two different Contexts for them? It looks to me like a single System and Context would work just as well.

Probably because we don't know what we are doing. Yes, if feasible and simpler, we can work with one System. But it would be preferable to keep using two inner Contexts for future use. We are considering protocols in which we go from a state (assigned to the first Context) with one set of displacements to a state (assigned to the second Context) with another set of displacements. This will allow chaining perturbations to build large transformations easily.

peastman:
> We decided that the existing expression should be used if one is not provided by the user. Do you have a plan/suggestion on how to achieve this?

I was figuring on having two constructors.  One would look like the constructor for CustomCVForce: you would provide an expression for the energy, and then you could call `addGlobalParameter()` for any parameters it needed.  The other would be exactly the same as the current constructor.  You would provide default values for the standard parameters, and it would set an appropriate energy expression and add the required parameters.

> We are considering protocols in which we go from a state (assigned to the first Context) with one set of displacements to a state (assigned to the second Context) with another set of displacements.

That shouldn't be a problem for a single Context.  Set one set of positions, compute the force and energy, set the other set of positions, compute them again.

egallicc:
> 
> I was figuring on having two constructors. One would look like the constructor for CustomCVForce: you would provide an expression for the energy, and then you could call `addGlobalParameter()` for any parameters it needed. The other would be exactly the same as the current constructor. You would provide default values for the standard parameters, and it would set an appropriate energy expression and add the required parameters.

That would be one way. Does SWIG support overloaded constructors?

An alternative maybe cleaner solution is to prepare the default energy expression for the user in a string returned by `ATMForce::DefaultEnergyExpression()`. The constructor would be called in the same way:

Default energy expression
```
ATMForce* atm = new ATMForce(ATMForce::DefaultEnergyExpression());
```

Custom energy expression:
```
ATMForce* atm = new ATMForce("lambda*u");
```

In either case, the parameters of the energy expression would be set using `setParameter()` rather than in the constructor. Providing the parameters of our energy expression (`lambda1`, `lambda2`, etc.) in the constructor did not turn out to be very useful because they inevitably get changed by `setParameter()` calls to place the system in the desired alchemical state. The constructor could just set the default parameters to zero, which corresponds to the initial alchemical state.

egallicc:
> That shouldn't be a problem for a single Context. Set one set of positions, compute the force and energy, set the other set of positions, compute them again.

Wouldn't displacing the atoms trigger the recalculation of the neighbor list? That would slow things down considerably since the displacement is applied at every time step.

In the current approach, the displaced atoms end up in more or less the same position from one step to another, and the neighbor list for the "displaced" Context is recomputed infrequently.

egallicc:
> Do you have any ideas for other tests?

The tests look good to me. I do not have clever ideas for additional tests other than testing actual molecular complexes.

egallicc:
> > An alternative maybe cleaner solution is to prepare the default energy expression for the user in a string returned by ATMForce::DefaultEnergyExpression().
> 
> That's also a possibility. It would be nice if it could automatically add the parameters for you.
> 

The parameters of the default energy expression are already added. And their names are already defined by `ATMForce::Lambda1()`, `ATMForce::Lambda2()`, etc. The user would have to only set their values.

The default energy expression would look something like:

```
(( ATMLambda2 -  ATMLambda1)/ATMAlpha) * log( 1 + exp(- ATMAlpha * (usc - ATMU0) ) ) + ATMLambda2*usc + ATMW0 ;
usc = select(step( u - ATMUbcore), u, ( ATMUmax - ATMUbcore)*fsc + ATMUbcore );
fsc = (z^ATMAcore - 1)/(z^ATMAcore + 1);
z = 1 + 2*(y/ATMAcore) + 2*(y/ATMAcore)^2;
y = (u - ATMUbcore)/( ATMUmax - ATMUbcore)
```

where `ATMLambda2`, etc are the global parameters and `u` is the perturbation energy, the difference of the potential energy before and after the displacement.

> Do all the parameters tend to change with time, or are there some that get set at the start and not changed?

Typically they all change all the time, except those that define the soft-core perturbation energy: `ATMUmax`, `ATMUbcore`, `ATMAcore`.

peastman:
I think the default parameters should only get added if you use the default energy expression.  Adding a bunch of unused parameters would be confusing and could potentially cause problems.

> Typically they all change all the time, except those that define the soft-core perturbation energy: ATMUmax, ATMUbcore, ATMAcore.

How about a constructor that takes values for those three parameters?

egallicc:
> I think the default parameters should only get added if you use the default energy expression. Adding a bunch of unused parameters would be confusing and could potentially cause problems.

How about an `ATMForce::setDefaultEnergyExpression()` utility method that defines and sets the parameters and returns the expression?

```
ATMForce* atm = new ATMForce(  ATMForce::setDefaultEnergyExpression(lambda1, lambda2, alpha, u0, w0, umax, ubcore, acore, direction)  );
```

On that note, the ATM plugin has a zoo of [utility functions](https://github.com/Gallicchio-Lab/openmm-atmmetaforce-plugin/blob/main/python/ATMMetaForceUtils.py), mostly to set specialized restraints often used in binding free energy calculations, that I do not plan to transfer to OpenMM. They will be available as part of the [AToM-OpenMM](https://github.com/Gallicchio-Lab/AToM-OpenMM) middleware.

> How about a constructor that takes values for those three parameters?

Sure, but it would not avoid the overloaded constructor.

peastman:
> ATMForce* atm = new ATMForce(  ATMForce::setDefaultEnergyExpression(lambda1, lambda2, alpha, u0, w0, umax, ubcore, acore, direction)  );

I don't think I understand.  What does `setDefaultEnergyExpression()` return?  What is the argument to the constructor?  How do the parameters get added to the Force?

> Sure, but it would not avoid the overloaded constructor.

There's nothing wrong with overloading constructors.  We do it in other places.

egallicc:
> I don't think I understand. What does `setDefaultEnergyExpression()` return? What is the argument to the constructor? How do the parameters get added to the Force?

Just like above, the argument of the constructor is the energy expression. I was envisioning a utility function that returns the default energy expression and adds/sets the parameters, but the example was confusing. Let me try again:

To use the default energy expression:
```
ATMForce* atm = new ATMForce( ATMForce::DefaultEnergyExpression() );
atm->addDefaultEnergyExpressionParameters();
system.addForce(atm);
...
Context context(system, integrator, platform);
double lambda1 = 0.5, lambda2 = 0.5, alpha = 0, u0 = 0, w0 = 0,  umax = 100, ubcore= 50, acore = 1./16., direction = 1;
atm->setDefaultEnergyExpressionParameters(context, lambda1, lambda2, alpha, u0, w0, umax, ubcore, acore, direction);
```

`DefaultEnergyExpression()` returns the string containing the default energy expression, `addDefaultEnergyExpressionParameters()`  adds the parameters it contains, and `setDefaultEnergyExpressionParameters()` sets their values.

Custom energy expression:
```
ATMForce* atm = new ATMForce("lambda*u");
atm->addGlobalParameter("lambda", 0);
system.addForce(atm);
...
Context context(system, integrator, platform);
context.setParameter("lambda", 0.5);
```

The idea is that the default energy expression is handled in the same way as the custom energy expression, but we provide utility functions to make it easier to set its parameters.

> There's nothing wrong with overloading constructors. We do it in other places.

Then maybe we just go with the current version that specifies all of the parameters of the default expression in the constructor. I find a constructor that specifies some parameters but not others a bit confusing.

As you mentioned, the version of the constructor that defines a custom expression should prevent the creation of the parameters of the default expression.

Currently, the default alchemical potential is hard-coded in C++. Do you recommend re-coding it using an expression? Either way, we should probably create a test that defines a custom expression equivalent to the default and check that it returns consistent results.

peastman:
> Currently, the default alchemical potential is hard-coded in C++. Do you recommend re-coding it using an expression? 

Yes, that's exactly what I was planning on.  I can get started on it.

peastman:
To make sure I'm clear on the design:

The user will provide an expression that depends on a single variable `u`.  The returned energy will be `u0 + f(u1-u0)`, where `f()` is the expression they defined and `u0` and `u1` are the energies of the non-displaced and displaced conformations, respectively.  Is that right?

egallicc:
> To make sure I'm clear on the design:
> 
> The user will provide an expression that depends on a single variable `u`. The returned energy will be `u0 + f(u1-u0)`, where `f()` is the expression they defined and `u0` and `u1` are the energies of the non-displaced and displaced conformations, respectively. Is that right?

Yes, if `direction > 0`. If `direction < 0`, it returns `u1 + f(u0 - u1)`.

peastman:
Got it.  So `direction` is kind of different from the other parameters.  It doesn't appear in the energy expression.  And even though it's represented as a floating point number, it's really a boolean.  All that matters is whether it's positive or negative.

I wonder whether it would make more sense to explicitly make it boolean?  It never changes in the middle of a simulation, right?  So it would be a property of the Force object.  You would first run the forward simulation.  Then when you were ready to run the reverse simulation, you would switch the direction by calling `force.setAlchemicalDirection()` and `context.reinitialize()`?

egallicc:
> Got it. So `direction` is kind of different from the other parameters. It doesn't appear in the energy expression. And even though it's represented as a floating point number, it's really a boolean. All that matters is whether it's positive or negative.
> 
> I wonder whether it would make more sense to explicitly make it boolean? It never changes in the middle of a simulation, right? So it would be a property of the Force object. You would first run the forward simulation. Then when you were ready to run the reverse simulation, you would switch the direction by calling `force.setAlchemicalDirection()` and `context.reinitialize()`?

Yes, `direction` could be a boolean parameter. Can a global parameter be a boolean?

Actually, in our applications `direction` changes just as frequently as the other parameters. Most of our free energy calculations consist of two legs, one starting from the initial alchemical state in the direction of the final state (`direction > 0`), and another starting from the final state and moving backward towards the initial state (`direction < 0`). The two legs meet midway, and the free energy of the process is calculated as the difference between the free energies of the two legs. Furthermore, we typically run parallel replica exchange in which replicas swap alchemical states dynamically. Some replicas are initially assigned to the first leg (`direction > 0`) and others to the second (`direction < 0`). Replicas move from one leg to the other by exchanging `direction`.

This is what we do to calculate binding free energies. Who knows how people will use ATMForce() for their purposes. Maybe they will have no use for `direction`. Or maybe it will be used statically as you describe. But in our current applications `direction` changes dynamically.

Could `direction` be incorporated into the default energy expression?

```
(( ATMLambda2 -  ATMLambda1)/ATMAlpha) * log( 1 + exp(- ATMAlpha * (usc - ATMU0) ) ) + ATMLambda2*usc + ATMW0 ;
usc = select(step( u - ATMUbcore), u, ( ATMUmax - ATMUbcore)*fsc + ATMUbcore );
fsc = (z^ATMAcore - 1)/(z^ATMAcore + 1);
z = 1 + 2*(y/ATMAcore) + 2*(y/ATMAcore)^2;
y = (u - ATMUbcore)/( ATMUmax - ATMUbcore);
u = direction * (u1 - u0)
```

egallicc:
> So `direction` is kind of different from the other parameters. 

I agree, `direction` is indeed different from the other parameters because it affects the behavior of `ATMForce()` in more ways than the parameters of the `f()` function. It also establishes the reference state of the alchemical process (`u0` or `u1`). Or in other words, whether to return `u0 + f(u)` or `u1 + f(-u)`. 

I wonder if we should rethink the meaning of the energy expression to include the whole expression of the alchemical potential as you suggested a while back, not just `f()`. Any expression that depends on the collective variables `u0` and `u1`.

The default expression would be of the form `u0 + f(u1 - u0)` (when `direction > 0`) for our particular `f()`. But the user would be free to define a custom expression with any combination of `u0` and `u1`, which could include a `direction` parameter or not depending on what they want to do.

peastman:
That could work.  In that case, what would `getPerturbationEnergy()` return?  Maybe it could take a boolean argument to say whether it should return `f(u0, u1) - u0` or `f(u0, u1) - u1`?  Or we could replace it with a function that returns all three of the values `u0`, `u1`, and `f(u0, u1)`?

egallicc:
> That could work. In that case, what would `getPerturbationEnergy()` return? Maybe it could take a boolean argument to say whether it should return `f(u0, u1) - u0` or `f(u0, u1) - u1`? Or we could replace it with a function that returns all three of the values `u0`, `u1`, and `f(u0, u1)`?

Good point. Essentially, the "perturbation energy" of a sample is the smallest set of quantities required to calculate the alchemical energy of the sample at all other alchemical states (given their parameters) for the multi-state free energy analysis (MBAR/UWHAM or similar).

So, if the most general alchemical potential is a function of `u0` and `u1`, we need an API that returns `u0` and `u1`. `f(u0, u1) - u0` alone won't do because it does not have enough information to calculate the alchemical energy for a different set of parameters for `f()`.

Returning all three: `u0`, `u1`, and `f(u0, u1)` appears unnecessary because the user (who provided `f()` and the parameters) presumably knows how to calculate`f(u0, u1)` given `u0` and `u1`.

In all physical applications I can think of, knowledge of `u = u1 - u0` is sufficient for the multi-state free energy analysis. Nevertheless, it is not a big ask for the user to retrieve `u1` and `u0` separately, and take the difference.

peastman:
Could it be useful to also return `f(u0, u1)`?  Yes, they presumably know how to calculate it themselves, but on the other hand, it's really easy for us to calculate it for them, given that we already have the expression.  If providing a return value for it would make the user's life easier, there's no difficulty with doing it.

egallicc:
> Could it be useful to also return `f(u0, u1)`? Yes, they presumably know how to calculate it themselves, but on the other hand, it's really easy for us to calculate it for them, given that we already have the expression. If providing a return value for it would make the user's life easier, there's no difficulty with doing it.

Sure, but I guess in this generalization `f()` refers to the entire alchemical energy expression, not just the shift from the reference energy, right?

The default energy expression is of the form `u0 + f(u0,u1)` and I expect that it would be so in most implementations. But in the most general case, the user would be able to specify any expression `f(u0,u1)`. Yes?

peastman:
Exactly.  It would be the total energy, not just the shift.

peastman:
Actually, I'm a little confused about it.  I assumed the "perturbation energy" meant `f(u)`, that is, the thing that gets added to `u0`.  But actually it looks like `getPerturbationEnergy()` returns something different.  Assuming the user defines a single expression `f(u0, u1)` for the total energy, what quantities should we return?

peastman:
Could you revert that last push?  I'm in the middle of making major changes that restructure the whole class.  That's going to create a horrible merge.

openmmapi/include/openmm/ATMForce.h
```diff
@@ -101,35 +101,46 @@ class OPENMM_EXPORT ATMForce : public OpenMM::Force {
...
+     * The displacement vector can have 3 or 6 elements. A displacement with 3 elements
+     * specifies the x, y, z coordinates of the displacement for the target state and 
+     * the particle is not displaced in the initial state (most common).
+     *
+     * If the displacement has 6 elements, the first 3 specify the x, y, z coordinates of 
+     * the displacement for the target state as above, and the last 3 elements specify the coordinates 
+     * of the displacement for the initial state.
      */
-    int addParticle(int particle, double dx, double dy, double dz);
+    int addParticle(const std::vector<double>& displacement=std::vector<double>());
```
peastman, line 115: If we want to make it a single vector instead of three numbers, let's use a Vec3.  That will be much cleaner.  A `vector<double>` can have any number of components.

egallicc:
I have cleaned up the interface to pass displacements as vectors of parameters.

From:
```
int addParticle( int particle, double dx, double dy, double dz);
```

To:
```
int addParticle(const std::vector<double>& displacement);
```

This API lets us pass two displacements for each particle (one for the target state and one for the initial state) and can be generalized to pass parameters for more complex coordinate transformations later on.

egallicc:
> Could you revert that last push? I'm in the middle of making major changes that restructure the whole class. That's going to create a horrible merge.

Yes, sorry, I should have warned you. How do I revert it?

egallicc:
Reverted.

openmmapi/include/openmm/ATMForce.h
egallicc: Yes, that's the intent. No problem. It can wait.

egallicc:
I added a test of ATMForce with the Nonbonded Force. The test adds a Nonbonded Force that was already added to the System to ATMForce and then removes it from the System. This is I think a common scenario in applications where users would want to transfer Forces to the ATMForce that were added to the System by setup utilities such as  AmberPrmtopFile.createSystem(). See [this post](https://github.com/openmm/openmm/issues/4040#issuecomment-1547029531)

However, there are some puzzling behaviors.

1. Adding a NonbondedForce that was not previously added to the System works fine

2. Adding a NonbondedForce already added to the System directly fails at runtime with a SEGV fault
```
system.addForce(nbforce);
atm->addForce(nbforce);
system.removeForce(0);
system.addForce(atm);
```
The fault occurs whether the NonbondedForce is removed from the System or not.

3. Adding a serialized copy of the NonbondedForce already added to the System works whether the NonbondedForce is later removed from the System or not
```
system.addForce(nbforce);
atm->addForce(XmlSerializer::clone<Force>(*nbforce))
system.removeForce(0);
system.addForce(atm);
```
However, having to add a copy seems unnecessary because the ATMForce() [already adds a copy of the force to the inner System](https://github.com/Gallicchio-Lab/openmm/blob/f76f37fc1bfdbddcf5ca6a49d416f84f2870ee40/openmmapi/src/ATMForceImpl.cpp#L76). 

4. Also, attempting to add a serialized copy of the NonbondedForce as above, but from Python, fails with a SEGV fault
```
atmforce.addForce(XmlSerializer.clone(nbforce))
system.removeForce(nbforce_index)
system.addForce(atmforce)
```
And, strangely, unlike from C++, adding the NonbondedForce directly in Python avoids the SEGV fault but only if the NonbondedForce is not removed from the System
```
atmforce.addForce(nbforce)
#system.removeForce(nbforce_index)
system.addForce(atmforce)
```

The SEGV faults seem to occur [at the point where the copy of the Force is made](https://github.com/Gallicchio-Lab/openmm/blob/f76f37fc1bfdbddcf5ca6a49d416f84f2870ee40/openmmapi/src/ATMForceImpl.cpp#L76), inside the XmlSerializer class.

Are the tests above valid? Is there a better way to transfer a Force from the System to ATMForce?

peastman:
That's the expected behavior.  The key is in the API docs for `addForce()`:

https://github.com/openmm/openmm/blob/b183e1b6e5147d16b2a88a1d2471c3b32a365d53/openmmapi/include/openmm/System.h#L178-L180

In C++, the concept of ownership is key to correct memory management.  Every object has an owner.  When you add a Force to a System, the System becomes the owner of it and is responsible for deleting it when it's no longer needed.  That means when you remove the Force from the System, it gets deleted.

ATMForce likewise assumes ownership of Forces you add to it.  That means you can't add a Force that already belongs to a System.  That would lead to two different objects both thinking they owned it, and hence to a variety of memory errors.

When doing it from Python, you can just use the standard `copy()` function.

egallicc:
> That's the expected behavior. The key is in the API docs for `addForce()`:
> 

Thank you. So to confirm, `atmforce.addForce(XmlSerializer.clone(force))` is the recommended way to add a Force to ATM that is already in the System?

> 
> When doing it from Python, you can just use the standard `copy()` function.

That is the first thing I tried. It failed with a SEGV fault similar to copying the NonbondedForce with the XmlSerializer.

peastman:
> So to confirm, atmforce.addForce(XmlSerializer.clone(force)) is the recommended way to add a Force to ATM that is already in the System?

Correct.

> It failed with a SEGV fault similar to copying the NonbondedForce with the XmlSerializer.

I'm not sure what's going on with that.  It ought to work.

egallicc:
> > It failed with a SEGV fault similar to copying the NonbondedForce with the XmlSerializer.
> 
> I'm not sure what's going on with that. It ought to work.

I will try to prepare a minimal example.

peastman:
Any thoughts on this question?

> Actually, I'm a little confused about it. I assumed the "perturbation energy" meant f(u), that is, the thing that gets added to u0. But actually it looks like getPerturbationEnergy() returns something different. Assuming the user defines a single expression f(u0, u1) for the total energy, what quantities should we return?

egallicc:
> Any thoughts on this question?
> 
> > Actually, I'm a little confused about it. I assumed the "perturbation energy" meant f(u), that is, the thing that gets added to u0. But actually it looks like getPerturbationEnergy() returns something different. 

It currently returns the perturbation energy `u = u1 - u0` (or `-u` if `direction < 0`). More specifically, it returns $u_{sc}$ which is `u` after it goes through the [soft-core mapping function](https://www.compmolbiophysbc.org/atom-openmm/the-science-behind-atom#h.1mde54zloic). This behavior is closely tied to our implementation and applications. 

> Assuming the user defines a single expression f(u0, u1) for the total energy, what quantities should we return?

Having  `getPerturbationEnergy()` return `u1`, `u0`, and `f(u1,u0)` as we discussed is a good generalization. We can process these outputs to get what we need for the free energy calculation.

peastman:
Ok, that's what I'll do then.  Thanks!

peastman:
I have an initial implementation of the custom expression done.

It looks like there's no serialization proxy for ATMForce?  Without that, `XmlSerializer.clone()` won't work.  Or `copy()`, which calls it.  I can add it next.

If you want to allow specifying displacements for both states, how about this API?

```python
int addParticle(Vec3 displacement1, Vec3 displacement2)
```

If you want to allow the displacement on the first state to be omitted, an alternative would be

```python
int addParticle(Vec3 displacement2, Vec3 displacement1=Vec3())
```

It might be confusing having `displacement2` come before `displacement1`, though.  Or we could just add an overload,

```python
int addParticle(Vec3 displacement2)
```

egallicc:
> I have an initial implementation of the custom expression done.

Thank you.

> It looks like there's no serialization proxy for ATMForce? Without that, `XmlSerializer.clone()` won't work. Or `copy()`, which calls it. I can add it next.

The serialization was added to the [plugin version](https://github.com/Gallicchio-Lab/openmm-atmmetaforce-plugin/tree/main/serialization), if it helps.

Would the absence of the serialization proxy for the ATMForce() affect the [issue](https://github.com/openmm/openmm/pull/4110#issuecomment-1656353667) I reported earlier? In one case it was the Force added to ATMForce() to be serialized. In the second case, serialization was not used.
 
> If you want to allow specifying displacements for both states, how about this API?

Yes, I like
```python
int addParticle(Vec3 displacement1, Vec3 displacement0=Vec3()))
```
where `displacement1` is applied to the target state (the one corresponding to `u1`) and `displacement0` is the optional displacement for the initial state (`u0`). Having `displacement1` appear before `displacement0` is good, I think. It mirrors the definition of the perturbation energy `u = u1 - u0`. 

`setParticleParameters()` would be similar:
```python
void setParticleParameters(int particle, Vec3 displacement1, Vec3 displacement0=Vec3()))
```
I will go ahead and implement this if it is okay.

Looking ahead, we might replace `setParticleParameters()` with individual `set` methods to introduce more complex coordinate transformations. For example, we could do
```python
void setParticleDisplacement(int particle, Vec3 displacement1, Vec3 displacement0=Vec3()))
```
for linear displacements. And 
```python
void setParticleRotation(int particle, Vec3 axis1, double angle1, Vec3 axis0 = Vec3(1,0,0), double angle0 = 0)
```
for rotations. 

A single `setParticleParameters()` method to set displacements, rotations, etc. would probably quickly become unmanageable.

peastman:
> int addParticle(Vec3 displacement1, Vec3 displacement0=Vec3()))

Looks good.  Go ahead.

egallicc:
I have modified the default energy expression to address NaN's in the gradients when `lambda1 = lambda2` and `alpha = 0`. When `lambda1 = lambda2` the alchemical perturbation potential should just be the linear potential `lambda2*u`, which is independent of `alpha`.

I also made it so that setting `acore = 0` turns off the soft-core potential.

egallicc:
The name of the `u0` parameter of the default energy expression clashed with the name we have given to the energy of the initial state. I changed the former to `uh`.

Also, if it is okay, I am going to clean up the naming of the variables that refer to the initial and target states. In this discussion, we refer to the initial state as state 0 and to the target state (the displaced state) as state 1. For example, the name of the energy of the initial state is `u0`. However, in many places in the code, the number 1 refers to the initial state and 2 to the final state. This leads to confusing statements such as
```
posq1[i] = posq[i] + d0;
posq2[i] = posq[i] + d1;
``` 
I intend to rename the variables consistently using 0 and 1, rather than 1 and 2. The changes are cosmetic but are going to affect many files.

Also, if it is okay, I will also swap the order of `u0` and `u1` in
```python
getPerturbationEnergy(context, u0, u1, energy)
```
to match the ordering of the displacements in
```python
addParticle( Vec3 displacement1, Vec3 displacement0=Vec3())
```

peastman:
Sounds good.

peastman:
I fixed the problem in the Python wrapper.  `addForce()` needed to be marked as taking ownership of the Force, so Python wouldn't try to delete it.

egallicc:
> I fixed the problem in the Python wrapper. 

Great. I added an initial Python test. It passes.

egallicc:
I added tests for the custom expression.

I think we are there. What's left? Serialization?

peastman:
The major elements still left to be done are derivatives with respect to global parameters, serialization, and documentation.  I'm working on them now.

egallicc:
> The major elements still left to be done are derivatives with respect to global parameters, serialization, and documentation. I'm working on them now.

The C++ API documentation is in. It's 90% complete, I think. The only omission is the displacement at the initial state.

Derivatives with respect to the global parameters are not needed. Maybe an option in some distant future if someone wants to implement a non-equilibrium protocol.

peastman:
> Derivatives with respect to the global parameters are not needed.

It's a potentially useful feature for various applications, and pretty easy to implement.  I'm almost done with it.

peastman:
Take a look at my latest commit and see if everything is ok.  I added more documentation, plus some general cleanup and a couple of bug fixes.

One change to particularly note is to the default energy expression.  `Direction` appears in two places.  In one place it was inside a `step()` function, so only its sign mattered.  In the other place it acted as a continuous variable, so if it was set to anything other than 1 or -1 you would get an invalid result.  I changed it to use a `step()` function in both places.

We also should consider a change to the behavior of `getPerturbationEnergy()`.  Currently it returns the energies calculated in the most recent call to `calcForcesAndEnergy()`.  That's a bit dangerous and could lead to unexpected results.  For example, if someone sets particle positions with `setPositions()` or modifies a global parameter with `setParameter()` and then immediately calls `getPerturbationEnergy()`, they'll get a result that is out of date.  It will reflect old particle positions and parameters, not current ones.

Would it be reasonable to make `getPerturbationEnergy()` recompute the energy to make sure the result is correct?  The disadvantage is extra computation.  How often does it normally get called?  If it's called after every step that would be expensive, but if it only gets called occasionally it shouldn't be a problem.

egallicc:
> Take a look at my latest commit and see if everything is ok. I added more documentation, plus some general cleanup and a couple of bug fixes.

It looks good, thank you. I fixed some minor typos.

> 
> One change to particularly note is to the default energy expression. 

Right. Yes, this is better.

> 
> We also should consider a change to the behavior of `getPerturbationEnergy()`
> 

Yes, that is a very good point. Right, `getPerturbationEnergy()` is called infrequently to sample the perturbation energy. It is okay to evaluate the energy before returning the energy values. Is it as simple as calling `ATMForceImpl::calcForcesAndEnergy()` before returning `u1` etc?

peastman:
> Is it as simple as calling ATMForceImpl::calcForcesAndEnergy() before returning u1 etc?

Pretty much.  A few other changes were needed to make it compile, and the change also uncovered a bug I needed to fix.

peastman:
I think this may be getting to the point where we can call it complete.  Do you have other things you want to be included?

egallicc:
> I think this may be getting to the point where we can call it complete. Do you have other things you want to be included?

Great. It's all in on my end.

egallicc:
> I added one more test and made a few minor edits. If they look ok to you, we can merge once CI finishes.

Yeah!

egallicc:
When ready and merged, I will prepare and test a version of AToM-OpenMM with the new API for host-guest and protein-ligand binding free energy applications.

peastman:
Merged!  Thanks so much for this.  It will be great to have.

egallicc:
> Merged! Thanks so much for this. It will be great to have.

Yes, thank you!

PR #4116, opened 2023-06-19
Add an option to force the timestep to be a specified value in `benchark.py`

### LINKED ISSUE #4111: Add timestep CLI option to `benchmark.py` example
Recently benchmarking OpenMM against the OpenMM folding@home core (Core 22) on a set of matching benchmark systems, I ran into the issue that the timestep for the `benchmark.py` script is auto selected based on the benchmark type.

 It may be useful to add an option to fix the timestep to a certain value in the CLI rather than having it auto-generated. Would people be open to this?

hmacdope:
Fixes #4111

Allows a timestep option to be given on the `benchmark.py` CLI which overrides the automatic timestep assignment if given. This was originally motivated by a comparison to the FAH openmm core which runs similar benchmark workflows with different timesteps.

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -312,31 +312,38 @@ def runOneTest(testName, options):
     
     # Create the integrator
     temperature = 300*unit.kelvin
+
+    # have an option to force a specific timestep
+    if options.force_timestep is not None:
+        dt_force = float(options.force_timestep) * unit.femtoseconds
+    else:
+        dt_force = None
+
     if explicit:
         friction = 1*(1/unit.picoseconds)
     else:
         friction = 91*(1/unit.picoseconds)
     if amoeba:
-        dt = 0.002*unit.picoseconds
+        dt = dt_force if dt_force else 0.002*unit.picoseconds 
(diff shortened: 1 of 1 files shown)


peastman:
I've been debating about this.  It seems different from our other options in an important way.  The script tries to make sure that it will only run reasonable benchmarks.  Some of them are esoteric, but it's at least plausible that someone might want to run them.  But this option lets you do things that are just completely wrong, such as using a timestep that causes the simulation to immediately blow up.  Also, there are good reasons that different benchmarks use different time steps.  Having a single option that overrides all of them and forces them to all use the same time step feels dangerous and wrong.

What is the use case for which you want this options?  If it's something specialized, it might be better to just create your own modified copy of the script but leave the main version as it is.

hmacdope:
No worries @peastman, feel free to close, I totally understand your perspective. The aim was to compare with some FAH benchmarks that use the same systems and general setting but different time steps. Perhaps it can just live in my branch, that seems like an perfectly acceptable solution.

peastman:
Ok, that sounds good.  Thanks.

Issue #4071, opened 2023-05-16
Differences in volume ensemble sampling


balazsfabian:
I compiled OpenMM from source following the [recommended steps](http://docs.openmm.org/latest/userguide/library/02_compiling.html) with CUDA support. I tried version v7.7.0 (git tag), ed049af1b34ca and some in between.

I simulated a small, 903 TIP3P water box using the `MonteCarloBarostat` and a pressure of 1 atmosphere. I saved the volume after every 1000 steps to a logfile.

When I compute the autocorrelation of volumes of neat water boxes (903 TIP3P waters) across the two versions, I get significantly different correlation times. To condense the differences in autocorrelation into a single number, I averaged the autocorrelation curve over the first 100 points (omitting the trivial 1 at the beginning). Using the same python script, the volumes produced by v7.7 are virtually immediately uncorrelated, while those of ed049af1b34ca show long-time correlations. Here are the versions that I tested along with the average volume and `acorr[1:100].mean()`.

| Commit    |   Volume   |   Mean of acorr.   |
| --------- | ------- | -------- |
| ed049af1b | 26.5706 |  0.24906 |
| 3c7eef3d0 | 27.4298 |  0.49866 |
| dbf25377e | 26.5994 |  0.00462 |
| ae42c911a | 26.5972 |  0.00277 |
| 48b6abc1e | 26.6088 |  0.00231 |
| v8.0rc1   | 26.6036 | -0.00026 |
| v8.0beta  | 26.5929 |  0.00117 |
| v7.7      | 26.5883 |  0.00045 |

The environment.yml:
```
name: /home/tb/bafabian/conda-envs/openmm-dev
channels:
  - conda-forge
  - bioconda
  - defaults
dependencies:
  - _libgcc_mutex=0.1=conda_forge
  - _openmp_mutex=4.5=2_gnu
  - bzip2=1.0.8=h7f98852_4
  - c-ares=1.18.1=h7f98852_0
  - ca-certificates=2022.9.24=ha878542_0
  - cmake=3.25.0=h816a3e0_0
  - cython=0.29.32=py310hd8f1fbe_1
  - doxygen=1.9.5=h583eb01_0
  - expat=2.5.0=h27087fc_0
  - keyutils=1.6.1=h166bdaf_0
  - krb5=1.19.3=h08a2579_0
  - ld_impl_linux-64=2.39=hc81fddc_0
  - libblas=3.9.0=16_linux64_openblas
[...]

PR #4119, opened 2023-06-20
another attempted fix for MonteCarloBarostat

### LINKED ISSUE #4117: Problem with jumping molecules
I alluded to an issue with "jumping" molecules in #4071 and wanted to post a follow up.  It seems that the problem still exists after the barostat fix was merged in #4106.  The system is the same as I have been describing in the previous issue (8000 iAMOEBA water molecules @ 184 K and 1800 atm with a Monte Carlo anisotropic barostat along the z-direction).

In the following movies, I'm showing the trajectories of the water molecule O atoms in one octant of the box only.  The intended behavior is very slow diffusion because the temperature is very low.  The trajectory frames are written once per 10 ps.

The omm_75.mp4 movie shows the behavior I have been getting with versions 7.4 and 7.5, which is the expected behavior.  However, in omm_80.mp4 you can see that at certain times in the trajectory all of the molecules "jump" relatively large distances to new positions.  These jumps do not happen at every single barostat move because there are 100 attempts per trajectory frame written.  However it's possible that with every barostat move there is the possibility of a jump.  I have not tested with a NVT run but I will do that next.

When I analyze the trajectories, the diffusion coefficients calculated using the version 8.0 trajectory are too high, presumably because of the jumping problem.  However, the thermodynamic properties seem okay, which makes me think it's just the atom indices that are getting scrambled.  Please let me know if you have thoughts on next steps for tracking this down.

https://www.dropbox.com/s/usrbf58q370nk35/omm_75.mp4?dl=0
https://www.dropbox.com/s/i1d7nfjrvfth4o5/omm_80.mp4?dl=0

sef43:
This aims to fix #4117 

In #4117 the atomic indices can end up incorrectly reordered.
In this PR the atomIndex vector is saved before an attempted MC move and  restored after a rejected one.

### DIFF
--- a/platforms/common/include/openmm/common/ComputeContext.h
+++ b/platforms/common/include/openmm/common/ComputeContext.h
@@ -388,11 +388,9 @@ public:
         return atomIndex;
     }
     /**
-     * Set the host-side vector which contains the index of each atom.
+     * Set the vector which contains the index of each atom.
      */
-    void setAtomIndex(std::vector<int>& index){
-        atomIndex = index;
-    }
+    void setAtomIndex(std::vector<int>& index);
     /**
      * Get the array which contains the index of each atom.
      */
--- a/platforms/common/include/openmm/common/ComputeContext.h
+++ b/platforms/common/include/openmm/common/ComputeContext.h
@@ -388,6 +388,12 @@ public:
         return atomIndex;
     }
     /**
+     * Set the host-side vector which contains the index of each atom.
+     */
+    void setAtomIndex(std::vector<int>& index){
+        atomIndex = index;
+    }
+    /**
      * Get the array which contains the index of each atom.
      */
     virtual ArrayInterface& getAtomIndexArray() = 0;
--- a/platforms/common/include/openmm/common/CommonKernels.h
+++ b/platforms/common/include/openmm/common/CommonKernels.h
@@ -1576,7 +1576,7 @@ public:
     void restoreCoordinates(ContextImpl& context);
 private:
     ComputeContext& cc;
-    bool hasInitializedKernels, rigidMolecules;
+    bool hasInitializedKernels, rigidMolecules, atomsWereReordered;
     int numMolecules;
     ComputeArray savedPositions, savedFloatForces, savedLongForces;
     ComputeArray moleculeAtoms;
--- a/platforms/common/src/CommonKernels.cpp
+++ b/platforms/common/src/CommonKernels.cpp
@@ -7720,6 +7720,7 @@ void CommonApplyMonteCarloBarostatKernel::saveCoordinates(ContextImpl& context) 
     if (savedFloatForces.isInitialized())
         cc.getFloatForceBuffer().copyTo(savedFloatForces);
     lastPosCellOffsets = cc.getPosCellOffsets();
+    lastAtomOrder = cc.getAtomIndex();
 }
 
 void CommonApplyMonteCarloBarostatKernel::scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ) {
@@ -7769,7 +7770,6 @@ void CommonApplyMonteCarloBarostatKernel::scaleCoordinates(ContextImpl& context,
     kernel->setArg(2, (float) scaleZ);
     setPeriodicBoxArgs(cc, kernel, 4);
     kernel->execute(cc.getNumAtoms());
-    lastAtomOrder = cc.getAtomIndex();
 }
 
 void CommonApplyMonteCarloBarostatKernel::restoreCoordinates(ContextImpl& context) {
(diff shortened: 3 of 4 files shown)


platforms/common/include/openmm/common/ComputeContext.h
```diff
@@ -387,6 +387,12 @@ class OPENMM_EXPORT_COMMON ComputeContext {
     const std::vector<int>& getAtomIndex() const {
         return atomIndex;
     }
+    /**
+     * Set the host-side vector which contains the index of each atom.
+     */
+    void setAtomIndex(std::vector<int>& index){
+        atomIndex = index;
+    }
```
peastman, line 395: The atom indices get stored on both host and device.  We need to update them in both places.  We also need to inform any ReorderListeners that the order has changed.  You can do those like this:

```c++
getAtomIndexArray().upload(atomIndex);
for (auto listener : reorderListeners)
    listener->execute();
```

And update the comment to remove "host-side".

peastman:
I've been worrying that this will add cost to the barostat.  ReorderListeners usually aren't real expensive to invoke, but they aren't free either.  And we only need to invoke them very occasionally.  In most cases, the atoms won't have been reordered so that's just wasted computation.  I think I've figured out a way to optimize it.

You can call `getAtomsWereReordered()` on the ComputeContext to find out whether reordering happened during the most recent force/energy evaluation.  The complication is that barostats do two energy evaluations.  The process is:

1. Compute the energy.
2. Call `scaleCoordinates()`.
3. Compute the energy again.
4. If the step is rejected, call `restoreCoordinates()`.

The solution is to check `getAtomsWereReordered()` in both `scaleCoordinates()` and `restoreCoordinates()`.  If either one returns true, we need to call `setAtomIndex()`.  If both of them return false, we know the atom order hasn't changed so we can skip `setAtomIndex()`.

sef43:
Yes that seems sensible! I have implemented it as you have described.

peastman:
Looks great.  @leeping can you confirm that this fixes your problem?

leeping:
Yes, I can confirm I'm not observing the jumps anymore!

jchodera:
Since this was present in the 8.0 release and it impacts a widely used feature, does it make sense to cut a bug release soon?

sef43:
I would like to double check the other relevant issues (#4032 , #4071) are still fixed by this before we merge and release

jchodera:
Should we have a test that could catch this issue? Wouldn't it show up in our [ideal molecular gas tests](https://github.com/openmm/openmm/blob/master/tests/TestMonteCarloBarostat.h#L86)?

sef43:
The reason it did not get caught by the existing test is because the volume of the box was still correct. We would need to compute a diffusion coefficient or something similar I think to catch the issue this fixes.

leeping:
Correct, I think whether the test could catch the problem would depend on how fast the diffusion is.  If the diffusion is normally fast, then a reordering of atomic indices might not register as being abnormal.  Also, the reordering does not appear to be completely random, as you can see from the movies.

sef43:
Thinking more on the testing side of things I think this would be a test more suitable as a regression test, as @leeping mentions it would require simulating for a reasonable amount of time to check the Barostat is working correctly.

I see that the userguide mentions there are system tests: http://docs.openmm.org/latest/userguide/library/07_testing_validation.html#system-tests
and links to a very old svn repo full of python 2.x code 
Are these actually ever run?
Is there an up to date suite of regression/system tests?

sef43:
> I would like to double check the other relevant issues (#4032 , #4071) are still fixed by this before we merge and release

I've tested both these issues are fixed and am happy for this to be merged

peastman:
Those tests are obsolete.  We should remove that section of the user guide.

Note that there has never been a version of OpenMM in which you could compute diffusion coefficients when using a barostat.  Until 8.0, it would translate every molecule into a single periodic box every time it attempted a barostat step.  8.0 changed it so it usually wouldn't, but due to #4032 it still would every once in a while.  8.1 will (hopefully!) be the first release in which it's possible.

peastman:
I'll go ahead and merge this.  If we want to add additional tests, that can be done in a separate PR.

leeping:
Thanks, Peter.  To compute the diffusion coefficient previously, I was "unwrapping" the molecules using the GROMACS trjconv program, but that was inaccurate because the trajectory doesn't store the exact times of the translations.  I'll use this version going forward.

jchodera:
@peastman : Was the barostat issue present in OpenMM 8.0 in a way that would cause incorrect results, or just inconvenient re-imaging? I'm trying to figure out if we need to hold on releasing the new FAH core until we can get a bugfix release out.

peastman:
I believe the same issue has existed in all past releases.  Looking at the 7.7 version of the code, I see that `scaleCoordinates()` saves the atom order, but `restoreCoordinates()` neglects to reset them.  But the issue there is moot because it wraps coordinates at every MC step.  Like I said, there has never been a release of OpenMM in which it was possible to correctly compute diffusion coefficients while using a barostat.  8.1 will be the first release where it's possible.

jchodera:
@peastman: Thanks! I concur that this does not sound like it warrants an urgent bugfix release. I just wanted to confirm that the barostat bug did not produce incorrect equilibrium probability densities, which _would_ require an urgent fix.

Issue #4135, opened 2023-07-06
Wrong formula for Langevin Integrator (at least in the documentation)


plainerman:
I believe that I have found an issue with the current documentation of the theory behind the integrators. I only looked more closely into the Langevin Integrator, but maybe this impacts the documentation of the other intergrators as well.

More specifically, it is about the equations of the documentation in this section.

https://github.com/openmm/openmm/blob/644dc1ecc9e95b3c8e831803bb3e2ac925999f74/docs-source/usersguide/theory/04_integrators.rst?plain=1#L41-L76
(Probably, it is easier to look [at the official documentation with equation rendering](http://docs.openmm.org/latest/userguide/theory/04_integrators.html#langevinintegator)).

I assume that the implementation in openmm of the `LangevinIntegrator` is as specified in [this paper](https://doi.org/10.1142%2F9789814295291_0026) (which is listed as reference).

There are a few discrepancies between the equations in the documentation and the paper
* Inconsistent timestep size $\Delta t$, which is sometimes assumed to be a half step, but in other instances, a full step
* Different variance in the random variable
* Unclear notation when dividing by $m$

This is the implementation suggested by the referenced paper
![image](https://github.com/openmm/openmm/assets/6825443/8f3d28c2-5a6d-47d5-b02e-b4c144ba3d03)

And this the formula in the documentation
$$\mathbf{v}_{i}(t+\Delta t/2) = \mathbf{v}_i(t-\Delta t/2)\alpha + \mathbf{f}_i(t)(1 - \alpha) / \gamma m_i + \sqrt{kT(1-\alpha^2)/m}R$$

If we assumed $R_i$ to have a variance of $2 m_i k_B T$ (as specified in the documentation) , then the last part of the equation with a standard-normal random variable $Z$ becomes

$$\sqrt{kT(1-\alpha^2)/m} \sqrt{2 m_i k_B T} Z$$

which would simplify to

$$\sqrt{2 k ^ 2 T^2(1-\alpha^2)} Z$$

This is very different from what we have in this paper, for example all the masses cancel out and the temperature is squared.

I believe the documentation should be updated to say the following.
Z is a random normal variable with zero mean and unit variance. Then the LangevinIntegrator is defined as

[...]

PR #4138, opened 2023-07-07
Improve documentation of LangevinIntegrator

plainerman:
This issue fixes https://github.com/openmm/openmm/issues/4135. 

It should make it clear that the random variable $R_i$ has unit variance. Further I made the variable notation of $m_i$ consistent and highlighted that we have to divide by the mass.


peastman:
I think you meant LangevinIntegrator, not VerletIntegrator?

plainerman:
Oh yes, you are right!

peastman:
Looks good, thanks!  I'll go ahead and merge since the docs build has successfully finished, and this doesn't affect anything other than docs.

Issue #3966, opened 2023-02-17
`findBlocksWithInteractions` optimization idea


bdenhollander:
`findBlocksWithInteractions` trails off differently than other kernels. When examining wavefront run times, I noticed that the total kernel run time is equal to the length of the longest wavefront. From the loop conditions, I get the impression that the kernel is running like this:

| Block1 | Block2 | Block2 | Block2 | Block2 | Block2 | Block2 | Block2 |
|--------|--------|--------|--------|--------|--------|--------|--------|
| 7      |        |        |        |        |        |        |        |
| 6      | 7      |        |        |        |        |        |        |
| 5      | 6      | 7      |        |        |        |        |        |
| 4      | 5      | 6      | 7      |        |        |        |        |
| 3      | 4      | 5      | 6      | 7      |        |        |        |
| 2      | 3      | 4      | 5      | 6      | 7      |        |        |
| 1      | 2      | 3      | 4      | 5      | 6      | 7      |        |
| 0      | 1      | 2      | 3      | 4      | 5      | 6      | 7      |

The wavefront that gets block 0 has the most comparisons to do and will finish last.

Would it be possible to spread the load more evenly by comparing each block against no more than half + 1 other blocks? This could theoretically speedup `findBlocksWithInteractions` by 40%.

| Block1 | Block2 | Block2 | Block2 | Block2 |
|--------|--------|--------|--------|--------|
| 7      |        | 0      | 1      | 2      |
| 6      | 7      |        | 0      | 1      |
| 5      | 6      | 7      |        | 0      |
| 4      | 5      | 6      | 7      |        |
| 3      | 4      | 5      | 6      | 7      |
| 2      | 3      | 4      | 5      | 6      |
| 1      | 2      | 3      | 4      | 5      |
| 0      | 1      | 2      | 3      | 4      |

<details><summary>Radeon GPU Profiler Screenshot</summary>
<p>
<img src="https://user-images.githubusercontent.com/44237618/219535552-0db96f87-dabb-4a23-8aa5-7040a185e9f3.png" />
</p>
</details>

peastman:
It shouldn't be as bad as that, though there could be a slight effect.  Here is the outer loop.

```c++
for (int block1 = startBlockIndex+warpIndex; block1 < startBlockIndex+numBlocks; block1 += totalWarps) {
```

Inside that is an inner loop.

```c++
    for (int block2Base = block1+1; block2Base < NUM_BLOCKS; block2Base += 32) {
```

[...]

PR #4147, opened 2023-07-17
Use large blocks to optimize building the neighbor list

peastman:
This is an attempt to improve the scaling on large systems.  As discussed in #4063, one part of building the neighbor list involves comparing the bounding boxes of "atom blocks" composed of 32 atoms each.  This scales as O(n^2).  For small to medium systems it takes very little time, but for systems with more than about a million atoms it can become a bottleneck.

This PR introduces "large blocks" composed of 32 atom blocks (1024 atoms).  Each atom block is checked against large blocks to see if we need to check against the individual blocks it contains.  For very large systems, this should reduce the cost of the O(n^2) loop by roughly a factor of 32.  In principle we could add still more levels (blocks of blocks of blocks) to create a full bounding volume hierarchy, which could potentially lead to O(n log n) scaling.

In the current code, atom blocks are sorted by size before building the neighbor list.  Due to the way neighbor lists are represented, this leads to a more efficient list and fewer interactions that have to be computed.  Unfortunately, that's incompatible with this optimization, since large blocks would not be spatially compact.  Therefore, this change is expected to make building the neighbor list faster, but computing interactions slightly slower.

To test this change, I simulated water boxes of varying sizes.  All simulations used reaction field with a cutoff of 1 nm, so that everything other than building the neighbor list should have O(n) scaling.  The graph show the time per 100,000 atoms to simulate 5000 steps.  For perfectly linear scaling this would be constant.

image

The scaling is greatly improved.  It's much faster for the largest systems.  Surprisingly, it also is faster even for the very smallest system tested (only 12,255 atoms).

I also tested it on some of our standard benchmarks, all of which use PME.

|Benchmark|Atoms|Original|New|Ratio|
|---|---|---|---|---|
pme|23558|2036.145|1694.8|0.83
apoa1pme|92224|602.358|627.88|1.04
amber20-cellulose|408609|153.6265|160.629|1.05
amber20-stmv|1067095|44.54215|48.19275|1.08

In this case, the very smallest benchmark became slower, while all others became faster.  Possibly that's because computing interactions is more expensive with PME than with reaction field?

If we automatically switch between the two methods at around 50,000 atoms, that would probably give good results in all cases.

cc @philipturner @bdenhollander @aizvorski

### DIFF
--- a/platforms/cuda/include/CudaNonbondedUtilities.h
+++ b/platforms/cuda/include/CudaNonbondedUtilities.h
@@ -338,6 +338,8 @@ private:
     CudaArray sortedBlocks;
     CudaArray sortedBlockCenter;
     CudaArray sortedBlockBoundingBox;
+    CudaArray largeBlockCenter;
+    CudaArray largeBlockBoundingBox;
     CudaArray oldPositions;
     CudaArray rebuildNeighborList;
     CudaSort* blockSorter;
@@ -351,7 +353,7 @@ private:
     std::map<int, double> groupCutoff;
     std::map<int, std::string> groupKernelSource;
     double lastCutoff;
-    bool useCutoff, usePeriodic, anyExclusions, usePadding, useNeighborList, forceRebuildNeighborList, canUsePairList;
+    bool useCutoff, usePeriodic, anyExclusions, usePadding, useNeighborList, forceRebuildNeighborList, canUsePairList, useLargeBlocks;
     int startTileIndex, startBlockIndex, numBlocks, maxExclusions, numForceThreadBlocks, forceThreadBlockSize, numAtoms, groupFlags;
     unsigned int maxTiles, maxSinglePairs, tilesAfterReorder;
     long long numTiles;
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -74,7 +74,13 @@ CudaNonbondedUtilities::CudaNonbondedUtilities(CudaContext& context) : context(c
     CHECK_RESULT(cuMemHostAlloc((void**) &pinnedCountBuffer, 2*sizeof(unsigned int), CU_MEMHOSTALLOC_PORTABLE));
     numForceThreadBlocks = 4*multiprocessors;
     forceThreadBlockSize = (context.getComputeCapability() < 2.0 ? 128 : 256);
-    useLargeBlocks = true;
+    
+    // When building the neighbor list, we can optionally use large blocks (1024 atoms) to
+    // accelerate the process.  This makes building the neighbor list faster, but it prevents
+    // us from sorting atom blocks by size, which leads to a slightly less efficient neighbor
+    // list.  We guess based on system size which will be faster.
+
+    useLargeBlocks = (context.getNumAtoms() > 90000);
     setKernelSource(CudaKernelSources::nonbonded);
 }
 
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -74,6 +74,7 @@ CudaNonbondedUtilities::CudaNonbondedUtilities(CudaContext& context) : context(c
     CHECK_RESULT(cuMemHostAlloc((void**) &pinnedCountBuffer, 2*sizeof(unsigned int), CU_MEMHOSTALLOC_PORTABLE));
     numForceThreadBlocks = 4*multiprocessors;
     forceThreadBlockSize = (context.getComputeCapability() < 2.0 ? 128 : 256);
+    useLargeBlocks = true;
     setKernelSource(CudaKernelSources::nonbonded);
 }
 
@@ -278,6 +279,8 @@ void CudaNonbondedUtilities::initialize(const System& system) {
         sortedBlocks.initialize(context, numAtomBlocks, 2*elementSize, "sortedBlocks");
         sortedBlockCenter.initialize(context, numAtomBlocks+1, 4*elementSize, "sortedBlockCenter");
         sortedBlockBoundingBox.initialize(context, numAtomBlocks+1, 4*elementSize, "sortedBlockBoundingBox");
+        largeBlockCenter.initialize(context, numAtomBlocks, 4*elementSize, "largeBlockCenter");
+        largeBlockBoundingBox.initialize(context, numAtomBlocks, 4*elementSize, "largeBlockBoundingBox");
         oldPositions.initialize(context, numAtoms, 4*elementSize, "oldPositions");
         rebuildNeighborList.initialize<int>(context, 1, "rebuildNeighborList");
         blockSorter = new CudaSort(context, new BlockSortTrait(context.getUseDoublePrecision()), numAtomBlocks, false);
@@ -333,6 +336,15 @@ void CudaNonbondedUtilities::initialize(const System& system) {
         sortBoxDataArgs.push_back(&blockBoundingBox.getDevicePointer());
         sortBoxDataArgs.push_back(&sortedBlockCenter.getDevicePointer());
         sortBoxDataArgs.push_back(&sortedBlockBoundingBox.getDevicePointer());
+        if (useLargeBlocks) {
+            sortBoxDataArgs.push_back(&largeBlockCenter.getDevicePointer());
+            sortBoxDataArgs.push_back(&largeBlockBoundingBox.getDevicePointer());
+            sortBoxDataArgs.push_back(context.getPeriodicBoxSizePointer());
+            sortBoxDataArgs.push_back(context.getInvPeriodicBoxSizePointer());
+            sortBoxDataArgs.push_back(context.getPeriodicBoxVecXPointer());
+            sortBoxDataArgs.push_back(context.getPeriodicBoxVecYPointer());
+            sortBoxDataArgs.push_back(context.getPeriodicBoxVecZPointer());
+        }
         sortBoxDataArgs.push_back(&context.getPosq().getDevicePointer());
         sortBoxDataArgs.push_back(&oldPositions.getDevicePointer());
         sortBoxDataArgs.push_back(&interactionCount.getDevicePointer());
@@ -355,6 +367,10 @@ void CudaNonbondedUtilities::initialize(const System& system) {
         findInteractingBlocksArgs.push_back(&sortedBlocks.getDevicePointer());
         findInteractingBlocksArgs.push_back(&sortedBlockCenter.getDevicePointer());
         findInteractingBlocksArgs.push_back(&sortedBlockBoundingBox.getDevicePointer());
+        if (useLargeBlocks) {
+            findInteractingBlocksArgs.push_back(&largeBlockCenter.getDevicePointer());
+            findInteractingBlocksArgs.push_back(&largeBlockBoundingBox.getDevicePointer());
+        }
         findInteractingBlocksArgs.push_back(&exclusionIndices.getDevicePointer());
         findInteractingBlocksArgs.push_back(&exclusionRowIndices.getDevicePointer());
         findInteractingBlocksArgs.push_back(&oldPositions.getDevicePointer());
@@ -396,7 +412,8 @@ void CudaNonbondedUtilities::prepareInteractions(int forceGroups) {
     if (lastCutoff != kernels.cutoffDistance)
         forceRebuildNeighborList = true;
     context.executeKernel(kernels.findBlockBoundsKernel, &findBlockBoundsArgs[0], context.getNumAtoms());
-    blockSorter->sort(sortedBlocks);
+    if (!useLargeBlocks)
+        blockSorter->sort(sortedBlocks);
     context.executeKernel(kernels.sortBoxDataKernel, &sortBoxDataArgs[0], context.getNumAtoms());
     context.executeKernel(kernels.findInteractingBlocksKernel, &findInteractingBlocksArgs[0], context.getNumAtoms(), 256);
     forceRebuildNeighborList = false;
@@ -503,6 +520,8 @@ void CudaNonbondedUtilities::createKernelsForGroups(int groups) {
             defines["USE_PERIODIC"] = "1";
         if (context.getBoxIsTriclinic())
             defines["TRICLINIC"] = "1";
+        if (useLargeBlocks)
+            defines["USE_LARGE_BLOCKS"] = "1";
         defines["MAX_EXCLUSIONS"] = context.intToString(maxExclusions);
         defines["MAX_BITS_FOR_PAIRS"] = (canUsePairList ? (context.getComputeCapability() < 8.0 ? "2" : "3") : "0");
         CUmodule interactingBlocksProgram = context.createModule(CudaKernelSources::vectorOps+CudaKernelSources::findInteractingBlocks, defines);
--- a/platforms/cuda/src/kernels/findInteractingBlocks.cu
+++ b/platforms/cuda/src/kernels/findInteractingBlocks.cu
@@ -86,15 +86,39 @@ extern "C" __global__ void findBlockBounds(int numAtoms, real4 periodicBoxSize, 
  * Sort the data about bounding boxes so it can be accessed more efficiently in the next kernel.
  */
 extern "C" __global__ void sortBoxData(const real2* __restrict__ sortedBlock, const real4* __restrict__ blockCenter,
-        const real4* __restrict__ blockBoundingBox, real4* __restrict__ sortedBlockCenter,
-        half3* __restrict__ sortedBlockBoundingBox, const real4* __restrict__ posq, const real4* __restrict__ oldPositions,
+        const real4* __restrict__ blockBoundingBox, real4* __restrict__ sortedBlockCenter, half3* __restrict__ sortedBlockBoundingBox,
+#ifdef USE_LARGE_BLOCKS
+        real4* __restrict__ largeBlockCenter, half3* __restrict__ largeBlockBoundingBox, real4 periodicBoxSize,
+        real4 invPeriodicBoxSize, real4 periodicBoxVecX, real4 periodicBoxVecY, real4 periodicBoxVecZ,
+#endif
+        const real4* __restrict__ posq, const real4* __restrict__ oldPositions,
         unsigned int* __restrict__ interactionCount, int* __restrict__ rebuildNeighborList, bool forceRebuild) {
     for (int i = threadIdx.x+blockIdx.x*blockDim.x; i < NUM_BLOCKS; i += blockDim.x*gridDim.x) {
         int index = (int) sortedBlock[i].y;
         sortedBlockCenter[i] = blockCenter[index];
         sortedBlockBoundingBox[i] = half3(trimTo3(blockBoundingBox[index]));
     }
+#ifdef USE_LARGE_BLOCKS
+    // Compute the sizes of large blocks (composed of 32 regular blocks) starting from each block.
     
+    for (int i = threadIdx.x+blockIdx.x*blockDim.x; i < NUM_BLOCKS; i += blockDim.x*gridDim.x) {
+        real4 minPos = blockCenter[i]-blockBoundingBox[i];
+        real4 maxPos = blockCenter[i]+blockBoundingBox[i];
+        int last = min(i+32, NUM_BLOCKS);
+        for (int j = i+1; j < last; j++) {
+            real4 blockPos = blockCenter[j];
+            real4 width = blockBoundingBox[j];
+#ifdef USE_PERIODIC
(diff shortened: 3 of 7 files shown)


philipturner:
> In the current code, atom blocks are sorted by size before building the neighbor list.

Would this change remove the GPU -> CPU dependency, so the neighbor list can be entirely GPU-driven? I'm doing something related with a ray tracing BVH, using 16384-atom ($4^3$ nm) blocks. The CPU pre-allocates a heap beforehand, which the GPU creates sub-allocations from. If the GPU runs out of memory, it simply stalls 3 frames to synchronize with the CPU.

Simulation Libraries Maybe

This kind of thinking is necessary for using OpenMM in real-time settings, where having the CPU wait on GPU results means UI hitches. For medium-small systems (the ones viable in real-time), CPU<->GPU latency can even dominate the execution time.

peastman:
It has no effect on GPU-CPU communication.  This change is an optimization for very large systems, not small ones.

philipturner:
> Unfortunately, that's incompatible with this optimization, since large blocks would not be spatially compact.

This might be possible if you could allocate memory at a finer granularity, GPU-side.

> Therefore, this change is expected to make building the neighbor list faster, but computing interactions slightly slower.

In the second level of my BVH, cells vary from 4/8 - 4/30 nm. That's 25-250 atom references/cell for solid diamond. However, less dense substances like water might be even cheaper.

peastman:
> This might be possible if you could allocate memory at a finer granularity, GPU-side.

There's no connection.  For this optimization to be useful, each large block (consisting of 32 regular atom blocks) must be spatially compact.  Otherwise they have huge bounding boxes, and aren't useful as a first level filter.  If the blocks are sorted by size instead of position, they aren't compact.

philipturner:
Could OpenMM incorporate a uniform grid BVH to cheaply sort atoms by position? For example, you could extract ~11 1024-blocks from a 4x4x4 voxel of solid diamond.

RaulPPelaez:
Since you are playing around with the idea of a full blown LBVH this one is the most performant one I am aware of: https://www.sciencedirect.com/science/article/pii/S092702561930206X?via%3Dihub

bdenhollander:
This looks promising! If this is ported to OpenCL, I can test on AMD.

philipturner:
> Since you are playing around with the idea of a full blown LBVH this one is the most performant one I am aware of:

An issue with uniform grids, is they are typically dense. I'm working to make them 2-level sparse. This approach unlocks optimizations like varying granularity of the second level, which helps balance compute cost for non-uniform distributions.

In the paper, the BVH took more time to construct than the grid. MD uses a different approach than RT (DDAs), so it would be helpful to understand why grid traversal is slower.

peastman:
The atoms are already sorted by position.  That's why this optimization works.  The existing neighbor list building method resorts the atom blocks by size, which would break it.  That's why I have to disable it when using this optimization:

```c++
if (!useLargeBlocks)
    blockSorter->sort(sortedBlocks);
```

> If this is ported to OpenCL, I can test on AMD.

I'm going to look into it.  The CUDA implementation makes use of a `ballot()` function, which isn't available in OpenCL.  The alternative is to use local memory for communicating between the threads, but that's higher overhead.  So this might end up being for CUDA only.  It would also work fine in HIP.

peastman:
> IIUC there is no code. You expect the user to pre-sort atoms at a 32-atom granularity.

We sort the atoms every 250 steps.  It's done in `ComputeContext::reorderAtoms()`.  Also see the discussion in the [Developer Guide](http://docs.openmm.org/latest/developerguide/06_opencl_platform.html#reordering-of-particles).

> There are OpenCL 2.0 extensions for SIMD ballots.

What are the extensions?

philipturner:
> What are the extensions?

https://github.com/philipturner/opencl-metal-stdlib#features

peastman:
Thanks!  Unfortunately, it doesn't look like the OpenCL implementations we care about support it.  Here are the extensions Apple supports.

> cl_APPLE_SetMemObjectDestructor cl_APPLE_ContextLoggingFunctions cl_APPLE_clut cl_APPLE_query_kernel_names cl_APPLE_gl_sharing cl_khr_gl_event cl_khr_byte_addressable_store cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_3d_image_writes cl_khr_image2d_from_buffer cl_khr_depth_images 

And here are the ones AMD supports.

> cl_khr_fp64 cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_int64_base_atomics cl_khr_int64_extended_atomics cl_khr_3d_image_writes cl_khr_byte_addressable_store cl_khr_fp16 cl_khr_gl_sharing cl_amd_device_attribute_query cl_amd_media_ops cl_amd_media_ops2 cl_khr_image2d_from_buffer cl_khr_subgroups cl_khr_depth_images cl_amd_copy_buffer_p2p cl_amd_assembly_program

The extension we would need is `cl_khr_subgroup_ballot`.

peastman:
I ported it to OpenCL.  Here are the times to simulate 500 steps for the water boxes on M1 Pro.

image

|Benchmark|Atoms|Original|New|Ratio|
|---|---|---|---|---|
pme|23558|157.54|139.637|0.886359019931446
apoa1pme|92224|38.4132|39.9721|1.0405824039653
amber20-cellulose|408609|7.45802|9.61445|1.28914242654217
amber20-stmv|1067095|1.93384|3.2664|1.68907458734952

It's similar to what we saw for CUDA on the RTX 4080, but the speedups are even larger.

peastman:
Here are the numbers for Radeon RX 5700 XT.

image

|Benchmark|Atoms|Original|New|Ratio|
|---|---|---|---|---|
pme|23558|385.581|248.548|0.644606451044009
apoa1pme|92224|110.795|105.366|0.950999593844487
amber20-cellulose|408609|20.9923|24.7377|1.17841780081268
amber20-stmv|1067095|5.11277|7.4702|1.46108665165849

The general picture is the same for all three GPUs.  The water boxes are faster across the board, with the speedup increasing steadily with system size for large systems.  For the PME benchmarks, DHFR becomes slower on all three GPUs, while Cellulose and STMV are faster on all three.  ApoA1 becomes about 4% faster on the 4080 and M1 Pro, but about 5% slower on the Radeon.  On all three GPUs, the water box that is roughly the same size as ApoA1 has minimal change in speed.  That suggests a cutoff of around 90,000 atoms would make sense for switching between the two behaviors.

philipturner:
> That suggests a cutoff of around 90,000 atoms would make sense for switching between the two behaviors.

I am not able to implement switching heuristics in my Metal plugin, because the number of atoms isn't known when the `MetalContext` constructor invokes. Code located inside `github.com/openmm/openmm` needs to execute such a heuristic before calling into `philipturner/openmm-metal`, just like with `useNeighborList`. In the meantime, both features are set manually through environment variables.

philipturner:
Reached a new performance maximum for apoa1rf: 152 -> 157 ns/day. With SIMD ballots and half precision, the overhead is so little, there’s no need for a switching heuristic. It varies between $\pm 4 \\%$ faster/slower for small systems, but <b>isn't consistently worse.</b>

<details>
<summary>Standard Benchmarks</summary>

Benchmark | Atoms | ns/day (Original) | ns/day (1024-blocks)
-- | -- | -- | --
gbsa (No NNL) | 2489 | 858.1 | n/a
gbsa (NNL) | 2489 | 632.0 | 659.5
rf | 23558 | 470.6 | 463.6
pme | 23558 | 267.8 | 262.1
apoa1rf | 92224 | 153.3 | 157.6
apoa1pme | 92224 | 80.1 | 80.7
amber20-cellulose | 408609 | 20.2 | 20.8
amber20-stmv | 1067095 | 7.23 | 7.51



</details>


<details>
<summary>Disabling Neighbor List (Reduces CPU Bottleneck)</summary>

Small Box Sizes (No NNL) | Atoms | ns/day (Original) | ns/day (1024-blocks)  
-- | -- | -- | --
2.0 nm | 774 | 1020 | n/a
2.5 nm | 1503 | 957 | n/a
3.0 nm | 2661 | 820 | n/a
3.5 nm | 4158 | 627 | n/a
4.0 nm | 6282 | 407 | n/a
4.5 nm | 8844 | 257 | n/a
5.0 nm | 12255 | 163 | n/a



</details>

<details>
<summary>Small Box Sizes</summary>

Small Box Sizes (NNL) | Atoms | ns/day (Original) | ns/day (1024-blocks) |
-- | -- | -- | -- 
2.0 nm | 774 | 877 | 892
2.5 nm | 1503 | 795 | 813
3.0 nm | 2661 | 671 | 689
3.5 nm | 4158 | 590 | 605
4.0 nm | 6282 | 490 | 493
4.5 nm | 8844 | 389 | 405
5.0 nm | 12255 | 330 | 337
6.0 nm | 21384 | 297 | 305
7.0 nm | 33840 | 213 | 216

</details>

<details>
<summary>Large Box Sizes</summary>

Large Box Sizes (NNL) | Atoms | s/100K/500 (Original) | s/100K/500 (1024-blocks) |
-- | -- | -- | --
5.0 nm / 5000 steps | 12255 | 4.29 | 4.21
10.0 nm / 5000 steps | 98880 | 2.02 | 2.02
15.0 nm / 2500 steps | 335025 | 1.95 | 1.90
20.0 nm / 1250 steps | 792450 | 2.03 | 1.92
25.0 nm / 600 steps | 1550160 | 2.35 | 2.09
30.0 nm / 300 steps | 2682600 | 2.61 | 2.32

</details>

It is strange that M1 Max gets the exact same speed as M1 Pro, despite having double the compute and bandwidth.

peastman:
I tested OpenCL on NVIDIA, and it also turned out to be slightly slower for ApoA1 when using the new code.  I therefore set the switchover at 90,000 atoms for CUDA, but 100,000 for OpenCL.

> With SIMD ballots and half precision, the overhead is so little, there’s no need for a switching heuristic.

It's not just overhead building the neighbor list.  It produces a less efficient neighbor list with more wasted interactions.

bdenhollander:
Here are results for Radeon RX 6600 OpenCL on Windows. 

Large blocks was set to true for all tests rather than based on 100000 atoms to confirm the cutoff point. It does appear to be a reasonable value.

| Benchmark          | Atoms   | Original | New     | Ratio  |
|--------------------|---------|----------|---------|--------|
| pme                | 23558   | 304.342  | 235.449 | 0.7736 |
| apoa1pme           | 92224   | 77.6963  | 81.0731 | 1.0435 |
| amber20-cellulose  | 408609  | 18.0175  | 22.7871 | 1.2647 |
| amber20-stmv       | 1067095 | 4.65864  | 6.968   | 1.4957 |

Water boxes

| Width (nm) | Atoms   | Original ns/day | New ns/day | Ratio  |
|------------|---------|----------|------|--------|
| 4          | 6282    | 499      | 506  | 1.0140 |
| 5          | 12255   | 381      | 393  | 1.0315 |
| 6          | 21384   | 341      | 342  | 1.0029 |
| 7          | 33840   | 221      | 217  | 0.9819 |
| 8          | 50442   | 151      | 151  | 1.0000 |
| 9          | 72279   | 109      | 110  | 1.0092 |
| 10         | 98880   | 85.7     | 87.2 | 1.0175 |
| 11         | 131445  | 69.2     | 74   | 1.0694 |
| 15         | 335025  | 24.5     | 29.7 | 1.2122 |
| 20         | 792450  | 7.47     | 11.9 | 1.5930 |
| 25         | 1550160 | 2.62     | 5.46 | 2.0840 |
| 30         | 2682600 | 1.05     | 2.93 | 2.7905 |

philipturner:
The elephant in the room is that factor of 0.60-0.80 decrease for PME. My code experiences no such thing - only 2% variation. If you remove that massive decrease, the other regressions are just noise.

| Processor | PME Before | PME After | Regression |
| --- | ---------- | --------- | ----- |
| CUDA GPU | 2036.145 | 1694.8 | -17% |
| RDNA 1 GPU | 385.581 | 248.548 | -36% |
| RDNA 2 GPU | 304.342 | 235.449 | -23% |
| Apple GPU (OpenCL) | 157.54 | 139.637 | -13% |
| Apple GPU (Metal) | 267.8 | 262.1 | <b>-2%</b> |

peastman:
Your code shows minimal benefit on any of the tests, which suggests something's wrong with it.  You should be getting a big speedup in the larger water boxes, not a small slowdown.

philipturner:
> You should be getting a big speedup in the larger water boxes, not a small slowdown.

Please look over the data in "Large Box Sizes" (lower is better): https://github.com/openmm/openmm/pull/4147#issuecomment-1644887309

The data does not show a small slowdown; it shows a small speedup. The difference is less than OpenCL because the $O(n^2)$ part was already quite efficient.

bdenhollander:
> The elephant in the room is that factor of 0.60-0.80 decrease for PME. 

Original - 360 us kernel runtime
image

New - 867 us kernel runtime
image

One wavefront in `findInteractingBlocks` during the PME benchmark is running a very long time compared to the rest and dragging down the kernel run time. The uneven work distribution to the earliest loop iteration may be twice as bad now that it does large blocks and then small blocks (see #3966).

philipturner:
I've done a lot of theoretical analysis with a similar problem regarding triangular matrices. The first thing to do: traverse the outer loop in reverse order. It is simple but causes the most expensive iterations to happen first. Cheaper ones happen last, at a time when there's many gaps in the work distribution.

You should combine that with some optimizations to the large-blocks part, such as using SIMD instrinsics instead of local memory. That decreases the total latency for any threadgroup.

peastman:
@bdenhollander what are those graphs of?  I don't know what I'm looking at.

> such as using SIMD instrinsics instead of local memory

Are you referring to using a `ballot()` function?  The CUDA implementation does that.  The OpenCL implementation can't because the necessary extensions aren't available.  Or do you mean some different intrinsic for something else?

philipturner:
There is a similar intrinsic in Metal, except there's dedicated circuitry that performs the SIMD ballot in 2 cycles.  You can access it in OpenCL through:

```opencl
__attribute__((__overloadable__)) ulong simd_ballot(bool expr)
  __asm("air.simd_ballot.i64")
```

CUDA GPUs just use circuitry from threadgroup memory, which takes 4 cycles. Because Apple GPUs can perform SIMD synchronizations so quickly, their threadgroup memory is slower than CUDA GPUs, taking 8 cycles.

bdenhollander:
> @bdenhollander what are those graphs of? I don't know what I'm looking at.

There are graphs of GPU utilization from Radeon GPU Profiler of a single execution of `findBlocksWithInteractions` during the pme benchmark. Use large blocks was set to true so that the 100000 atom cutoff would not apply. The first half of the new/after graph is similar to the execution without this change applied. The second half of the graph, where there is a series of short yellow pillars, is when only one wavefront is active. That single wavefront is drastically extending the runtime of the kernel and thus reducing overall ns/day.

6 nm and 6.1 nm water boxes with a similar number of atoms as the pme benchmark do not exhibit a similar regression. There must be something within the pme test itself that causes the performance decrease rather than the specific number of atoms. Perhaps there is very dense or very sparse block that has many neighbours and the extra large block calculations filter out very few or 0 blocks, adding execution time rather than reducing it.

I tried adding some counters for skipped and processed large blocks and printing that out at the end of the function. During pme, 0-30% of large blocks were skipped while amber20-cellulose reported over 90%. Could skip percentage be calculated during initialization of the system and used to set whether to enable large blocks rather than the atom count?

peastman:
I tried modifying it to use an atomic counter for assigning atom blocks rather than just looping over a fixed set.  That allows each warp to grab work as it becomes available and should improve balancing.  On Mac it produced a tiny speedup, a few percent or less.  On NVIDIA there was no consistent benefit, and it often became slower.

> That single wavefront is drastically extending the runtime of the kernel and thus reducing overall ns/day.

Not necessarily.  Modern GPUs can overlap kernels.  If there's no direct dependency between successive kernels, it can start running the next one even if a few thread blocks of the previous one are still running.  Also, when using PME we run the direct and reciprocal space parts on different streams, which lets them run in parallel.  If a few thread blocks take longer than the rest to finish, it doesn't mean the rest of the GPU is idle.

philipturner:
> On Mac it produced a tiny speedup, a few percent or less.

Could you show me the commit, so I can implement the algorithm in my Metal plugin?

peastman:
I haven't checked it in.  Here's a diff of the changes.

```
diff --git a/platforms/opencl/src/OpenCLNonbondedUtilities.cpp b/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
index 2c8181b..63a187c 100644
--- a/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
+++ b/platforms/opencl/src/OpenCLNonbondedUtilities.cpp
@@ -293,7 +293,7 @@
         int numAtoms = context.getNumAtoms();
         interactingTiles.initialize<cl_int>(context, maxTiles, "interactingTiles");
         interactingAtoms.initialize<cl_int>(context, OpenCLContext::TileSize*maxTiles, "interactingAtoms");
-        interactionCount.initialize<cl_uint>(context, 1, "interactionCount");
+        interactionCount.initialize<cl_uint>(context, 2, "interactionCount");
         int elementSize = (context.getUseDoublePrecision() ? sizeof(cl_double) : sizeof(cl_float));
         blockCenter.initialize(context, numAtomBlocks, 4*elementSize, "blockCenter");
         blockBoundingBox.initialize(context, numAtomBlocks, 4*elementSize, "blockBoundingBox");
@@ -305,9 +305,9 @@
         oldPositions.initialize(context, numAtoms, 4*elementSize, "oldPositions");
         rebuildNeighborList.initialize<int>(context, 1, "rebuildNeighborList");
         blockSorter = new OpenCLSort(context, new BlockSortTrait(context.getUseDoublePrecision()), numAtomBlocks, false);
-        vector<cl_uint> count(1, 0);
+        vector<cl_uint> count(2, 0);
         interactionCount.upload(count);
-        rebuildNeighborList.upload(count);
+        rebuildNeighborList.upload(&count[0]);
     }
 }
 
diff --git a/platforms/opencl/src/kernels/findInteractingBlocks.cl b/platforms/opencl/src/kernels/findInteractingBlocks.cl
index 61a5d0c..735ba91 100644
--- a/platforms/opencl/src/kernels/findInteractingBlocks.cl
+++ b/platforms/opencl/src/kernels/findInteractingBlocks.cl
@@ -97,6 +97,7 @@
     if (rebuild) {
         rebuildNeighborList[0] = 1;
         interactionCount[0] = 0;
+        interactionCount[1] = 0;
     }
 }
 
@@ -138,7 +139,15 @@
 
     // Loop over blocks.
 
-    for (int block1 = startBlockIndex+warpIndex; block1 < startBlockIndex+numBlocks; block1 += totalWarps) {
+    while (true) {
+        SYNC_WARPS;
+        if (indexInWarp == 0)
+            buffer[0] = ATOMIC_ADD(&interactionCount[1], 1);
+        SYNC_WARPS;
+        int block1 = startBlockIndex+buffer[0];
+        if (block1 >= startBlockIndex+numBlocks)
+            break;
+
         // Load data for this block.  Note that all threads in a warp are processing the same block.
         
         real2 sortedKey = sortedBlocks[block1];
```

bdenhollander:
I tested the atomic counter with a slight modification to only use it when also using large blocks. There needs to be a sufficient number of outer loop iterations for the improved balancing to impact ns/day.

```C++
#ifdef USE_LARGE_BLOCKS
    while (true) {
        SYNC_WARPS;
        if (indexInWarp == 0)
            buffer[0] = ATOMIC_ADD(&interactionCount[1], 1);
        SYNC_WARPS;
        int block1 = startBlockIndex + buffer[0];
        if (block1 >= startBlockIndex + numBlocks)
            break;
#else
    for (int block1 = startBlockIndex+warpIndex; block1 < startBlockIndex+numBlocks; block1 += totalWarps) {
#endif
```

Radeon RX 6600 OpenCL on Windows (* indicates no expected improvement since atoms < 100000)

| Benchmark          | Atoms   | Original | New     | Ratio  |
|--------------------|---------|----------|---------|--------|
| pme                | 23558   | 308.417  | 304.067 | 0.9859 * |
| apoa1pme           | 92224   | 79.4715  | 79.718  | 1.0031 * |
| amber20-cellulose  | 408609  | 22.9065  | 23.478  | 1.0249 |
| amber20-stmv       | 1067095 | 6.99547  | 7.27253 | 1.0396 |

Water boxes
| Width (nm) | Atoms   | Original | New  | Ratio  |
|------------|---------|----------|------|--------|
| 10         | 98880   | 84.4     | 84.2 | 0.9976 * |
| 11         | 131445  | 68       | 68.3 | 1.0044 |
| 15         | 335025  | 29.8     | 30.4 | 1.0201 |
| 20         | 792450  | 11.9     | 12.1 | 1.0168 |
| 25         | 1550160 | 5.46     | 5.69 | 1.0421 |

PR #4148, opened 2023-07-20
Fixed numeric overflow in computing interaction groups

### LINKED ISSUE #4127: Energy is NaN after very few time steps for large systems using interaction groups
Dear OpenMM community,

I hope this finds you well. 

Recently, I am using DrudeLangevinIntegrator (timestep=1fs, frictioncoefficient=1/ps, systemtemp=300K, drudetemp=1K) to run a ionic liquid and water mixture system with polarizable force fields. The system contains 1000 ion pairs and 250 water molecules. There is vacuum above and below the liquid since I am studying the air-liquid interface. But I keep getting "Energy is NaN" error and observing extremely high potential energy (I have output the energy from difference forces, and it occured to be the harmonic bond energy that is extremely high after tens of time steps). I first think it maybe the problem with my force field (I use a SAPT force field downloaded from here: https://github.com/jmcdaniel43/SAPT_force_field_OpenMM). But I checked my scripts many times and didn't found any mistake. 

Then I though I should make a small system for debug, so I started running with a system containing 100 ion pairs and 25 water molecules with exact the same scripts I used above. Surprisingly, I got the small system running for 10 ns without any error. I have also tested systems with "500 ion pairs and 125 water" and "800 ion pairs and 200 water". Both of them can survive at least for more than 1 ns. But then I run the system with "900 ion pairs and 225 water", the energy becames extremely high again after tens of time steps.

This problem seems very weird for me. From my understanding, if it is because of a bad force field, then even for a small system, it should crash very soon. But the system with 800 ion pairs and 200 water works for more than 1 ns, the system with 900 ion pairs and 225 water crashes very soon. I am not sure if this is the problem with the force field itself or there is some bug with the integrator. 

Any help would be appreciated! Please inspire me.

Best,
Sijia

Exmaple input files are attached with systems "800 ion pairs + 200 water" and "900 ion pair and 225 water". Part of the output files are also included. My initial configuration were generated by packmol.

[example.tar.gz](https://github.com/openmm/openmm/files/11888474/example.tar.gz)

peastman:
Fixes #4127.

The error was caused by a combination of two factors:

1. The system had a CustomNonbondedForce that used very large interaction groups.  In the issue that was reported, the interaction groups contained nearly 3.5 billion interactions.
2. You were running on a GPU with a very large number of SMs.  It happened on RTX 3090 (82 SMs) and H100 (114 SMs), but not on RTX 3080 (68 SMs) or RTX 4080 (76 SMs).

The combination of those two factors led to a number that exceeded the range of a 32 bit integer, leading to some interactions being included more than once.

### DIFF
--- a/platforms/common/src/kernels/customNonbondedGroups.cc
+++ b/platforms/common/src/kernels/customNonbondedGroups.cc
@@ -50,8 +50,9 @@ KERNEL void computeInteractionGroups(
     LOCAL AtomData localData[LOCAL_MEMORY_SIZE];
     LOCAL int reductionBuffer[LOCAL_MEMORY_SIZE];
 
-    const unsigned int startTile = (useNeighborList ? warp*numGroupTiles[0]/totalWarps : FIRST_TILE+warp*(LAST_TILE-FIRST_TILE)/totalWarps);
-    const unsigned int endTile = (useNeighborList ? (warp+1)*numGroupTiles[0]/totalWarps : FIRST_TILE+(warp+1)*(LAST_TILE-FIRST_TILE)/totalWarps);
+    mm_ulong wl = warp;
+    const unsigned int startTile = (unsigned int) (useNeighborList ? wl*numGroupTiles[0]/totalWarps : FIRST_TILE+wl*(LAST_TILE-FIRST_TILE)/totalWarps);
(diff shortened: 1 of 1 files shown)


sef43:
Amazing how often integer overflow crops up. Nice catch!

PR #4153, opened 2023-07-25
Update to a more modern version of pymbar

dwilliams-nobias:
It appears that this example is using an older (and likely considered obsolete) version of pymbar.


peastman:
Thanks!

PR #4155, opened 2023-07-28
Add in cstddef fix for pocket FFT

### LINKED ISSUE #4154: Compiling on very old compilers needs <cstddef> include for pocketfft support
When compiling on a very old compiler `gcc-4.8.5` in a docker container as part of exploring some GLIBC symbol  versioning issues, I came across the problem that `openmm/libraries/pocketfft/include/pocketfft_hdronly.h` fails with 
` 'max_align_t' was not declared in this scope`

This is because std:max_align wasn't a thing in gcc versions that old. Can be fixed with a simple #include <cstddef>.

hmacdope:
Fixes #4154

### DIFF
--- a/libraries/pocketfft/include/pocketfft_hdronly.h
+++ b/libraries/pocketfft/include/pocketfft_hdronly.h
@@ -52,8 +52,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #include <cmath>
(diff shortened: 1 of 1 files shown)


peastman:
@hmacdope is it ok to close this, since I think we've agreed that it only affects very old compilers that aren't supported?

hmacdope:
Yes for sure @peastman, eventually I will contribute something useful :laughing:

PR #4161, opened 2023-07-31
Fix for patches that affect only bonds, not atoms

### LINKED ISSUE #4152: Parametrize protein with uncapped residues
I'd like to parametrize a (fragment of) a protein where the backbone has been cut at certain residues but no caps have been added: it would look like NH-CA1-C(=O)- ... -NH-CAn-C(=O) 

The first and last residue here aren't ionized (NH3+ and COO-) and also aren't capped with ACE/NME, and aren't capped with hydrogens either.  This would be very useful in many cases where it's necessary to simulate a partial protein but adding ACE/NME can introduce clashes.

Desired behavior: treat these like normal (middle of chain, not terminal) residues, and add any terms where all participating atoms exist; so for example in the first residue there would be a H-N-CA1 angle term, but not a previousC-N-H angle term.

Is it possible to do this with the Patches mechanism?

Using the CTER/NTER patches in charmm36.xml as example, would something really simple like this work?

```
<Patch name="NCUT">
    <RemoveExternalBond atomName="N"/>
</Patch>

<Patch name="CCUT">
    <RemoveExternalBond atomName="C"/>
</Patch>
```

(I'm mainly interested in doing this on top of amber14-all.xml)

peastman:
Fix #4152.

cc @aizvorski

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -391,9 +391,11 @@ class ForceField(object):
                     for bond in patch.findall('RemoveExternalBond'):
                         atom = ForceField._PatchAtomData(bond.attrib['atomName'])
                         patchData.deletedExternalBonds.append(atom)
-                    atomIndices = dict((atom.name, i) for i, atom in enumerate(patchData.addedAtoms[atomDescription.residue]+patchData.changedAtoms[atomDescription.residue]))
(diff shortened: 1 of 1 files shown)


aizvorski:
@peastman Thanks!  I tried this, it seems to work with my xml file

peastman:
Thanks!

PR #4195, opened 2023-08-18
Amoeba and ATM Force minor cleanup

bdenhollander:
- Fix variable name in string
- Remove odd space between variable and period that is inconsistently styled
- Replace random tabs with spaces in new ATM Force


peastman:
Looks good.  Thanks!

Issue #1155, opened 2015-09-28
Reporting on minimisation progress


mj-harvey:
Hi,

I'd like to be able to monitor the progress of an energy minimisation - ie get a report of the iteration # and system pe.  Reporters attached  via simulation.reporters.append() don't fire during minimisation, only dynamics - is there some other way that i have missed?

Matt

peastman:
Not currently, but it would be a useful feature.  I'll tag this as an enhancement request.

mj-harvey:
Thanks Peter, appreciated!

Matt

swails:
+1

My thought here is to treat this kind of like `Simulation.step` works, and successively call `minimizeEnergy` with some maximum number of steps equal to the report interval until the criteria is met or the real maximum is reached.  I'd thought about this before and that's what I came up with.  If it sounds good, I can try and draft a PR for it.

peastman:
The problem is that the steps aren't independent.  LBFGS uses information from previous iterations to improve its behavior on future iterations.  So if you call it ten times with maxIterations=10, that doesn't work nearly as well as calling it once with maxIterations=100.

Constraints will also cause problems for it.  LBFGS doesn't support constrained optimization, so instead we replace all the constraints with harmonic forces.  We run the whole minimization that way, then check the constrained distances.  If any is violated by more than the allowed tolerance, we increase the strength of the harmonic forces and minimize again.  If the constraints are only violated by a little bit, we continue minimizing from where we left off.  But if they're too far off, then we go back to the beginning and minimize from scratch.  That's important to avoid getting stuck in a bad state and having the minimization completely fail.

swails:
Glad I didn't try this before, then :)

ajasja:
I would also like to monitor minimisation (especially useful when debugging an exploding system:). 

In NAMD the default is that minimisation gets written to the trajectory file. Personally I would argue that the reporters should fire during energy minimization.

MJfadeaway:
I wonder what is the real positions after running energy minimization

peastman:
You can query the positions with `context.getState(getPositions=True).getPositions()`.

aizvorski:
@peastman Continuing discussion here:

> The process works by replacing all constraints with harmonic restraints, then minimizing, then checking whether the constraints are satisfied, then tightening the restraints if they aren't, then possibly doing a bit more minimization or possibly going back to the beginning and restarting the whole minimization from scratch. So what do you report and when?

[...]

PR #4207, opened 2023-08-28
Report status during minimization

peastman:
Implements #1155.

This adds a virtual MinimizationReporter class.  You subclass it, implementing the `report()` method, and pass it to `minimize()`.  It gets called after each iteration.

The Python API was a little challenging.  This is the first time we've had an interface in which C++ code invokes Python code, instead of the other way around.  I ran into some SWIG issues (https://github.com/swig/swig/issues/2670 and https://github.com/swig/swig/issues/2671), but I think I have workarounds for them.

### DIFF
--- a/docs-source/api-c++/extras.rst
+++ b/docs-source/api-c++/extras.rst
@@ -62,6 +62,7 @@ that they aren't important!
     :maxdepth: 2
 
     generated/LocalEnergyMinimizer
+    generated/MinimizationReporter
     generated/NoseHooverChain
     generated/OpenMMException
     generated/Vec3
--- a/openmmapi/include/openmm/LocalEnergyMinimizer.h
+++ b/openmmapi/include/openmm/LocalEnergyMinimizer.h
@@ -76,12 +76,13 @@ public:
      * objective function with respect to them, and a set of useful statistics.  In particular,
      * args contains these values:
      * 
-     * <ul>
-     * <li>"system energy": the current potential energy of the system</li>
-     * <li>"restraint energy": the energy of the harmonic restraints</li>
-     * <li>"restraint strength": the force constant of the restraints (in kJ/mol/nm^2)</li>
-     * <li>"max constraint error": the maximum relative error in the length of any constraint</li>
-     * </ul>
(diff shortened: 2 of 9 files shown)


peastman:
@swails I believe the jenkins failure is because it's using a very old version of SWIG.  Can we upgrade it?

sef43:
One problem I have found when testing this out is that if I use the Python API and have a syntax error in the report method it will not be caught and the minimizer will just silently stop. This could be very problematic for users having to blindly debug their python report code. E.g. this code results in the minimizer stopping with no errors reported.
```python
class Reporter(MinimizationReporter):
    def report(self, iteration, x, grad, args):
        # purposely incorrect python code with a syntax error
        # prnt instead of print
        prnt(iteration)

reporter = Reporter()


# Perform local energy minimization
print("Minimizing energy...")
simulation.minimizeEnergy(reporter=reporter)
```

bdenhollander:
> It's a bigger problem for C++. GCC has a "feature" that it happily lets you omit the return statement from a function that declares a return value, and then it just returns whatever garbage happens to be on the stack. In practice, though, most reporters are likely to be written in Python rather than C++.

Enabling `-Werror=return-type` should catch a [missing return statement](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=43943).

peastman:
I switched to the other design.  Overall I think it will cause fewer problems for users.  Exceptions thrown inside the reporter get propagated back to the caller as expected.  If a reporter implemented in Python omits a return value, SWIG produces the exception

```
TypeError: SWIG director type mismatch in output value of type 'bool'
```

Hopefully that's clear enough.

sef43:
Yes this is much better. Can we explicitly mention this error in the Docs for minimizationReporter? I am not sure if it is common for a user to come across a SWIG error using the Python API in other parts of OpenMM so it might not be very obvious what the problem is.

peastman:
I'm not sure if there's any clean way of doing that.  The API docs are written for the C++ API.  Then they automatically get copied over to the Python API.  We don't have any good mechanism for customizing them at the Python level.

PR #4208, opened 2023-08-28
Remove amber20-factorix from benchmarks

### LINKED ISSUE #3391: amber20-factorix benchmark fails with "PeriodicTorsionForce: periodicity must be positive"
It fails in this check (recently added): 
https://github.com/openmm/openmm/blame/610e92fa5f79fbc8cc48ea96d221b0583f817839/openmmapi/src/PeriodicTorsionForceImpl.cpp#L61

bdenhollander:
Test no longer runs and prints out a warning when running benchmark.py with `--verbose`. Fixes #3391


peastman:
Thanks!

PR #4214, opened 2023-09-01
Update IntegrationUtilities.cpp to include <functional>

### LINKED ISSUE #4213: Issue with include of `binary_function` on MSVC 22 build
```
\openmm\platforms\common\src\IntegrationUtilities.cpp(78,57): error C2504: 'binary_function': base class undefined [C:\Users\hugom\WORK\folding_at_home\openmm\build\platform
s\cuda\sharedTarget\OpenMMCUDA.vcxproj]
```

Occurs from `IntegrationUtilities.cpp`

Likely because the `<functional>` header is not included. 

Regardless the `std::binary_function` class has been removed in C++17

hmacdope:
Fixes #4213 

Including the `functional` header as per its standard location per: https://en.cppreference.com/w/cpp/utility/functional/binary_function

### DIFF
--- a/platforms/common/src/IntegrationUtilities.cpp
+++ b/platforms/common/src/IntegrationUtilities.cpp
@@ -36,6 +36,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdlib>
+#include <functional>
 #include <map>
 #include <set>
 


peastman:
Looks good.  Tests are running now.

peastman:
I see that binary_function is deprecated in C++11 and removed in C++17.  If you're inclined to replace it with whatever the recommended alternative is, that would also be great.

PR #4217, opened 2023-09-04
Remove std::binary_function from ConstraintOrderer functor

### LINKED ISSUE #4216: Replace `std::binary_function` as it is deprecated in C++11 and removed in C++17
I see that binary_function is deprecated in C++11 and removed in C++17.  If you're inclined to replace it with whatever the recommended alternative is, that would also be great.

_Originally posted by @peastman in https://github.com/openmm/openmm/issues/4214#issuecomment-1703041173_

hmacdope:
Fixes #4216 

Removes deprecated `std::binary_function` inheritance from `ConstraintOrderer` as it is removed in C++17.

### DIFF
--- a/platforms/common/src/IntegrationUtilities.cpp
+++ b/platforms/common/src/IntegrationUtilities.cpp
@@ -36,7 +36,6 @@
 #include <algorithm>
 #include <cmath>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


peastman:
Looks good.  Thanks!

PR #4245, opened 2023-09-26
Replace distutils with setuptools

peastman:
Distutils has been deprecated in favor of setuptools for quite a while.  Python 3.12 removes it altogether.  This switches the imports, which I think is the only change necessary.

### DIFF
--- a/wrappers/python/openmm/app/gromacstopfile.py
+++ b/wrappers/python/openmm/app/gromacstopfile.py
@@ -42,7 +42,7 @@ import openmm as mm
 import math
 import os
 import re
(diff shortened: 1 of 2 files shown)


bdenhollander:
There are a few more usages of `distutils` in `gromacstopfile.py` that should probably be addressed as well.

https://github.com/openmm/openmm/blob/b247e7d8cf772f22e9ab559047d1721d42660673/wrappers/python/openmm/app/gromacstopfile.py#L45

peastman:
Thanks!  I fixed them.

PR #4261, opened 2023-10-10
WIP - looking for a way to optimise performance of creating contexts by removing temporary arrays (and their associated mallocs/frees)

chryswoods:
Thanks for OpenMM - I've really enjoyed getting to know the code more as we integrate it more fully as part of the somd2 project.

As part of that project, I noticed that context creation could become very slow for larger systems which used `CustomNonbondedForce`, especially when constraints were added. Profiling indicated that this was because of a lot of mallocs/frees that were incurred by the creation of the temporary `std::vector<double>` arrays that held the results of `getParticleParameters`. This function was called for most atom-pairs as part of the loop that checks if atoms are the same (`areParticlesIdentical`).

This isn't an issue for the standard Force objects as their `getParticleParameters` functions use plain double references to return the parameter values.

There are different ways that this could be solved. I wanted to raise this as a WIP PR to get an idea of what approach would be best, as I would like to roll out a solution across all of the Custom Forces where this would be an issue.

The root problem is that the `X::ForceInfo` objects need to use the data contained within their matching `Force` object as part of context setup. For speed, they really would want privileged access to the data held in the `Force` object, either via API functions, or by getting direct access via `friend` or similar. In contrast, the `Force` objects want to present a simple (and consistent) API to end-users, as the `Force` objects form the public C++/Python API that end-users see. Thus, you likely don't want this API confused with hooks that the `X::ForceInfo` objects would need to get optimised access.

I've implemented Option 1 for `CustomNonbondedForce` in this PR, but am very open to being told to go down a different option, or to try something else.

## Option 1

We avoid the copy by adding a function to `CustomNonbondedForce` (in this case called `haveSameParameters`) that does the parameter comparison internally on the parameter arrays directly.

Benefit - the modifications are minimal and we don't need to use `friend` or other means to get privileged access to `CustomNonbondedForce`'s data.

Drawback - we now need to add functions to the `CustomNonbondedForce` public API that may be confusing to users. While this one function is not too bad, further optimisation may require more such functions.

## Option 1a

Do option 1, but make the `haveSameParameters` (and possible future other methods) be `private` or `protected`. Then use `friend` to give `CommonCalcCustomNonbondedForceKernel::ForceInfo` privileged access to these functions.

Benefit - all the benefits of option 1, but without polluting the public API of `CustomNonbondedForce`

Drawback - `friend` is a bit icky and code can become messy if this is over-used. But, there is clearly quite tight coupling between `X::ForceInfo` and its corresponding `Force`.

## Option 2

We add an overload `getParticleParameters(int index, double **ptr, int &size)` to get the raw pointer and size of the particle array, so that calling functions can get privileged access to the data and thus avoid the copy.

Benefit - simplest route that would work for anything that needs rapid access to this data which avoids a copy.

Drawback - lots - race conditions, potential for segfaults if memory is deallocated, invitation to developers to mis-use this privileged access to modify memory directly behind the back of the `CustomNonbondedForce` API etc etc

## Option 3

We create an implicitly shared vector type that does copy-on-write, so that we can pass back shallow copies of vectors that will automatically be deep-copied if they are edited.

Benefit - gives the benefit of Option 2 without the risk of being misused, race conditions, memory errors etc

Drawback - this would be a large amount of work, as OpenMM is heavily based on using std::vector. This would also be a big change of API.

==

There's almost certainly other options too! ;-)

What would be the best approach?

p.s. I've also included a diff for `CustomNonbondedForce::getParticleParameters` that replaces the copy of std::vector with a resize and memcpy. This allows the memory of the std::vector argument to be re-used, e.g. if this function is called several times. This way, if you loop over all atoms and want to get parameters, you can allocate the std::vector once to the right size before entering the loop, and then call `getParticleParameters` using that vector, do what you need, then keep that vector as the space for the next iterations `getParticleParameter` call. I'd like also to widen this approach out to other places where std::vector is used, using std::memcpy and std::memmove to maximise memory reuse and avoid malloc/free where possible.

p.p.s. This PR doesn't functionally change OpenMM - I've checked that all of the tests still pass. The change is that contexts involving `CustomNonbondedForce` are now created more quickly.

### DIFF
--- a/platforms/common/src/CommonKernels.cpp
+++ b/platforms/common/src/CommonKernels.cpp
@@ -1778,13 +1778,13 @@ public:
         }
     }
     bool areParticlesIdentical(int particle1, int particle2) {
-        // call a function that checks for parameter equivalence on 
-        // the force's arrays directly, so we avoid the malloc/free
-        // of copying the parameters into temporary vectors
-        // for the comparison
-        if (not force.haveSameParameters(particle1, particle2))
-            return false;
-
+        vector<double> params1;
+        vector<double> params2;
+        force.getParticleParameters(particle1, params1);
+        force.getParticleParameters(particle2, params2);
+        for (int i = 0; i < (int) params1.size(); i++)
+            if (params1[i] != params2[i])
+                return false;
         if (groupsForParticle.size() > 0 && groupsForParticle[particle1] != groupsForParticle[particle2])
             return false;
         return true;
--- a/platforms/common/src/CommonKernels.cpp
+++ b/platforms/common/src/CommonKernels.cpp
@@ -1080,7 +1080,7 @@ void CommonCalcCMAPTorsionForceKernel::copyParametersToContext(ContextImpl& cont
 class CommonCalcCustomExternalForceKernel::ForceInfo : public ComputeForceInfo {
 public:
     ForceInfo(const CustomExternalForce& force, int numParticles) : force(force), indices(numParticles, -1) {
-        thread_local static vector<double> params;
+        vector<double> params;
         for (int i = 0; i < force.getNumParticles(); i++) {
             int particle;
             force.getParticleParameters(i, particle, params);
--- a/platforms/common/src/CommonKernels.cpp
+++ b/platforms/common/src/CommonKernels.cpp
@@ -1125,8 +1125,8 @@ void CommonCalcCustomExternalForceKernel::initialize(const System& system, const
     vector<vector<int> > atoms(numParticles, vector<int>(1));
     params = new ComputeParameterSet(cc, force.getNumPerParticleParameters(), numParticles, "customExternalParams");
     vector<vector<float> > paramVector(numParticles);
-    vector<double> parameters;
     for (int i = 0; i < numParticles; i++) {
+        vector<double> parameters;
         force.getParticleParameters(startIndex+i, atoms[i][0], parameters);
         paramVector[i].resize(parameters.size());
         for (int j = 0; j < (int) parameters.size(); j++)
@@ -1876,8 +1876,8 @@ void CommonCalcCustomNonbondedForceKernel::initialize(const System& system, cons
         globals.initialize<float>(cc, force.getNumGlobalParameters(), "customNonbondedGlobals");
     vector<vector<float> > paramVector(paddedNumParticles, vector<float>(numParams, 0));
     vector<vector<int> > exclusionList(numParticles);
-    vector<double> parameters;
     for (int i = 0; i < numParticles; i++) {
+        vector<double> parameters;
         force.getParticleParameters(i, parameters);
         paramVector[i].resize(parameters.size());
         for (int j = 0; j < (int) parameters.size(); j++)
@@ -2758,8 +2758,8 @@ void CommonCalcCustomGBForceKernel::initialize(const System& system, const Custo
         globals.initialize<float>(cc, force.getNumGlobalParameters(), "customGBGlobals");
     vector<vector<float> > paramVector(paddedNumParticles, vector<float>(numParams, 0));
     vector<vector<int> > exclusionList(numParticles);
-    vector<double> parameters;
     for (int i = 0; i < numParticles; i++) {
+        vector<double> parameters;
         force.getParticleParameters(i, parameters);
         for (int j = 0; j < (int) parameters.size(); j++)
             paramVector[i][j] = (float) parameters[j];
@@ -4239,8 +4239,8 @@ void CommonCalcCustomManyParticleForceKernel::initialize(const System& system, c
     
     params = new ComputeParameterSet(cc, force.getNumPerParticleParameters(), numParticles, "customManyParticleParameters");
     vector<vector<float> > paramVector(numParticles);
-    vector<double> parameters;
     for (int i = 0; i < numParticles; i++) {
+        vector<double> parameters;
         int type;
         force.getParticleParameters(i, parameters, type);
         paramVector[i].resize(parameters.size());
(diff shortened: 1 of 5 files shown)


peastman:
Thanks!  Do you have a test case I can use to reproduce it?  I'd like to try profiling and analyzing it to see exactly what's going on.

chryswoods:
I've created an XML file of a system plus written a small timing test script (test_openmm.py). You can download them from here: https://drive.google.com/drive/folders/1BsVlT5BDmrGZ0s102YSeUf58UW68udz-?usp=share_link

The XML file was created from a context created from perturbation between two small ligands in water. It was created by the create_xml.py script.

Here's a screen grab I get from the profiler result (MacOS, M1, ARM64 standard conda package, 8.0)

image

peastman:
Perfect, thanks.

chryswoods:
Interesting. I agree that `thread_local static` is an easy fix. I suspect the cost will be O(N2) because this is called on (most?) pairs of atoms so it is worth doing. I agree it is worth balancing optimisation, code complexity and robustness.

I also find it surprising that a vector copy is slower than a manual copy in a loop. I'd have hoped that the vector copy was doing a memcpy or similar underneath? Maybe it has some extra aliasing checking logic? (not that I can see how you could create two std::vector objects that could be aliased)

I'll keep digging locally as well. I think I have done something in the setup of the context that is causing OpenMM to work harder than is needed in `findMoleculeGroups` which is then making so many called to `areParticlesIdentical` and then `getParticleParameters`.

chryswoods:
Yes - looking deeper I am making `findMoleculeGroups` work harder. I've added bond constraints to water molecules that are slightly different because they are based on the input coordinates. This means that every water molecule looks like a different molecule because its constraints are different. I'll keep debugging and will let you know how I get on.

peastman:
In normal situations the cost is only O(N).  It compares every water molecule to the first water molecule.  You've managed to create a kind of pathological situation by having a very large number of molecules, all with the same numbers of atoms and constraints, but with every molecule being unique.  That's a very unusual situation.

I tested on a different computer with an x86 processor running Linux, and compiling with gcc instead of clang.  Here are the results.

Original: 3.14
static: 2.06
thread_local static: 2.98
thread_local static, manual copy: 3.00

The ordering of the first three is the same, but the amount of speedup is different.  Eliminating the vector creation speeds it up by only 34% rather than the 52% speedup on the Mac.  And thread_local static is only a very minor 5% speedup rather than the 40% speedup on the Mac.  This time the manual copy is no faster than the standard vector assignment.

peastman:
I agree, `thread_local static` is an easy and worthwhile optimization.  You can do it everywhere a vector appears as a local variable inside one of the ForceInfo methods: `areParticlesIdentical()`, `getParticlesInGroup()`, and `areGroupsIdentical()`.

chryswoods:
I've added `thread_local_static` and removed the `memcpy`s I'd added. The changes are now just in `CommonKernels.cpp`.

With a non-pathological `System` (newly uploaded to `system.xml.bz2` in the above google drive), the `Context` takes only about 0.12s to create on my machine (Mac Studio M1 Max). For the pathological `System` I got timings similar to you (1.7s).

Thanks for your help :-)

peastman:
Let's make the change only in the ForceInfo methods.  You made similar changes in lots of other places that aren't performance sensitive and have no benefit.

platforms/common/src/CommonKernels.cpp
```diff
@@ -236,15 +236,16 @@ class CommonCalcCustomBondForceKernel::ForceInfo : public ComputeForceInfo {
...
         force.getBondParameters(index, particle1, particle2, parameters);
         particles.resize(2);
         particles[0] = particle1;
         particles[1] = particle2;
     }
     bool areGroupsIdentical(int group1, int group2) {
         int particle1, particle2;
-        vector<double> parameters1, parameters2;
+        thread_local static vector<double> parameters1;
+        thread_local static vector<double> parameters2;
```
peastman, line 248: You can keep it as one line: `thread_local static vector<double> parameters1, parameters2;`.

chryswoods:
I've moved things back to a single line and removed `thread_local static` from everything except the ForceInfo methods.

platforms/common/src/CommonKernels.cpp
```diff
@@ -1080,7 +1080,7 @@ void CommonCalcCMAPTorsionForceKernel::copyParametersToContext(ContextImpl& cont
 class CommonCalcCustomExternalForceKernel::ForceInfo : public ComputeForceInfo {
 public:
     ForceInfo(const CustomExternalForce& force, int numParticles) : force(force), indices(numParticles, -1) {
-        vector<double> params;
+        thread_local static vector<double> params;
```
peastman, line 1083: This is in a constructor.  It's only executed once.  There's no benefit to making it static.

chryswoods: Thanks for spotting this - removed in the latest commit

platforms/common/src/CommonKernels.cpp
```diff
@@ -2761,8 +2758,8 @@ void CommonCalcCustomGBForceKernel::initialize(const System& system, const Custo
         globals.initialize<float>(cc, force.getNumGlobalParameters(), "customGBGlobals");
     vector<vector<float> > paramVector(paddedNumParticles, vector<float>(numParams, 0));
     vector<vector<int> > exclusionList(numParticles);
+    vector<double> parameters;
```
peastman, line 2761: I think moving this out of the block will create a naming conflict.  Later in the same method (line 3438) we declare another variable called `parameters`.

chryswoods: I've changed this back (and all of the other changes to the constructors). The only changes now are the `thread_local statics` added to the ForceInfo methods.

peastman:
Looks good.  Thanks!

Issue #4262, opened 2023-10-11
Units of virtual site weights


mattwthompson:
The [critical term](http://docs.openmm.org/latest/userguide/theory/05_other_features.html?highlight=threeparticleaveragesite#virtual-sites) in `outOfPlaneSite` takes the cross product of two vectors $\vec{r_{12}}$ and $\vec{r_{13}}$, multiplies them through by a weighting factor, and adds that result (say $\vec{r_{cross}} = \vec{r_{12}} \times \vec{r_{13}}$) to a running sum that computes the position of the virtual site. If my understanding of vectors is correct[^1][^2] the units of  $\vec{r_{cross}}$ are distance squared. This would imply that $w_{cross}$ must have units of inverse distance for this term to safely be added to the equation it's defined in. The documentation doesn't necessarily state that the other weight factors are unitless, but I think it's fair for the user to assume this after looking at the simpler equations for the other virtual site types. The [API documentation](http://docs.openmm.org/latest/api-python/generated/openmm.openmm.OutOfPlaneSite.html?highlight=outofplanesite#outofplanesite) for the class doesn't mention that $w_{cross}$ might have units of inverse distance. The [TIP5P](https://github.com/openmm/openmm/blob/8.0.0/wrappers/python/openmm/app/data/tip5p.xml#L14-L15) shipped with OpenMM doesn't include units, but I guess that's to be expected since most (everything?) else is in implicit units.

I came across this while working on something else that benefited from me re-deriving TIP5P parameters in OpenMM's models. I was getting something off by a factor of 10 because the [page I was following](https://docs.lammps.org/Howto_tip5p.html) happened to use Angstroms. When everything internally and externally uses the same units (nanometers for TIP5P) I don't think this would be an issue. I can post my back-of-the-envelope math in its ugly detail if that's useful - I figure nobody here is that much of a masochist.

I realize I could be the only person in the world who cares about this ... I'm posting just as much as a sanity check as an offer to update the documentation to mention this.

[^1]: Mostly that a vector has, or acts like it has, position units. So you can have a particle at a point in space, add a vector to it, and the result is a new point in space. It makes sense to me that the vector 
[^2]: Also that taking the magnitude of a vector preserves its units in some way, or at least the magnitude of a distance vector has the same units (of distance).

peastman:
I believe your understanding is correct.  It would be worth adding a sentence to the documentation, something like, "Note that while w12 and w13 are unitless, wcross has units of inverse distance."

[...]

PR #4263, opened 2023-10-11
Note that w_cross has units of inverse distance

mattwthompson:
https://github.com/openmm/openmm/issues/4262#issuecomment-1758274887

This is the smallest change that would prevent future users from catching the tripwire I've found myself on. I didn't update the user guide or other classes since I think that's somewhat more obvious - it's really this one weight in this one class that's not intuitive - but I can do that easily if we think that's useful.

### DIFF
--- a/openmmapi/include/openmm/VirtualSite.h
+++ b/openmmapi/include/openmm/VirtualSite.h
@@ -136,7 +136,9 @@ private:
  * r<sub>1</sub> + w<sub>12</sub>r<sub>12</sub> + w<sub>13</sub>r<sub>13</sub> + w<sub>cross</sub>(r<sub>12</sub> x r<sub>13</sub>)
  * 
  * The three weight factors are user-specified.  This allows the virtual site location
- * to be out of the plane of the three particles.
+ * to be out of the plane of the three particles. Note that while
+ * w<sub>12</sub> and w<sub>13</sub> are unitless and w<sub>cross</sub> has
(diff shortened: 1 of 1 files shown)


mattwthompson:
I don't think this CI setup gives me a view into the built docs? I see they built successfully.

Not a complaint, just a nice-to-have that I take for granted with RTD.

openmmapi/include/openmm/VirtualSite.h
```diff
@@ -136,7 +136,9 @@ class OPENMM_EXPORT ThreeParticleAverageSite : public VirtualSite {
  * r<sub>1</sub> + w<sub>12</sub>r<sub>12</sub> + w<sub>13</sub>r<sub>13</sub> + w<sub>cross</sub>(r<sub>12</sub> x r<sub>13</sub>)
  * 
  * The three weight factors are user-specified.  This allows the virtual site location
- * to be out of the plane of the three particles.
+ * to be out of the plane of the three particles. Note that while
+ * w<sub>12</sub> and w<sub>13</sub> are unitless and w<sub>cross</sub> has
```
peastman, line 140: "are unitless, w<sub>cross</sub> has"

peastman:
You can test the docs by building locally: `make C++ApiDocs`.

peastman:
Looks good.  Thanks!

mattwthompson:
Thank you!

PR #4267, opened 2023-10-14
Make TestCudaATMForce an actual test by renaming to .cpp

ex-rzr:
See https://github.com/openmm/openmm/pull/4110

The OpenCL test is correct.


peastman:
Thanks!

Issue #3462, opened 2022-02-15
compareInt2LargeSIMD can't be used without cutoff


ex-rzr:
https://github.com/openmm/openmm/blob/master/platforms/opencl/src/OpenCLNonbondedUtilities.cpp#L230

The OpenCL platform uses a different order of tiles with exclusions on AMD GPUs with warps of 64 (GCN, CDNA). It groups diagonal tiles in `exclusionTiles` together to eliminate cases when one half of a wrap processes a diagonal tile and another precesses an off-diagonal tile, improving divergence in the first loop of kernels like nonbonded.

However, `exclusionTiles` is also used later in these kernels: if interactions don't use cutoff, the kernel needs to process all tiles except the tiles with exclusions processed in the first loop.

The code doing it depends on a correct order of tiles in `exclusionTiles`:
https://github.com/openmm/openmm/blob/master/platforms/opencl/src/kernels/nonbonded.cl#L245-L259

`compareInt2` really sorts all tiles so this code works correctly, but `compareInt2LargeSIMD` basically splits the list in two (diagonal and off-diaonal tiles) and sorts them independently.

So there are situations when some tiles are processed twice: as a tile with exclusions and as a tile without exclusions (without considering that some interactions within the tile must be excluded).

I added this piece of code to all files where such filtering with skipTiles is used:

```cpp
        // after includeTile = (skipTiles[currentSkipIndex] != pos);

        if (get_local_id(0) % 32 == 0) {
            bool found = false;
            for (int ii = 0; ii < NUM_TILES_WITH_EXCLUSIONS; ii++) {
                int2 tile = exclusionTiles[ii];
                int p = tile.x + tile.y*NUM_BLOCKS - tile.y*(tile.y+1)/2;
                if (pos == p) {
                    found = true;
                }
            }
            if (includeTile) {
                if (found) {
                    printf("!!! incorrectly included %d (%d, %d)\n", pos, x, y);
                }
            }
            else {
                if (!found) {
                    printf("!!! incorrectly excluded %d (%d, %d)\n", pos, x, y);
                }
            }
        }
```

There are 13 such files (.cc and .cl). I ran OpenCL tests on AMD GPU (ROCm OpenCL), "!!! incorrectly included" is reported for several tiles by 3 tests:
* TestOpenCLAmoebaGeneralizedKirkwoodForce
* TestOpenCLCustomNonbondedForce
* TestOpenCLNonbondedForce

and one benchmark: amoebagk.

[...]

PR #4268, opened 2023-10-14
Fix exclusion tiles sorting on AMD GCN/CDNA (64 threads per wave) in OpenCL

ex-rzr:
With changes introduced in https://github.com/openmm/openmm/pull/4070 the nonbonded kernel uses USE_NEIGHBOR_LIST (useNeighborList) so host code also must check it instead of useCutoff.

See also https://github.com/openmm/openmm/issues/3462

(I discovered this issue when I encountered test failures in OpenMM-HIP after merging https://github.com/amd/openmm-hip/pull/9)


peastman:
Thanks!

PR #4271, opened 2023-10-16
fix the time unit in XTC trajectory file

z-gong:
The recently added `XTCReporter` is very useful. It's space efficient. And it records time together with positions, making the calculation of time-related properties more convenient.

The time unit used by GROMACS is picosecond. `XTCFile` writes femtosecond into the file by accident.

### DIFF
--- a/wrappers/python/openmm/app/xtcfile.py
+++ b/wrappers/python/openmm/app/xtcfile.py
@@ -12,7 +12,7 @@ from openmm.app.internal.xtc_utils import (
 import numpy as np
 import os
 from openmm import Vec3
-from openmm.unit import nanometers, femtoseconds, is_quantity, norm
+from openmm.unit import nanometers, picoseconds, is_quantity, norm
 import math
 import tempfile
 import shutil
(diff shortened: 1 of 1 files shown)


peastman:
cc @RaulPPelaez

RaulPPelaez:
As far as I can tell the XTC format does not impose any units for time:
https://manual.gromacs.org/current/reference-manual/file-formats.html#xtc

However the GROMACS time units are indeed picoseconds:
https://manual.gromacs.org/2019/reference-manual/definitions.html

LAMMPS ran into the same realization time ago:
https://matsci.org/t/xtc-trajectory-time-step/22677/5
They also write in picosecond:
https://github.com/lammps/lammps/blob/0fe621886ef85d316475f871a241f8d7df77c744/src/EXTRA-DUMP/dump_xtc.cpp#L81-L86

Tests are passing either way because they create the XTCFile with a number, not a Quantity, in which case XTCFile just writes that number as time step. 

This looks good to me and it should be merged.
Thanks @z-gong!

peastman:
Thanks!

Issue #4112, opened 2023-06-14
Modeller.add ignores bond orders and types


zanebeckwith:
In Modeller.add ([modeller.py line 120](https://github.com/openmm/openmm/blob/add2bbeee88933f4ffe9066fb8a41fce561c8086/wrappers/python/openmm/app/modeller.py#L120)), the new bond ignores the old bond's `type` and `order`.

Would it be as easy as making this `newTopology.addBond(newAtoms[bond[0]], newAtoms[bond[1]], bond.type, bond.order)`?

Or is there something subtle I'm missing?

If it is that simple, I'm more than happy to submit a PR!

Thanks for the great project!

peastman:
A PR would be great!  Yes, it's as easy as that.  Probably we just missed that line when we added support for bond types and orders

zanebeckwith:
Ah, great!

> Probably we just missed that line when we added support for bond types and orders

I was hoping that was all it was!

I just opened PR #4113 with the little fix

Thanks!

PR #4323, opened 2023-11-22
Preserve bond type and order in all `Modeller` operations

nicoladmSB:
This is not linked to an open issue (as far as I can tell), but it falls in the same category as #4112 (issue), #4113 (PR), completing the work of the latter. 

----

When generating a new topology for a `Modeller` as a result of any operation, bonds that had `type` and `order` attributes in the original topology should retain them, instead of implicitly setting them to `None`.

This PR

- ensures that  `delete()` (therefore implicitly `deleteWater()`), `addSolvent()`, `addHydrogens()` and `addExtraParticles()` behave in the desirable way, while #4112 did it for `add()` (and implicitly for `addMembrane()`).

- adds regression tests either as new tests method, or as assertions in existing test methods

### DIFF
--- a/wrappers/python/openmm/app/modeller.py
+++ b/wrappers/python/openmm/app/modeller.py
@@ -178,7 +178,7 @@ class Modeller(object):
         for bond in self.topology.bonds():
             if bond[0] in newAtoms and bond[1] in newAtoms:
                 if bond not in deleteSet and (bond[1], bond[0]) not in deleteSet:
-                    newTopology.addBond(newAtoms[bond[0]], newAtoms[bond[1]])
+                    newTopology.addBond(newAtoms[bond[0]], newAtoms[bond[1]], bond.type, bond.order)
         self.topology = newTopology
         self.positions = newPositions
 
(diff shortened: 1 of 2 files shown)


peastman:
Thanks!  The change looks good.

I think the tests could be simplified.  Rather than modifying a bunch of existing tests, how about adding a single new test that runs through a series of manipulations?

- Create a Topology with a small molecule that includes bond types.
- Call `addSolvent()` to add some water.
- Call `convertWater()` to convert it to TIP4P.
- Call `addExtraParticles()` to add the virtual sites.
- Call `delete()` to remove some of the water.
- Verify that the bond information for the small molecule has survived all those calls intact.

nicoladmSB:
Thanks for the review!

I've reverted the changes to the test module, and replaced them with a single test that does more or less what you recommended. I've just added a couple of steps to it, to ensure that `add()`, `addHydrogens()` and `deleteWater()` are also exercised.

peastman:
Looks good.  That's a really nice test case.

PR #4324, opened 2023-11-23
Add support for non-(amd64,Arm,PPC) using SIMD Everywhere

mr-c:
Hello, this enables building and running openmm on riscv64 and other non-(amd64,arm,ppc) systems.

It uses the portable (and sometimes optimized) implementations of x86 intrinsics from https://github.com/simd-everywhere/simde

Once riscv64 hardware with support for version 1.0 of their vector extension becomes more widely available, then a proper riscv64 implementation could be written.

If you want to bundle the SIMDe library instead of making it an external dependency, then I'm happy to do so. Just let me know how you would like it: 1) single-file; 2) folder copied in; or 3) git sub module.

### DIFF
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -41,6 +41,7 @@ jobs:
       # Note: conda activate doesn't work here, because it creates a new shell!
       - script: |
           conda install cmake ^
+                        simde ^
                         cython ^
                         ninja ^
                         numpy ^
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ addons:
       - python-numpy
       - python-scipy
       - libfftw3-dev
+      - libsimde-dev
 
 env:
   global:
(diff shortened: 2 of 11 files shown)


peastman:
That's really cool that you have it working on RISC-V.  How is the performance?

We don't need a separate library for SIMD.  We have a version of the vectorization routines that uses clang/gcc portable vectors, so it works on any hardware they support.  I even have a few improvements to it I've been half meaning to check in, but haven't bothered since we aren't currently using it anywhere.  But I can check them in if you want to try it out.

mr-c:
Oh, for Debian we had build failures on every arch except amd64, arm64, ppc64el: https://buildd.debian.org/status/package.php?p=openmm

I don't have performance numbers, sorry. 

Using my patch, I was able to build a Debian package of 8.0.0 (and the tests  passed) on Debian's [Hifive Unmatched](https://db.debian.org/machines.cgi?host=debian-riscv64-porterbox-01) porterbox.

peastman:
I'm going to close this, since #4328 now provides support for other architectures.  If you want to try it out, I'd be interested to hear how it works on RISC-V.

PR #4330, opened 2023-11-29
Optimizations to reporters

peastman:
It turned out that the reporters for writing trajectories in OpenMM had much higher overhead than was really necessary.  See https://github.com/mdtraj/mdtraj/issues/1840, which found that the new XTCReporter is many times slower than the corresponding one in MDTraj.

The first major source of overhead was that they retrieved the positions from the State by calling `getPositions()`, which involves constructing a Vec3 object for every particle, rather than `getPositions(asNumpy=True)`, which just requires a simple memcpy.  It was especially bad for XTCReporter, which then internally coverts it to a Numpy array anyway.

To test this, I ran the simulatePdb.py example, which simulates a fairly small system, while having it write a frame every 10 steps in a few different formats.  That's much more often than people would typically want, of course, but it's useful to emphasize the cost of the reporter.  Here are the total times in seconds for 10,000 steps.

No reporter: 13.6
PDB: 68.5
DCD: 41.7
XTC: 50.7

Here are the times after adding `asNumpy=True`.

PDB: 62.9
DCD: 39.2
XTC: 31.7

It's only a small speedup for PDB and DCD, but a much larger speedup for XTC, presumably because it avoids the extra conversion.

The next bottleneck came from these lines:

https://github.com/openmm/openmm/blob/127a373381a35ca6abc7a4966795bb7b4a87ae37/wrappers/python/openmm/app/xtcfile.py#L95-L102

That's a very inefficient way of doing the error check.  I replaced them with calls to `np.isnan(positions)` and `np.isinf(positions)`.  Here are the times after doing that.

PDB: 43.1
DCD: 19.3
XTC: 22.5

The result is a moderate speedup to PDB, and quite large speedups to DCD and XTC.

### DIFF
--- a/wrappers/python/openmm/app/dcdfile.py
+++ b/wrappers/python/openmm/app/dcdfile.py
@@ -121,9 +121,10 @@ class DCDFile(object):
             raise ValueError('The number of positions must match the number of atoms')
         if is_quantity(positions):
             positions = positions.value_in_unit(nanometers)
-        if any(math.isnan(norm(pos)) for pos in positions):
+        import numpy as np
+        if np.isnan(positions).any():
             raise ValueError('Particle position is NaN.  For more information, see https://github.com/openmm/openmm/wiki/Frequently-Asked-Questions#nan')
-        if any(math.isinf(norm(pos)) for pos in positions):
+        if np.isinf(positions).any():
             raise ValueError('Particle position is infinite.  For more information, see https://github.com/openmm/openmm/wiki/Frequently-Asked-Questions#nan')
         file = self._file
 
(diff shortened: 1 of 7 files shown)


sukritsingh:
Cross posting here from the mdtraj/mdtraj#1840 thread, but this does appear to create a massive speedup for the XTC reporter compared to before. Results summarized below for a ~60K atom system with 4fs timesteps and `LangevinMiddleIntegrator`. 

Summary table with all the results for the `XTCReporter` (I didn't test the DCDReporter as this issue arose due to comparison with MDTraj)

| XTC Reporter used | Save frequency (steps) | Speed (ns/day) |
|----------------|------------------------|------------------------|
| OpenMM 8.1 (original)           | 10                    | 66.4           |
| OpenMM 8.1 (original)           | 500                    | 584            |
| OpenMM 8.1  (with just `asNumpy=True`)          | 10                     | 90.6           |
| OpenMM 8.1  (with just `asNumpy=True`)          | 500                   | 625            |
| MDTraj            | 10                    | 596            |
| MDTraj            | 500                    | 670            |
| OpenMM 8.1 (this PR)          | 10                     |567          |
| OpenMM 8.1 (this PR)            | 500                    | 682            |

Looks like a major speedup across the board!

peastman:
Thanks!

PR #4340, opened 2023-12-05
Fix for tabulated functions with CustomCentroidBondForce

### LINKED ISSUE #4339: Segmentation faults when using tabulated functions with CustomCentroidBondForce
CustomCentroidBondForce can cause segmentation faults when tabulated functions are used.  A minimal example is

```
import openmm
force = openmm.CustomCentroidBondForce(2, "0")
force.addTabulatedFunction("f", openmm.Continuous1DFunction((0, 1), 0, 1))
print(force.getTabulatedFunction(0))
```

which causes a segmentation fault, whereas this same example having `CustomCentroidBondForce` replaced by `CustomCompoundBondForce` produces the expected output

```
<openmm.openmm.Continuous1DFunction; proxy of <Swig Object of type 'OpenMM::Continuous1DFunction *' at 0x[...]> >
```

The crash occurs within the call to `getTabulatedFunction()`.  I believe that the `Continuous1DFunction` might be getting incorrectly destroyed before then when its Python object gets garbage collected because `CustomCentroidBondForce.addTabulatedFunction()` in the `openmm.py` generated at build time fails to set `function.thisown=0` as other `addTabulatedFunction()` methods do.

It appears that this is caused by a missing entry from the `STEAL_OWNERSHIP` table in `wrappers/python/src/swig_doxygen/swigInputConfig.py`.  In particular, adding

```
("CustomCentroidBondForce", "addTabulatedFunction") : [1],
```

to this table and recompiling OpenMM resolves this problem.

If this makes sense and seems good, I can submit a pull request.

epretti:
Should fix #4339.  Adds `addTabulatedFunction` for `CustomCentroidBondForce` to `STEAL_OWNERSHIP` table in `swigInputConfig.py`.


peastman:
Looks good, thanks!

PR #4343, opened 2023-12-06
Improved sorting of blocks when building neighbor list

### LINKED ISSUE #4334: Performance regression on some models
In #4147 we added an optimization to reduce the O(n^2) cost of building the neighbor list for very large systems.  Unfortunately, it was incompatible with another optimization we were already doing: sorting atom blocks by size to produce a more compact neighbor list.  We have to choose between one optimization and the other, one reducing the cost of building the neighbor list and one reducing the cost of computing interactions.  Since one scales as O(n^2) and the other as O(n), which one is more important depends on the size of the system.  We ran a bunch of tests and selected 90,000 atoms as the cutoff for selecting between the two.

It turns out this is very sensitive to the details of the system.  Even for some systems much larger than that, you're much better off sorting the blocks rather than trying to optimize building the neighbor list.  A common situation is when you have a mix of water and ions.  The blocks containing the ions are much larger than the ones containing the water (many fewer atoms scattered over the same volume), so it's very important for them to be sorted to the end.  When OpenMM builds water boxes it puts the ions at the end, right where we want them, so additional sorting isn't needed.  But when Amber builds water boxes, it puts the ions before the water, not after.

To show the impact of this, I ran benchmarks of water boxes of varying sizes using the script from https://github.com/openmm/openmm/issues/4063#issuecomment-1536768880, only specifying `ionicStrength=0.1*molar`.  I then did the same, only reversing the order of the water and ions:

```python
modeller = Modeller(Topology(), [])
modeller.addSolvent(ff, boxSize=Vec3(width, width, width)*nanometers, ionicStrength=0.1*molar)
m2 = Modeller(modeller.topology, modeller.positions)
m2.delete([r for r in m2.topology.residues() if r.name != 'HOH'])
modeller.deleteWater()
modeller.add(m2.topology, m2.positions)
```

Here is the time per 100,000 atoms to run 5000 steps.  Lower is better.

|Box Width|Atoms|w/o large blocks|Large blocks|Large blocks, salt reversed|
|---|---|---|---|---|
5|12227|7.44|7.35|12.15
10|98644|2.63|2.61|4.68
15|334217|2.67|2.60|3.31
20|790542|3.06|2.70|2.95
25|1546428|4.57|3.79|3.56

The results for the standard water box are very similar to what we found in #4147: the large blocks optimization improves performance in all cases, with the amount of improvement growing as the system grows.  But reversing the order to put the ions before the water dramatically changes it.  The smaller systems get much slower.  Even at 334,000 atoms, using the large blocks optimization slows things down significantly.  You have to get up to around 700,000 atoms before it becomes faster than omitting the large blocks optimization and instead sorting blocks by size.

peastman:
Fixes #4334.  The large blocks optimization added in #4147 creates a dilemma about how to sort atom blocks.  We want them to be sorted by size, because that produces the most compact neighbor list and makes computing interactions faster.  But we also want them sorted spatially, because otherwise the large blocks optimization isn't effective.  We compromised by disabling the large blocks optimization for small systems.  We sort spatially when using it, and by size when not using it.

This turned out to hurt performance in some cases, as described in #4334.  This PR uses a different compromise: it splits the difference between the two sorting methods.  It divides blocks by size into a small number of bins, while keeping the blocks within each bin spatially sorted.  That gives most of the benefit of each one.

Here are some benchmarks for simulating water boxes with 0.1 molar salt.  I show the speed for the existing code with the large blocks optimization disabled or enabled, and for the version in this PR.  The numbers are the time per 100,000 atoms to take 5000 steps, so smaller is better.

|Box Width|Atoms|w/o large blocks|Large blocks|This PR|
|---|---|---|---|---|
5|12227|7.93|7.96|7.43
10|98644|2.65|2.68|2.68
15|334217|2.71|2.72|2.60
20|790542|3.07|3.15|2.72
25|1546428|4.59|4.76|3.80

The new version is faster in almost all cases, and for the largest systems it's quite a bit faster.  Now here is the same thing, only reversing the order of the atoms so the salt comes before the water rather than after.

|Box Width|Atoms|w/o large blocks|Large blocks|This PR|
|---|---|---|---|---|
5|12227|7.55|12.15|7.67
10|98644|2.66|4.68|2.66
15|334217|2.73|3.31|2.62
20|790542|3.09|2.95|2.73
25|1546428|4.64|3.56|3.78

Previously that had a big effect on performance, and hurt performance a lot for the smaller systems.  With this PR it no longer has a significant effect.

I also tested it on more typical systems.  I included all our standard benchmark systems as well as FactorIX from the Amber benchmarks.  We don't usually include that one due to problems with the file: see #3391.  But it turned out to have a big performance regression in 8.1.  I had to hack the benchmark script to get it to run without errors.  These numbers are speeds in ns/day, so higher is better.

|Test|w/o large blocks|Large blocks|This PR|
|---|---|---|---|
DHFR|1,941.7|1,643.4|1,926.6
ApoA1|581.2|581.9|578.2
FactorIX|653.6|481.3|659.1
Cellulose|142.9|148.4|147.3
STMV|39.8|43.6|46.7

It's now similar to or faster than 8.1 in all cases.  I still want to tune a few parameters to see if I can get it slightly faster, and then I'll port the changes over to OpenCL as well.

### DIFF
--- a/platforms/cuda/include/CudaNonbondedUtilities.h
+++ b/platforms/cuda/include/CudaNonbondedUtilities.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2009-2022 Stanford University and the Authors.      *
+ * Portions copyright (c) 2009-2023 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
@@ -338,6 +338,7 @@ private:
     CudaArray sortedBlocks;
     CudaArray sortedBlockCenter;
     CudaArray sortedBlockBoundingBox;
+    CudaArray blockSizeRange;
     CudaArray largeBlockCenter;
     CudaArray largeBlockBoundingBox;
     CudaArray oldPositions;
@@ -345,7 +346,7 @@ private:
     CudaSort* blockSorter;
     CUevent downloadCountEvent;
     unsigned int* pinnedCountBuffer;
-    std::vector<void*> forceArgs, findBlockBoundsArgs, sortBoxDataArgs, findInteractingBlocksArgs;
+    std::vector<void*> forceArgs, findBlockBoundsArgs, computeSortKeysArgs, sortBoxDataArgs, findInteractingBlocksArgs;
     std::vector<std::vector<int> > atomExclusions;
     std::vector<ParameterInfo> parameters;
     std::vector<ParameterInfo> arguments;
@@ -354,7 +355,7 @@ private:
     std::map<int, std::string> groupKernelSource;
     double lastCutoff;
     bool useCutoff, usePeriodic, anyExclusions, usePadding, useNeighborList, forceRebuildNeighborList, canUsePairList, useLargeBlocks;
-    int startTileIndex, startBlockIndex, numBlocks, maxExclusions, numForceThreadBlocks, forceThreadBlockSize, numAtoms, groupFlags;
+    int startTileIndex, startBlockIndex, numBlocks, maxExclusions, numForceThreadBlocks, forceThreadBlockSize, numAtoms, groupFlags, numBlockSizes;
     unsigned int maxTiles, maxSinglePairs, tilesAfterReorder;
     long long numTiles;
     std::string kernelSource;
@@ -371,6 +372,7 @@ public:
     std::string source;
     CUfunction forceKernel, energyKernel, forceEnergyKernel;
     CUfunction findBlockBoundsKernel;
+    CUfunction computeSortKeysKernel;
     CUfunction sortBoxDataKernel;
     CUfunction findInteractingBlocksKernel;
     CUfunction findInteractionsWithinBlocksKernel;
--- a/platforms/cuda/src/CudaNonbondedUtilities.cpp
+++ b/platforms/cuda/src/CudaNonbondedUtilities.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2009-2022 Stanford University and the Authors.      *
+ * Portions copyright (c) 2009-2023 Stanford University and the Authors.      *
  * Authors: Peter Eastman                                                     *
  * Contributors:                                                              *
  *                                                                            *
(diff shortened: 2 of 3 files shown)


peastman:
I copied the changes over to OpenCL.  Here are the speeds on RTX 4080.

|Test|8.1|This PR|
|---|---|---|
DHFR|1455.68|1454.31
ApoA1|471.653|467.47
FactorIX|495.984|507.72
Cellulose|113.765|115.686
STMV|27.899|39.2594

The change is small for all systems except STMV, which has a huge speedup.

Here are the speeds on M1 Pro.

|Test|8.1|This PR|
|---|---|---|
DHFR|156.535|155.514
ApoA1|38.426|38.5397
Cellulose|9.56219|9.5409
STMV|3.25286|3.40891

The pattern is similar although the changes are smaller overall.  STMV is the only one that has a measurable speedup, and even that is only about 5%.

bdenhollander:
I'm seeing 0.3-6.75% improvement with this change on the FAH projects I previously benchmarked. There is still a 15% regression for FAH project [17308](https://github.com/openmm/openmm/files/13621815/17308.zip) compared to OpenMM 8.0. I attached the system in case that helps to track down the source of the slowdown.

peastman:
I timed simulating your system for 5000 steps on a RTX 4080.  Here are the times in seconds:

8.0: 5.1
8.1: 7.6
This PR: 5.5

Is that about the same as what you're seeing?  Do you have a PDB file for it?  It's a bit hard to tell how it's composed with only the System to look at.

bdenhollander:
I accidentally missed a step when building and wasn't testing the correct branch. Here are ns/day on RX 6600:

| | ns/day |
|--------|--------|
| 8.0 | 49.1856 |
| 8.1 | 41.3582 |
| 8.1 w/out large blocks | 50.7176 |
| This PR | 53.348 |

The performance regression is remediated. Unfortunately, a FAH project that was previously stable is throwing consistent `Particle coordinate is NaNs` on this branch. Here's the [system](https://github.com/openmm/openmm/files/13622590/17424.zip) to see if you can reproduce it on any of your GPUs.

I reached out to the FAH researcher to see if there's a PDB available for 17308.

ex-rzr:
@peastman 
How did you choose `int numSizeBins = 20;`? By experiments?

What bothers me is that `sortedBlocks[i] = (((unsigned int) bin)<<24) + i;` makes very weird distribution of keys. Basically we have 20 groups of (almost) consecutive values with HUGE gaps in between (up to 2^24 long).
I wonder if `assignElementsToBuckets2` is able to handle this or most of buckets are empty and few have too many values to use the fast path with sorting in shared memory.
I guess I will check it because I need to port `assignElementsToBuckets2` to HIP as it's an essential optimization to make this change work fast in HIP.

peastman:
>  How did you choose int numSizeBins = 20;? By experiments?

Correct.  I've been running tests on it, and while no single value is optimal in all cases, 20 seems to be a pretty good compromise.

> What bothers me is that sortedBlocks[i] = (((unsigned int) bin)<<24) + i; makes very weird distribution of keys.

If it causes problems, we could try replacing 24 with the minimum shift needed given the actual number of atom blocks.

> I guess I will check it because I need to port assignElementsToBuckets2 to HIP

I suspect that will be important.  You can test to be sure, but as you say the distribution may be very nonuniform.

peastman:
I changed it to use the smallest shift possible.  That does seem to provide a small additional speedup, though usually only on the order of 1%.

peastman:
@bdenhollander I think I found the problem with the system blowing up you reported above.  It only happens for systems with triclinic boxes.  When a large block spans two bins, it can end up being very very large.  In that case, the distance check isn't sufficient, and some atom blocks can get skipped.  I want to do a few more tests to make sure I've completely fixed the problem and then I'll open a new PR with the change.

peastman:
It will be in 8.2.  No estimate yet on when that will be released, but I hope not too long.

PR #4345, opened 2023-12-07
Support `auth_seq_id` in `_struct_conn` PDBx section

### LINKED ISSUE #4341: Bond orders for HETATM-like entries should also serialize a `struct_conn.ptnr{1,2}_auth_seq_id`.
https://github.com/openmm/openmm/blob/921992685ec890cf0a6b91b844918dadee3e6946/wrappers/python/openmm/app/pdbxfile.py#L337-L344

See the discussion at https://github.com/molstar/molstar/pull/999#issuecomment-1843020131. Apparently, `label_seq_id` for HETATM like entities is poorly deifned.

pechersky:
Both in reading and writing.
For reading, we get residue numbers preferentially from
`auth_seq_id` in the `_atom_site` section, so we should use
the same in `_struct_conn` if it is available.
For writing, we duplicate the residue number in `_struct_conn`
the same way we do so in `_atom_site`.

Closes #4341.

### DIFF
--- a/wrappers/python/openmm/app/pdbxfile.py
+++ b/wrappers/python/openmm/app/pdbxfile.py
@@ -92,6 +92,7 @@ class PDBxFile(object):
         resNameCol = atomData.getAttributeIndex('auth_comp_id')
         if resNameCol == -1:
             resNameCol = atomData.getAttributeIndex('label_comp_id')
+        # number the residues preferrably by `auth_seq_id`, fallback to `label_seq_id`
         resNumCol = atomData.getAttributeIndex('auth_seq_id')
         if resNumCol == -1:
             resNumCol = atomData.getAttributeIndex('label_seq_id')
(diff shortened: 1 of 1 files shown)


peastman:
I don't think the changes to reading files are a good idea.  The `label` fields are the recommended ones to use for references between fields.  The spec sets no requirements for the values in the `auth` fields, not even that they be unique.  The `label` fields are required to be present and to be sequential positive integers.  In case of an inconsistency between the two sets of fields, those are the ones that should take precedence.

I really think this is just coming from a misunderstanding of the spec by the molstar developers.

pechersky:
I can remove the "reading" change, but currently, the existing implementation on master gets the `resNumCol` preferentially from `auth_seq_id` as it is. Why is it not preferrable to be consistent in which field is relied on in the `_struct_conn` as well, when reading? I am asking earnestly to try to better understand.

PR #4351, opened 2023-12-12
Fixed bug in large blocks optimization with triclinic boxes

peastman:
See https://github.com/openmm/openmm/pull/4343#issuecomment-1848221093 and https://github.com/openmm/openmm/pull/4343#issuecomment-1851122718.  The fix is the section marked off with `#ifdef TRICLINIC`.  I also made some minor changes to `sortBoxData()` while I was at it, merging two loops together.  That wasn't to fix anything, just to make the code slightly simpler and perhaps slightly faster.

@bdenhollander can you confirm this fixes your problem?

### DIFF
--- a/platforms/cuda/src/kernels/findInteractingBlocks.cu
+++ b/platforms/cuda/src/kernels/findInteractingBlocks.cu
@@ -146,19 +146,17 @@ extern "C" __global__ void sortBoxData(const unsigned int* __restrict__ sortedBl
         unsigned int index = sortedBlocks[i] & BLOCK_INDEX_MASK;
         sortedBlockCenter[i] = blockCenter[index];
         sortedBlockBoundingBox[i] = half3(trimTo3(blockBoundingBox[index]));
-    }
+
 #ifdef USE_LARGE_BLOCKS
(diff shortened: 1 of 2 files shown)


bdenhollander:
5 runs of 50000 steps on the project in the linked comment and another system I found later that was even less stable ran to completion without any NaNs. This fix does the trick.

Would it be possible to cherry pick some of the recent fixes and make an 8.1.1 release?

peastman:
Did you experience this problem with 8.1, or was it just with the changes from #4343?  This PR is just a minor change that would be easy to backport if needed.  That one was a much larger change that I wouldn't feel safe putting in a patch release.

bdenhollander:
I went back to 8.1 to retest and 17424 was stable there. It became unstable with #4343.

The other system I found later, 12271, is unstable in 8.1 when `useLargeBlocks` is hard-coded to true. It flew under the radar since it only has 42065 atoms.

peastman:
In that case, we should probably plan to do a 8.1.1 patch and include this fix.  Let's wait a little longer to see if anything else comes up that should be included.

Issue #4296, opened 2023-11-03
Custom virtual sites


selimsami:
Hi,

Is there a way to add custom virtual sites, just like you add custom forces? If not, could you point out which parts of the code I would need to change to add my custom virtual site?

To give you an example, I would like the virtual site placement to depend on the r_ij bond vector times the deviation from the equilibrium bond lengths (r_ij-r0) on top of the common virtual site placement rules like weighted center of geometry.

(sounds a little crazy, I know)

Thanks.
Selim

peastman:
The most general virtual site type is [LocalCoordinatesSite](http://docs.openmm.org/latest/api-c++/generated/LocalCoordinatesSite.html).  It doesn't let you specify arbitrary functions, but it does have a fairly flexible form.  A fully custom virtual site would be challenging to implement, since the calculation has to be invertible (to propagate forces on the virtual site back to the original particles).

selimsami:
Right, thought so. LocalCoordinatesSite unfortunately won't work for what I have in mind.
Could you perhaps point me at the right direction for implementing my own virtual site placement & force propagation?
Thanks

peastman:
That could be challenging depending on what you have in mind, but you're certainly welcome to try!  The first step is to add the API for it in VirtualSite.h and VirtualSite.cpp.  Next you need to write the calculations for it.  Start with ReferenceVirtualSites.cpp.  That's the implementation used by the Reference and CPU platforms.  Add another block in `computePositions()` that computes the position for your new virtual site class, and another block in `distributeForces()` that propagates the forces from it.

You'll then need to implement it for the CUDA and OpenCL platforms, which is a little more complicated, but start with the reference version.  If you can get that working, the other platforms will be easy.

selimsami:
That's very clear, thank you. I will give it a try and update here.

peastman:
The CPU platform always uses mixed precision.  If that's good enough for your purposes, you can start testing it.

To implement the CUDA and OpenCL versions (they're both implemented with the same code), there are two parts.  First, there's C++ code that records the parameters for all the virtual sites and uploads them to the GPU.  It starts at [this line](https://github.com/openmm/openmm/blob/796ffaaa97decdf1593ed75afa5fbb21b6b32e9c/platforms/common/src/IntegrationUtilities.cpp#L422).  You should be able to add the code for your virtual site class by copying what the other classes do.

[...]

PR #4356, opened 2023-12-14
setVelocitiesToTemperature() works with massless Drude particles

peastman:
If a Drude particle had zero mass, `setVelocitiesToTemperature()` would set the velocity of its parent particle to zero.  This was an issue when using DrudeSCFIntegrator.

### DIFF
--- a/plugins/drude/openmmapi/src/DrudeHelpers.cpp
+++ b/plugins/drude/openmmapi/src/DrudeHelpers.cpp
@@ -6,7 +6,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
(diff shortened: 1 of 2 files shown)


peastman:
@selimsami posted at https://github.com/openmm/openmm/issues/4296#issuecomment-1856988708:

> No longer getting all zeros but I get a temperature of 600 K now when I set it to 300K.

Can you show exactly what you're doing?  The test case in TestDrudeSCFIntegrator.h now tests `setVelocitiesToTemperature()` with both zero and nonzero Drude mass, and it comes out correct in both cases.

selimsami:
My bad - it works fine.
What got me was that the reporters never seems to print step 0.
With a small enough timestep, the temperature is indeed 300K. 
It then jumps to 600, then gradually back to 300. But that says more about my system!

peastman:
Thanks!

PR #4362, opened 2023-12-18
[WIP] Implementation of vectorangle function

cabb99:
Implements the vectorangle function to use Custom Forces that depend on the angle of vectors defined by four particles.

### DIFF
--- a/openmmapi/include/openmm/CustomCentroidBondForce.h
+++ b/openmmapi/include/openmm/CustomCentroidBondForce.h
@@ -112,7 +112,7 @@ namespace OpenMM {
  * select(x,y,z) = z if x = 0, y otherwise.
  * 
  * This class also supports the functions pointdistance(x1, y1, z1, x2, y2, z2),
- * pointangle(x1, y1, z1, x2, y2, z2, x3, y3, z3), and pointdihedral(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4).
+ * pointangle(x1, y1, z1, x2, y2, z2, x3, y3, z3), pointvectorangle(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4), and pointdihedral(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4).
  * These functions are similar to distance(), angle(), and dihedral(), but the arguments are the
  * coordinates of points to perform the calculation based on rather than the names of groups.
  * This enables more flexible geometric calculations.  For example, the following computes the distance
--- a/openmmapi/include/openmm/CustomCompoundBondForce.h
+++ b/openmmapi/include/openmm/CustomCompoundBondForce.h
@@ -106,7 +106,7 @@ namespace OpenMM {
  * select(x,y,z) = z if x = 0, y otherwise.
  * 
  * This class also supports the functions pointdistance(x1, y1, z1, x2, y2, z2),
- * pointangle(x1, y1, z1, x2, y2, z2, x3, y3, z3), and pointdihedral(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4).
+ * pointangle(x1, y1, z1, x2, y2, z2, x3, y3, z3),  pointvectorangle(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4), and pointdihedral(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4).
  * These functions are similar to distance(), angle(), and dihedral(), but the arguments are the
  * coordinates of points to perform the calculation based on rather than the names of particles.
  * This enables more flexible geometric calculations.  For example, the following computes the distance
--- a/openmmapi/include/openmm/CustomManyParticleForce.h
+++ b/openmmapi/include/openmm/CustomManyParticleForce.h
@@ -170,7 +170,7 @@ namespace OpenMM {
  * As seen above, the expression may also involve intermediate quantities that are defined following the main expression, using ";" as a separator.
  * 
  * This class also supports the functions pointdistance(x1, y1, z1, x2, y2, z2),
- * pointangle(x1, y1, z1, x2, y2, z2, x3, y3, z3), and pointdihedral(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4).
+ * pointangle(x1, y1, z1, x2, y2, z2, x3, y3, z3), pointvectorangle(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4), and pointdihedral(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4).
  * These functions are similar to distance(), angle(), and dihedral(), but the arguments are the
  * coordinates of points to perform the calculation based on rather than the names of particles.
  * This enables more flexible geometric calculations.  For example, the following computes the distance
--- a/openmmapi/include/openmm/internal/CustomHbondForceImpl.h
+++ b/openmmapi/include/openmm/internal/CustomHbondForceImpl.h
@@ -69,22 +69,24 @@ public:
      * in it, and replaces them with variables.  The particle indices returned in the maps are defined
      * as follows: 0=a1, 1=a2, 2=a3, 3=d1, 4=d2, 5=d3.
      *
-     * @param force     the CustomHbondForce to process
-     * @param functions definitions of custom function that may appear in the expression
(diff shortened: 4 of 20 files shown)


peastman:
Very impressive!  Let me know when it's ready for a full review.

I have a few questions in the mean time.  It looks like you've actually implemented two different functions, with some forces supporting only one and some supporting both.  `vectorangle(a, b, c, d)` computes the angle between `a-b` and `c-d`, where a, b, c, and d are four specified particles.  `pointvectorangle(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4)` is similar but with four arbitrary points.  I can see why you designed the latter one that way, since it parallels the `pointdistance()`, `pointangle()`, etc. functions.  But in practice that probably isn't the most convenient signature.  Would it be better to just take six arguments `x1, y1, z1, x2, y2, z2` specifying the two vectors to compute the angle between?  We would probably need a different name for it then.

cabb99:
I added another function to compute the angle between two vectors comparable to pointvectorangle called arrayvectorangle that only depends on 6 parameters as you suggested. However, I think it can only be used to replace pointvectorangle for periodic systems. Unless there is a function to calculate the periodic distance between the coordinates x1 and x2. Is there a function like that implemented?

Also, while writing the tests, I noticed that the angle function is a particular case of vector angle. The following forces are equivalent:
    CustomCentroidBondForce* angle = new CustomCentroidBondForce(3, "angle(g1,g2,g3)");
    CustomCentroidBondForce* vectorangle = new CustomCentroidBondForce(4, "vectorangle(g1,g2,g3,g2)");
    CustomCentroidBondForce* pointvectorangle = new CustomCentroidBondForce(4, "pointvectorangle(x1,y1,z1,x2,y2,z2,x3,y3,z3,x2,y2,z2)");
    CustomCentroidBondForce* arrayvectorangle = new CustomCentroidBondForce(4, "arrayvectorangle(x2-x1,y2-y1,z2-z1,x2-x3,y2-y3,z2-z3)");
Would it be a good idea to overload the angle function to accept 3 or 4 particles? It could be overloaded so that angle(p1,p2,p3) is automatically defined under the hood as vectorangle(p1,p2,p3,p2). Similarly writing angle(p1,p2,p3,p4) would call vectorangle(p1,p2,p3,p4).

cabb99:
@peastman This PR is ready for review.

peastman:
Great, thanks!

I suggest we simplify the set of functions.  There's no need for both `pointvectorangle()` and `arrayvectorangle()`.  The following are exactly equivalent to each other:

```
pointvectorangle(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4)
arrayvectorangle(x2-x1, y2-y1, z2-z1, x4-x3, y4-y3, z4-z3)
```

The `pointvectorangle()` function doesn't add anything.  It's no more general, no more concise, and it's slightly ambiguous which makes it more confusing (is it `x1-x2` or `x2-x1`)?  So I suggest removing it.

The third function, which is also exactly equivalent to the two above, is

```
vectorangle(g1, g2, g3, g4)
```

This has the benefit of being a little more concise, but otherwise doesn't add anything.  And really, a function called `vectorangle()` *should* take vectors as its arguments, not points.  Anything else is potentially confusing.  So I suggest having only the single function that takes two vectors (six arguments), and calling it just `vectorangle()`.

Does that sound reasonable?

cabb99:
@peastman, thank you for your response. I agree with your suggestion to simplify the functions and make the names clearer. I'll eliminate the pointvectorangle function and rename arrayvectorangle to simply vectorangle.

However, I'd prefer to keep the vectorangle(g1, g2, g3, g4) function, as it appears to be the most straightforward approach for practical use. I would also prefer to use it in that way. I recognize the potential confusion with two "vectorangle" functions, though. 

What do you think about modifying the existing angle function to accommodate this? For instance, we could have angle(g1, g2, g3) and an overloaded version, angle(g1, g2, g3, g4). This change seems logical, especially since the current angle(g1, g2, g3) is effectively equivalent to vectorangle(g1, g2, g3, g2). Do you think this would be a good solution?

peastman:
The difficulty is that the machinery for custom expressions doesn't support functions with variable numbers of arguments.  A CustomFunction needs to implement `getNumArguments()` to return the exact number of arguments that are expected.

peastman:
It would be nice to get this finished up and merged.  @cabb99 what do you think about the changes suggested above?

cabb99:
Sorry for the delay in answering this thread. I encountered some challenges implementing the `vectorangle(x2-x1, y2-y1, z2-z1, x4-x3, y4-y3, z4-z3)` (currently `arrayvectorangle`) under periodic boundary conditions.

Currently this commit has implementations of vector angle for `CustomHBondForce`, `CustomCentroidBondForce`, `CustomManyParticleForce` and `CustomCompoundBondForce`. While `CustomExternalForce` has an implementation of `periodicdistance(x1, y1, z1, x2, y2, z2)`, the forces where I implemented `vectorangle` don't include that implementation. Implementing `periodicdistance` in a similar manner for these forces would also make defining a simple vector angle not straightforward.

```
vectorangle(v1_x, v1_y, v1_z, v2_x, v2_y, v2_z);
v1x = periodicdistance(x2, 0, 0, x1, 0, 0);
v1y = periodicdistance(0, y2, 0, 0, y1, 0);
v1z = periodicdistance(0, 0, z2, 0, 0, z1);
v2x = periodicdistance(x3, 0, 0, x4, 0, 0);
v2y = periodicdistance(0, y3, 0, 0, y4, 0);
v2z = periodicdistance(0, 0, z3, 0, 0, z4);
```

Additionally, for the case of `CustomHBond` I would also have to expose the coordinates of the donor and acceptors, since only the particles d1, d2, d3, a1, a2 and a3 are exposed. 

I think the `vectorangle(p1,p2,p3,p4)` function is necessary for a simple implementation. Would you consider maybe renaming instead `vectorangle(x2-x1, y2-y1, z2-z1, x4-x3, y4-y3, z4-z3)` (currently `arrayvectorangle`)? or maybe you can give me other suggestions on how to implement `vectorangle(x2-x1, y2-y1, z2-z1, x4-x3, y4-y3, z4-z3)` under periodic boundary conditions.

peastman:
Periodic boundary conditions are a pain!

> v1x = periodicdistance(x2, 0, 0, x1, 0, 0);
> v1y = periodicdistance(0, y2, 0, 0, y1, 0);
> v1z = periodicdistance(0, 0, z2, 0, 0, z1);

That will only work correctly for rectangular boxes.  In systems with triclinic boxes, you need to apply periodic boundary conditions to the entire displacement vector as a whole, not to its components individually.

For CustomCompoundBondForce and CustomCentroidBondForce, periodic boundary conditions aren't an issue.  All the particles are guaranteed to be in the same periodic copy, so you can just take the difference.

CustomHbondForce and CustomManyParticleForce are nonbonded forces, so the interacting particles might have been wrapped differently.  In that case, you do need to apply periodic boundary conditions to the displacement vector.  That's easy to do as long as the distance between them is less than half the smallest side of the periodic box.  If it's larger than that, finding the nearest periodic copy is hard.  You have to actually loop over all 27 neighboring boxes and check each one.  That's why we require the cutoff distance to be less than half the box width.  Since both of these forces tend to have very short cutoffs, that shouldn't be a problem.

Of course, there's another issue here: the vectors aren't guaranteed to be displacements between particles.  For example, you might want a force that aligns a molecule to a fixed direction in space.  So ideally you want to be able to specify separately for each of the two vectors whether or not periodic boundary conditions should be applied to it.  But that's probably not worth worrying about.  The case I just described would be handled with a CustomCompoundBondForce, which never needs periodic boundary conditions applied.  It's hard to think of a case where you would want a CustomManyParticleForce or CustomHbondForce to depend on a fixed direction in space.

cabb99:
@peastman, I renamed the functions to:
```
angle4(g1,g2,g3,g4)
vectorangle(x2-x1, y2-y1, z2-z1, x4-x3, y4-y3, z4-z3)
pointvectorangle(x1, y1, z1, x2, y2, z2, x3, y3, z3, x4, y4, z4)
```
I don't know how to merge the functions into a single `vectorangle` function without doing major changes to the code.
`angle4` is needed for the `CustomHBondForce` implementation, since `CustomHBondForce` doesn't expose the particle coordinates.  `pointvectorangle` is required for the implementation of `angle4` in `CustomCentroidBondForce`, `CustomManyParticleForce` and `CustomCompoundBondForce` since there is a transformation under the hood of `angle4` to `pointvectorangle`. The periodic boundary conditions on this implementation should work as they are based on the `angle` and `pointangle` functions.

PR #4366, opened 2023-12-20
Use Intel OpenCL for CI

peastman:
To test the OpenCL platform on CI, we use AMD's CPU based OpenCL implementation.  This involves using a very old SDK that isn't available from AMD anymore.  Their modern frameworks don't support running OpenCL on CPUs.

I'd like to switch to something still actively supported.  This is an experiment to see if I can get it working with Intel's OpenCL.


peastman:
Thanks!  I've been working on this and I have fixes for almost all the failures.  There's just one more I'm working on.  I hope to have an update in later today.

yafshar:
@peastman, did you finish your fixes? I would be happy to help if you need a new pair of eyes or extra help.

peastman:
I was able to fix almost everything.  The one remaining problem is the failure in TestCustomHbondForce.  I tracked it down to a clear bug in the OpenCL compiler/runtime.  It happens when constructing the index into a tabulated function.  At this line

https://github.com/openmm/openmm/blob/162b7c37eda7a4faaa47264688683608fe449b3a/platforms/common/src/ExpressionUtilities.cpp#L517

if I change the type to `volatile int` the error goes away.  Marking variables as volatile tends to disable compiler optimizations, and in this case it seems to prevent the compiler from incorrectly optimizing.  Unfortunately, that hurts performance on other platforms, so it isn't an acceptable workaround.

yafshar:
I will check this issue. I found a similar issue at https://github.com/openmm/openmm/blob/162b7c37eda7a4faaa47264688683608fe449b3a/platforms/common/src/kernels/pme.cc#L61 where I used volatile to avoid wrong values in the data array.

bdenhollander:
Does adding braces to wrap the `if` statement make any difference? If it helps then I would recommend the same for lines 534-535.

```C++
out << "if (index >= 0 && index < xsize*ysize) {\n";
out << nodeNames[j] << suffix << " = " << functionNames[i].second << "[index];\n";
out << "}";
```

peastman:
It still gets the same error.  Braces are handled by the compiler front end, but most optimizations are done by the back end.  All record of whether there were braces has probably been forgotten by the time it gets to the part that does the incorrect optimization.

bdenhollander:
What about hoisting the variables outside of the `for` loop so that they're not re-declared when `nodes.size() > 1`? Re-declaration in the same scope normally works but most compilers will throw a warning.

```C++
out << "int x, y, xsize, ysize, index;\n";
for (int j = 0; j < nodes.size(); j++) {
    const vector<int>& derivOrder = dynamic_cast<const Operation::Custom*>(&nodes[j]->getOperation())->getDerivOrder();
    if (derivOrder[0] == 0 && derivOrder[1] == 0) {
        out << "x = (int) floor(" << getTempName(node.getChildren()[0], temps) << suffix << "+0.5f);\n";
        out << "y = (int) floor(" << getTempName(node.getChildren()[1], temps) << suffix << "+0.5f);\n";
        out << "xsize = (int) " << paramsInt[0] << ";\n";
        out << "ysize = (int) " << paramsInt[1] << ";\n";
        out << "index = x+y*xsize;\n";
        out << "if (index >= 0 && index < xsize*ysize) {\n";
        out << nodeNames[j] << suffix << " = " << functionNames[i].second << "[index];\n";
    }
}
```

peastman:
Good idea.  I tried adding braces around the block so they would be clearly seen as separate variables.  It didn't help though.

peastman:
I just tried simulating ApoA1 with CHARMM36.  That force field is sensitive to this, since it uses a CustomNonbondedForce with a Discrete2DFunction to implement Lennard-Jones interactions.  Running on a RTX 4080, adding the `volatile` keyword there decreases the speed from 536 to 525 ns/day.

peastman:
@yafshar just wondering if you ever made any progress on this?

yafshar:
@peastman, yes the compiler team found the issue and the solution will be available soon. I will update here

peastman:
Thanks!

yafshar:
@peastman the issue has been fixed and will be available publicly with the release of oneAPI v2025 probably in Nov or Q4 2024 time frame. Thanks for reporting this issue and I will update here as soon as I get the public release notification.

peastman:
Thanks!  That's good to know.

peastman:
The new version of Intel OpenCL released last month seems to have fixed the failure in TestCustomHbondForce.  On the other hand, there was a new failure in TestOpenCLCustomCPPForce.  It passed when it reran that test, but it ought to be deterministic, so something isn't right.  I'll investigate.

peastman:
I can't reproduce the error locally.  I'll go ahead and merge this, then keep an eye out for any further errors in future builds.

PR #4375, opened 2024-01-02
Fixes formatting issue for "sub" tag conversion in Python API documentation

cabb99:
This pull requests fixes a formatting issue for the LocalCoordinatesSite Python API documentation . Previously, the \<sub\> tag in the Python API documentation was not being processed correctly, leading to display issues.

Before the fix (Incorrect formatting with \<sub\> tag):
Screenshot from 2024-01-01 21-33-57

After the fix (Corrected formatting):
Screenshot from 2024-01-01 21-33-01

### DIFF
--- a/docs-source/api-python/process-docstring.py
+++ b/docs-source/api-python/process-docstring.py
@@ -51,6 +51,9 @@ def process_docstring(app, what, name, obj, options, lines):
                 s = linesep + s
             newline = '.. verbatim::' + linesep
             return newline + '    ' + s.replace(linesep, linesep + '    ')
+    def replace_subscript(m):
+        """ Replace subscript tags. """
+        return r'\ :sub:`{}`\ '.format(m.group(1))
 
     linesep = '|LINEBREAK|'
     joined = linesep.join(lines)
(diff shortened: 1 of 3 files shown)


peastman:
Thanks!  How about doing the same for `<sup>` tags?  Those lines contain both.  The superscripts seem to just be getting deleted.

https://github.com/openmm/openmm/blob/abadc8211158900209a0893876e69ba7fe65968c/openmmapi/include/openmm/VirtualSite.h#L177-L181

cabb99:
The sup tags were not being passed correctly to the python documentation. It seems they were omitted by the swigInputBuilder.
```
>>> openmm.LocalCoordinatesSite.__doc__
'\n    This is a VirtualSite that uses the locations of several other particles to compute a local coordinate system, then places the virtual site at a fixed location in that coordinate system. The origin of the coordinate system and the directions of its x and y axes are each specified as a weighted sum of the locations of the other particles:\n\n\n    origin = w<sub>1</sub>r<sub>1</sub> + w<sub>2</sub>r<sub>2</sub> + w<sub>3</sub>r<sub>3</sub> + ...\n\n\n    xdir = w<sub>1</sub>r<sub>1</sub> + w<sub>2</sub>r<sub>2</sub> + w<sub>3</sub>r<sub>3</sub> + ...\n\n\n    ydir = w<sub>1</sub>r<sub>1</sub> + w<sub>2</sub>r<sub>2</sub> + w<sub>3</sub>r<sub>3</sub> + ...\n\n\n
```
It should be corrected now:
LocalCoordinateSite_corrected

peastman:
That looks great.  Thank you!

PR #4402, opened 2024-01-22
Fix unqualified std::move instances

jhenin:
that are causing warnings by Clang. This is just touching one file from the Lepton library.
```
lepton/src/ExpressionTreeNode.cpp:66:105: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
```


jhenin:
Tests fail with memory errors, but only on AMD OpenCL:
```
  7/166 Test   #8: TestOpenCLBrownianIntegratorMixed ...............Subprocess aborted***Exception:   1.20 sec
corrupted double-linked list
        Start   9: TestOpenCLBrownianIntegratorDouble
  8/166 Test   #9: TestOpenCLBrownianIntegratorDouble ..............Subprocess aborted***Exception:   1.11 sec
malloc_consolidate(): invalid chunk size
```
...and more. I doubt all of these tests use the Lepton code paths that this PR touches.

peastman:
Thanks!

Issue #4395, opened 2024-01-16
Poor ATMForce() performance due to lack of atom reordering


egallicc:
I am investigating poor ATMForce performance for the common case where `ATMForce()` is given a NonbondedForce that is removed from the System. 

I tracked the issue down to ComputeContext skipping `reorderAtomsImpl()` because `getNonbondedUtilities().getUseCutoff()` is `false` (in addition, `getNonbondedUtilities().getUsePeriodic()` is also `false`). I verified that the performance improves by a factor of two by forcing atom reordering.

Based on my tests, `ComputeContext::reorderAtoms()` is only called in the outer Context. Does this reflect the design? If so, it would make sense that reordering is not carried out because the outer context does not manage a NonbondedForce.

Assuming that this analysis is correct, is there a clean way to force atom reordering in the outer Context even when it does not manage a NonbondedForce?

egallicc:
We found a [workaround](https://github.com/Gallicchio-Lab/AToM-OpenMM/commit/be867de26bbf638576dc8c7db07167abf0f7e64a) for AToM-OpenMM that disables the NonbondedForce using force groups rather than removing it from the outer System. As a result, atom reordering appears to be triggered correctly resulting in a ~2x speed-up for protein-sized systems. This setup is similar to the one we had in place with the ATMMetaForce() plugin.

peastman:
I'm confused about how the code works, especially the ReorderListener:

https://github.com/openmm/openmm/blob/71b2a93e09d9d5f110310448c9ff503eb2e71d55/platforms/common/src/CommonKernels.cpp#L8154-L8164

It seems to be forcing the inner context to work with unsorted coordinates?  That will be very inefficient.

Compare to CustomCVForce which lets the inner context order the particles however it wants and just maintains a mapping between them for use when copying data:

https://github.com/openmm/openmm/blob/71b2a93e09d9d5f110310448c9ff503eb2e71d55/platforms/common/src/CommonKernels.cpp#L5443-L5449

egallicc:
I also have been doubtful about the current strategy. However, I tried the reordering strategy in CustomCVForce() and did not resolve the poor performance. After some testing, I almost convinced myself that the atom reordering routine is never called in the inner Contexts. Is this possible? If so, it would explain why the different strategies lead to the same outcome.

ATMForce()'s copyState() function:

[...]

PR #4408, opened 2024-01-23
Additional test case

egallicc:
We investigated random crashes of ATMForce() with a NonbondedForce and the default softplus alchemical potential during replica exchange binding free energy calculations with [AToM-OpenMM](https://github.com/Gallicchio-Lab/AToM-OpenMM). We found that many of these crashes occur at the physical endpoints where Lambda1=Lambda2=0. In these cases, when Direction=1, the ATM potential is u0 and does not depend on u1 (and vice versa for Direction=-1). As a result, u1 is unbound and can overflow when, for example, atoms clash.

This PR excludes from hybridForce() the forces from u1 (or u0) when the corresponding potential is not included in the energy function, avoiding "0*NaN = NaN" occurrences. Similarly, the u1 and u0 energies are checked for NaN values and normalized to a large value.

The PR also includes a new test of ATMForce() with a NonbondedForce() using two approaches, one that triggers atom reordering and the other that does not. It checks that the two approaches return the same perturbation energy.

 Comments/suggestions are welcome.

### DIFF
--- a/libraries/lepton/src/ExpressionTreeNode.cpp
+++ b/libraries/lepton/src/ExpressionTreeNode.cpp
@@ -63,7 +63,7 @@ ExpressionTreeNode::ExpressionTreeNode(Operation* operation) : operation(operati
 ExpressionTreeNode::ExpressionTreeNode(const ExpressionTreeNode& node) : operation(node.operation == NULL ? NULL : node.operation->clone()), children(node.getChildren()) {
 }
 
-ExpressionTreeNode::ExpressionTreeNode(ExpressionTreeNode&& node) : operation(node.operation), children(move(node.children)) {
+ExpressionTreeNode::ExpressionTreeNode(ExpressionTreeNode&& node) : operation(node.operation), children(std::move(node.children)) {
     node.operation = NULL;
     node.children.clear();
 }
@@ -108,7 +108,7 @@ ExpressionTreeNode& ExpressionTreeNode::operator=(ExpressionTreeNode&& node) {
     if (operation != NULL)
         delete operation;
     operation = node.operation;
-    children = move(node.children);
+    children = std::move(node.children);
     node.operation = NULL;
     node.children.clear();
     return *this;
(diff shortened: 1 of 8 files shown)


peastman:
I think the changes to the kernel aren't necessary.  The forces are represented in fixed point, not floating point.  Nans aren't an issue.

How did you pick 1e12 as the value to set the force to when it's nan?  That seems rather arbitrary.  Is it specific to the default energy expression, or can we be certain it will always be appropriate for all expressions?

egallicc:
Good point. I guess an overflow in the forces would be reflected in ways other than NaN's.  Would the overflow cause a sign change? Or something else? Empirically, this approach has eliminated all crashes at the endpoints during hundreds of ns of testing.

Right, I was not sure about the cutoff in general. With the default potential and near the endpoint in question, dEdu1 or dEdu0 (depending on the direction) are equal to Lambda2. At the endpoint, Lambda2 should be exactly zero. Because Lambda2 varies in reasonable increments near zero (0.0, 0.1, 0.2, ...), any small cutoff value, maybe including zero, would work fine. But I am not sure in general.

egallicc:
Sorry, you were asking about the maxEnergy value. The default expression includes a softCoreExpression tapering function that brings large perturbation energies down to a set value (Umax). Any large energy value then yields Umax regardless of the specific value. But, yes, it is not clear how to pick the maxEnergy value in general.

What we know is that crashes were not caused by state0Energy or state1Energy overflowing. The crashes were due to numerical instabilities of the forces of the "unused" energy component (u1 or u0) at the endpoint.

peastman:
The maximum possible force that can be represented is about +/- 2e9 kJ/mol/nm.  Your change only omits it when the scale factor <1e-12.  That means the contribution being added to the force will be no more than 2e-3 kJ/mol/nm, orders of magnitude smaller than typical forces.  That shouldn't make any difference.

egallicc:
Could there be some internal conversion to floats during the calculation? And that results in NaNs?

What happens to a long that overflows? Does it stay at the maximum representable value?

egallicc:
We will run more tests following your suggestions to understand the problem better. It will take a while though. The crashes are annoying but rare.

egallicc:
It looks like the crashes in our tests were confounded by mistakenly using two different versions. With 8.1.0 we observed crashes when using PBC's with atom reordering in the main Context. With 8.1.1, no crashes were observed under all tested conditions. (Possibly because of the #4351 fix?). 

Anyway, It is all good. False alarm. I undid the changes in the hybridForce() kernel. 

I left the new NonbondedForce test in `TestATMForce.h` in this PR. It stress-tests ATMForce() by handling a NonbondedForce in two different ways. It can be useful when working on #4395. Let me know if this is acceptable. Otherwise, I will close it.

peastman:
Glad to know it's already fixed.  The new test is fine.  I'll merge it once the builds finish.

Issue #4405, opened 2024-01-22
Problem reading unit cell data from OpenMM 8.x DCD files into CHARMM


RickVenable:
After successfully testing the performance of version 8.1 on a new H100 system, I tried some production runs and ran into trouble for analyses in CHARMM that required the unit cell information. OpenMM 7.x versions do not not present this problem, but both 8.0 and 8.1 are afflicted.

I reviewed the CHARMM code that reads DCD files and used the Linux od command to analyze the DCD files and found that while the unit cell data itself was okay, there was a setting in the initial 80 byte DCD header that was triggering the issue.  Specifically, the 20th 4-byte field is supposed to be an integer version number, which is checked to determine the storage form of the unit cell data.  Versions 21 and earlier stored the angles, instead of their cosines; OpenMM 7.x versions I checked were putting the number 24 in that field, while OpenMM 8.x versions are putting 21 in that field, which triggers the error, aborting the calculation.

I've patched a local CHARMM build to disable the version check, and the analyses can now be run without difficulty.  However, it would be appreciated if this could be corrected in the OpenMM source code.

peastman:
It was changed in #3521 to fix the problem described in https://github.com/openmm/openmm/issues/3507#issuecomment-1065439558 and the following discussion.  It was necessary to get correct behavior with cpptraj.  Can you read through that discussion and suggest if there's a better behavior to use?

RickVenable:
I'd like to first point out that the DCD format originated with CHARMM, and not NAMD or cpptraj.  As such, making changes that break CHARMM compatibility seem ill-considered in my view.  

NAMD and VMD in particular  have taken great liberties with the information in the file header.  It wouldn't be the first time I've made changes to the CHARMM trajectory reader to allow for the incorrect behavior of other programs, and probably won't be the last.

peastman:
In principle I agree.  But CHARMM doesn't provide documentation on the DCD format, and the source code isn't public.  If there were a clear spec we'd follow it, but since there isn't everyone tries to reverse engineer it as best they can.

RickVenable:
CHARMM has been freely available to academic and non-profits for quite a while, but does require registration.

If the the version number is < 22, the layout of cell  terms should be different; one should not change the version number without changing that as well.

peastman:
Is there public documentation for the format?

drroe:
[...]

PR #4410, opened 2024-01-23
Write DCD version as 24 instead of 21 to fix ability of CHARMM to read DCD trajectories written by OpenMM

drroe:
Revert #3521; see #4405.

Essentially OpenMM is writing values that are compatible with the symmetric shape matrix, even if it is not calculated the same way. For a version of 21 in the DCD file to be valid, OpenMM would need to store X Y Z and alpha beta gamma (in degrees). Therefore, a version of 24 is more appropriate.

Eventually it may be better to actually calculate the symmetric shape matrix, but this PR should fix compatibility with CHARMM. 

Tagging @peastman and @RickVenable.

### DIFF
--- a/wrappers/python/openmm/app/dcdfile.py
+++ b/wrappers/python/openmm/app/dcdfile.py
@@ -92,7 +92,7 @@ class DCDFile(object):
                 raise ValueError('Cannot append to a DCD file that contains a different number of atoms')
         else:
             header = struct.pack('<i4c9if', 84, b'C', b'O', b'R', b'D', 0, firstStep, interval, 0, 0, 0, 0, 0, 0, dt)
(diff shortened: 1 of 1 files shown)


drroe:
Looks like Linux AMD OpenCL has been failing for a bit, so I don't think it's from this PR.

peastman:
Right, you can ignore it, and also the Jenkins CPU build.

peastman:
Is this ready to merge?  @rickvenable do you agree this is correct?

RickVenable:
Yes, changing the version back to 24 should fix the unit cell reading issue in CHARMM.

peastman:
Ok, thanks!

PR #4438, opened 2024-02-14
DO NOT MERGE - Checking if the new docker images fix the Jenkins issues.

swails:


swails:
CI is working now, and I've upgraded the CUDA compiler as much as my CI machines will allow me to (which is the 11.4.x branch).  Attempts to go newer than that fails because the CUDA drivers on the host machine are versioned to the 11.4 CUDA runtime and fail with a `CUDA_ERROR_COMPAT_NOT_SUPPORTED_ON_DEVICE` error.

For the time being, I've pushed those images to the `latest` tag while I get the Amber builds working.  The Jenkins builds should be working again (unfortunately I'll need newer hardware or an OS/driver update before I can update the base images again).

----

Note that even though the builds are dockerized, the Nvidia runtime basically just passes through the host CUDA driver, so you're still restricted to the host configuration.

PR #4440, opened 2024-02-14
Use LF-Middle for LangevinIntegrator and VariableLangevinIntegrator

### LINKED ISSUE #4435: Use LF-Middle for LangevinIntegrator
It has been over three years since we introduced LangevinMiddleIntegrator in version 7.5.  It's a drop in replacement for LangevinIntegrator, but uses a more accurate discretization.  There's really no reason for anyone to still use the old LangevinIntegrator, but lots of people still do.  It frequently appears in scripts people post here, and every time I have to point out to them they should switch to the newer class.

At the time, I objected to changing the algorithm while keeping the class name the same, on the grounds of backward compatibility.  That was largely because we thought at the time we were going to switch from a leapfrog to a velocity verlet formulation.  That would fundamentally change the definition of velocities and introduce subtle errors in a lot of code.  But we ended up sticking with leapfrog instead, just calculating it in a more accurate way.  That's a much safer change.  And really, there is no reason for anyone to still be using the old discretization.

So I suggest the time has come to switch LangevinIntegrator to use LF-Middle as well.  It and LangevinMiddleIntegrator can become synonyms.  In fact, one of them can just subclass the other without adding or changing anything.

While we're at it, we should also switch VariableLangevinIntegrator to use the better discretization.

peastman:
Fixes #4435.

@tristanic you might want to try this out since you use VariableLangevinIntegrator.  This will hopefully allow it to be stable with larger time steps.  At the very least it shouldn't make it any worse.  It would be good to get some independent confirmation of that.


swails:
image

tristanic:
Thanks - this looks like it will be useful, particularly for ISOLDE users on less-powerful machines. I'm a little time-poor at the moment so can't promise to test it immediately, but will make certain to do so when I get the chance.

PR #4449, opened 2024-02-22
Add LoongArch architecture LSX support.

zangruochen:
LoongArch is a new architecture, already supported by linux-6.1, gcc-12.

Test result：
100% tests passed, 0 tests failed out of 118

Total Test time (real) = 310.05 sec
[openmm_test.log](https://github.com/openmm/openmm/files/14367770/openmm_test.log)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,11 @@ if ("${TARGET_ARCH}" MATCHES "ppc")
     set(PPC ON)
     add_definitions(-D__PPC__=1)
 endif()
+if ("${TARGET_ARCH}" MATCHES "loongarch64")
+    set(LOONGARCH64 ON)
+    add_definitions(-D__LOONGARCH64__=1 -mlsx)
+endif()
+
 
 # Where to install
 IF(WIN32)
--- a/cmake_modules/TargetArch.cmake
+++ b/cmake_modules/TargetArch.cmake
@@ -87,6 +87,8 @@ set(archdetect_c_code "
     #else
         #error cmake_ARCH ppc
     #endif
+#elif defined(__loongarch64)
+    #error cmake_ARCH loongarch64
 #endif
 
 #error cmake_ARCH unknown
--- a/openmmapi/include/openmm/internal/hardware.h
+++ b/openmmapi/include/openmm/internal/hardware.h
@@ -94,7 +94,7 @@ static int getNumProcessors() {
 #define cpuid __cpuid
 #else
 #if !defined(__ANDROID__) && !defined(__PNACL__) && !defined(__PPC__) \
-    && !defined(__ARM__) && !defined(__ARM64__)
+    && !defined(__ARM__) && !defined(__ARM64__) && !defined(__LOONGARCH64__)
     static void cpuid(int cpuInfo[4], int infoType) {
     #ifdef __LP64__
         __asm__ __volatile__ (


swails:
@peastman - a counterargument is that loongarch is [already supported by QEMU](https://www.qemu.org/docs/master/about/emulation.html#id2).

We could likely add a CI job that exercises this code.  I also don't think "lots of code" is a fair description of this PR.  At least a quarter of the added lines are comments or blank, and much of the rest is pretty typical boilerplate.

These vectorize headers also seem to be some of the more stable parts of the code base.  If you count the total number of commits that have touched each file, the number is incredibly small:

```bash
-> % for f in openmmapi/include/openmm/internal/vectorize*; do
for> echo "$f - `git log $f | grep commit | wc -l`"
for> done
openmmapi/include/openmm/internal/vectorize.h -       16
openmmapi/include/openmm/internal/vectorizeAvx.h -        1
openmmapi/include/openmm/internal/vectorizeAvx2.h -        2
openmmapi/include/openmm/internal/vectorize_neon.h -       21
openmmapi/include/openmm/internal/vectorize_portable.h -        1
openmmapi/include/openmm/internal/vectorize_ppc.h -        9
openmmapi/include/openmm/internal/vectorize_sse.h -       13
```

vectorize_neon is the outlier here, and seems to have pulled in efforts ranging from android optimization to macOS ARM optimization.  Regardless, these files are rarely touched, so I suspect the maintenance burden would be pretty minimal.

That said, I do think it's worth measuring the performance impact of this code vs. `vectorize_portable.h`.

peastman:
> That said, I do think it's worth measuring the performance impact of this code vs. vectorize_portable.h.

Good idea.  @zangruochen can you run some of the benchmarks and compare the speed with your vectorized code compared to the portable version?  Run this in the `examples` directory:

```
python benchmark.py --platform=CPU --test=gbsa,pme,apoa1pme --style=table
```

That assumes you're planning to use the CPU platform.  If you'll be running simulations on a GPU, the vectorization speed is irrelevant.

zangruochen:
@peastman @swails Thank you very much for your suggestions, here are the test results:


pci id for fd 4: 0014:7a06, driver (null)
pci id for fd 4: 0014:7a06, driver (null)
pci id for fd 5: 0014:7a06, driver (null)
/home/loongarch/openmm-la/examples/benchmark.py:496: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  system_info['timestamp'] = datetime.utcnow().isoformat()
/bin/sh: 行 1: nvidia-smi: 未找到命令
hostname: Sunhaiyong
timestamp: 2024-02-26T09:38:33.779225
openmm_version: 8.1.0.dev-0e79473
cpuinfo: 
cpuarch: 
system: Linux

gbsa              mixed       HBonds        1.5            4         NVT        CPU        3.83294
pme               mixed       HBonds        1.5            4         NVT        CPU        4.79662
apoa1pme          mixed       HBonds        1.5            4         NVT        CPU        1.18133

peastman:
Which version is that?  The point is to compare the two implementations of vectorization (vectorize_portable.h and vectorize_lsx.h) and see if one is significantly faster than the other.

zangruochen:
@peastman 
The above test results were obtained with vectorize_lsx.h enabled.
I tried compiling with vectorize_portable.h, but there is a type conversion error.
[compile.log](https://github.com/openmm/openmm/files/14416612/compile.log)

peastman:
Here is the error message.

```
/home/loongarch/openmm-la/openmmapi/include/openmm/internal/vectorize_portable.h:61:26: error: cannot convert ‘<brace-enclosed initializer list>’ to ‘__m128’ {aka ‘__vector(4) float’} in assignment
   61 |         val = {v, v, v, v};
```

I don't know why it would object to that line.  I haven't had any trouble compiling it before, and that's exactly the standard syntax for initializing vectors.  See https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html.  Here is the complete method.

```c++
    fvec4(float v) {
        val = {v, v, v, v};
    }
```

`val` is declared as `__m128`, which is defined as

```
typedef float __m128 __attribute__((vector_size(16), aligned(4)));
```

swails:
@zangruochen - what compilers (and versions) are you using?

zangruochen:
@peastman @swails 
I tested it in the following two gcc version environments and got the same error.

$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/loongarch64-loongson-linux-gnu/8/lto-wrapper
Target: loongarch64-loongson-linux-gnu
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,fortran,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.loongnix.com/bugzilla --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --with-linker-hash-style=gnu --enable-plugin --enable-initfini-array --with-isl --disable-libmpx --with-arch=loongarch64 --with-abi=lp64 --enable-tls --with-long-double-128 --disable-multilib --enable-initfini-array --enable-gnu-indirect-function --disable-host-shared --disable-emultls --build=loongarch64-loongson-linux-gnu
Thread model: posix
gcc version 8.3.0 20190222 (Loongson 8.3.0-35 vec) (GCC) 

$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/loongarch64-unknown-linux-gnu/14.0.1/lto-wrapper
Target: loongarch64-unknown-linux-gnu
Configured with: ../configure --prefix=/usr --libdir=/usr/lib64 --build=x86_64-cross-linux-gnu --host=loongarch64-unknown-linux-gnu --target=loongarch64-unknown-linux-gnu --enable-__cxa_atexit --enable-threads=posix --with-system-zlib --enable-libstdcxx-time --enable-checking=release --with-build-sysroot=/opt/mylaos/Sunhaiyong/distro/Yongbao/loongarch64/workbase/sysroot --enable-linker-build-id --with-linker-hash-style=gnu --enable-gnu-unique-object --enable-initfini-array --enable-gnu-indirect-function --disable-multilib --with-simd=lsx --enable-languages=c,c++,fortran,objc,obj-c++,lto
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 14.0.1 20240128 (experimental) (GCC)

peastman:
It seems that gcc's handling of the vector API is much more finicky than llvm's.  Take the line

```c++
        val = {v, v, v, v};
```

That compiles fine on llvm, but gcc reports an error.  On the other hand, it has no problem with

```c++
        __m128 vv = {v, v, v, v};
        val = vv;
```

or with

```c++
        val = (__m128) {v, v, v, v};
```

Give me a little time and I'll see if I can get it compiling correctly under both compilers.  Meanwhile, can you test on llvm?

zangruochen:
@peastman @swails 
I compiled it using clang19.

The vectorize_portable.h test result is as follows:
gbsa              mixed       HBonds        1.5            4         NVT        CPU        1.31195
pme               mixed       HBonds        1.5            4         NVT        CPU        1.62123
apoa1pme          mixed       HBonds        1.5            4         NVT        CPU        0.429308

The vectorize_lsx.h test result is as follows:
gbsa              mixed       HBonds        1.5            4         NVT        CPU        3.82635
pme               mixed       HBonds        1.5            4         NVT        CPU        4.73292
apoa1pme          mixed       HBonds        1.5            4         NVT        CPU        1.18535

swails:
It looks like the `vectorize_lsx.h` is 3x faster! That's a fair bit higher than I had expected.

peastman:
That's a big enough difference that it probably means clang isn't really vectorizing it, and is just implementing the vector code with scalar instructions.

Here's an updated version that should (hopefully!) compile on gcc.  Can you try it out and see if gcc gives better performance?

[vectorize_portable.h.zip](https://github.com/openmm/openmm/files/14437703/vectorize_portable.h.zip)

peastman:
The other question is how this computer will be used?  Even the LSX speeds are still really slow, probably too slow to be useful for much of anything.  Are there faster computers with this architecture?  Or will you be running simulations on a GPU, in which case the speed of the CPU platform doesn't matter?

For comparison, here is what I get running those benchmarks on a 2+ year old midrange laptop with an embedded GPU.

```
gbsa              single      HBonds        1.5            4         NVT        OpenCL     395.697
pme               single      HBonds        1.5            4         NVT        OpenCL     156.274
apoa1pme          single      HBonds        1.5            4         NVT        OpenCL     38.6842
```

Running on a good discrete GPU is roughly 10x faster than that.

zangruochen:
@peastman @swails 

Maybe clang didn't vectorize it correctly, when I compile the new vectorize_portable.h with gcc, it compiles and passes properly. I tested it on cpu (3A5000) and cpu (3A6000), and the result is that vectorize_portable.h and vectorize_lsx.h perform the same.

3a5000 test results:
gbsa              mixed       HBonds        1.5            4         NVT        CPU        3.83028
pme               mixed       HBonds        1.5            4         NVT        CPU        4.75194
apoa1pme          mixed       HBonds        1.5            4         NVT        CPU        1.18163

3a6000 test results:
gbsa              mixed       HBonds        1.5            4         NVT        CPU        8.77684
pme               mixed       HBonds        1.5            4         NVT        CPU        11.6979
apoa1pme          mixed       HBonds        1.5            4         NVT        CPU        2.76064

I'll try to continue optimizing vectorize_lsx.h later to see if I can improve performance.

peastman:
Sorry, my mistake.  Try this version.

[vectorize_portable.h.zip](https://github.com/openmm/openmm/files/14453362/vectorize_portable.h.zip)

zangruochen:
@peastman @swails 
I tested the new code with the following results

1. The following error occurs when compiling
```
/home/loongson/python/workspace/openmm/openmmapi/include/openmm/internal/hardware.h:100:9: error: output number 1 not directly addressable
 100 | __asm__ __volatile__ (
 | ^~~~~~~
/home/loongson/python/workspace/openmm/openmmapi/include/openmm/internal/hardware.h:100:9: error: output number 2 not directly addressable
```
2、Use gcc to compile, the test passes.
3、Use clang to compile, the test are reported the following errors
```
Start 1: TestReferenceATMForce
 1/118 Test #1: TestReferenceATMForce ......................... ***Exception: SegFault 0.59 sec.
 Start 2: TestReferenceAndersenThermostat
 2/118 Test #2: TestReferenceAndersenThermostat ............... ***Exception: SegFault 0.62 sec.
```

peastman:
Did you change something else?  That error is in hardware.h.  I only sent changes to vectorize_portable.h.

And none of this should have any effect on the reference platform tests.

zangruochen:
@peastman 
Perhaps the judgment here in hardware.h needs some tweaking
```
/**
 * Get a description of the CPU's capabilities.
 */
#ifdef WIN32
#define cpuid __cpuid
#else
**#if !defined(__ANDROID__) && !defined(__PNACL__) && !defined(__PPC__) \
    && !defined(__ARM__) && !defined(__ARM64__)**
    static void cpuid(int cpuInfo[4], int infoType) {
    #ifdef __LP64__

```

The gcc and clang tests are based on the same source code, and I've only made changes to hardware.h to fix the error above.

peastman:
But [up above](https://github.com/openmm/openmm/pull/4449#issuecomment-1970503354) you said it compiled and ran.  What changed?

zangruochen:
I used the following file modifications from this PR.
CMakeLists.txt
cmake_modules/TargetArch.cmake
openmmapi/include/openmm/internal/hardware.h

peastman:
My question is, what's different between https://github.com/openmm/openmm/pull/4449#issuecomment-1970503354 where you said it compiled and ran, and https://github.com/openmm/openmm/pull/4449#issuecomment-1978596718 where you said it didn't?

zangruochen:
Sorry, I didn't describe it very well.

When I compile after getting the source code directly and replacing vectorize_portable.h, it errors out.

When I get the source and modify the files CMakeLists.txt, cmake_modules/TargetArch.cmake and openmmapi/include/openmm/internal/hardware.h and replace vectorize_portable .h, and after that compiled with gcc, it compiled and tested successfully.

When I get the source and modify the files CMakeLists.txt, cmake_modules/TargetArch.cmake and openmmapi/include/openmm/internal/hardware.h and replace vectorize_portable .h, and then use clang to compile, it compiles successfully but the test fails.

See #4449 for how to modify CMakeLists.txt, cmake_modules/TargetArch.cmake and openmmapi/include/openmm/internal/hardware.h

peastman:
Ok, I think I understand.  The updated vectorize_portable.h is in #4466.  If I understand correctly, under gcc it compiles and runs correctly and is just as fast as vectorize_lsx.h.  Is that right?  If so then we can use it.  You can remove the LSX code from this PR and use the portable version.  All the other changes are minor and I don't see any issue with merging them.

Issue #4452, opened 2024-02-22
Large time taken to initialize `OpenMM::Context`


philipturner:
I was working on a large structural piece (600,000 atoms, 50 nm x 120 nm), and found a bottleneck in my workflow. I am trying to utilize a GPU cluster with ~360 TFLOPS, but there's a 10-second latency bottleneck that comes from single-core CPU. Perhaps this could motivate an enhancement to multithread the C++ code in OpenMM. I already multithread the Swift code external to OpenMM.

![Screenshot 2024-02-22 at 6 16 48 PM](https://github.com/openmm/openmm/assets/71743241/2afac583-942b-496c-953b-862dba41ec4e)

Below is the workflow and where the bottleneck appears.

```swift
  // t = 0 ms
  
  /* compile the lattice of carbon and silicon atoms */

  // t = 115 ms
  
  var topology = Topology()
  topology.insert(atoms: backBoardAtoms)
  var reconstruction = SurfaceReconstruction()
  reconstruction.material = .checkerboard(.silicon, .carbon)
  reconstruction.topology = topology
  reconstruction.removePathologicalAtoms()
  reconstruction.createBulkAtomBonds()
  reconstruction.createHydrogenSites()
  reconstruction.resolveCollisions()
  reconstruction.createHydrogenBonds()
  topology = reconstruction.topology
  topology.sort()

  // t = 468 ms
  
  var paramsDesc = MM4ParametersDescriptor()
  paramsDesc.atomicNumbers = topology.atoms.map(\.atomicNumber)
  paramsDesc.bonds = topology.bonds
  let parameters = try! MM4Parameters(descriptor: paramsDesc)

  // t = 2717 ms
  
  var forceFieldDesc = MM4ForceFieldDescriptor()
  forceFieldDesc.parameters = parameters
  forceFieldDesc.integrator = .multipleTimeStep
  forceFieldDesc.platform = OpenMM_Platform.platforms.first(where: { $0.name == "HIP" })!
  let forceField = try! MM4ForceField(descriptor: forceFieldDesc)
  
  // t = 11750 ms
  
  forceField.positions = topology.atoms.map(\.position)
  forceField.minimize()
  for i in topology.atoms.indices {
    topology.atoms[i].position = forceField.positions[i]
  }

  // t = 11874 ms - set maxIterations argument to 1
  // t = 17745 ms - full minimization on M1 Max (10 TFLOPS)
  
  return topology.atoms
```

[...]

PR #4453, opened 2024-02-23
Minor optimization to validating exclusions

peastman:
See #4452.  This slightly reduces context creation time for systems with very large numbers of exclusions.

cc @philipturner.  On your system with two CustomNonbondedForces and 4.8 million exclusions, this reduces context creation time by about 1.5 seconds.


peastman:
I added some optimizations to `findMoleculeGroups()`.  They shave about another second off the time.

Issue #4460, opened 2024-03-04
`error: HWCAP_NEON was not declared in this scope` on NVIDIA Grace-Hopper


migueldiascosta:
it looks like `openmmapi/include/openmm/internal/vectorize_neon.h` is looking for `HWCAP_NEON` instead of `HWCAP_ASIMD`

because  `CMakeLists.txt` is setting `-D__ARM__=1` instead of `-D__ARM64__=1`, 

because `cmake_modules/TargetArch.cmake` is outputing `arm` instead of `armv8`;

forcing `TARGET_ARCH` to be `armv8` in  `openmm-8.0.0/CMakeLists.txt` fixed the issue for me, maybe `TargetArch.cmake` needs to be updated to support grace-hopper/armv9?

peastman:
I'm not familiar with the difference between those flags.  When should it check `HWCAP_NEON` and when should it check `HWCAP_ASIMD`?

A quick search didn't turn up any explanation.  I found ARM documentation that describes `HWCAP_NEON` but doesn't mention `HWCAP_ASIMD`.  And I found Linux kernel documentation that describes `HWCAP_ASIMD` but doesn't mention `HWCAP_NEON`.

migueldiascosta:
I'm certainly not an expert but it looks like the advanced simd instructions are called `neon` in Arm 32-bit (Arm v7 and below, referred to as simply `ARM`) hwcaps (cpu features) and `asimd` in Arm 64-bit (Arm v8 and above, referred to as `ARM64`), so only `HWCAP_NEON` is defined in 32-bit kernels and only `HWCAP_ASIMD` in 64-bit kernels 

What I don't get is why `cmake_modules/TargetArch.cmake` is not returning `armv8` anyway in this system, at first I thought it was because it was explicitly looking for v8 and NVIDIA Grace (Arm Neoverse V2) is v9, but it's also looking for `__aarch64__` in that `ifdef`, so it should still work...

peastman:
We don't support any 32 bit ARM platforms, so if that's the case, we can simplify

```c++
#elif defined(__ARM__)
    unsigned long features = getauxval(AT_HWCAP);
    return (features & HWCAP_NEON) != 0;
#else
    unsigned long features = getauxval(AT_HWCAP);
    return (features & HWCAP_ASIMD) != 0;
#endif
```

to just


```c++
#else
    unsigned long features = getauxval(AT_HWCAP);
    return (features & HWCAP_ASIMD) != 0;
#endif
[...]

PR #4462, opened 2024-03-04
remove possibility of defining (32-bit) __ARM__ instead of __ARM64__

migueldiascosta:
fixes https://github.com/openmm/openmm/issues/4460


peastman:
Thanks!

PR #4467, opened 2024-03-08
Add LoongArch architecture LSX support.

zangruochen:
LoongArch is a new architecture, already supported by linux-6.1, gcc-12.


peastman:
This is a revised version of #4449.  See the discussion there for background.

Looks good!

PR #4470, opened 2024-03-09
Revised code from installing CUDA on CI

peastman:
CUDA builds have recently begun failing.  I'm not sure why, though I think it might be that github has started using a newer version of Ubuntu.  I'm attempting to get it working again.

I also updated the versions of CUDA and Python in various builds to better reflect the versions we currently support.


swails:
Maybe you could try using the same docker images that Jenkins runs on as the build agent?

peastman:
Got it!

PR #4480, opened 2024-03-18
Update references to "openmmforcefields" project

mattwthompson:
Small thing (the link redirects after both name change and organization transfer) but it stood out to my eyes while doing some questionable things with combinations of force fields ...

image


peastman:
Thanks!

PR #4485, opened 2024-03-27
Avoid overflow in large XTC files

RaulPPelaez:
The XTC wrappers store some indexes as int, which can overflow when the trajectories are sufficiently large (a.i. 2e6 atoms in frame 1000).
https://github.com/openmm/openmm/blob/f82876f05d0a3d87927ba32428ee165e056829ab/wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp#L137-L139

This PR changes these calculations so that they are stored as size_t instead.


peastman:

RaulPPelaez:
Yes, thanks Peter!.

PR #4495, opened 2024-04-03
ATMForce reorders inner contexts for better performance

### LINKED ISSUE #4395: Poor ATMForce() performance due to lack of atom reordering

peastman:
Fixes #4395.  The main context and the inner contexts are reordered independently now, so each one can pick an appropriate atom order based on its forces.

@egallicc can you try this out?  You should now get good performance without needing the workaround.

### DIFF
--- a/openmmapi/src/ATMForceImpl.cpp
+++ b/openmmapi/src/ATMForceImpl.cpp
@@ -115,9 +115,6 @@ void ATMForceImpl::initialize(ContextImpl& context) {
 
     innerContext0 = context.createLinkedContext(innerSystem0, innerIntegrator0);
     innerContext1 = context.createLinkedContext(innerSystem1, innerIntegrator1);
-    vector<Vec3> positions(system.getNumParticles(), Vec3());
-    innerContext0->setPositions(positions);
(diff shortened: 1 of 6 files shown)


egallicc:
Thank you! I will test it this weekend if it's okay.

egallicc:
All good on our end. This will make our life a lot easier. Thanks!

peastman:
Thanks!

PR #4506, opened 2024-04-09
added type checking for Simulation.step()

### LINKED ISSUE #4503: simulation.step(n) will raise a TypeError when n is a float when n%1 != 0 after all but the last step
Actual behavior:
When calling step on a simulation with the amount of steps 'n' being a float, the simulation will run until the last 10 steps and only then throw a TypeError if n%1 is not 0.

Expected behaviour:
Either throw a TypeError right away if a float is passed to simulation.step(), or convert the float to an integer number.

MWE:
(Code taken from the cookbook)
```
from openmm.app import *
from openmm import *
from openmm.unit import *
from sys import stdout

# Load in the PDB strucure
pdb = PDBFile('villin.pdb')

# Specifiy the forcefield
forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')


# Combine the molecular topology and the forcefield
system = forcefield.createSystem(pdb.topology, nonbondedMethod=PME,
        nonbondedCutoff=1*nanometer, constraints=HBonds)

# Create the integrator to use for advacing the equations of motion.
# It specifies a Langevin integrator.
# The paramters set are temperature, friction coefficient, and timestep.
integrator = LangevinMiddleIntegrator(300*kelvin, 1/picosecond, 0.004*picoseconds)

# Combines the molecular topology, system, and integrator
# to begin a new simulation.
simulation = Simulation(pdb.topology, system, integrator)
simulation.context.setPositions(pdb.positions)

# Perform local energy minimization
print("Minimizing energy...")
simulation.minimizeEnergy(maxIterations=100)

# Write the trajectory to a file called "output.pdb"
simulation.reporters.append(PDBReporter('output.pdb', 1000))

# Report infomation to the screen as the simulation runs
simulation.reporters.append(StateDataReporter(stdout, 100, step=True,
        potentialEnergy=True, temperature=True))

#Run the simulation for 1000 timsteps
print("Running simulation...")
simulation.step(1000.4)
```

Output:
```
Minimizing energy...
Running simulation...
#"Step","Potential Energy (kJ/mole)","Temperature (K)"
100,-143311.46522934374,257.8203757521306
200,-143070.9745066875,278.06338493020644
300,-142160.01112778124,283.8685100712398
400,-141118.346577,284.75547186722747
500,-140812.1034129375,292.94962307842
600,-140912.8231395,296.31750958703674
700,-141149.50429184374,301.20267256578234
800,-140853.22108871874,297.52991215247675
900,-140940.55507309374,296.56129305815483
1000,-140816.0643504375,303.43163527463173
  File "/xxx/debug.py", line 40, in <module>
    simulation.step(1000.4)
  File "/xxx/anaconda3/envs/openMM/lib/python3.10/site-packages/openmm/app/simulation.py", line 147, in step
    self._simulate(endStep=self.currentStep+steps)
  File "/xxx/anaconda3/envs/openMM/lib/python3.10/site-packages/openmm/app/simulation.py", line 216, in _simulate
    self.integrator.step(stepsToGo)
  File "/xxx/anaconda3/envs/openMM/lib/python3.10/site-packages/openmm/openmm.py", line 15287, in step
    return _openmm.LangevinMiddleIntegrator_step(self, steps)
TypeError: in method 'LangevinMiddleIntegrator_step', argument 2 of type 'int'
```

Proposed fix:
Add a type hint in simulation.step(), or convert any float passed to the function to an integer.

This is a relatively small change. I can implement a fix and make a PR if desired.

schuhmc:
Fixes #4503 

Calling Simulation.step() will now raise a TypeError when passed a non-integer value. A type hint was added as well.

A more lenient approach would be to to convert a passed float to an int and provide the user with a warning instead. In my opinion, a fractional step does not really make a lot of sense and therefore restricting the number of steps to be taken to an integer is a better option.

### DIFF
--- a/wrappers/python/openmm/app/simulation.py
+++ b/wrappers/python/openmm/app/simulation.py
@@ -142,8 +142,11 @@ class Simulation(object):
         """
         mm.LocalEnergyMinimizer.minimize(self.context, tolerance, maxIterations, reporter)
 
-    def step(self, steps):
+    def step(self, steps: int):
         """Advance the simulation by integrating a specified number of time steps."""
+        if not isinstance(steps, int):
+            raise TypeError(f'Expected type int for steps, got {type(steps)}')
+            
         self._simulate(endStep=self.currentStep+steps)
 
     def runForClockTime(self, time, checkpointFile=None, stateFile=None, checkpointInterval=None):
--- a/wrappers/python/openmm/app/simulation.py
+++ b/wrappers/python/openmm/app/simulation.py
@@ -144,8 +144,8 @@ class Simulation(object):
 
     def step(self, steps: int):
         """Advance the simulation by integrating a specified number of time steps."""
-        if not isinstance(steps, int):
-            raise TypeError(f'Expected type int for steps, got {type(steps)}')
+        if int(steps) != steps:
+            raise TypeError(f'Expected an integer for steps, got {type(steps)}')
             
         self._simulate(endStep=self.currentStep+steps)
 


wrappers/python/openmm/app/simulation.py
```diff
@@ -142,8 +142,11 @@ def minimizeEnergy(self, tolerance=10*unit.kilojoules_per_mole/unit.nanometer, m
         """
         mm.LocalEnergyMinimizer.minimize(self.context, tolerance, maxIterations, reporter)
 
-    def step(self, steps):
+    def step(self, steps: int):
         """Advance the simulation by integrating a specified number of time steps."""
+        if not isinstance(steps, int):
```
peastman, line 147: How about changing the check to `if int(steps) != steps`?  That way it will accept any integer type, for example `np.int64`, not just `int`.

schuhmc: Good point. I also like this approach as it will not break old simulation setups where a float f%1=0 was passed.

peastman:
Looks good.  Thanks!

Issue #4524, opened 2024-04-26
modeller.addSolvent error with residue templates and non-standard residues


FloLangenfeld:
I'm working on a system with a non-canonical residue that binds to a cystein residue. I computed a set of parameters for these residues (including the cystein, which has different partial charges compared to the standard CYX residue, but has the same atoms).

Then, I tried to solvate my system. I provided the `Modeller.addSolvent` method a dictionary (where the keys are `Residue`s from my system's `Topology`, and the values are the residue names) using the `residueTemplates` keyword, but it throws an error, saying that there is an ambiguity between my cystein-like residue andthe standard cystein residue (despite the residue templates are provided). The trace points to the `Modeller._addIons` method.

Looking at the code, the `Modeller.addSolvent` method replaces the system's initial `Topology` by a new one before adding ions, which breaks the references to the `Residue` objects stored in the residueTemplates dictionary (as they are pointing to `Residue`s of the `Topology` that were replaced). 

Creating a new residueTemplates dictionary while creating the new `Topology` solved the issue and allowed me to solvate and ionize the system. I may submit a PR with this fix if you wish.

peastman:
A PR would be great, thank you!

FloLangenfeld:
Didn't notice that a few other methods also used resideuTemplates... Good to know!
A test is a good idea, agreed.

Thank you very much @peastman

PR #4525, opened 2024-04-29
Update residueTemplates dictionary when solvating a system

FloLangenfeld:
During the solvation process, a new `Topology` is created when adding the water molecules. Providing a residueTemplates to the `Modeller.addSolvent` method results in an error when adding the ions, as the residueTemplates' keys point to the old topology.

This PR creates an updated newResidueTemplates with the updated keys, that is then provided to the `Modeller._addIons` method.

In my case, this fix solves the issue described in #4524


peastman:
Looks good.  We should also add a test case for it, but I can do that in a separate PR.  Thanks!

PR #4531, opened 2024-05-02
Allow multiple registrations of the same atom type if definitions identical

mattwthompson:
This implements a suggestion that `openmforcefields` appears to need for current versions of GAFF and/or `prmchk2`: https://github.com/openmm/openmmforcefields/issues/281#issuecomment-1599598920

I have no idea how to test this here and I haven't thought of unintended consequences. It certainly seems like the short of change that could have side effects.

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -438,14 +438,9 @@ class ForceField(object):
         if name in self._atomTypes:
             #  allow multiple registrations of the same atom type provided the definitions are identical
             existing = self._atomTypes[name]
-            if (
-                existing.atomClass != parameters['class']
-            ) or (
-                existing.mass != float(parameters['mass'])
-            ) or (
-                existing.element.symbol != parameters['element']
-            ):
-                raise ValueError('Found multiple definitions for atom type: '+name)
+            if existing.atomClass == parameters['class'] and existing.mass == float(parameters['mass']) and existing.element.symbol == parameters['element']:
+                return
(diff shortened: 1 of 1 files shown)


wrappers/python/openmm/app/forcefield.py
```diff
@@ -436,7 +436,16 @@ def registerAtomType(self, parameters):
...
-            raise ValueError('Found multiple definitions for atom type: '+name)
+            #  allow multiple registrations of the same atom type provided the definitions are identical
+            existing = self._atomTypes[name]
+            if (
+                existing.atomClass != parameters['class']
+            ) or (
+                existing.mass != float(parameters['mass'])
+            ) or (
+                existing.element.symbol != parameters['element']
+            ):
```
mattwthompson, line 447: I tried to keep the change minimal, but there isn't a really pretty way to string together three expressions like is needed here. Python should short-circuit on the first one that evaluates falsy, so maybe there's a better order than what I happened to think of.

peastman: It might be safer to write it like this:

```python
if existing.atomClass == parameters['class'] and existing.mass == float(parameters['mass']) and existing.element.symbol == parameters['element']:
    return
```

That way it will exit leaving all existing definitions intact, instead of overwriting them with new definitions.  That seems less likely to cause problems, especially if code related to atom types changes in the future.

peastman:
Looks good.

mattwthompson:
Thanks!

PR #4533, opened 2024-05-06
Implementation of Girsanov reweighting factor reporter

anyschaefer:
# Description

We added a reweighting reporter to the  openmm app module, such that the reweighting factors are reported on-the-fly and, at regular intervals, to a reweighting factor trajectory file.

# Status

We demonstrated the correct functioning and error-free applicability of the newly implemented
reporter class. The corresponding reference "Implementation of Girsanov reweighting in OpenMM and Deeptime" is currently under review. However, the modified openmm package cannot be installed without further changes due to import errors in various `..._bindings.cpython-311-x86_64-linux-gnu.so` files. 

The required input files and a setup for the MD simulation with Girsanov reweighting are available at [reweightingtools](https://github.com/bkellerlab/reweightingtools)

# Reference

Joana-Lysiane Schäfer, Bettina G. Keller, Implementation of Girsanov reweighting in OpenMM and Deeptime

### DIFF
--- a/wrappers/python/openmm/app/reweightingreporter.py
+++ b/wrappers/python/openmm/app/reweightingreporter.py
@@ -1,0 +1,213 @@
+"""
+reweightingreporter.py: Outputs reweighting factors evaluated within simulation using a LangevinSplittingGirsanov integrator implemented in openmmtools.
+Author: Joana-Lysiane Schaefer, Freie Universität Berlin
+"""
+
+import openmm as mm
+import openmm.unit as unit
+import math
+import time
+
+## ToDo. enable more pertubations
+##       include remaining time see State reporter script
+
+class ReweightingReporter(object):
+    """ReweightingReporter outputs the energy of the perturbation U.
+    
+    To use it, create a PertubationEnergyReporter, then add it to the Simulation's list of reporters.  The set of
+    data to write is configurable using boolean flags passed to the constructor.  By default the data is
+    written in comma-separated-value (CSV) format, but you can specify a different separator to use.
+    """
+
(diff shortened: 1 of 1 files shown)


peastman:
I'm not familiar with this method.  Can you describe what it's for and what it does?

anyschaefer:
> I'm not familiar with this method. Can you describe what it's for and what it does?

Girsanov reweighting is a dynamic reweighting technique for restoring the unbiased dynamics of a molecular system from simulations performed with a biased potential. A detailed discussion about the theory is given with [Girsanov reweighting for simulations of underdamped Langevin dynamics. Theory]( 	
https://doi.org/10.48550/arXiv.2303.14696)

peastman:
How is this class intended to be used?  It seems to expect that you'll be using a specific integrator that isn't included with OpenMM?

anyschaefer:
> How is this class intended to be used? It seems to expect that you'll be using a specific integrator that isn't included with OpenMM?

Exactly, we have integrated Langevin integrators that are suitable for Girsanov reweighting in the openmmtools package with  [GirsanovSplittingIntegrator](https://github.com/choderalab/openmmtools/pull/729). For the analysis of a e.g. reweighted Markov state model with [deeptime](https://github.com/deeptime-ml/deeptime/pull/290), a trajectory of the reweighting factors generated on-the-fly is required. This could be obtained with the `ReweightingReporter` presented here.

peastman:
In that case, wouldn't it make more sense to put this class in openmmtools along with the rest of the implementation of the method?

peastman:
Closing this due to lack of response.

PR #4537, opened 2024-05-11
add install_requires with runtime deps to setup.py

mikemhenry:
not sure if we have an upper or lower bound to consider here, but things JustWorked with a `pip install numpy` which pulled down `numpy==1.26.4`

### DIFF
--- a/wrappers/python/setup.py
+++ b/wrappers/python/setup.py
@@ -164,6 +164,7 @@ def buildKeywordDictionary(major_version_num=MAJOR_VERSION_NUM,
     (through custom forces and integrators), openness, and high performance
     (especially on recent GPUs) that make it truly unique among simulation codes.
     """
+    setupKeywords["install_requires"]  = ["numpy"]
 
     define_macros = [('MAJOR_VERSION', major_version_num),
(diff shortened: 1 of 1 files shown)


mikemhenry:
>  It should be up to the library you're depending on to report the range of versions over which it maintains binary compatibility

:100: Yes, this is something that `conda build` can handle when but I am not sure the best way to handle this here. I do agree in a perfect world, at build time, we would inject the appropriate `install_requires` version range(s) for numpy. This could get complex quite quickly but I think if we only need to worry about `numpy`, I wouldn't be surpirsed if there is even a numpy function you can call that spits out the abi version ranges covered by whatever you have installed/just built against.

I will investigate the correct way to do this, perhaps for now we can put constraints on the "current" abi range while we work on a more robust solution. 

So I will figure out the correct upper and lower versions to pin here and update the PR + build patch.

mikemhenry:
Just some references 
https://numpy.org/devdocs/dev/depending_on_numpy.html#build-time-dependency 
https://github.com/scipy/oldest-supported-numpy/

peastman:
This change was included in #4662.

Issue #4199, opened 2023-08-21
Make HIP a standard platform


peastman:
AMD GPUs seem to be getting more attention these days, and the HIP platform has much better performance than OpenCL.  We should consider whether at some point we want to make it into a standard platform and bundle it with OpenMM.

In technical terms, the biggest obstacle is that conda-forge doesn't support building or distributing packages that use HIP.  Several months ago I asked about adding support (https://github.com/conda-forge/conda-forge.github.io/issues/1923).  The answer was that they have no plans, but would be open to it if someone else does the work.  AMD would of course be the obvious party to take on implementing and supporting HIP on conda-forge.  No idea if that's something they're interested in.

cc @ex-rzr @jdmaia

jdmaia:
Hey Peter, 
I'm touching base with our system software team again to see where we stand on adding ROCm on conda-forge, and will send an update once I get a definitive answer.

peastman:
It looks like there's finally some motion toward getting it supported on conda-forge.  The feedstock at https://github.com/conda-forge/hip-feedstock has been updated to version 6.0.2 (not quite the latest release, but they're working on that).  That means we can start thinking about moving forward with this.

peastman:
I'm trying to get it working using the conda-forge HIP packages.  Here are the code changes I've made so far.

- I had to bring it up to date with the latest changes in OpenMM, particularly #4346 and #4435.
- There isn't currently a package for hipFFT, so I removed all code related to it.  I think it would be fine to always use vkFFT and remove the support for alternate FFTs.  It's the default anyway.
- It wasn't finding any of the HIP header files.  The fix was just to add this line to CMakeLists.txt:

```
INCLUDE_DIRECTORIES("${HIP_INCLUDE_DIR}")
```

- The following line in HipContext.cpp failed to compile, so I removed it.  `gcnArch` seems to no longer exist in the latest version of HIP.

```c++
int major = thisDevice.gcnArch;
```

Once I made those changes it was able to compile, but running a test case failed with this error:

```
exception: Error launching HIP compiler: 256
clang: error: cannot find ROCm device library; provide its path via '--rocm-path' or '--rocm-device-lib-path', or pass '-nogpulib' to build without ROCm device library
```

I'm not sure whether this means I'm missing a library or something has to be done to let it find the libraries.  Here are the relevant conda packages installed in my environment.

```
hip-devel                 6.0.2                ha770c72_2    conda-forge
[...]

PR #4577, opened 2024-06-30
Adding updateSomeParametersInContext to optimise parameter updates

chryswoods:
This PR adds in `updateSomeParametersInContext` to complement the existing `updateParametersInContext` functions of the Force classes (with a corresponding `copySomeParametersToContext` added to the force kernels, complementing `copyParametersToContext`).

The use case for this is to optimise when only a small number of parameters are changed, i.e. changing forcefield parameters of just the ligand during a free energy simulation. Currently, `updateParametersInContext` would update the parameters of all atoms, copying all of them via `copyParametersToContext`. This is O(n_atoms), and was a bottleneck for how we've implemented free energy calculations in [somd2](https://github.com/openbiosim/somd2).

The `updateSomeParametersInContext` function lets the caller provide the hint of which contiguous range of parameters have changed, enabling the implementation to optimise what needs to be copied (via `copySomeParametersInContext`) and what needs updating.

This reduces the cost of updating parameters to O(n_updated_atoms), which significantly speeds things up.

To reduce code duplication, this PR changes `updateParametersInContext` and `copyParametersInContext` to be special cases, calling `updateSomeParametersInContext` and `copySomeParametersInContext` hinting that all parameters have changed (e.g. passing 0 to natoms as arguments for the atom-based forces).

I have implemented the optimised logic of `updateSomeParametersInContext` for most of the forces for the major platforms (CPU, OpenCL, CUDA). I have provided a stub for the missing forces that ignores the hint (i.e. calls the original `updateAllParametersInContext`). I can fill in the missing forces if desired - I didn't want to go too far in case you wanted structural changes made to this optimisation.

Of note, the range given to `updateSomeParametersInContext` is a hint - i.e. the implementation is free to still update all parameters. The aim is that the code has an idea of what has changed, so that it can optimise the update if that makes sense.

We've been using this in testing and development for about six months, and it has been very stable.

Thanks for taking a look :-)


peastman:
Interesting.  Do you have any benchmarks showing the effect of the change?

I have some concerns about the implementation.  For one thing, it puts a lot of burden on the user.  It introduces new functions in lots of classes, and users have to change their code to use them or they don't get any benefit.  If they do use the new functions, that creates new opportunities to make mistakes.  The design of the functions makes it very easy to get things wrong, if the values you pass for `start` and `count` don't match the changes you just made.  And if you make a mistake, the results are particularly bad: it silently behaves incorrectly in ways that may be very difficult to detect.

It also puts a lot of burden on us, or rather on anyone implementing a force.  It creates new opportunities to make mistakes, and again the consequence of a mistake is that you silently get wrong results.  That isn't just a hypothetical.  I've only done a quick skim through the code changes, but I already spotted a bug of this sort.  Your implementation of `copySomeParametersToContext()` for NonbondedForce only calls `invalidateMoleculesByAtom()`, not `invalidateMoleculesByGroup()`.  So if the user modifies exception parameters, the molecules don't get invalidated and they get wrong results.  This design makes it very easy for the programmer to get things wrong.

So I'd like to see if we can find a design that gets some of the benefit without the downsides.  There are three distinct pieces that you're optimizing:

1. Spend less time in `copyParametersToContext()` by only retrieving parameters for a subset of atoms or bonds.
2. Spend less time in `invalidateMolecules()` by only examining a subset of parameters.
3. Reduce communication bandwidth by only uploading part of the array to the GPU.

How much of the benefit comes from each of these?  Are all of them important, or does most of it come from just one or two?

chryswoods:
Thanks for looking so quickly at this - and sorry for my slow reply (work on Isambard kept me busy all week, so I've only had this weekend to respond).

Yes, @akalpokas has been running lots of benchmarks in the context of somd2 free energy simulations. I still need to create a pure OpenMM benchmark so that you can test this more easily.

For example, on a modern CUDA GPU (@akalpokas can give details), for a 66k atom solvated protein-ligand system we see a significant speed up of 2-4x in nanoseconds per day when calling `updateSomeParametersInContext` versus `updateParametersInContext`. This is best shown in these tables (@akalpokas can send more with more detail).

Here, the columns show the number of lambda windows evaluated at each energy evaluation, while the rows are the frequency of energy evaluations. For example, 0.1ps / 3 would be evaluating the energy at 3 lambda values every 0.1ps. As you move towards lower numbers of lambda windows and higher frequencies, you would expect the simulation speed to trend towards the raw "pure MD" speed.

For this system, the pure MD speed (no lambda, just a standard set of forces) is 1070 ns/day.

First, using standard OpenMM with `updateParametersInContext` using RF.

| Frequency | 3     | 5     | 7     | 19    |  49   | 99    |
|-----------|-------|-------|-------|-------|-------|-------|
| 0.1ps     | 60.2  | 38.1  | 27.5  | 9.8   | 4.0   | 2.0   |
| 1 ps      | 398.4 | 282.5 | 220.5 | 90.1  | 38.8  | 19.7  |
| 10 ps     | 905.1 | 824.9 | 765.3 | 509.5 | 291.0 | 168.9 |

Now using the optimised code with `updateSomeParametersInContext` using RF.

| Frequency | 3     | 5     | 7     | 19    |  49   | 99    |
|-----------|-------|-------|-------|-------|-------|-------|
| 0.1ps     | 244.0 | 165.1 | 124.4 | 48.0  | 20.4  | 9.9   |
| 1 ps      | 785.0 | 679.4 | 599.1 | 337.1 | 171.6 | 94.0  |
| 10 ps     | 1016  | 994.1 | 977.2 | 867.8 | 694.2 | 518.2 |

And now using standard OpenMM using PME (raw MD speed here is 303 ns / day).

| Frequency | 3     | 5     | 7     | 19    |  49   | 99    |
|-----------|-------|-------|-------|-------|-------|-------|
| 0.1ps     | 52.9  | 33.9  | 25.3  | 9.8   | 3.9   | 1.9   |
| 1 ps      | 205.2 | 169.1 | 143.3 | 73.3  | 34.7  | 18.2  |
| 10 ps     | 288.5 | 279.8 | 272.0 | 230.0 | 171.3 | 118.7 |

Finally, using the optimised code using PME.

| Frequency | 3     | 5     | 7     | 19    |  49   | 99    |
|-----------|-------|-------|-------|-------|-------|-------|
| 0.1ps     | 149.2 | 112.9 | 89.7  | 38.9  | 17.2  | 8.8   |
| 1 ps      | 273.1 | 257.4 | 243.7 | 181.2 | 113.8 | 70.3  |
| 10 ps     | 299.1 | 297.2 | 295.0 | 282.5 | 259.3 | 266.2 |

This is particularly useful at the regimes where we would be running free energy calculations, e.g. update every picosecond with 3-5 lambda windows. In this regime, we only lose ~10-20% of speed running a free energy simulation versus a pure MD simulation.

This also make higher frequency updates practical, i.e. 244 ns/day for a 0.1ps frequency is usable, while 60 ns/day is too big a performance hit.

Because the optimisation turns an O(n_total_parameters) update into a O(n_changed_parameters) update, we expect it be even more necessary as system sizes get larger.

Yes, there is an implementation burden, but I have tried to minimise this by making `updateSomeParametersInContext` effectively optional - i.e. the vanilla implementation is to update all parameters regardless of the hint. This PR updates all the forces so that this is the implementation. A force developer only has to do work if they want to add an optimised implementation of `updateSomeParametersInContext`. I have already done this work for all of the common forces. As such, this PR doesn't add any extra work now for any force developers of the bundled forces. For users, these are new functions so don't force any changes on existing developers. I was keen not to change behaviour or risk any change in behaviour of existing code.

I agree, it does add a risk that a user of `updateSomeParametersInContext` could get things wrong, which could silently cause issues. However, the hint is simple to add - just the start index and number of parameters to update - so this is an easy thing for a developer to track if they are going to make use of this function (the developer is in control of which parameters they update). The documentation for this function could include warnings that this is an optimisation function, and it is important to correctly specify the range of parameters to update. Given the large speed-ups possible, and balancing the environmental cost of running simulations for 2-4x longer on a GPU without this optimisation, I think this risk is manageable.

Longer term, the risk could be mitigated by adding in code that tracked when parameters were changed, and automatically kept track of the contiguous range of parameters that would need to be updated. Then, `updateParametersInContext` could call `updateSomeParametersInContext` automatically with the right values. But, this would be a larger change and would have more of an implementation burden. It would also require a lot of testing to be sure that the contiguous range of parameters to update is correctly automatically detected.

All three things highlighted are tackled by this optimisation, and all three were quite slow (and all O(n_parameters) rather than O(n_changed_parameters)). The slowest was anything that involved moving data between CPU and GPU memory, and so only updating part of the array was the biggest speed improvement. But, fetching the parameters and invalidating parameters was also pretty slow when I did the initial benchmarking, and would become a bottleneck as system size increases (due to their O(n_parameters) scaling). When I looked at different options, letting the user provide a hint to limit the update to only a contiguous block of parameters was the most efficient and required the least disruption to other code.

Yes, I did get it wrong with `invalidateMoleculesByAtom()` rather than `invalidateMoleculesByGroup()` and will change that.

peastman:
I suspect a lot of the speedup comes from not validating the exceptions.  That part is still linear in the total number of parameters.  And there tend to be more exceptions than particles, so that's the more expensive part.  Conversely, if we can find a way to properly optimize that too (see below), there should be room for further speedups.

> Longer term, the risk could be mitigated by adding in code that tracked when parameters were changed, and automatically kept track of the contiguous range of parameters that would need to be updated.

Those are the lines I was thinking along.  It has to be done carefully: it shouldn't start tracking changes until you create a Context, and then it needs to account for the possibility of multiple Contexts being created from the same System.  But I believe it can be done safely, and it avoids adding any new opportunities for the user to make mistakes.  That's very important: we can test our code thoroughly, but we can't assume users will do the same for their code.

It also has the benefit that we can automatically track multiple types of changes.  For example, NonbondedForce can track changes to both particles and exceptions.  Then we just need to add four new arguments to `copyParametersToContext()`: `firstParticle`, `lastParticle`, `firstException`, and `lastException`.

peastman:
I decided to see if I could come up with a more automated approach.  I have a proof of concept.  It's only implemented for CustomNonbondedForce, and it doesn't optimize that much yet, but it's enough to give the idea.  Take a look at https://github.com/peastman/openmm/commit/2b56a184c1b160e9bd392178a22b66e4a96488d4 and see what you think about the approach.

Each CustomNonbondedForce keeps count of how many Contexts have been created for it by incrementing a counter in `createImpl()`.  Once the first one has been created, `setParticleParameters()` starts tracking the range of particles that have been modified.  `updateParametersInContext()` passes that information along for whatever optimizations the platform chooses to do with it.  At that point, if only one context has been created, it resets the information about modified particles since we know the only existing context is up to date.  If multiple contexts have been created for the force it can't do that, but situations where that matters are likely to be rare.

chryswoods:
That's a good idea - it prevents you having to try and track this at the context level (which would be difficult). It gives you a nice way to special case when there is a one-to-one mapping from Force to context.

peastman:
I added some more optimizations, so it now optimizes retrieving the changed values and uploading them to the GPU.  If this looks ok to you, I'll copy the changes to some of the other widely used force classes and open a PR for it.

I want to put off making any changes to NonbondedForce until #4199 is done.  It will require changes to platform specific code, and I don't want to create conflicts that will further complicate merging the HIP platform.

peastman:
This is superseded by #4610, which I just merged.

PR #4585, opened 2024-07-06
addHydrogens() allows specifying exactly what hydrogens to add

peastman:
When calling `addHydrogens()`, there is a new option to specify exactly what hydrogens should be present for a residue.  This is needed for some new features I'll be adding to PDBFixer.


peastman:
The build failures are unrelated to this PR.  The PPC one is caused by CentOS 7 reaching end of life.  The docs one is from the URL for the GLYCAM docs having changed.  I'm not sure what's going on with the Mac one.  I'll debug them in another PR.

PR #4588, opened 2024-07-09
Debug CI failures

peastman:
Some of the builds have started failing.  I'm trying to get them working again.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -471,7 +471,7 @@ jobs:
             compilers: compilers
 
           - name: PowerPC CPU Python 3.11 devtoolset-7 compilers
-            docker-image: quay.io/condaforge/linux-anvil-ppc64le-cuda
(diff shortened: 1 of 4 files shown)


peastman:
Am I going crazy?  The line

```
sudo yum install -y centos-release-scl
```

runs without problem.  But the very next line

```
sudo sed -i.bak 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-*
```

fails with the error

> sudo: no tty present and no askpass program specified

Why is sudo happy to run without asking for a password on the first line, but fails on the second one if it can't ask for a password?  Am I misinterpreting what the error message means?

peastman:
@mikemhenry do you have any insight on what's going on?

peastman:
I decided to just remove that build.  The proper fix would be to switch to a different docker container.  But given how little PowerPC is used these days, I don't think it's worth the effort.  We don't really need to run two different PowerPC builds with different compilers.

Issue #4601, opened 2024-07-25
scipy 1.14 breaks loading of AMBER NetCDF restart files


timpalpant:
`NetCDFFile` is no longer exported in scipy 1.14 (as a result of https://github.com/scipy/scipy/pull/19904). This breaks the import [here](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/internal/amber_file_parser.py#L1443) when attempting to load an AMBER NetCDF restart file. It may need to be updated with an additional case such as:
```python
from scipy.io import netcdf_file as NetCDFFile
```

A simple repro is to run the benchmark script:
```bash
~/src/openmm/examples$ conda create -n openmm openmm scipy
~/src/openmm/examples$ conda activate openmm
~/src/openmm/examples$ python benchmark.py --platform=Reference --test amber20-dhfr

timestamp: 2024-07-25T13:54:25.497144
openmm_version: 8.1.2.dev-440a9c7
cpuinfo: AMD Ryzen Threadripper PRO 5975WX 32-Cores
cpuarch: x86_64
system: Linux
Traceback (most recent call last):
  File "/home/tim/conda/envs/openmm/lib/python3.12/site-packages/openmm/app/internal/amber_file_parser.py", line 1443, in __init__
    from scipy.io import NetCDFFile
ImportError: cannot import name 'NetCDFFile' from 'scipy.io' (/home/tim/conda/envs/openmm/lib/python3.12/site-packages/scipy/io/__init__.py). Did you mean: 'netcdf_file'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/tim/conda/envs/openmm/lib/python3.12/site-packages/openmm/app/internal/amber_file_parser.py", line 1447, in __init__
    from scipy.io.netcdf import NetCDFFile
ImportError: cannot import name 'NetCDFFile' from 'scipy.io.netcdf' (/home/tim/conda/envs/openmm/lib/python3.12/site-packages/scipy/io/netcdf.py). Did you mean: 'netcdf_file'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/tim/conda/envs/openmm/lib/python3.12/site-packages/openmm/app/internal/amber_file_parser.py", line 1537, in readAmberCoordinates
    crdfile = AmberNetcdfRestart(filename)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4602, opened 2024-07-25
Fix import of netcdf_file for compatibility with scipy 1.14

timpalpant:
In scipy=1.14 the `NetCDFFile` alias is no longer exported. Instead, use `from scipy.io import netcdf_file` which looks to be compatible [as far back as scipy=0.6](https://github.com/scipy/scipy/blob/maintenance/0.6.x/scipy/io/__init__.py#L13). (I tested as far back as scipy=1.0 which is the earliest I could install with conda).

Fixes https://github.com/openmm/openmm/issues/4601


peastman:
Thanks!  I started the CI running.

peastman:
Looks good and tests passed.  Thanks!

Issue #4421, opened 2024-02-02
Monte Carlo simulations


peastman:
A question that occasionally comes up is how to implement Monte Carlo simulations.  That's a broad category that can mean a lot of different things.  It's worth considering whether we can add features or create a generic MC framework that makes at least some useful subset of MC methods easier.

The simplest set of MC methods are those that modify only particle positions.  For example, you might sample dihedral rotations.  Or you might translate water molecules into and out of the binding site.  Move sets can be complex, involving multiple distinct types of moves that can be attempted.  I don't know whether it would be useful to support just a few common move types, or whether we really need to allow users to define arbitrary moves.

Another class of MC methods are ones that modify global parameters, either in isolation or together with positions.  That's what MonteCarloBarostat does: it modifies both the box volume and the particle positions.  Another example would be an alchemical method that uses MC to vary the lambda parameter.

A more complicated class is methods that modify force field parameters or other aspects of the System.  An example would be constant pH methods, which modify per-particle parameters at a large set of variable sites.  (Note that we already have two open issues about constant pH, #172 and #1278.  That case may be unique enough that it's worth handling through its own mechanism.)

Getting still more complicated, a single Monte Carlo move can actually be a complex process involving many internal steps.  An example is NCMC methods that modify the simulation in some way, perform a short simulation, and then if the step is rejected roll back to the start of the whole process.

In some cases, Monte Carlo is the complete simulation protocol.  You perform a series of steps to sample a distribution.  In other cases, it is combined with MD for a compound simulation.  That is the case in a constant pH simulation, for example: you perform a standard MD simulation, but also intersperse dynamics with MC moves.

So there are a lot of possibilities to consider.  Just about any of these things can already be implemented in a custom Python script.  The question is whether it would be beneficial to provide standardized tools for some of them.  There are two main reasons we might want to do it.  One is to make it easier.  If lots of users are writing similar code to do similar things, we can save them effort by providing the code.  The other reason would be to improve efficiency.  Doing lots of work in Python is slow.  If we can perform the moves right on the GPU, it could potentially make simulations a lot faster.

Does anyone have thoughts on this?  Would you be interested in Monte Carlo features if they existed?  What applications would you use them for?  What features would be useful to you?

Gowindyer:
A great topic.For me, It would be useful to change the system topology, such as forming or breaking bond connections through Monte Carlo (MC) features.

amin-sagar:
[...]

PR #4610, opened 2024-07-31
Optimize updateParametersInContext()

peastman:
This is an alternative to #4577.  It makes similar optimizations but in a way that requires nothing from the user.  It automatically tracks what parameters have changed and optimizes updates based on that information.

This involves changes to platform-specific code.  To avoid causing problems for @ex-rzr, I'd like to hold off on merging it until the HIP platform has been merged (see #4199).

To test the effect of this, I simulated ApoA1 on a RTX 4080.  I tested three different versions:

- Not updating any parameters
- Updating the parameters for 20 particles in the NonbondedForce every 10 steps
- Updating the parameters for 20 bonds in the HarmonicBondForce every 10 steps.

Here are the speeds in ns/day for the original code and with this optimization.

|Test|Original|This PR|
|---|---|---|
No updates|604|604|
Update particles|261|373
Update bonds|262|427

cc @chryswoods


chryswoods:
Thanks - we'll try this out and let you know how we get on.

chryswoods:
Thanks @peastman - the new code is really fast and works perfectly. It is faster than the approach I'd proposed, at least on my Mac Studio. Audrius will run some benchmarks on an Nvidia GPU once he is back from leave, but I expect that will also show your code is faster.

It's also great that this is automatic, and doesn't need any code support from the user side.

Here's the results of the benchmarks I ran - this was using RF on a 98.5k atom system, with a 48 atom ligand that was being perturbed, with a 4fs timestep - units all ns/day

devel = current OpenMM devel branch
4610 = this PR
4577 = updateSome PR

Calculate dE every 10ps

| Code   | 3 windows | 7 windows | 9 windows |
|--------|-----------|-----------|-----------|
| devel  |   125.9   |   124.4   |   123.5   |
| 4610   |   132.5   |   132.0   |   131.9   |
| 4577   |   132.5   |   132.4   |   129.6   |

Calculate dE every 1ps

| Code   | 3 windows | 7 windows | 9 windows |
|--------|-----------|-----------|-----------|
| devel  |   116.6   |   102.6   |    97.3   |
| 4610   |   125.6   |   112.2   |   109.3   |
| 4577   |   131.8   |   123.1   |   114.2   |

Calculate dE every 0.1ps

| Code   | 3 windows | 7 windows | 9 windows |
|--------|-----------|-----------|-----------|
| devel  |    60.9   |    35.9   |    29.2   |
| 4610   |   103.8   |    80.8   |    72.1   |
| 4577   |   118.4   |   103.5   |    97.2   |

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -1079,18 +1079,32 @@ void CudaCalcNonbondedForceKernel::copyParametersToContext(ContextImpl& context,
...
+        ewaldSelfEnergy = 0.0;
+        if (nonbondedMethod == Ewald || nonbondedMethod == PME || nonbondedMethod == LJPME) {
+            if (cu.getContextIndex() == 0) {
+                for (int i = 0; i < force.getNumParticles(); i++) {
+                    ewaldSelfEnergy -= baseParticleParamVec[i].x*baseParticleParamVec[i].x*ONE_4PI_EPS0*alpha/sqrt(M_PI);
+                    if (doLJPME)
+                        ewaldSelfEnergy += baseParticleParamVec[i].z*pow(baseParticleParamVec[i].y*dispersionAlpha, 6)/3.0;
+                }
+            }
+        }
```
ex-rzr, line 1102: > but was not able to see any significant differences between the two codes

At first glance this could be the main difference with #4577 

The whole `baseParticleParamVec` is allocated and filled, then only a part of it used for uploading but the whole vector  - to calculate energies if needed.

On the other hand #4577 creates a small vector only for uploading and calculates energies by calling force.getParticleParameters for all atoms.

Considering that the allocation and filling of the whole vector takes time, but the benchmark (done by @akalpokas) uses RF (`(nonbondedMethod == Ewald || nonbondedMethod == PME || nonbondedMethod == LJPME)` is false, right?), this time is wasted.

peastman: Be aware that #4577 has a bug: it doesn't revalidate the exceptions for a NonbondedForce.  I suspect that's where a lot of its apparent speedup comes from.  At the end of `copySomeParametersToContext()` where it calls `invalidateMoleculesByAtom()`, add a call to `invalidateMoleculesByGroup()` immediately after it.  How does that affect the above numbers?

akalpokas: I have reran the benchmark with your suggested additions and the performance between the two implementations is now essentially identical (please see my updated original comment) - initial addition of `invalidateMoleculesByGroup()` to `platforms/cuda/src/CudaKernels.cpp` only regressed the performance of 4577 by around 3%.  Once I updated `platforms/common/src/CommonKernels.cpp` as well, the performance between both implementations essentially equalized.

WillGPoole:
Sorry I am a bit late. I've not done too much extensive benchmarking with it but I can see a marked improvement in my workflows. Great job! (#4421)

PR #4622, opened 2024-08-20
Support numpy 2

### LINKED ISSUE #4494: NumPy 2.0 support
I didn't find another discussion on this - are there plans to support 2.0? AFAICT, OpenMM doesn't hook deeply into the low-level stuff, so it might not be an onerous transition. 
More info: https://numpy.org/devdocs/numpy_2_0_migration_guide.html

There is now a 2.0.0rc1 releases and built (`conda-forge/label/numpy_dev` and https://pypi.org/project/numpy/2.0.0rc1/). I could blindly fire off the CI runs into the void, but I want to ask first.

peastman:
Fixes #4494.

### DIFF
--- a/devtools/ci/gh-actions/conda-envs/build-macos-13.yml
+++ b/devtools/ci/gh-actions/conda-envs/build-macos-13.yml
@@ -10,7 +10,7 @@ dependencies:
 - python
 - cython
 - swig
-- numpy <2.0
+- numpy
 - doxygen 1.8.14
 # test
 - pytest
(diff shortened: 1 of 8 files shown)


peastman:
The Pypy build is also failing on other PRs, so it isn't related to this change.

mattwthompson:
Not sure if you have a large Pypy userbase (nor if that's something that can be known) but it's being phased out of `conda-forge` world soon, so the testing/packaging side of things would need to be updated at some point in the future. I don't think the project itself is dying off, so some PyPI-based tooling should still work

https://github.com/conda-forge/conda-forge.github.io/issues/2255

peastman:
Thanks for the FYI.  I keep hoping Pypy will become more widely used, but mostly the community seems to be sticking with CPython.  It's also possible the need for it will decrease with time, if the performance of CPython gets good enough.

PR #4630, opened 2024-08-24
Fix formal charge reading from pdb files; store formal charges in topology

aizvorski:
This pull request fixes some bugs with formal charge handling in pdb files, and also stores the charges in the resulting topologies.

How charges are stored, from the PDB format spec https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM

> Columns 79 - 80 indicate any charge on the atom, e.g., 2+, 1-. In most cases, these are blank

The charges are not stored as +1 or -1, rather the sign follows the number.  I have confirmed this using some example files saved from popular software.

The existing code before this patch tries to parse this with `try: self.formal_charge = int(pdb_line[78:80])` which always fails, both for missing charges and for present, correctly formatted charges, because `int('1+')` fails.  The correspnding catch statement sets the charge to None.

The current fix reverses the string before trying to parse it as int.

The second part of this pull request makes it possible to store formal charges in the topology, as follows:
- add an optional formal_charge argument to the Atom constructor
- add an optional formal_charge argument to topology.newAtom()
- pass charges from the PdbStructure to the Topology

Test case, using the attached file:

```
pdb_file = 'formal_charge.pdb'
# check that PdbStructure reads the charges correctly - this always prints None before this patch
from openmm.app.internal.pdbstructure import PdbStructure
fh = open(pdb_file, 'r')
pdb = PdbStructure(fh, load_all_models=True, extraParticleIdentifier='EP')
for chain in pdb.iter_chains():
    for residue in chain.iter_residues():
        for atom in residue.atoms_by_name.values():
            print(atom.get_name(), atom.formal_charge)

# check that Topology gets the charges - this would raise an exception before this patch, because a.formal_charge doesn't exist
import openmm as mm
import openmm.app as app
pdb = app.PDBFile(pdb_file)
for a in topology.atoms():
    print(a, a.formal_charge)
```


aizvorski:
Use the following test input formal_charges.pdb
```
ATOM    320  N   LYS A  39      -4.106  14.368  -3.912  1.00 11.33           N  
ATOM    321  CA  LYS A  39      -4.916  15.220  -3.037  1.00 11.28           C  
ATOM    322  C   LYS A  39      -4.213  15.550  -1.713  1.00 10.47           C  
ATOM    323  O   LYS A  39      -4.744  16.309  -0.904  1.00 10.67           O  
ATOM    324  CB  LYS A  39      -6.270  14.555  -2.753  1.00 11.52           C  
ATOM    325  CG  LYS A  39      -7.053  14.165  -3.999  1.00 12.35           C  
ATOM    326  CD  LYS A  39      -8.418  13.593  -3.654  1.00 12.90           C  
ATOM    327  CE  LYS A  39      -9.133  13.063  -4.889  1.00 13.68           C  
ATOM    328  NZ  LYS A  39      -9.339  14.124  -5.910  1.00 14.91           N1+
```

peastman:
I guess just reversing the string is a way of doing it!  It assumes the charge is only a single digit, but since OpenMM isn't generally used to simulate superheated plasmas of heavy elements, I think that's a reasonable assumption.

I wonder if we should also accept `-1`, since it's very likely some people will write files that format it that way?  Check whether the string ends with `+` or `-`, and if it does move that character to the start.

Can you add a test case that reads a PDB file and checks the formal charges?

What about also supporting this for PDBx/mmCIF?  It's stored in [atom_site.pdbx_formal_charge](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.pdbx_formal_charge.html).

The attribute of the Atom class should be called `formalCharge` rather than `formal_charge`.  The public API uses camel case for consistency with the C++ API.  (I know, PdbStructure uses a different convention, but since it's an internal class I haven't bothered to change it.)

aizvorski:
@peastman Could you have a look when you have a chance?  I made a few changes, it should be about ready now

> I guess just reversing the string is a way of doing it! It assumes the charge is only a single digit, but since OpenMM isn't generally used to simulate superheated plasmas of heavy elements, I think that's a reasonable assumption.

Yes, the spec doesn't explicitly say so but the examples are two characters with a sign at the end, so I think what it can handle is only '9-' to '9+'

> I wonder if we should also accept `-1`, since it's very likely some people will write files that format it that way? Check whether the string ends with `+` or `-`, and if it does move that character to the start.

I modified the reader to support any flavor of '+1', '1+' or '1 ' or ' 1'

The writer is spec-compliant

> Can you add a test case that reads a PDB file and checks the formal charges?

Added test :+1: 

> What about also supporting this for PDBx/mmCIF? It's stored in [atom_site.pdbx_formal_charge](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.pdbx_formal_charge.html).

Definitely, that would be good.  I'll have to find a knonw-good example file like that.  Maybe separate PR?

> The attribute of the Atom class should be called `formalCharge` rather than `formal_charge`. The public API uses camel case for consistency with the C++ API. (I know, PdbStructure uses a different convention, but since it's an internal class I haven't bothered to change it.)

Changed :+1:

peastman:
Looks great.  Thank you!

Issue #4623, opened 2024-08-21
Wave64 of `findInteractingBlocks` is missing support for `USE_LARGE_BLOCKS`


bdenhollander:
It looks Wave64 support was accidentally missed during #4147 so the number of parameters expected by OpenCL is different when running on GCN or CDNA GPUs. I don't see any code in OpenCLNonbondedUtilities.cpp that checks SIMD width and disables `useLargeBlocks` so C++ is generating parameters that do not exist in the kernel code.

Wave32 used by RDNA:

https://github.com/openmm/openmm/blob/75d4f2996a053e444cc8aa4230f1fece1a3df7a1/platforms/opencl/src/kernels/findInteractingBlocks.cl#L157-L166

Wave64 used by GCN and CDNA:

https://github.com/openmm/openmm/blob/75d4f2996a053e444cc8aa4230f1fece1a3df7a1/platforms/opencl/src/kernels/findInteractingBlocks.cl#L575-L580

This was noticed by a FAH tester on a Radeon VII GPU. If needed, I can test a fix on an RX 460.

peastman:
Thanks for reporting it.  That should be an easy fix.

peastman:
The fix is in #4629.  If you can test it, that would be great.  I don't have a GPU with width 64 to test on.

bdenhollander:
Adding the arguments fixes the `clSetKernelArg` error. However, I'm seeing a lot of particle coordinate is NaN errors on master. I tried git blame twice from different starting points and it gave different commits as the first bad one, neither of which made sense. 8.1.0 seems to be running fine so I will test 8.1.2 and retry git blame next week.

bdenhollander:
I confirmed 8.1.2 runs fine on GCN. Running git bisect again, from the 8.1.0 tag on master, https://github.com/openmm/openmm/commit/796ffaaa97decdf1593ed75afa5fbb21b6b32e9c is the first bad commit. If you have any suggestions to test, please let me know. Also, you may be able to run your 5700 XT in Wave64 mode by setting the environment variable `GPU_ENABLE_WAVE32_MODE=0`.

peastman:
> Also, you may be able to run your 5700 XT in Wave64 mode by setting the environment variable GPU_ENABLE_WAVE32_MODE=0.

Do you know what effect that variable is supposed to have?  `device.getInfo<CL_DEVICE_WAVEFRONT_WIDTH_AMD>()` still returns 32.

bdenhollander:
[...]

PR #4632, opened 2024-08-28
HIP platform

ex-rzr:
Related issue: #4199

I have rebased commit of the HIP plugin, reordered and combined commits by their purpose:
* the first commits contain mainly "hipification" of the CUDA platform and plugins with some compatibility changes;
* commits with various optimizations combined by the subsystem (nonbonded, find interactions etc.);
* commits with changes ported from the CUDA/Common platform;
* the last 3 commits with optimizations of PME and bonded forces (I want to discuss them further, see below).

I think this is a good compromise between a bunch of tiny-to-large commits and one huge commit (please do not squash if possible:))

Back to 3 last commits.

The PME optimizations consists of two commit. "Copy 3 kernels from pme.cc to HIP for future optimizations" copies the kernels without modifications, so the changes of "Optimize PME spread and interpolate kernels" are better visible. The main idea is to process one atom with 5 (PME_ORDER) threads using cross-lane operations, from the commit message:
* PME_ORDER lanes process one atom;
* PME_ORDER lanes access consecutive addresses;
* No need to permute z indices with zindexTable;
* Each component (x, y, z) of thetas is calculated by one lane;

This change improves overall performance of benchmarks with PME by 4-14% (depending on the GPU and the benchmark).
I think, this optimization can be ported to CUDA but I haven't tried this yet.
@peastman, are interested in it? Can we have it as it is or should it be moved to common if it will work well on CUDA too (and what if it will not)? Currently it has kind of an opposite direction than your work of moving code from all platforms to the Common platform.
This change is independent and can be moved to another pull request and discussed there.

The last commit is with an optimization of bonded forces. Atom indices for many forces are usually grouped so threads of one warp can write forces to the same index, so there are atomic conflicts within the warp. 
My idea is to do a reduction of consecutive segments of forces for the same atom index (using head segmented reduction). See [conflicts-with-atoms.txt](https://github.com/user-attachments/files/16785576/conflicts-with-atoms.txt), there are atom indices of 128 bonds for each force and some statistics: avgConflicts is the number of such conflicts that can be eliminated with head segmented sum.
This optimization is implemented for HIP and then ported to CUDA. On AMD GPUs it improves overall performance by 2-5% (except for gbsa and rf with small drop of ~0-1%).
For CUDA I've tested on Titan V, 3090, 4070 Ti Super: 3090 has +1-2% but -2% for gbsa, Titan V  has +2-3% but also -2% for gbsa, 4070 Ti Super has the biggest change +2-6% (6% for apoa1). I don't have access to older architectures or more low-end GPUs (2060 on my desktop show +-3% random performance fluctuations so it's impossible to benchmark on it reproducibly).
I don't know if it's worth having this optimization for CUDA (the `#elif defined(__CUDA_ARCH__) ...` part can be easily removed/commented), on AMD it seems to provide a bit more consistent improvement, it would be great to keep it. Also I'm not sure if you are ok with having a platform specific kernel code in the Common platform. I look forward to your decision.
This change is also independent and can be moved to another pull request and discussed there.

Another change that can be ported to other platforms (OpenCL) is caching of VkFFT kernels. In the HIP platform I use the code for hashing from HipContext that is used for the regular kernels. This change significantly improves "performance" of tests and initialization of benchmarks with PME.


peastman:
Fantastic, thanks!  I'll start going through the code.  16,000 lines in 140 files may take a little while. :)

I'm definitely interested in the optimizations, but can we move them into separate PRs?  Just reviewing the HIP platform on its own is going to be a mammoth task.  Adding in novel optimizations to review, understand, and test on various hardware will add even more complexity.  If we can separate them out, that will make it more manageable.

ex-rzr:
Done: #4633 and #4634

peastman:
Perfect, thanks!  I'll review this PR first, then move on to those ones after this is merged.

peastman:
What do you think about removing the support for alternate FFTs, and just making it always use VkFFT?  That would eliminate a lot of complexity.  And if we support multiple FFT libraries, that means we need to test all of them, which complicates the test matrix.

peastman:
All the new files have the copyright statement

```
* Portions copyright (C) 2020-2022 Advanced Micro Devices, Inc. All Rights
* Reserved. 
```

"All Rights Reserved" directly contradicts the license that appears immediately below it.  Can we fix this?  Most of these files are derived from LGPL licensed code, which means the new file has to be open source and AMD can't reserve all rights.

ex-rzr:
> What do you think about removing the support for alternate FFTs, and just making it always use VkFFT? That would eliminate a lot of complexity. And if we support multiple FFT libraries, that means we need to test all of them, which complicates the test matrix.

Ok, I don't mind, this sounds reasonable considering that VkFFT is the default one (I'll evaluate performance of the hipFFT backend, just to be sure, it was slower than VkFFT but I don't know its current state).

> "All Rights Reserved" directly contradicts the license that appears immediately below it. Can we fix this?

Do I understand this correctly: "All Rights Reserved" should be removed but "... copyright ...Advanced Micro Devices..." stays? (I didn't add the copyrights, they were in the early hipified version by AMD, hence my question).

Another question, how do you prefer to see the fixes (these two and future ones):
(option 1) Should I add new commits?
(option 2) Or should I edit the existing commits?

peastman:
> Do I understand this correctly: "All Rights Reserved" should be removed but "... copyright ...Advanced Micro Devices..." stays?

Exactly.  They hold a copyright, but license it under the LGPL.

> Another question, how do you prefer to see the fixes (these two and future ones):

Normally I would do a squash and merge, but you want to avoid that, so probably editing the exist ones is better.  Or you could add new commits for the moment, and squash them by hand before we merge.

platforms/hip/src/HipContext.cpp
```diff
@@ -0,0 +1,942 @@
...
+    // HIP-TODO: find a good value here
+    int numThreadBlocksPerComputeUnit = 6;
+
+    if (gpuArchitecture.find("gfx908") == 0 ||
+        gpuArchitecture.find("gfx90a") == 0 ||
+        gpuArchitecture.find("gfx94") == 0) {
+        // MI100 and newer CDNA support 32 bit float atomic add
+        // HIP-TODO: evaluate performance on RDNA 3 (gfx11* and gfx12*), where it's supported too
+        this->supportsHardwareFloatGlobalAtomicAdd = true;
+    }
```
peastman, line 184: Will this need to be regularly updated as new GPUs get released?  I wonder if there's a more future-proof way of determining whether it's supported?

platforms/hip/src/HipContext.cpp
```diff
@@ -0,0 +1,942 @@
...
+    reduceEnergyKernel = getKernel(utilities, "reduceEnergy");
+    setChargesKernel = getKernel(utilities, "setCharges");
+
+    // Set defines based on the requested precision.
+
+    compilationDefines["SQRT"] = useDoublePrecision ? "sqrt" : "__fsqrt_rn";
+    compilationDefines["RSQRT"] = useDoublePrecision ? "rsqrt" : "__frsqrt_rn";
+    compilationDefines["RECIP(x)"] = useDoublePrecision ? "(1.0/(x))" : "(1.0f/(x))";
+    compilationDefines["EXP"] = useDoublePrecision ? "exp" : "__expf";
+    compilationDefines["LOG"] = useDoublePrecision ? "log" : "__logf";
```
peastman, line 276: According to [the documentation](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/math_api.html#floating-point-intrinsics), `__expf()` and `__logf()` return a "fast approximate" value, but it doesn't say what approximation it uses or how accurate it is.  If you haven't already, we should validate them to make sure the accuracy is good enough.

platforms/hip/src/HipNonbondedUtilities.cpp
```diff
@@ -0,0 +1,749 @@
...
+                }
+                else {
+                    // RDNA
+                    if (context.getNumAtoms() < 100000)
+                        maxBits = 4;
+                    else if (context.getNumAtoms() < 500000)
+                        maxBits = 2;
+                    else // Very large systems
+                        maxBits = 0;
+                }
```
peastman, line 580: I assume this was based on profiling to see what was fastest on different GPUs?

platforms/hip/src/kernels/common.hip
```diff
@@ -0,0 +1,50 @@
...
+__device__ inline long long realToFixedPoint(float x) {
+    // More effective way to do static_cast<long long>(x * 0x100000000)
+    int hi = __float2int_rz(x);
+    float rem = (x - __float2int_rz(x)) * 0x100000000;
+    bool isRemNeg = rem < -0.5f;
+    int lo = __float2uint_rn(fabsf(rem));
+    return static_cast<long long>(
+        (static_cast<unsigned long long>(static_cast<unsigned int>(isRemNeg ? hi - 1 : hi)) << 32) |
+        static_cast<unsigned long long>(static_cast<unsigned int>(isRemNeg ? -lo : +lo)));
+}
```
peastman, line 48: Curious.  Is this meant to be more accurate?  Where did it come from?

peastman:
This looks great!  I haven't tried building it yet, but based on a first pass through the code I don't see anything that looks like it will be a problem for merging.  Most of it is a straightforward port of the CUDA platform.  The one big exception is findInteractingBlocks, which is fairly different.  Was that done to get good performance on GPUs with wavefront sizes of both 32 and 64?

We'll need to think about how to test that.  Since both sizes are supported, we'll need to test regularly on both.  Will that require testing on both RDNA and CDNA GPUs, or is it possible to test both modes on a single GPU?  This relates to the discussion in #4623.

Going through this, I notice there's still a lot of code duplicated between the platforms, especially in multi-GPU support and test cases.  Once this is merged, I'll see if I can clean that up and eliminate some more duplicate code.

peastman:
I'm having trouble compiling.  When I tell CMake to generate makefiles, it fails with this error:

```
 CMake Error in platforms/hip/CMakeLists.txt:
   Imported target "hip::host" includes non-existent path

     "/include"

   in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

   * The path was deleted, renamed, or moved to another location.

   * An install or uninstall procedure did not complete successfully.

   * The installation package was faulty and references files it does not
   provide.
```

Am I missing a package?  Here's a list of all the relevant packages in my conda environment.

```
hip-devel                 6.0.2                ha770c72_2    conda-forge
hip-runtime-amd           6.0.2                he02047a_2    conda-forge
hipcc                     6.0.2                he02047a_2    conda-forge
rocm-comgr                6.0.2                h1566cd9_0    conda-forge
rocm-device-libs          6.1.0                ha6fb4c9_0    conda-forge
rocminfo                  6.0.2                hd3aeb46_1    conda-forge
```

ex-rzr:
* Fixed copyrights
* Removed hipFFT and the builtin FFT. VkFFT is faster than or on par with hipFFT in benchmarks, so it's worth using only it, fewer dependencies too

> I'm having trouble compiling. When I tell CMake to generate makefiles, it fails with this error:

Looks like the old issue we had: https://github.com/openmm/openmm/issues/4199#issuecomment-2078606830
Though it should be fixed: https://github.com/conda-forge/hip-feedstock/blob/main/recipe/0004-Fix-hip-config.cmake-when-it-is-not-a-symlink.patch
There are builds `_3` for these packages, I think yours don't have the fix.
I haven't tried to build this branch with conda environment, I'll do it later but please let me know if the new packages work.

I'll answer other questions later.

peastman:
Updating the packages fixed it.  Thanks!  I'm building now.

peastman:
It compiled successfully, and all tests passed.

platforms/hip/src/HipContext.cpp
ex-rzr: Changed

platforms/hip/src/kernels/common.hip
ex-rzr: I've updated the code, added more comments to explain the magic.
It's not more accurate, but a bit faster. AMD GPUs don't have hardware instructions for float-to-(u)int64 conversion, so it's implemented as a few instructions with float-to-(u)int32 conversion (there are hardware instructions for it).

However, it seems that the difference with `static_cast<long long>(x * 0x100000000)` is much smaller with the current compiler than before (this implementation was significantly faster than the code generated by the compiler, supposedly, this part of the compiler was optimized):
+2-3% for cellulose and stmv on MI100 (gfx908), less on RDNA1 (gfx1030).

peastman:
That all sounds good.  I think we might be ready to merge!  Shall I go ahead?  We can clean up details later.

ex-rzr:
Let's do it! :drum:

peastman:
And it is merged!  Thank you so much for your hard work on this.  It's amazing!

PR #4633, opened 2024-08-28
Optimize PME spread charges: process 1 atom by 5 threads

ex-rzr:
The PME optimizations consists of two commit. "Copy 3 kernels from pme.cc to HIP for future optimizations" copies the kernels without modifications, so the changes of "Optimize PME spread and interpolate kernels" are better visible. The main idea is to process one atom with 5 (PME_ORDER) threads using cross-lane operations, from the commit message:
* PME_ORDER lanes process one atom;
* PME_ORDER lanes access consecutive addresses;
* No need to permute z indices with zindexTable;
* Each component (x, y, z) of thetas is calculated by one lane;

This change improves overall performance of benchmarks with PME by 4-14% (depending on the GPU and the benchmark).
I think, this optimization can be ported to CUDA but I haven't tried this yet.
@peastman, are interested in it? Can we have it as it is or should it be moved to common if it will work well on CUDA too (and what if it will not)? Currently it has kind of an opposite direction than your work of moving code from all platforms to the Common platform.

### DIFF
--- a/platforms/cuda/src/CudaKernels.cpp
+++ b/platforms/cuda/src/CudaKernels.cpp
@@ -427,7 +427,8 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
             pmeDefines["GRID_SIZE_Z"] = cu.intToString(gridSizeZ);
             pmeDefines["EPSILON_FACTOR"] = cu.doubleToString(sqrt(ONE_4PI_EPS0));
             pmeDefines["M_PI"] = cu.doubleToString(M_PI);
-            if (cu.getUseDoublePrecision() || cu.getPlatformData().deterministicForces)
+            useFixedPointChargeSpreading = cu.getUseDoublePrecision() || cu.getPlatformData().deterministicForces;
+            if (useFixedPointChargeSpreading)
                 pmeDefines["USE_FIXED_POINT_CHARGE_SPREADING"] = "1";
             if (usePmeStream)
                 pmeDefines["USE_PME_STREAM"] = "1";
@@ -455,7 +456,8 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
                 pmeConvolutionKernel = cu.getKernel(module, "reciprocalConvolution");
                 pmeInterpolateForceKernel = cu.getKernel(module, "gridInterpolateForce");
                 pmeEvalEnergyKernel = cu.getKernel(module, "gridEvaluateEnergy");
-                pmeFinishSpreadChargeKernel = cu.getKernel(module, "finishSpreadCharge");
+                if (useFixedPointChargeSpreading)
+                    pmeFinishSpreadChargeKernel = cu.getKernel(module, "finishSpreadCharge");
                 cuFuncSetCacheConfig(pmeSpreadChargeKernel, CU_FUNC_CACHE_PREFER_SHARED);
                 cuFuncSetCacheConfig(pmeInterpolateForceKernel, CU_FUNC_CACHE_PREFER_L1);
                 if (doLJPME) {
@@ -466,10 +468,9 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
                     pmeDefines["RECIP_EXP_FACTOR"] = cu.doubleToString(M_PI*M_PI/(dispersionAlpha*dispersionAlpha));
                     pmeDefines["USE_LJPME"] = "1";
                     pmeDefines["CHARGE_FROM_SIGEPS"] = "1";
-                    if (cu.getUseDoublePrecision() || cu.getPlatformData().deterministicForces)
-                        pmeDefines["USE_FIXED_POINT_CHARGE_SPREADING"] = "1";
                     module = cu.createModule(CudaKernelSources::vectorOps+CommonKernelSources::pme, pmeDefines);
-                    pmeDispersionFinishSpreadChargeKernel = cu.getKernel(module, "finishSpreadCharge");
+                    if (useFixedPointChargeSpreading)
+                        pmeDispersionFinishSpreadChargeKernel = cu.getKernel(module, "finishSpreadCharge");
                     pmeDispersionGridIndexKernel = cu.getKernel(module, "findAtomGridIndex");
                     pmeDispersionSpreadChargeKernel = cu.getKernel(module, "gridSpreadCharge");
                     pmeDispersionConvolutionKernel = cu.getKernel(module, "reciprocalConvolution");
@@ -481,15 +482,16 @@ void CudaCalcNonbondedForceKernel::initialize(const System& system, const Nonbon
                 // Create required data structures.
 
                 int elementSize = (cu.getUseDoublePrecision() ? sizeof(double) : sizeof(float));
-                int roundedZSize = PmeOrder*(int) ceil(gridSizeZ/(double) PmeOrder);
-                int gridElements = gridSizeX*gridSizeY*roundedZSize;
+                int gridElements = gridSizeX*gridSizeY*gridSizeZ;
                 if (doLJPME) {
-                    roundedZSize = PmeOrder*(int) ceil(dispersionGridSizeZ/(double) PmeOrder);
-                    gridElements = max(gridElements, dispersionGridSizeX*dispersionGridSizeY*roundedZSize);
+                    gridElements = max(gridElements, dispersionGridSizeX*dispersionGridSizeY*dispersionGridSizeZ);
                 }
                 pmeGrid1.initialize(cu, gridElements, 2*elementSize, "pmeGrid1");
                 pmeGrid2.initialize(cu, gridElements, 2*elementSize, "pmeGrid2");
-                cu.addAutoclearBuffer(pmeGrid2);
+                if (useFixedPointChargeSpreading)
+                    cu.addAutoclearBuffer(pmeGrid2);
+                else
+                    cu.addAutoclearBuffer(pmeGrid1);
                 pmeBsplineModuliX.initialize(cu, gridSizeX, elementSize, "pmeBsplineModuliX");
                 pmeBsplineModuliY.initialize(cu, gridSizeY, elementSize, "pmeBsplineModuliY");
                 pmeBsplineModuliZ.initialize(cu, gridSizeZ, elementSize, "pmeBsplineModuliZ");
@@ -890,6 +892,10 @@ double CudaCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeF
 
         // Execute the reciprocal space kernels.
 
+        // With fixed point charge spreading, finishSpreadCharge kernel is not needed as
+        // gridSpreadCharge can write directly to pmeGrid1.
+        CudaArray& pmeSpreadDstGrid = useFixedPointChargeSpreading ? pmeGrid2 : pmeGrid1;
+
         if (hasCoulomb) {
             void* gridIndexArgs[] = {&cu.getPosq().getDevicePointer(), &pmeAtomGridIndex.getDevicePointer(), cu.getPeriodicBoxSizePointer(),
                     cu.getInvPeriodicBoxSizePointer(), cu.getPeriodicBoxVecXPointer(), cu.getPeriodicBoxVecYPointer(), cu.getPeriodicBoxVecZPointer(),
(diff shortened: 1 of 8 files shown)


peastman:
That looks really interesting.  Let's try it for CUDA and see if it's faster there too.  For OpenCL, I guess the difficulty would be that there aren't shuffles, so you would need to use more expensive shared memory operations to transfer values between threads.

What are the reasons for the C++ code changes?  Particularly changing the order of bonded and nonbonded interactions, and changing the priority of the PME stream?

peastman:
I'll try profiling and see if I can figure out what's going on.

peastman:
I needed to force it to regenerate the code for the python wrapper.  I have it working now.  Here's the performance I get on RTX 4080.  For the main branch:

```
pme               single      HBonds        1.5            4         NVT        CUDA       2050.13
apoa1pme          single      HBonds        1.5            4         NVT        CUDA       598.781
amber20-cellulose single      HBonds        1.5            4         NVT        CUDA       157.109
```

And for this PR:

```
pme               single      HBonds        1.5            4         NVT        CUDA       2131.45
apoa1pme          single      HBonds        1.5            4         NVT        CUDA       638.484
amber20-cellulose single      HBonds        1.5            4         NVT        CUDA       159.681
```

It's a small but measurable speedup for the first two (4% for pme, 7% for apoa1pme), but a negligible difference for the larger system (<2% speedup, which is within the noise level).  I'll profile and see if I can tell what's going on.

peastman:
On the other hand, STMV gets slightly slower with the new version (by about 2%).

Profiling shows that this change makes `gridSpreadCharge()` faster (from 153 to 92 us for ApoA1), but makes `gridInterpolateForce()` slower (from 34 to 54 us for ApoA1).

Looking further into the reasons `gridInterpolateForce()` is slower, the memory throughput decreases from 46% to 25%.  The compute throughput increases from 20% to 46%, which seems like a good thing until you look at the instruction statistics and see that the new version executes almost four times as many instructions (4.4 million in the original version, 16.6 million in the new version).  It executes twice as many instructions in each cycle, but has to get through four times as many.

Compare that to `gridSpreadCharge()`.  It has a roughly 2x increase in the number of instructions (from 3.4 million to 7 million), but that's balanced by a 2x increase in compute throughput (from 5.2% to 11.5%).  In any case, this kernel is dominated by memory access rather than compute.  Looking at the memory analysis, the big difference is in the total amount of data transferred to L2: 183 MB in the original version, decreasing to 100 MB in the new version.

peastman:
Here's what I get using the new version of `gridSpreadCharge()` and the old version of `gridInterpolateForce()`.

```
pme               single      HBonds        1.5            4         NVT        CUDA       2151.03
apoa1pme          single      HBonds        1.5            4         NVT        CUDA       658.778
amber20-cellulose single      HBonds        1.5            4         NVT        CUDA       168.298
amber20-stmv      single      HBonds        1.5            4         NVT        CUDA       51.1136
```

In all cases it's faster than either the current code or the version in this PR.  At least for NVIDIA, it looks like that's the way to go.

peastman:
Here is what I get with HIP.

Current code:

```
pme               single      HBonds        1.5            4         NVT        HIP        762.675
apoa1pme          single      HBonds        1.5            4         NVT        HIP        201.408
amber20-cellulose single      HBonds        1.5            4         NVT        HIP        43.2503
amber20-stmv      single      HBonds        1.5            4         NVT        HIP        15.1513
```

This PR:

```
pme               single      HBonds        1.5            4         NVT        HIP        818.492
apoa1pme          single      HBonds        1.5            4         NVT        HIP        216.304
amber20-cellulose single      HBonds        1.5            4         NVT        HIP        47.1901
amber20-stmv      single      HBonds        1.5            4         NVT        HIP        17.1939
```

New `gridSpreadCharge()`, old `gridInterpolateForce()`:

```
pme               single      HBonds        1.5            4         NVT        HIP        818.197
apoa1pme          single      HBonds        1.5            4         NVT        HIP        218.901
amber20-cellulose single      HBonds        1.5            4         NVT        HIP        47.4872
amber20-stmv      single      HBonds        1.5            4         NVT        HIP        17.2838
```

The last version seems to be the fastest for HIP as well, though the difference is smaller than for CUDA.  That's with Radeon RX 5700 XT.

platforms/cuda/src/CudaKernels.cpp
```diff
@@ -902,8 +907,9 @@ double CudaCalcNonbondedForceKernel::execute(ContextImpl& context, bool includeF
                     cu.getInvPeriodicBoxSizePointer(), cu.getPeriodicBoxVecXPointer(), cu.getPeriodicBoxVecYPointer(), cu.getPeriodicBoxVecZPointer(),
                     recipBoxVectorPointer[0], recipBoxVectorPointer[1], recipBoxVectorPointer[2], &pmeAtomGridIndex.getDevicePointer(),
                     &charges.getDevicePointer()};
-            cu.executeKernel(pmeSpreadChargeKernel, spreadArgs, cu.getNumAtoms(), 128);
+            cu.executeKernel(pmeSpreadChargeKernel, spreadArgs, pmeNumThreadBlocks * pmeThreadBlockSize, pmeThreadBlockSize);
```
peastman, line 910: I think a better value for the third argument is `6*cu.getNumAtoms()`.  The goal of that argument is to set a limit on the number of threads, so it doesn't create lots of threads that have nothing to do for very small systems.  The new version of the kernel creates more threads (32 for every 6 atoms, rather than one per atom), so the limit should be higher, but it should still be limited for sufficiently small systems.

peastman:
What do you think about just removing the changes to `gridInterpolateForce()`, since it seems to be slower in all cases?

ex-rzr:
I agree, I've seen your comment with profiling results and analysis -
really interesting, some numbers are even counter-intuitive for me (I would
like to do the same on HIP using omniperf, of course, after the release).
So it should be removed, at least for now.

But unfortunately I can finalize this pull request only in a few days, as I
will return home from my vacation by the end of this week.
Can you wait with the 8.2 release till Monday?

вт, 17 сент. 2024 г., 01:04 Peter Eastman ***@***.***>:

peastman:
Ok, that's fine.  Thanks!

bdenhollander:
Here's another data point from RX 6600 HIP on Windows.

Current code:
```
pme               single      HBonds        1.5            4         NVT        HIP        459.848
apoa1pme          single      HBonds        1.5            4         NVT        HIP        149.004
amber20-cellulose single      HBonds        1.5            4         NVT        HIP        37.9571
amber20-stmv      single      HBonds        1.5            4         NVT        HIP        11.8502
```

This PR:
```
pme               single      HBonds        1.5            4         NVT        HIP        481.151
apoa1pme          single      HBonds        1.5            4         NVT        HIP        159.096
amber20-cellulose single      HBonds        1.5            4         NVT        HIP        39.7321
amber20-stmv      single      HBonds        1.5            4         NVT        HIP        12.9074
```

New `gridSpreadCharge()`, old `gridInterpolateForce()`:
```
pme               single      HBonds        1.5            4         NVT        HIP        483.419
apoa1pme          single      HBonds        1.5            4         NVT        HIP        159.624
amber20-cellulose single      HBonds        1.5            4         NVT        HIP        39.527
amber20-stmv      single      HBonds        1.5            4         NVT        HIP        12.8845
```

The PR version is slightly better on larger systems while the alternate is slightly better on smaller systems. Either version is a 3-8% improvement.

ex-rzr:
Unfortunately, I didn't manage to use omniperf to gather performance counters on AMD (I'll try later again).
I reverted the changes in gridInterpolateForce.
Also I did some experiments with gridSpreadCharge: considering that it is memory bound, there is no significant performance improvement when thetas for x, y and z are calculated by different lanes. I've returned the old code so all threads do the same calculations. As the result, we don't need shuffles (or shared memory) anymore. This version seems to provide a similar performance.

I've run CUDA+OpenCL and AMD+OpenCL tests - no regressions.

@peastman @bdenhollander 
Could you rerun the benchmarks with the new branch? (I've force-pushed it)
If you'll get similar results, we can merge it.

For reference, the old version of this branch is here: https://github.com/ex-rzr/openmm/commits/optimize-pme-backup/

However, I think there are still some opportunities of optimizations. gridInterpolateForce works faster with low occupancy (on AMD). When I change something in the kernel that improves occupancy (like reduce the number of registers), performance drops! This is likely caused by cache thrashing (but this is just my assumption): too many active warps need access different memory areas, so when one thread loads the first of 5 consecutive values, the next values will not be in cache anymore (replaced by values loaded by another thread).
Later I'll try to prove/disprove this hypothesis by gathering some statistics with omniperf. But this will be a separate investigation.

peastman:
Here are the results on an Apple M1 Pro.  It also gets a speedup.

Current code:

```
pme               single      HBonds        1.5            4         NVT        OpenCL     156.729
apoa1pme          single      HBonds        1.5            4         NVT        OpenCL     38.9667
amber20-cellulose single      HBonds        1.5            4         NVT        OpenCL     9.59166
amber20-stmv      single      HBonds        1.5            4         NVT        OpenCL     3.46364
```

This PR:

```
pme               single      HBonds        1.5            4         NVT        OpenCL     179.783
apoa1pme          single      HBonds        1.5            4         NVT        OpenCL     45.0727
amber20-cellulose single      HBonds        1.5            4         NVT        OpenCL     10.7347
amber20-stmv      single      HBonds        1.5            4         NVT        OpenCL     3.8491
```

If you're happy with the code, mark this PR as no longer a draft and I'll merge it.

DaveCerutti:
This is a very interesting result you've got.  I wish you had included me in this discussion.  If you check [the STORMM paper](https://pubs.aip.org/aip/jcp/article/161/3/032501/3303330/STORMM-Structure-and-topology-replica-molecular), in Figure 7 I've got a plot that may look in some ways familiar to what you're reporting.  The optimization you're doing makes OpenMM's implementation more similar to STORMM's "general purpose" charge spreading kernel with fifth-order interpolation.  I don't know how exactly you got five threads working on one atom--what I did was to devote five threads as they appear in sequence, and have each thread issue 25 `atomicAdd()` operations to `__global__`.  In that way, the threads are well occupied up to (number of atoms) x 5, but the work of spreading every 7th atom or so will be split across two warps.  I didn't do your fancy atom sorting before and after, but as you can see with the solid blue dotted line STORMM's 5th order charge spreading (which is the lion's share of the work, as you've noted in this discussion) is performing better than OpenMM's 8.1 release (and 8.2 prior to this PR), with the delta diminishing as the total number of atoms climbs.

Another thing that Figure 7 helps to show is the effect of a small global cache size.  The cellulose and STMV systems start to thrash with the 6MB caches on Titan-V and RTX 3090.  The much larger global caches on RTX 4080 and A100 are  much less prone to such effects.

Finally, and @ex-rzr take a look at the method STORMM actually uses (the global cache method is a fallback, if the user requests some bizarre PME density).  The localized method is making work units to do atomics to `__shared__` rather than `__global__`.  This is MUCH faster, especially on the latest architectures, immune to global cache effects on the prior architectures, and reaches asymptotic performance at around 500k to 1M total atoms.

Here is Figure 7, for convenience (apologies to AIP if this violates policies, but the article is open access):
image

DaveCerutti:
In the grand scheme of things, these days I look at the _total amount of work_.  In a heterogeneous environment that's going to be all that matters.  In that sense, the "local work units" are doing 50%-100% more B-spline computations, to my recollection, because the work units require that they compute B splines for atoms in a "halo" around each region of the grid that the work unit is given exclusive rights to fill (that's how the work can happen in `__shared__`, not L2).  My guess is that, if the new OpenMM method work like my "global accumulation" approach, then there are 400% more B-spline computations going on.  All methods are issuing the same number of `atomicAdd()` operations, which is probably the lion's share of the compute.  But in an environment where there are multiple streams and perhaps MPS or some other method for stacking multiple systems on the same card (see... purpose of STORMM) extra computations are just waste heat, even if they make the one case a little faster.

How would OpenMM implement the local work units approach?  Take the current neighbor list and replace your existing sort with a routine that "sorts" all particles, at each step in their instantaneous positions, into a grid of cells aligned to the PME grid, 4 x 4 x 4 grid points per cell.  It forces you to include a factor of four in your PME grid dimension, but that's what pmemd.cuda already does and nothing wrong in terms of FFT efficiency (it also keeps you away from grid  dimensions of 75, which is very evil, to the tune of +100% FFT compute time).  The work units get write permissions to blocks of grid points in cells Ai...Af x Bi...Bf x Ci...Cf, and must compute B-splines and loop over atoms in cells Ai-1...Af x Bi-1...Bf x Ci-1...Cf.  Works well if you cut the work units to be longest along whatever axis gives you atoms contiguous in memory proceeding from one cell to the next, however you did the sort.

DaveCerutti:
@peastman To orient my reading, I know that "apoa1pme" is the Apo-A1 system with 92,224 atoms,  but is "pme" the familiar DHFR / JAC benchmark (23,558 atoms)?

peastman:
>  In my view, the local work units are pushing towards compute-bound.

The compute throughput is only 11.5%.  It spends almost all its time waiting on memory.

> is "pme" the familiar DHFR / JAC benchmark (23,558 atoms)?

Correct.

DaveCerutti:
The "local work units" are a STORMM thing, not an OpenMM thing, but it just kind of confirms that we're talking past each other here.  I'm here to beat a dead horse on a closed PR, and I'm not here to advertise STORMM, rather to offer an exchange of ideas.  I believe you about the OpenMM method, and by extension STORMM's similar implementation issuing contributions to accumulators in `__global__` cache.  What I'm saying is that there are other ways to set up the problem that make the traffic to and from `__global__` cache effectively zero, which could be really good from the perspective of concurrency with a process that does need that resource (like the non-bonded pairs).

ex-rzr:
@DaveCerutti 

> The optimization you're doing makes OpenMM's implementation more similar to STORMM's "general purpose" charge spreading kernel with fifth-order interpolation. I don't know how exactly you got five threads working on one atom--what I did was to devote five threads as they appear in sequence, and have each thread issue 25 atomicAdd() operations to __global__. In that way, the threads are well occupied up to (number of atoms) x 5, but the work of spreading every 7th atom or so will be split across two warps.

Yes, this PR uses the same approach.

> My guess is that, if the new OpenMM method work like my "global accumulation" approach, then there are 400% more B-spline computations going on.

Yes. Initially the kernel also had a trick to do less computations:
https://github.com/ex-rzr/openmm/commit/8415119a63b58eae7bcbf35ed0f63fa87a83293c#diff-4f10c0a9cd23bf8937b61c814f0be872a1e69e50c7b3b8bf6e953547974803b2R460-R486
3 threads of 5 calculate x, y, z of values of B-splines, then all threads use them using shuffles (yes, additional overhead). But it provided only a very small improvement in performance and couldn't work on OpenCL (however I don't know performance with exchanging values B-splines using shared memory) so I chose the current, simpler code.

I suppose that with this PR OpenMM's yellow line will move close to "STORMM, Global Accumulation, Order 5", considering that they use the same approach and that `gridSpreadCharge` became ~40% faster.

Thanks for the link to your  paper. It's really interesting (though I haven't read all of it yet).
I'll try to understand your "local accumulation". I also had some (untested) ideas about doing atomics in shared memory first, but not as advanced as yours. Perhaps I'll find strength to think about it again :)

Btw, how much shared memory do you use? All available (i.e. one thread blocks per SM) or less? What is block size?

PR #4634, opened 2024-08-28
Optimize bonded forces by reducing intra-warp atomic conflicts

ex-rzr:
A follow-up of #4632

It can be tested on the HIP (though it needs commits from #4632 ) or CUDA platforms.

An optimization of bonded forces: atom indices for many forces are usually grouped so threads of one warp can write forces to the same index, so there are atomic conflicts within the warp. 
My idea is to do a reduction of consecutive segments of forces for the same atom index (using head segmented reduction). See [conflicts-with-atoms.txt](https://github.com/user-attachments/files/16785576/conflicts-with-atoms.txt), there are atom indices of 128 bonds for each force and some statistics: avgConflicts is the number of such conflicts that can be eliminated with head segmented sum.
This optimization is implemented for HIP and then ported to CUDA. On AMD GPUs it improves overall performance by 2-5% (except for gbsa and rf with small drop of ~0-1%).
For CUDA I've tested on Titan V, 3090, 4070 Ti Super: 3090 has +1-2% but -2% for gbsa, Titan V  has +2-3% but also -2% for gbsa, 4070 Ti Super has the biggest change +2-6% (6% for apoa1). I don't have access to older architectures or more low-end GPUs (2060 on my desktop show +-3% random performance fluctuations so it's impossible to benchmark on it reproducibly).
I don't know if it's worth having this optimization for CUDA (the `#elif defined(__CUDA_ARCH__) ...` part can be easily removed/commented), on AMD it seems to provide a bit more consistent improvement, it would be great to keep it. Also I'm not sure if you are ok with having a platform specific kernel code in the Common platform. I look forward to your decision.

### DIFF
--- a/platforms/common/src/BondedUtilities.cpp
+++ b/platforms/common/src/BondedUtilities.cpp
@@ -27,6 +27,7 @@
 #include "openmm/common/BondedUtilities.h"
 #include "openmm/common/ComputeContext.h"
 #include "openmm/OpenMMException.h"
+#include "CommonKernelSources.h"
 #include <iostream>
 
 using namespace OpenMM;
@@ -102,6 +103,7 @@ void BondedUtilities::initialize(const System& system) {
     // Create the kernel.
 
     stringstream s;
+    s<<CommonKernelSources::bonded;
     for (int i = 0; i < (int) prefixCode.size(); i++)
         s<<prefixCode[i];
     s<<"KERNEL void computeBondedForces(GLOBAL mm_ulong* RESTRICT forceBuffer, GLOBAL mixed* RESTRICT energyBuffer, GLOBAL const real4* RESTRICT posq, int groups, real4 periodicBoxSize, real4 invPeriodicBoxSize, real4 periodicBoxVecX, real4 periodicBoxVecY, real4 periodicBoxVecZ";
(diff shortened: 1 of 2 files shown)


peastman:
This is an interesting change.  I wonder why it's faster for PME but slower for RF and GBSA?  This only changes the bonded forces, and they're identical in all versions.

Here is my guess: this isn't really about avoiding conflicts, just about reducing the number of writes.  Running on its own, the bonded kernel doesn't saturate the bus, so avoiding writes has no benefit.  Instead, the extra work of combining forces makes it slower.  But if it runs at the same time as PME charge spreading, which is entirely dominated by the cost of all the writes, this change reduces the conflict between them leading to a benefit.

Here's a way we can test it.  When [creating the Context](https://github.com/openmm/openmm/blob/78902bed7e7b65d1aeb63216772c6099eb0fb6fd/examples/benchmark.py#L367) in the benchmark script, add another property:

```python
properties['DisablePmeStream'] = 'true'
```

If I'm right, this change will also cause PME to become slower when the separate stream is disabled.

ex-rzr:
Good idea, I'll benchmark with DisablePmeStream = true. But I can say the in my old profiling results (I profile with DisablePmeStream = true, so kernels are not running concurrently) on AMD this optimization makes computeBondedForces faster but not is small systems like RF and GBSA.

DaveCerutti:
My guess here is that you are re-ordering the series of instructions for doing valence interactions (dihedrals, in particular) so that the set of 32 (or 64 on legacy AMD) instructions contains as few of the same atoms as possible among the various threads.  Surprising that it's up to 5% on AMD, but when I tried this in Amber on CUDA I wasn't seeing anything.  Then again, the Amber kernel was hard memory bound, and when tying together interactions that can affect the same clusters of atoms one can find "candlesticks" (to use a Backgammon term) of 11 or more.  I talked about it on the NVIDIA forums with Norbert Juffa back in 2018, and his prediction was that spreading out the candlesticks wouldn't result in much optimization.  I was doing this with atomics to `__shared__` in specialized work units, much as I am doing in STORMM today, but perhaps there is a significant effect when you consider atomics to `__global__` like OpenMM is doing.

PR #4647, opened 2024-09-05
Updates to AmoebaVdwForce, AmoebaGeneralizedKirkwoodForce and AmoebaWcaDispersionForce

mjschnie:
The updates to the AmoebaGeneralizedKirkwoodForce and AmoebaWcaDispersionForce will align OpenMM with recent updates to the corresponding forces in Tinker and FFX (further details are available here https://doi.org/10.1063/5.0158914).

The update to the AmoebaVdwForce adds a per-particle scale factor needed for continuous AMOEBA CpHMD (described here https://doi.org/10.1021/acs.jctc.3c01180) to support many hydrogen atoms being turned on / off.


peastman:
Thanks!  I made a first quick pass through the code.  A detailed review will take longer.

How does this affect compatibility?  Will existing code now produce different results?  Or will code continue to produce the same results as long as it doesn't use the new features?

I noticed in particular that you changed the values of some parameters.  In AmoebaWcaDispersionForce, `shctd` used to be 0.81, and you changed it to 0.75.  In AmoebaGeneralizedKirkwoodForce, you changed `dielectricOffset` from 0.009 to 0.0.

Do we need changes to ForceField to support the new features?  Do we need updated versions of the forcefields?

For AmoebaGeneralizedKirkwoodForce, there are several hundred lines defining tabulated constants that are duplicated, appearing both in AmoebaCommonKernels.h and AmoebaReferenceGeneralizedKirkwoodForce.cpp.  Can we remove the duplication?  I suggest making them static constants in AmoebaGeneralizedKirkwoodForceImpl.  The definitions should be in the cpp file, not the header.  That way the compiler only has to process them once, not repeatedly in every file that includes the header.

There are places where parameters were packed into a float2 for efficiency.  You increased the number of parameters, but instead of expanding the float2 to a float4, you changed it into just an array of floats with more elements.  That will hurt performance.  Loading a single float4 is more efficient than loading four separate floats.

The serialization proxy for AmoebaGeneralizedKirkwoodForce will throw an exception if you try to load an old file, because it requires the new properties to be present.  When loading older files, it should supply default values for them.

Is "integratl" a misspelling for "integrate" or "integral"?  It appears several times.

mjschnie:
@peastman  -- Thanks for your awesome feedback. I'm working to incorporate your comments this afternoon. I'll make my best effort to complete things for the 8.2 beta given other constraints on my time this week.

mjschnie:
@peastman  -- Here are responses to the questions you've posed:
The parameter "shctd" compensates for computing an integration of the Weeks-Chandler-Andersen dispersion energy over solute spheres that overlap (an shctd of 1.0 would be used with atomic volumes that don't overlap at all, whereas 0.75 scales down the integration to 75% of the computed value to prevent overcounting due to the overlaps). I'm not sure where the value of 0.82 originated, while the value of 0.75 was fit against small molecules dispersion energies from explicit solvent simulations. This parameter can be set by the user, so backwards compatibility is supported. I'm fine with keeping the default value as 0.82, and then setting it to 0.75 for our updated model.

The delectricOffset parameter of "0.09" goes back to Clark Still's original 1990 generalized Born paper -- they defined an offset that was added to the OPLS-AA radii to define the beginning of the dielectric continuum boundary and give useful electrostatic solvation energies. For our AMOEBA / GK model, we're using "0.3". I'll revert the default to value from Still et al. 

We should eventually update the force field files, but perhaps this can wait until after the 8.2 beta? Perhaps this update should involve a consultation with the AMOEBA force field developers? 

I've moved the descreening "neck" constants to the class "AmoebaGeneralizedKirkwoodForceImpl" and made them static constants. Could you suggest the syntax / approach for passing these constants within the AmoebaReferenceGeneralizedKirkwoodForce.cpp and  AmoebaCommonKernels.cpp classes?

I've added backwards compatibility to the serialization proxy for AmoebaGeneralizedKirkwoodForce.

Thanks for catching the typo on "integratl" -- it should be "integral" and this is fixed.

peastman:
Thanks!  I don't know where those values came from, but they were validated against Tinker, and we want existing code to keep producing the same results.

> We should eventually update the force field files, but perhaps this can wait until after the 8.2 beta? Perhaps this update should involve a consultation with the AMOEBA force field developers?

That sounds fine.

> Could you suggest the syntax / approach for passing these constants within the AmoebaReferenceGeneralizedKirkwoodForce.cpp and AmoebaCommonKernels.cpp classes?

You should be able to just access them directly: `AmoebaGeneralizedKirkwoodForceImpl::AIJ`.  Or if you prefer, you can put them behind a static method that returns a const reference to them: `AmoebaGeneralizedKirkwoodForceImpl::getAij()`.

peastman:
Let me know when you have all the changes in.  Then I can start on a detailed review of the code.

mjschnie:
The change I've not had a chance to finish is using float4 instead of passing an array of floats in the updated Born radii routines. I'll have time to work on this tomorrow.

mjschnie:
@peastman  -- Thanks for all your feedback. Please go ahead with a detailed code review when you're ready!

plugins/amoeba/openmmapi/include/openmm/AmoebaGeneralizedKirkwoodForce.h
```diff
@@ -76,25 +76,42 @@ class OPENMM_EXPORT_AMOEBA AmoebaGeneralizedKirkwoodForce : public Force {
      */
     int addParticle(double charge, double radius, double scalingFactor);
 
+    /**
+     * Add the  parameters for a particle.  This should be called once for each particle
+     * in the System.  When it is called for the i'th time, it specifies the parameters for the i'th particle.
```
peastman, line 81: Since there are two different versions of `addParticle()`, maybe document what the difference between them is?  Especially what values get filled in for `descreenRadius` and `neckFactor` when they aren't specified.

plugins/amoeba/platforms/common/src/AmoebaCommonKernels.cpp
```diff
@@ -1814,11 +1820,28 @@ void CommonCalcAmoebaGeneralizedKirkwoodForceKernel::initialize(const System& sy
         defines["MUTUAL_POLARIZATION"] = "";
     else if (polarizationType == AmoebaMultipoleForce::Extrapolated)
         defines["EXTRAPOLATED_POLARIZATION"] = "";
+    defines["DIELECTRIC_OFFSET"] = cc.doubleToString(force.getDielectricOffset());
+    tanhRescaling = force.getTanhRescaling();
+    // Max Born radius is 3 nm (30 A).
+    defines["BIG_RADIUS"] = cc.doubleToString(3.0);
+    if (tanhRescaling) {
+        defines["TANH_RESCALING"] = cc.intToString(1);
```
peastman, line 1828: That's an interesting way of writing "1". :)  Since the kernel interprets it as a bool, it would be clearer to set it to "true" or "false" rather than "1" or "0".

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -3,27 +3,57 @@
...
 
         // Now calculate Born radius.
+        sum = PI4_3 * ir3 - sum;
 
-        float radius = params[index].x;
-        radius = RECIP(radius*radius*radius);
-        sum = radius-sum;
-        sum = (sum <= 0 ? (real) 1000 : POW(sum, -1/(real) 3));
+        // If the sum is less than zero, set the Born radius to 50.0 Angstroms.
+        sum = (sum <= 0 ? BIG_RADIUS : POW(INVERSE_PI4_3 * sum, -ONE_THIRD));
```
peastman, line 37: `BIG_RADIUS` is set to 3, not 5.

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -46,51 +76,184 @@ KERNEL void computeSurfaceAreaForce(GLOBAL mm_long* RESTRICT bornForce, GLOBAL m
...
-DEVICE real computeBornSumOneInteraction(AtomData1 atom1, AtomData1 atom2) {
+DEVICE real interpolate2D(real x1, real x2, real y1, real y2, real x, real y,
+                            real fx1y1, real fx2y1, real fx1y2, real fx2y2) {
+    real fxy1 = (x2 - x) / (x2 - x1) * fx1y1 + (x - x1) / (x2 - x1) * fx2y1;
+    real fxy2 = (x2 - x) / (x2 - x1) * fx1y2 + (x - x1) / (x2 - x1) * fx2y2;
+
+    return (y2 - y) / (y2 - y1) * fxy1 + (y - y1) / (y2 - y1) * fxy2;
+}
+
+DEVICE void getBounds(real rho, int bounds[]) {
```
peastman, line 90: Same comments here about using an array to return values.  It's more confusing, leads to more complicated code, and will be less efficient if the function doesn't get inlined.

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -3,27 +3,57 @@
 /**
  * Reduce the Born sums to compute the Born radii.
  */
-KERNEL void reduceBornSum(GLOBAL const mm_long* RESTRICT bornSum, GLOBAL const float2* RESTRICT params, GLOBAL real* RESTRICT bornRadii) {
+KERNEL void reduceBornSum(GLOBAL const mm_long* RESTRICT bornSum, GLOBAL const float4* RESTRICT params, GLOBAL real* RESTRICT bornRadii) {
+
+    real RECIP_MAX_RADIUS3 = POW(BIG_RADIUS, (real) -3.0);
+    real PI4_3 = 4.0 / 3.0 * M_PI;
+    real INVERSE_PI4_3 = 1.0 / PI4_3;
+    real ONE_THIRD = 1.0 / 3.0;
```
peastman, line 11: Writing double precision constants in kernel code leads to the GPU performing double precision operations, even if it's just to cast the value back to single precision.  This leads to much slower code.  When possible prefer integer constants, or if not then single precision.  This takes a lot of care in how you write expressions.  I would write these lines as

```c++
    real RECIP_MAX_RADIUS3 = POW(BIG_RADIUS, (real) -3);
    real PI4_3 = 4 * M_PI / 3;
    real INVERSE_PI4_3 = 1 / PI4_3;
    real ONE_THIRD = 1 / (real) 3;
```

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -46,51 +76,184 @@ KERNEL void computeSurfaceAreaForce(GLOBAL mm_long* RESTRICT bornForce, GLOBAL m
...
+    real fxy1 = (x2 - x) / (x2 - x1) * fx1y1 + (x - x1) / (x2 - x1) * fx2y1;
+    real fxy2 = (x2 - x) / (x2 - x1) * fx1y2 + (x - x1) / (x2 - x1) * fx2y2;
+
+    return (y2 - y) / (y2 - y1) * fxy1 + (y - y1) / (y2 - y1) * fxy2;
+}
+
+DEVICE void getBounds(real rho, int bounds[]) {
+    // Note this function is in Angstroms.
+    real MINIMUM_RADIUS = 0.80;
+    real SPACING = 0.05;
```
peastman, line 93: More double precision constants.  Instead write `0.8f` and `0.05f`.  In the unlikely case you're using double precision mode, they'll get converted with no loss in accuracy.

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -46,51 +76,184 @@ KERNEL void computeSurfaceAreaForce(GLOBAL mm_long* RESTRICT bornForce, GLOBAL m
...
+}
+
+DEVICE void getNeckConstants(real rhoDescreened, real rhoDescreening, real constants[],
+    GLOBAL const float* RESTRICT neckRadii,
+    GLOBAL const float* RESTRICT neckA,
+    GLOBAL const float* RESTRICT neckB) {
+
+    // Convert the input radii from nm to A.
+    rhoDescreened *= 10.0;
+    rhoDescreening *= 10.0;
```
peastman, line 120: Write `10` instead of `10.0` to avoid double precision math.

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -46,51 +76,184 @@ KERNEL void computeSurfaceAreaForce(GLOBAL mm_long* RESTRICT bornForce, GLOBAL m
...
+
+    real bij = interpolate2D(neckRadii[lowI], neckRadii[highI],neckRadii[lowJ], neckRadii[highJ],
+                               rhoDescreened, rhoDescreening,
+                               neckB[lowOff + lowJ], neckB[highOff + lowJ],
+                               neckB[lowOff + highJ], neckB[highOff + highJ]);
+
+    // Never let Aij be negative.
+    if (aij < 0.0) {
+        aij = 0.0;
+    }
```
peastman, line 151: More double precision constants.  I won't keep flagging every one.  There's lots more below, so go through carefully.

peastman (commented):
Since there's a lot of code and it will take me a while to get through all of it, I'll break my review into a few pieces.  That way you can get started on the changes.  I'll try to get through all of it by the end of tomorrow, or Sunday at the latest.

plugins/amoeba/tests/TestAmoebaGeneralizedKirkwoodForce.h
```diff
@@ -75,7 +75,8 @@ static void checkFiniteDifferences(vector<Vec3> forces, Context &context, vector
     State state2 = context.getState(State::Energy);
     context.setPositions(positions3);
     State state3 = context.getState(State::Energy);
-    ASSERT_EQUAL_TOL(state3.getPotentialEnergy()+norm*stepSize, state2.getPotentialEnergy(), 1e-4);
+
+    ASSERT_EQUAL_TOL(state3.getPotentialEnergy()+norm*stepSize, state2.getPotentialEnergy(), 1e-3);
```
peastman, line 79: Is the change in tolerance really needed?  Make sure this isn't masking a bug.

plugins/amoeba/tests/TestAmoebaGeneralizedKirkwoodForce.h
```diff
@@ -96,63 +97,57 @@ static void setupMultipoleAmmonia(System& system, AmoebaGeneralizedKirkwoodForce
...
-    nitrogenMolecularQuadrupole[2] =  -2.6883671e-04;
-    nitrogenMolecularQuadrupole[3] =   0.0000000e+00;
-    nitrogenMolecularQuadrupole[4] =   2.5463927e-04;
-    nitrogenMolecularQuadrupole[5] =   0.0000000e+00;
-    nitrogenMolecularQuadrupole[6] =  -2.6883671e-04;
-    nitrogenMolecularQuadrupole[7] =   0.0000000e+00;
-    nitrogenMolecularQuadrupole[8] =   1.4942322e-04;
+    nitrogenMolecularDipole[0]     =  0.0;
+    nitrogenMolecularDipole[1]     =  0.0;
+    nitrogenMolecularDipole[2]     =  0.01555516;
```
peastman, line 102: I would feel better about adding a new test rather than replacing the existing one.  We need to make sure the existing code continues to produce the same results it did before.

plugins/amoeba/tests/TestAmoebaGeneralizedKirkwoodForce.h
```diff
@@ -7064,19 +7061,19 @@ static void testGeneralizedKirkwoodAmmoniaDirectPolarization() {
...
+    expectedForces[2]         = Vec3( 5.22373867e+01, -1.22133406e+01, 2.63868093e+00);
+    expectedForces[3]         = Vec3( 7.28501357e+01, 5.23520969e-02, -2.11572152e+00);
+    expectedForces[4]         = Vec3( -1.16737706e+02, 1.82676163e-01, -9.15644064e+01);
+    expectedForces[5]         = Vec3( 7.15433778e+00, -8.75935967e+00, 3.20920876e+01);
+    expectedForces[6]         = Vec3( 2.80388964e+02, 1.86448609e-02, 2.79161533e+01);
+    expectedForces[7]         = Vec3( 7.13149260e+00, 8.64144649e+00, 3.21405304e+01);
 
     double tolerance          = 1.0e-04;
-    compareForcesEnergy(testName, expectedEnergy, energy, expectedForces, forces, tolerance);
+    // compareForcesEnergy(testName, expectedEnergy, energy, expectedForces, forces, tolerance);
```
peastman, line 7076: Why did you remove this here and in the functions below?  That means nothing is actually being tested.

wrappers/python/tests/systems/alanine-dipeptide-amoeba-forces.xml
```diff
@@ -1,32 +1,32 @@
...
+		<Force x="839.6940799021637" y="56.97466842408522" z="-318.8521475939689"/>
+		<Force x="362.3272180197968" y="-706.8489409774945" z="-264.4258472168631"/>
+		<Force x="-505.8214032425471" y="-723.0346848237261" z="324.9594864383717"/>
+		<Force x="-568.8744161828374" y="266.5494507684997" z="-30.478132146722988"/>
+		<Force x="-485.2241615788016" y="939.5060141587863" z="114.86701913403719"/>
+		<Force x="762.0358996664957" y="-84.19661079197815" z="105.88781416394265"/>
+		<Force x="349.4477297990101" y="-1429.4764613260852" z="-5.030937078796498"/>
+		<Force x="-567.6667253076068" y="689.2675715661583" z="-.6763489861962313"/>
+		<Force x="313.27996277450796" y="454.69440232536095" z="551.5915187153109"/>
+		<Force x="313.05157212187447" y="452.19258398876815" z="-552.5454929317236"/>
```
peastman, line 30: Here again we need to add new tests, not change the existing ones.  We need to verify that existing code still produces the same results.

peastman (commented):
Whew!  I got through all of it.

I think it's a stretch to try to get this into 8.2.  There's still a lot that needs to be done.  But if you'll make the changes in the next couple of days, I'll find the time to do another review pass and we can try to get it into shape in time.

mjschnie:
@peastman  -- I appreciate all your hard work on the weekend! I believe I've addressed all your great feedback. There are two testing issues we should consider more deeply.
1) In the current code (prior to this PR) the "dielectric offset" parameter was only applied to the (approximate) nonpolar cavity term, but were not being applied when the Born radii were calculated. In this PR, the dielectric offset is applied in both places. For this reason, backwards compatibility for tests where BOTH the dielectric offset is greater than zero and the cavity term is being used is really difficult. We'd have to "hack" some way for this parameter to only be applied in one place, or we'd have to create support for "two" dielectric offset parameters (one for Born radii and one for the nonpolar cavity). This issue is the reason I updated the "wrappers/python/tests/systems/alanine-dipeptide-amoeba-forces.xml".
2) I noticed that for a test labeled "testGeneralizedKirkwoodAmmoniaExtrapolatedPolarization", the "direct" polarization model was being used instead of "extrapolated". If extrapolated is used, the Reference force passes, but the Common/OpenCL force does not? Currently I've left the test the way I encountered it (i.e. using "direct"), but it would be good to understand why "direct" was being used and perhaps you could test how "extrapolated" behaves for Common/OpenCL in your hands. The good news is the GeneralizedKirkwoodVillinExtrapolatedPolarization passes for me using both the Reference Force and Common/OpenCL. I've not worked through the theory for GeneralizedKirkwood plus Extrapolated polarization, but I will as time permits.

peastman:
Thanks!  I'll take a look.

Do you know if that behavior has changed in Tinker?  The OpenMM implementation was based on the Tinker code, and the tests were written to make sure forces and energy matched.  How does Tinker currently handle the dielectric offset?

mjschnie:
To the best of my knowledge, Tinker has never implemented a Born radii based non-polar cavity term. I think this idea may have come from another code or paper?
This PR updates the OpenMM GeneralizedKirkwoodForce Born radii code to exactly match the current version of Tinker   (https://github.com/TinkerTools/tinker/blob/242c2a116dd578b1d39601a4a093c5ebdb316837/source/born.f#L329)
Rather than using the Born radii to approximate surface area, I'd advocate for moving toward GaussVol (https://doi.org/10.1002/jcc.24745) or AlphaMol (https://doi.org/10.1021/acs.jcim.2c01346) for the cavitation force. Perhaps eventually deprecation of the "includeCavityTerm" could be considered?

plugins/amoeba/openmmapi/include/openmm/internal/AmoebaGeneralizedKirkwoodForceImpl.h
```diff
@@ -37,6 +37,8 @@
 #include "openmm/Kernel.h"
 #include <string>
 
+using std::vector;
```
peastman, line 40: Never use the `using` keyword in a header file, only in a cpp file.  It forces it on every file that includes it.

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -87,102 +87,87 @@ DEVICE real interpolate2D(real x1, real x2, real y1, real y2, real x, real y,
...
+    if (*aij < 0.0f) {
+        *aij = 0.0f;
     }
 
     // Convert aij from A^(-11) to nm^(-11);
-    constants[0] = aij * 1.0e11;
+    *aij *= 1.0e11f;
     // Convert bij from A to nm.
-    constants[1] = bij * 0.1;
+    *bij *= 0.1f;
```
peastman, line 145: It's better to write this as `*bij /= 10`.  The reason is that 0.1 can't be exactly represented in floating point.  In double precision mode, you're multiplying by a single precision approximation to 0.1.  If you instead divide by an integer constant (exact in all modes), the compiler will automatically convert it to multiplying by an appropriate constant whose precision matches the precision of bij.

peastman:
> To the best of my knowledge, Tinker has never implemented a Born radii based non-polar cavity term.

[Here is the code](https://github.com/TinkerTools/tinker/blob/242c2a116dd578b1d39601a4a093c5ebdb316837/source/esolv.f#L80-L95) where Tinker calculates that term.

mjschnie:
Thanks for finding the relevant Tinker code for the Born radii based nonpolar energy. Just above that code, the intended nonpolar terms for the Generalized Kirkwood model are called (i.e. Weeks-Chandler-Andersen Dispersion plus Cavitation from solvent excluded volume and solvent accessible surface area). The WCA Dispersion is already part of OpenMM and it would be great to have SEV and SASA at some point.

Regarding the dielectricOffset parameter -- to maintain backwards compatibility I can create a second offset parameter. The current dielectricOffset parameter will behave as it always has (i.e. only used for the Born radii based nonpolar cavity term). A new parameter that I'll call "descreenOffset" will be used during Born radii calculation.

Does this sound ok?

peastman:
That sounds good.  Thanks!

peastman:
I ran the benchmarks on a RTX 4080, and it looks like there's a bit of a performance regression.

Current code:

```
amoebapme         single      None          1              2         NVT        CUDA       36.8552
amoebagk          single      None          1              2         NVT        CUDA       50.156
```

This PR:

```
amoebapme         single      None          1              2         NVT        CUDA       36.5141
amoebagk          single      None          1              2         NVT        CUDA       48.6586
```

I'm profiling now to see where it's coming from.

peastman:
There are still a couple of double precision constants, at lines 619 and 622 of amoebaGk.cc.

Even after fixing those, the profiler still reports a bit of activity in the FP64 unit for both `computeGkForces()` and `computeWCAForce()`.  I'll try to figure out where that's coming from.

peastman:
The profile for `computeGKForces()` is mostly unchanged.  I think the slowdown there is mostly just coming from having to load two extra parameters.  There's not much to do about that.

`computeWCAForce()` has much bigger differences in the profile.  The compute throughput is a lot lower (down from 56% to 43%), and the warp state statistics show that for every instruction executed, it spends about one extra clock cycle stalled waiting on a local dependency.  The code now has a lot more `if` blocks than it used to, which likely are the reason.  Can you describe the reasons for the changes?  Perhaps there's some way we could consolidate the code to reduce the number of branches.

mjschnie:
"Can you describe the reasons for the changes?"

The original WCA Dispersion code was part of my thesis, but wasn't formally published until Rae Corrigan led working through the pairwise integration edge cases as described in the following publication: https://pubs.acs.org/doi/full/10.1021/acs.jctc.0c01286

The if statements handle the various geometrical cases (see Figure 6, panels B, C & D).

peastman:
Does that mean AmoebaWCADispersionForce now produces different results than it used to, even for an identical system?  You didn't make any changes to its API, such as new parameters that would trigger a different computation.

If so, how did Tinker handle this transition?  It used to produce identical results to what OpenMM currently produces.  At some point did that change?

mjschnie:
The concordance between OpenMM and Tinker for WCA Dispersion diverged here:
https://github.com/TinkerTools/tinker/commit/9c4245dd644494de62ef87aeeef566c07e3deefc

The intent of this PR is for OpenMM and Tinker to concord exactly again.

peastman:
How did they handle the transition?  Was there any attempt to provide backward compatibility?  I see that PR involved changes to both code and parameter files.  Will our existing parameter files for AMOEBA GK now produce incorrect results?

peastman:
I discussed this offline with @mjschnie, and I have a better understanding of the context.  A recent version of Tinker changed how it computes implicit solvent.  This PR includes some, but not all, of the changes necessary to match Tinker.  In addition to these changes, it introduced a new method of representing the cavity energy and new force field parameters.

We decided not to include this in 8.2.  Merging it now would lead to results that don't match any version of Tinker, either the old implicit solvent or the new one.  So we'll hold off on merging this until 8.2 is released.  In the mean time, we can start looking into implementing the new cavity term.

peastman:
Now that 8.2 is out, let's move forward with getting this merged.  I'll make another pass through the code, and also see if I can come up with ways to improve the performance issues mentioned above.

plugins/amoeba/openmmapi/include/openmm/internal/AmoebaGeneralizedKirkwoodForceImpl.h
```diff
@@ -62,6 +62,46 @@ class AmoebaGeneralizedKirkwoodForceImpl : public ForceImpl {
...
+     * @param r separation distance.
+     * @param radius Radius of the current atom.
+     * @param radiusK Radius of the descreening atom.
+     * @param sneck The neck scale factor.
+     * @return
+     */
+    static double neckDescreen(double r, double radius, double radiusK, double sneck);
+
+    /**
+     * Get Neck Aij and Bij constants.
```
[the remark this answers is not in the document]
peastman: This still needs to be addressed.  I *think* the difference is that this method performs a computation for a single atom and returns the results (meaning that the returned values are *not* constants), whereas `getAij()` and `getBij()` return the constants used to compute them?  Anyway, the documentation is confusing.

plugins/amoeba/platforms/common/src/kernels/amoebaGk.cc
```diff
@@ -432,65 +586,153 @@ KERNEL void computeGKForces(
...
+ * @param radius base radius of descreened atom.
+ * @param scaledRadius scaled radius descreening atom.
+ * @return the derivative.
+ */
+DEVICE real pairIntegralDerivative(real r, real r2, real radius, real scaledRadius) {
+
+        real de = 0.0;
+        // Descreen only if the scaledRadius is greater than zero.
+        // and atom I does not engulf atom K.
+        if (scaledRadius > 0.0 && (radius < r + scaledRadius)) {
```
peastman, line 622: A couple of double precision constants that got missed.

peastman:
I attempted some optimizations, but didn't find anything that consistently helped.  Can you make the remaining changes from the review comments?  Then we can merge.  I might do some cleanup in another PR, but that's just for minor things that don't need to delay this.

plugins/amoeba/platforms/reference/src/AmoebaReferenceKernels.cpp
```diff
@@ -573,6 +592,16 @@ void ReferenceCalcAmoebaGeneralizedKirkwoodForceKernel::getCharges(vector<double
     copy(charges.begin(), charges.end(), outputCharges.begin());
 }
 
+void ReferenceCalcAmoebaGeneralizedKirkwoodForceKernel::getDescreenRadii(vector<double>& outputDescreenRadii) const {
+    outputDescreenRadii.resize(descreenRadii.size());
+    copy(descreenRadii.begin(), descreenRadii.end(), outputDescreenRadii.begin());
```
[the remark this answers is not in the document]
peastman: This still needs to be fixed.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceGeneralizedKirkwoodForce.h
```diff
@@ -81,6 +81,31 @@ class AmoebaReferenceGeneralizedKirkwoodForce {
...
+    int getTanhRescaling() const;
+
+    /**
+     * Get Tanh parameters beta0, beta1 and beta2.
+    */
+    void getTanhParameters(double& b0, double& b1, double& b2) const;
+
+    /**
+     * Set the flag signaling whether the solute integral is rescaled by a Tanh function
+     * to account for interstitial spaces.
```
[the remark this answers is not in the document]
peastman: This still needs to be fixed.

plugins/amoeba/platforms/reference/src/SimTKReference/AmoebaReferenceMultipoleForce.h
```diff
@@ -1269,6 +1276,7 @@ class AmoebaReferenceGeneralizedKirkwoodMultipoleForce : public AmoebaReferenceM
     double _probeRadius;
     double _surfaceAreaFactor;
     double _dielectricOffset;
+    double _tanhRescaling;
```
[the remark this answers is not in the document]
peastman: This still needs to be fixed.

plugins/amoeba/serialization/src/AmoebaGeneralizedKirkwoodForceProxy.cpp
```diff
@@ -42,46 +42,67 @@ AmoebaGeneralizedKirkwoodForceProxy::AmoebaGeneralizedKirkwoodForceProxy() : Ser
...
     node.setStringProperty("name", force.getName());
     node.setDoubleProperty("GeneralizedKirkwoodSolventDielectric", force.getSolventDielectric());
     node.setDoubleProperty("GeneralizedKirkwoodSoluteDielectric",  force.getSoluteDielectric());
-    //node.setDoubleProperty("GeneralizedKirkwoodDielectricOffset",  force.getDielectricOffset());
+    node.setDoubleProperty("GeneralizedKirkwoodDielectricOffset",  force.getDielectricOffset());
     node.setDoubleProperty("GeneralizedKirkwoodProbeRadius",       force.getProbeRadius());
     node.setDoubleProperty("GeneralizedKirkwoodSurfaceAreaFactor", force.getSurfaceAreaFactor());
     node.setIntProperty(  "GeneralizedKirkwoodIncludeCavityTerm", force.getIncludeCavityTerm());
-
+    node.setIntProperty("GeneralizedKirkwoodTanhRescaling", force.getTanhRescaling());
```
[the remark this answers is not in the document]
peastman: This still needs to be fixed.

plugins/amoeba/platforms/reference/src/AmoebaReferenceKernels.cpp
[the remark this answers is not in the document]
peastman: If you want to change them, that would be great.  If you don't want to, I'll fix them in a followup PR.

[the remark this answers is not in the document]
mjschnie: Thanks - I fixed all five instances of this.

plugins/amoeba/platforms/reference/src/AmoebaReferenceKernels.cpp
```diff
@@ -573,6 +592,16 @@ void ReferenceCalcAmoebaGeneralizedKirkwoodForceKernel::getCharges(vector<double
...
+}
+
+void ReferenceCalcAmoebaGeneralizedKirkwoodForceKernel::getNeckFactors(std::vector<double> &outputNeckFactors) const {
+    outputNeckFactors.resize(neckFactors.size());
+    copy(neckFactors.begin(), neckFactors.end(), outputNeckFactors.begin());
```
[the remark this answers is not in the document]
mjschnie: This is fixed now.

plugins/amoeba/openmmapi/include/openmm/AmoebaGeneralizedKirkwoodForce.h
```diff
@@ -124,6 +141,41 @@ class OPENMM_EXPORT_AMOEBA AmoebaGeneralizedKirkwoodForce : public Force {
...
+     */
+    void getTanhParameters(double& b0, double& b1, double& b2) const {
+        b0 = this->beta0;
+        b1 = this->beta1;
+        b2 = this->beta2;
+    }
+
+    /**
+     * Set the flag signaling whether the solute integral is rescaled by a Tanh function
+     * to account for interstitial spaces.
```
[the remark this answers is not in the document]
peastman: You fixed the corresponding line in AmoebaReferenceGeneralizedKirkwoodForce.h, but not this one.

mjschnie:
Yes - from my shoes it's ready!

peastman:
Great!  I think we can merge it then!

Issue #4651, opened 2024-09-10
Inconsistent Behavior of XTCReporter Compared to Other Reporters When Target File Already Exists


z-gong:
When using PDBReporter or DCDReporter, the target file is silently overwritten if it already exists upon initialization. However, with XTCReporter, the behavior is different. The existence of the target .xtc file is only checked when the first set of positions is written, not during initialization. If a file already exists at that point, an exception is raised, and the simulation is terminated.

This behavior can be problematic when a large reporting interval (e.g., 1 million steps) is used. The simulation run for a significant period only to crash once it tries to write to an existing .xtc file.

Proposed Fixes:
- Overwrite the .xtc file if it already exists, making the behavior consistent with PDBReporter and DCDReporter.
- Check for the existence of the .xtc file during the initialization of XTCReporter. If the file exists, raise an exception immediately.

Do you think one of these fixes is reasonable? If so, I’d be happy to draft a pull request.

peastman:
Agreed, we should make the behavior consistent with the other reporters.  A PR would be great!

peastman:
Fixed by #4657.

PR #4657, opened 2024-09-13
XTCReporter overwrites existing xtc file

z-gong:
This pull request addresses issue #4651.

Currently, if an XTC file already exists, the XTCReporter raises an exception the first time it tries to write a frame. With this pull request, it will instead overwrite the existing file, consistent with the behavior of the PDB and DCD reporters.

Additionally, a check on the positions has been added. The XTCFile scales up positions by 1000 and store them as integers. If a position value exceeds 2E6, the xtc file becomes corrupted and unreadable by an xtc parser. This fix introduces a validation step that raises an exception if the positions exceed the limit that xtc format can handle.


peastman:
This looks great.  Thanks!

PR #4671, opened 2024-09-25
Increase describeNextReport readability and allow requesting more values

### LINKED ISSUE #4663: Allow to request all parameters context.getState hast to offer from a reporters describeNextReport() function
At the moment, custom reporters are not able to request parameter derivatives or integrator parameters for their next report.

https://github.com/openmm/openmm/blob/fed44c579e9ac3136fe38056968a9c6dc1b6931e/wrappers/python/openmm/app/simulation.py#L254-L272

Currently, to get parameter derivatives using a custom reporter, the state needs to be fetched separately.

I can make the changes and create a pull request if desired.

schuhmc:
Fixes #4663 
Also harmonizes _generate_reports with #4437 

describeNextReport can now return a dict instead of a tuple with the format
`return {'steps':steps, 'periodic':False, 'include':['positions', 'forces']}`

The old tuple format is still supported.

This makes it mainly possible to also request parameterDerivatives for TI calculations without having to get the state for that reporter separately.
It also makes the output of describeNextReport much more readable than the previous tuple.

---
# Example
Following is an example reporter utilizing the new format

```python
class dUdLReporter(object):
    """
    Reporter for printing out current dUdL values.

    Calculate derivative needs to be set for this to work.

    Based on openMM source code for StateDataReporter, and openMM user guide http://docs.openmm.org/latest/userguide/application/04_advanced_sim_examples.html#extracting-and-reporting-forces-and-other-data

    Args:
        object (_type_): _description_
    """
    def __init__(self, file, reportInterval: int, derivativeNames: list, append: bool =False, timeUnit: unit = picosecond):
        """
        Create reporter for writing out energies of individual force groups

        Args:
            file (str or stdout): where to print the values
            reportInterval (int): report every n steps
            derivativeNames (list, optional): List of str defining the names of derivatives to report.
            append (bool, optional): append to file. Defaults to False.
            timeUnit (mm.unit, optional): time unit to use. Defaults to unit.picosecond.
        """

        self._append = append
        self._derivativeNames = derivativeNames
        self._reportInterval = reportInterval
        self._timeUnit = timeUnit

        # Open the file if it is a file, otherwise write directly to it (probably stdout)
        self._openedFile = isinstance(file, str)
        if self._openedFile:
            self._out = open(file, 'a' if append else 'w')
        else:
            self._out = file

        self._hasInitialized = False

    def __del__(self):
        # Close the file upon deletion
        if self._openedFile:
            self._out.close()

    def describeNextReport(self, simulation):
        # returns in how many steps the next report is due, and which kind of data it needs
        steps = self._reportInterval - simulation.currentStep%self._reportInterval
        return {'steps': steps, 'periodic': None, 'include':['parameterDerivatives']}

    def report(self, simulation, state=None):
        # Generates the report. Simulation and state is passed through by the openMM engine.

        # print header if it is the first time this object has been called and is not set to append.
        if not self._hasInitialized:
            self._hasInitialized = True
            if not self._append:
                header = ','.join(self._derivativeNames)
                header = "time," + header

                print(header, file=self._out)
                try:
                    self._out.flush()
                except AttributeError:
                    pass

        derivatives = []
        if state is None:
            state = simulation.context.getState(getParameterDerivatives=True)
        
        for derivativeName in self._derivativeNames:
            derivatives.append(f"{state.getEnergyParameterDerivatives()[derivativeName]: .8f}")

        values = ','.join(derivatives)
        values = f"{state.getTime().value_in_unit(self._timeUnit): .2f},{values}"

        print(values, file=self._out)
        try:
            self._out.flush()
        except AttributeError:
            pass
```

---
# Todo

- [x] write documentation

---
# Additional thoughts

It might also make sense to include a reporter similar as the one above.

Should existing reporters be adjusted to match the proposed format?


wrappers/python/openmm/app/simulation.py
```diff
@@ -207,8 +207,14 @@ def _simulate(self, endStep=None, endTime=None):
             anyReport = False
             for i, reporter in enumerate(self.reporters):
                 nextReport[i] = reporter.describeNextReport(self)
-                if nextReport[i][0] > 0 and nextReport[i][0] <= nextSteps:
-                    nextSteps = nextReport[i][0]
+                # Allow to pass either a dictionary, or a tuple from the reporter for backwards compatibility
```
peastman, line 210: If the return value is a tuple, let's immediately convert it to the new format.  Then we don't have to worry about alternate formats in anything that follows.  Something like this (not tested, I'm just writing this in the browser).

```python
report = reporter.describeNextReport(self)
if isinstance(report, tuple):
    # Convert an old format tuple to the new dict format
    includes = ['positions', 'velocities', 'forces', 'energy']
    include = [includes[i] for i in range(4) if report[i+1]]
    newReport = {'steps': report[0], 'include':include}
    if len(report) > 5:
        newReport['periodic'] = report[5]
    report = newReport
nextReport[i] = report
```

wrappers/python/openmm/app/simulation.py
```diff
@@ -252,23 +272,43 @@ def _simulate(self, endStep=None, endTime=None):
...
+                oldReports.append(report)
+
+        # Check if any reporter needs positions, velocities,...
+        # will return False if newReports is empty
+        positions = any(['positions' in report[1]['include'] for report in newReports])
+        velocities = any(['velocities' in report[1]['include'] for report in newReports])
+        forces = any(['forces' in report[1]['include'] for report in newReports])
+        energy = any(['energy' in report[1]['include'] for report in newReports])
+        integratorParameters = any(['integratorParameters' in report[1]['include'] for report in newReports])
+        parameterDerivatives = any(['parameterDerivatives' in report[1]['include'] for report in newReports])
```
peastman, line 296: I suggest doing this a little differently.

```python
includes = set.union(*[set(report['include']) for report in reports])
includeArgs = {include:True for include in includes}
```

And then in the call to `getState()`, pass `**includeArgs` as an argument list.  Advantages of this are 1) it's less code, 2) it's forward compatible if we add other options to `getState()` in the future, and 3) if a reporter returns an invalid option in `include`, the error will be caught and produce an exception.

peastman (commented):
That's the idea.  I made some suggestions for changes to the code.

Once we're satisfied with the format, we also need to update [the documentation](https://github.com/openmm/openmm/blob/0f67a646a93a7fa600ae238b6485c54cd3c553db/docs-source/usersguide/application/04_advanced_sim_examples.rst?plain=1#L116-L193), and we probably should convert the built in reporters to the new format.

wrappers/python/openmm/app/simulation.py
schuhmc: Good suggestion, this is certainly more readable. I made a slight modification to use a set comprehension and it now reads
`includes = {property for report in reports for property in report[1].get('include', [])}`

peastman: I'm staring at that comprehension and totally failing to comprehend it. :)  It looks to me as if you're using `report` before defining it?  Perhaps it should be `{property for property in (report[1].get('include', []) for report in reports)}`?

wrappers/python/openmm/app/simulation.py
```diff
@@ -252,23 +263,17 @@ def _simulate(self, endStep=None, endTime=None):
...
-                getEnergy = True
-        state = self.context.getState(positions=getPositions, velocities=getVelocities, forces=getForces,
-                                      energy=getEnergy, parameters=True, enforcePeriodicBox=periodic,
-                                      groups=self.context.getIntegrator().getIntegrationForceGroups())
-        for reporter, next in reports:
+        '''
+        reports is a tuple of (reporter, nextReport)
+        reporter is the corresponding reporter object
+        nextReport is a description on what the reporter needs in the context. This can either be a dict, or a tuple (old).
+        '''
```
peastman, line 270: Can you make the docstring follow numpy format?  See any of the other methods in the file for examples.

wrappers/python/openmm/app/simulation.py
schuhmc: Nested list comprehensions need some getting used to

```python
includes = set()
for report in reports:
   for property in report[1].get('include', []):
      includes.add(property)
```

We start with the last expression, but then we need to go from top to bottom:

`{property for report in reports for property in report[1].get('include', [])}`

See [here](https://stackoverflow.com/questions/18072759/how-can-i-use-list-comprehensions-to-process-a-nested-list) for reference.

peastman: Let's write it in a way that doesn't need getting used to. :)  Making code clear and easy to understand is important!  An advantage of `union()` is that it tells you at a glance what the code is doing: taking the union of the sets.

peastman:
This is looking good, and I think we've converged on the format.  Do you want to make the other changes described above?

schuhmc:
Yes, I will do that in the next few days when I find time.

Should I also add a reporter for the derivative, like the one in the first comment? I think it could be quite useful and would negate the need of writing one yourself if you wanted to do TI.

Is there any documentation on how to write a custom reporter besides [here](https://github.com/openmm/openmm/blob/master/docs-source/usersguide/application/04_advanced_sim_examples.rst) that would also need to be updated?

wrappers/python/openmm/app/simulation.py
```diff
@@ -226,19 +232,33 @@ def _simulate(self, endStep=None, endTime=None):
                 unwrapped = []
                 either = []
                 for reporter, report in zip(self.reporters, nextReport):
-                    if report[0] == nextSteps:
-                        if len(report) > 5:
-                            wantWrap = report[5]
-                            if wantWrap is None:
+                    if isinstance(report, dict):
+                        if report['steps'] == nextSteps:
+                            wantWrap = self._usesPBC if report['periodic'] is None else report['periodic']
```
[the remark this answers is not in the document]
schuhmc: This could have unintended side effects. Currently, some reporters use None as value to get the default behavior (e.g. the [dcdreporter](https://github.com/openmm/openmm/blob/f48eb17ba3a537e93beb1a60bb3e8c7b4099d3dc/wrappers/python/openmm/app/dcdreporter.py)).

Changing it in the included reporters should not be an issue, but it could change how custom reporters behave.

I also think we should make it possible to explicitly request to use the system default.

I would propose the following:

`wantWrap = self._usesPBC if report.get('periodic') is None else report['periodic']`

That would catch all possibilities and return the default value in both cases, when report['periodic'] is undefined and when it is set to `None`, explicitly requesting the sytem default.

peastman:
I think that's the only documentation that needs updating.

schuhmc:
Should be gtg from my side. Maybe you want to look over the reporter still. Note that the CI is failing today because of #4677

peastman:
I told it to rerun the builds, but it still used mambaforge.  Maybe try merging changes from the master branch?

peastman:
> Should I also add a reporter for the derivative, like the one in the first comment?

I think let's not do that.  Parameter derivatives tend to be used for specialized purposes.  A class that just prints out their values isn't going to be generally useful.

peastman:
This looks good and the tests are working again.  Is it ready to merge?

schuhmc:
> This looks good and the tests are working again. Is it ready to merge?

From my side, yes.

peastman:
Great, thanks!

PR #4672, opened 2024-09-25
Added documentation on HIP

peastman:
This adds more documentation about the HIP platform.

@ex-rzr am I right that it's currently only supported on Linux?  I didn't add anything about it in the section on compiling for Windows.

### DIFF
--- a/docs-source/developerguide/07_cuda_platform.rst
+++ b/docs-source/developerguide/07_cuda_platform.rst
@@ -8,23 +8,23 @@
 
 .. _the-cuda-platform:
 
-The CUDA Platform
-#################
+The CUDA and HIP Platforms
+##########################
 
-The CUDA platform is very similar to the OpenCL platform, and most of the
-previous chapter applies equally well to it, just changing “OpenCL” to “Cuda” in
-class names.  There are a few differences worth noting.
+The CUDA and HIP platforms are very similar to the OpenCL platform, and most of the
+previous chapter applies equally well to them, just changing “OpenCL” to “Cuda” or
+"Hip" in class names.  There are a few differences worth noting.
 
 Caching Kernels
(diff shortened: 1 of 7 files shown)


ex-rzr:
The HIP platform works on Windows as well.
Of course, there is no conda packages for Windows yet, so HIP SDK must be installed via the official installer:
https://www.amd.com/en/developer/resources/rocm-hub/hip-sdk.html 
https://rocm.docs.amd.com/projects/install-on-windows/en/develop/

In my case the only thing I needed to build it was `-D CMAKE_PREFIX_PATH="C:\Program Files\AMD\ROCm\6.1"`.
I also have `C:\Program Files\AMD\ROCm\6.1\bin` in the `PATH` environment variable to run tests/examples, but I'm not sure this is required everywhere.

@bdenhollander Do you use something special to build and run on Windows?

bdenhollander:
I wrote [these instructions]([url](https://github.com/bdenhollander/openmm-hip/blob/windows-compatibility/README-Windows.md)) for building on Windows when HIP was a separate plugin. Now that it's merged I was able to set all the variables via the CMake GUI and get the generated solution compiling in Visual Studio. I manually copy OpenMMHip.dll from the Release folder to the plugins directory of a conda environment. There are a few DLLs mentioned in the document that need to be copied to the plugins folder or made available via `PATH` like @ex-rzr did.

HIP 6.1 SDK should be recommended as the minimum for Windows to avoid the AMOEBA bug in 5.5 & 5.7 and the header issues in 5.7.

bdenhollander:
Looks good to me. It's definitely a lot easier now that HIP is a project within the same VS solution and VkFFT is always used.

peastman:
Thanks!

PR #4683, opened 2024-10-03
updated units in energy minimisation user guide

WillGPoole:
Units for energy minimization have changed recently(ish). The units in the API don't match the user guide.
http://docs.openmm.org/latest/api-python/generated/openmm.app.simulation.Simulation.html#openmm.app.simulation.Simulation.minimizeEnergy

### DIFF
--- a/docs-source/usersguide/application/02_running_sims.rst
+++ b/docs-source/usersguide/application/02_running_sims.rst
@@ -1333,9 +1333,9 @@ specify if you want further control over the minimization.  First, you can
 specify a tolerance for when the energy should be considered to have converged:
 ::
 
-    simulation.minimizeEnergy(tolerance=5*kilojoule/mole)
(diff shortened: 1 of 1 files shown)


peastman:
Thanks!  They haven't actually changed.  They were just documented incorrectly.

In all three places, can you list the units as `kilojoules/mole/nanometer`?  That's clearest.  Energy is measured in kJ/mole, and force has units of energy/distance.

WillGPoole:
Strange. I have many old scripts that used to run fine, but I had also erroneously put units of kJ/mole. Recently they correctly fail with the following: `TypeError: Unit "kilojoule/mole" is not compatible with Unit "kilojoule/(nanometer*mole)".` 

Looking at the `swigInputConfig` from 7.5 vs. master, there is an additional unit check!

I've made the units more readable in the docs :)

peastman:
Looks good.  Thanks!

PR #4688, opened 2024-10-08
Prevent linking to nvrtc-builtins

peastman:
This is a workaround for a bug in CMake.  When compiling against CUDA 12, it incorrectly links to libnvrtc-builtins.so.  That's supposed to be only for internal use by the toolkit, not something programs link to directly.  Doing so breaks binary compatibility between different 12.x versions.


peastman:
See [this discussion](https://discourse.cmake.org/t/cmake-incorrectly-links-to-nvrtc-builtins/12723/4) for more details.

PR #4695, opened 2024-10-15
Benchmark enhancements

bdenhollander:
- Add `--disable-pme-stream` to benchmark.py to easily compare performance without having to build with different default values for `DisablePmeSteam`
- Replace utcnow to remove DeprecationWarning
- Check for nvidia-smi before attempting to launch it
- Mention HIP where appropriate

### DIFF
--- a/examples/benchmark.py
+++ b/examples/benchmark.py
@@ -352,8 +352,8 @@ def runOneTest(testName, options):
         properties['DeviceIndex'] = options.device
         if ',' in options.device or ' ' in options.device:
             initialSteps = 250
-    if options.disable_pme_stream is not None:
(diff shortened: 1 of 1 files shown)


peastman:
This looks good, thanks!

Would it be better to define the argument like this?

```python
parser.add_argument('--disable-pme-stream', default=False, action='store_true', help='disable use of a separate GPU stream for PME')
```

That better reflects how the options works.  The default is always false, which means to leave it up to the platform.  This option is already a way to override the default, so the idea of overriding whether to override the default is a bit confusing!

bdenhollander:
That's a good idea and avoids needing to specify a value. I updated the description and pushed the change.

peastman:
Thanks!

Issue #4707, opened 2024-10-25
Suggestions on how to implement COM-centered flat-bottom restraint which applies to individual particles


stefdoerr:
So, I have the following problem. I want to restrain individual particles to be within X nanometers from the center of mass of a different molecule.

I have tried to come up with a solution with the given OpenMM forces but I cannot.  `CustomCentroidBondForce` is the closest I can find but it applies restaints between two COMs while I want to apply the restraint on the individual atoms of the first set.

I guess ideally I would want an `addParticle` option on `CustomCentroidBondForce` which adds individual particles and then access their `xyz` coordinates in the energy expression with `x`, `y`, `z` while the groups are numbered as they are already with `x1`, `x2` etc.

Is there any way to implement this with the existing forces though?

stefdoerr:
The only other way I have come up with is to use a `CustomExternalForce` and define the COM as a global variable. But then I need to update the COM every few steps in the simulation which requires me to retrieve the coordinates from the GPU context which will slow down the simulation significantly if I do this frequently

peastman:
In the CustomCentroidBondForce, make the individual particles each be its own group.  You'll have one group that's a whole molecule, and a lot of other groups for the individual atoms you want to restrain relative to its center of mass.

peastman:
You write an energy expression that connects a single atom to the COM of the molecule.  Then you call `addBond()` once for each atom you want to restrain.

stefdoerr:
Ok, I'll give it a try. 

Is there any way to get the periodic distance of a single axis though? I see that it has the `distance(g1, g2)` function but it doesn't seem to have the `periodicdistance(x1, y1, z1, x2, y2, z2)` function available so that I can use it like `periodicdistance(x1, 0, 0, x2, 0, 0)`.

peastman:
Use the `pointdistance()` function.  If you call `setUsesPeriodicBoundaryConditions(True)`, it will be periodic.  Remember that's only valid for rectangular boxes.  For triclinic boxes, you can't separate out one axis from the others.

stefdoerr:
Hi Peter, sorry to bother again, but could you provide us with an example for this? We are a bit stuck on how to write the energy function to "connect a single atom to the COM of the molecule". If you can provide a full example with how to add the bonds as well it would be super helpful. Thanks!

peastman:
That looks right.  Think carefully about your energy function though.  What you've written has a singularity when x1=x2.

stefdoerr:
[...]

PR #4713, opened 2024-10-31
Reduced memory use while identifying molecule groups

peastman:
This should fix the problems reported in #4707.  Instead of considering every particle in a group to be directly bonded to every other, it treats them as bonded in a linear chain.  When a group contains a very large number of particles, this uses much less memory.

### DIFF
--- a/platforms/common/src/ComputeContext.cpp
+++ b/platforms/common/src/ComputeContext.cpp
@@ -35,6 +35,7 @@
 #include <cmath>
 #include <set>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


stefdoerr:
Would it be possible to create a minor release of OpenMM for this? Or is there some timeline known for the next release?

peastman:
No current timeline.  We just released 8.2.  Unless a critical bug comes up, we'll want to wait a bit before doing a patch to give more problems time to show up so we can fix them all at once.

PR #4755, opened 2024-12-19
Fix residueTemplates argument in addMembrane

james-gardner-cresset:
This is similar to https://github.com/openmm/openmm/pull/4528 but was missed out. When the membrane and protein are combined a new topology is created. This invalidates the residueTemplates dict so it must be recreated before calling addSolvent.


peastman:
Looks good.  Thanks!

Issue #4750, opened 2024-12-12
modeller.addSolvent() for AMOEBA


chabi-fin:
Hello!

I have parameterized a small molecule for AMOEBA and want to simulate in explicit solvent in OpenMM. I see that modeller.addSolvent() can't be used with the AMOEBA ff--are there plans to expand this capability?

Otherwise, what is the current protocol for explicit solvation in OpenMM with AMOEBA? I suppose external tools are needed--any advice?

peastman:
It will work fine if you use `model='tip3p'`.  They're both three point water models.  It will use a water box that was pre-equilibrated with a different force field, but since you should always start by energy minimizing the system and then equilibrating it, that won't make any difference.

chabi-fin:
Hey, thanks for the reply. 
Yes, I see your point about using 3 point models to initiate a solvated system. 
I have some trouble to use this approach still. If I set `model="tip3p"`, then I have the error 
> ValueError: The ForceField does not specify a NonbondedForce

while if I include tip3p in my force field as `forcefield = app.ForceField(f"{ input_path }/AMOEBA-BIO-2018.xml",  "amber/tip3p_standard.xml")` then I get a conflict between the forcefields,

> ValueError: Residue template CA with the same override level 0 already exists.

Is there an inbuilt .xml for just tip3p which doesn't conflict with the AMOEBA?

peastman:
Since you closed this, I assume you figured it out?  Anyway, don't include tip3p.xml in the force field.  AMOEBA provides its own water model.

chabi-fin:
Hey, so I did figure it out, but using the above approach, because when I set model="tip3p", then I have the error

>    ValueError: The ForceField does not specify a NonbondedForce

due to using AMOEBA's water model

peastman:
You shouldn't edit the file.  Simply call `addSolvent()` as usual to add the water box and create your force field with `ForceField('amoeba2018.xml')`.

chabi-fin:
OK, must be an issue with my forcefield then.

Erikna00:
Hi,

I can add that i have the same problem as the previous speaker where modeller does not work with amoeba.
It seems that modeller does not recognize the nonbonded amoeba term as it complains of such a error.
I was able to bypass the issue by first creating a amber force field, adding solvent and then remaking my openff system generator with amoeba.

Best, Erik

peastman:
[...]

Issue #4735, opened 2024-11-22
Load Tinker files


peastman:
We should consider adding classes to load Tinker files.  This is related to #4661, which is about converting Tinker files to OpenMM force fields, but it would be even better to be able to load the files directly, the same way we do for Amber, CHARMM, and Gromacs files.  A main goal is to be able to simulate small molecules parameterized with poltype2.

Tinker divides information between files differently from other programs.  Amber and Gromacs each have a file that contains everything we need to make a Topology or a System (prmtop for Amber, top for Gromacs).  Then there's another file (inpcrd or gro) that provides coordinates and box vectors.  In contrast, Tinker requires all files to get any of these things.

There are three types of files: xyz, key, and prm.  There isn't really much difference between the key and prm files, though.  Exactly the same keywords can appear in either one.  Typically prm files define force fields and key files contain options specific to one simulation, but that's more a matter of convention.  The xyz file contains coordinates, but also bonds and atom types.  And periodic box vectors go in the key file, not the xyz file.

Given that, it might be best to have a single TinkerFiles class.  You would provide all the input files to the constructor.  It would expect there to be exactly one xyz file, and one or more key or prm files.  It would set `topology` and `positions` fields, and also provide a `createSystem()` method.

Tinker supports lots of force fields.  We don't need to support all of them.  The main priority is AMOEBA.  In the future we'll also want to support HIPPO.

cc @jmorado

JMorado:
I think this is a great idea, and thank you for putting together a plan. Initially, I wasn't sure if the model generator approach would be better, but I think its main disadvantage is that, in addition to focusing only on small molecules, it usually doesn't assume that any files are provided, since all the parameterization steps are handled internally.

I like this approach. I currently have a working converter from XYZ and key/prm files to OpenMM-readable XML files. I’ll take care of this and open a PR once it’s ready.

peastman:
This was implemented in #4769.

PR #4769, opened 2025-01-15
Reader of Tinker files

JMorado:
Hi,

Here is a draft PR of the class to load Tinker files discussed in #4735. The class is called `TinkerFiles`, and the constructor accepts a single `.xyz` file and one or more `.key`/`.prm` files. Upon initialisation, the topology is created, and OpenMM-compatible `.xml` files are stored in memory (and optionally written to disk). The topology can be accessed in a similar way as is done in related readers (e.g., `AmberPrmtopFile`), and the class also provides a `createSystem()` method that calls this method's counterpart in the `ForceField` class and passes to it the created `.xml` files. 

I decided to follow the approach of creating `.xml` files in `TinkerFiles` and passing them to the `ForceField` class because, in my view, this:

1. provides a way of writing `.xml` files from Tinker files, which is quite useful for portability.
2. retains all the logic used to create forces in one place (`ForceField`), avoiding code duplication and all the issues that this can cause.

Here are some usage examples (the files used are in attachment):

### Poltype ligand

```python
tinker = TinkerFiles('final.xyz', 'final.key')
forcefield = ForceField(*tinker.forceFields, "amoeba2018.xml")
modeller = Modeller(tinker.topology, tinker.positions)
modeller.addSolvent(forcefield, boxSize=Vec3(3.0, 3.0, 3.0)*nanometers, model='tip3p')
system = forcefield.createSystem(modeller.topology, nonbondedMethod=PME)
```

This example reads in the Tinker files of a ligand in vacuum parameterised using Poltype2 and solvates it using `Modeller`. For this example to work, I had to make small changes to `Modeller` so that it works with AMOEBA ForceFields. See the associated commits. I think these commits also solve issue #4750.

### Protein+ligand in explicit solvent

```python
tinker = TinkerFiles('tinker.xyz', ['tinker.key', 'amoebabio18.prm'])
system = tinker.createSystem(polarization="mutual", mutualInducedTargetEpsilon=1e-5, nonbondedMethod=PME, useDispersionCorrection=True)
```

This example builds the entire ligand+protein+water system using only Tinker Files. I obtained the definition of AMOEBA 2018 from [here](https://github.com/TinkerTools/tinker/tree/release/params). 
### Ligand in implicit solvent

```python
tinker = TinkerFiles('final.xyz', 'final.key')
system = tinker.createSystem(nonbondedMethod=NoCutoff, implicitSolvent=True)
```

This example can be used to simulate in implicit solvent. As I've coded it, the `implicitSolvent=True` parameter must be passed during the creation of the system. Please let me know if you think there are better ways to handle this.

## Unresolved issues / questions

1. I think the main issue in the implementation is the definition of the residues. As I've programmed it, the static method `_getResiduesFromBonds` forms entire residues by traversing bonded atoms. This means entire molecules, whether they are proteins, ligands, or water molecules, are considered a single residue. This is not a problem from the perspective of running simulations, but I understand that if one needs to write `.pdb` files, the residue names will not look good. This is the issue that is taking me the longest to solve, and I am running out of ideas on how to correctly set the residue names (and respective abbreviations) based solely on the information contained in the `.xyz` and `.key` files. I've attempted several approaches so far, some involving more complex mappings to the residues defined [here](https://github.com/openmm/openmm/blob/master/devtools/forcefield-scripts/residuesFinal.xml), but I've encountered several Tinker files I tested that contain residues not defined in this `.xml` file, and nothing seems to be robust. Any ideas on how to improve this would be greatly appreciated!
2. I am reading all the information on `biotypes` from `.key` files, but I am not using it. I read the Tinker manual but could not fully understand how use this information when writing the `.xml` files. 
3. I have a suspicion that the `AmoebaAngleTorsionForce` parameters defined [in the amoeba2018.xml file](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/data/amoeba2018.xml#L7121-L7134) are incorrect. At least, I cannot reproduce these values using the latest definition of the AMOEBA 2018 force field, and an energy decomposition comparison indicates my values are correct. Could these have been generated using a previous version? Can someone confirm if this is the case?
4. I have a couple of questions on the way implicit solvent works in OpenMM for the AMOEBA FF:
	1. The AMOEBA parameter files contain lines like `solute 51 3.3967 3.2550 4.7385` (atom type, diameter for APBS, diameter for ddCosmo, diameter for GK). I believe OpenMM only supports the GK model and the Grycuk Born radius. I assume that’s why the radius type is always set to `Bondi` [here](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/forcefield.py#L5859-L5868)? Would it make sense to read in the GK diameters and use those instead? I've come across some systems with K+ ions for which there's no radius definition in the current implementation.
	2. The overlap scaling factor is always set to 0.69. Is this the intended behaviour? Can't we support a more general implementation, like [this one](https://github.com/TinkerTools/tinker/blob/bdd412fcb752860edae49799737293cde40031c2/openmm/ommstuf.cpp#L3392-L3469)?

--- 

I added some tests and examples to this PR. I'll write the documentation once we've got the final implementation.

Any comments/suggestions to improve the implementation are welcome.
Many thanks in advance.

cc @jmichel80

[TinkerReaderExamples.tar.gz](https://github.com/user-attachments/files/18429261/TinkerReaderExamples.tar.gz)


peastman (commented):
Thanks!  I made a first quick pass through the code and added a few comments.

I'm not convinced of the approach of generating a force field internally.  I understand why you implemented it that way, but it's going to be very limiting.  It means the file loader can never support any feature that isn't supported by ForceField.  That's one of the main reasons for having the file importers, especially for features that don't fit well in the parameterization model used by ForceField.  For example, ForceField can't currently support Amber ff19 because of the way it determines some parameters, but the Amber prmtop file reader does support it.  The Tinker loader is likely to have similar applications.

The extra levels of translation also create more opportunities for things to go wrong.  For example, the xyz file tells us the exact type of every atom.  But then you throw that information out and instead ask the ForceField to redetermine the types based on bonds and elements.  There can be cases where multiple assignments are possible and the result may not match.

> This means entire molecules, whether they are proteins, ligands, or water molecules, are considered a single residue.

We need to find a better solution to that.  A Topology *must* be properly broken into residues, and it *must* follow standard PDB naming.  That's part of the definition of the Topology class.  Any Topology that doesn't is invalid and will break lots of things downstream.  For example, you can't parameterize it with a ForceField, or use Modeller to add or remove hydrogens, or write a valid PDB or PDBx/mmCIF file.

Tinker includes the XYZPDB program which converts from xyz format back to PDB.  It makes use of the seq file to get the sequence.  Perhaps we need to require it as well.  We should look at that program to understand exactly how it works, then do the same thing.

JMorado:
Thanks for the comments.

The idea then is to create the forces and add them to the system in this loader, like it's done [here](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/internal/amber_file_parser.py#L674) for AMBER prmtop files, right? I'm happy to change the implementation to this. I think it will take a bit longer for me to program it than the approach of generating a force field internally, but I'll keep committing to this PR, and we can discuss things as they progress.

I agree that the current approach to creating the residues is not ideal. I'll look into that program. I think we'll indeed need to require that sequence file.

In any case, I could leave the XML file utilities in this PR for now, perhaps in a separate class. Would it be worth having a general converter to OpenMM XML files?

peastman:
That sounds reasonable.  We can figure out what to do with the XML code later.

Most of the code needed for creating the forces is already in ForceField.  Rather than duplicate it, we can extract any shared code into a new `openmm/app/internal/amoebaforces.py` file that gets used in both places.

JMorado:
I have been making progress in recreating the topology of biomolecules using .xyz, .key, and .seq Tinker files. As mentioned above, we essentially need to recreate the logic of the `xyzpdb` script from Tinker ([source code here](https://github.com/TinkerTools/tinker/blob/release/source/xyzpdb.f)). 

Peptide-like and nucleic-like residues follow strict rules in Tinker files. I am leaving my notes on this in this comment as I make progress, and I will continue editing it, as it will be quite useful for future reference, testing, and debugging purposes.

## Peptide-like residues

Atoms highlighted in red are automatically found via an adapted BFS algorithm. This is more flexible than directly assigning the indices of hydrogen and side chain atoms, as some of these might be missing in united atom FFs supported by Tinker.

image


| **Residue**   | **Reference Atom** | **Requirements for Reference Atom**                                                                 | **Assigned Indices**               |
|---------------|--------------------|-----------------------------------------------------------------------------------|------------------------------------|
| BACKBONE      | N                  | Oxygen 3 bonds away                                                               | N(i), CA(i+1), C(i+2), O(i+3)      |
| H2N           | N                  | Oxygen 3 bonds away                                                               | N(i), CA(i+1), C(i+2), O(i+3)      |
| FOR           | C                  | Nitrogen directly connected<br>Oxygen directly connected                          | CA(i), O(i+1)                   |
| ACE           | C                  | Nitrogen 2 bonds away<br>Oxygen 2 bonds away                                      | CA(i), C(i+1), O                   |
| COH           | O                  | More than one neghbour atom<br>Oxygen 2 bonds away<br>Nitrogen 3 bonds away | O(i)                               |
| NH2           | N                  | Oxygen 2 bonds away<br>Nitrogen 3 bonds away                              | N(i)                               |
| NME           | N                  | Oxygen 2 bonds away<br>Nitrogen 3 bonds away                              | N(i), CA(i+1)                      |

## Nucleic-like residues

TODO

JMorado:
I've now added several methods that enable us to recreate the correct topology from Tinker files (the .seq file must be provided, otherwise each residue corresponds to an entire molecule). 

I've tested the current implementation with several examples from this folder: https://github.com/TinkerTools/tinker/tree/release/example, as well as with outputs from poltype, and the resulting topologies are consistently correct. There may still be some edge cases, but identifying those will require specific examples, which I'm not aware of at the moment.

I'm now starting to implement a `createSystem` method that is decoupled from its `ForceField` counterpart. Regarding the shared code, would it make sense to move the entire set of Amoeba generators from forcefield.py to a new openmm/app/internal/amoebaforces.py file? Or would it be better to somehow leave the parser functions of those classes in forcefield.py?

JMorado:
I've greatly simplified the implementation, as I found out recently at a conference that Tinker can now automatically determine sequences of amino acids and nucleotides by analysing the connectivity [TinkerTools/tinker@653c5bf119721ad024954e8a89b5f96c1555d1db). I've replicated the logic of [findnuc.f](https://github.com/TinkerTools/tinker/blob/release/source/findnuc.f) and [findpro.f](https://github.com/TinkerTools/tinker/blob/release/source/findpro.f) from Tinker, and now this loader doesn't require the sequence to be provided as input. I think this will make the UX much smoother. The only downside is that only all-atom topologies are supported, but it probably wouldn't be recommended to use this loader with Tinker input files for force fields other than AMOEBA anyway.

Let me know if there's any issue or feedback on this.

peastman (commented):
Sorry I took so long.  There was a lot of code to review!

It looks like this isn't finished yet, and there are still a bunch of forces to implement?

I didn't review TinkerXmlWriter, because we don't support creating a force field from a single parameterized system.  I can't think of many situations where that would make sense, but a lot of situations where it could lead people to do things they shouldn't.  I suggest just removing that code.

All the tests are failing because the xyz files aren't getting copied to the build directory.  Add them to the list of file types [here](https://github.com/openmm/openmm/blob/1ce5d91d9dedfdc273066fafa1a618bf05c25b85/wrappers/python/CMakeLists.txt#L50).

JMorado:
Hi @peastman, many thanks for reviewing! There are still all the nonbonded forces to implement, but I hope to get this into the code soon. I’ll also address the issues you’ve raised. This reader will support all all-atom force fields supported by Tinker. The only reason united-atom force fields won’t be supported for now is because of the automated way of infering the topology, which looks for explicit hydrogens. If there’s a lot of demand for supporting united-atom FFs, we can always add support later. The previous implementation I had supported that but with the requirement of passing the sequence file. I'll also remove the class for writing the XML files.

peastman:
Would you like me to implement some of the missing forces?  I can try AmoebaMultipoleForce and see how that goes.  I think at this point you've done all the hard parts, and what's left is mostly mechanical.

peastman:
I implemented AmoebaMultipoleForce and created a PR at https://github.com/JMorado/openmm/pull/1 with the changes.  I'll continue on to the other missing classes.

peastman:
I have AmoebaMultipoleForce and AmoebaVdwForce implemented.  Let's merge https://github.com/JMorado/openmm/pull/1 and then split up the remaining work.  What would it be most helpful for me to do?

JMorado:
Many thanks, @peastman.

I’m now working on implementing the remaining forces: `AmoebaTorsionTorsion`, `AmoebaWcaDispersion`, and `AmoebaGeneralizedKirkwood`. I think those are the only ones still missing.

peastman:
Sounds good.  I'll work on the documentation updates for it.

JMorado:
I've added the missing forces and refactored the previous implementation to follow the builder pattern you used.

I'll now address some of the issues you raised in the review and add various tests focusing on matching the energies output by Tinker.

peastman:
Thanks!  Let me know when it's ready for another review.

Skimming through the changes I did notice a couple of things.  You removed all the `@private` comments in forcefield.py.  Those need to be there.  Otherwise all the internal classes will get included in the API documentation.  And in `AmoebaVdwForceBuilder.__init__()`, you removed the checks for settings that OpenMM doesn't support.  Those also need to be there.

JMorado:
OK, makes sense, I’ll revert that. I removed the checks for settings in `AmoebaVdwForceBuilder.__init__()`  because in some of the tests I was using Tinker files for OPLS, which applies different scalings (e.g. https://github.com/TinkerTools/tinker/blob/release/params/oplsaa08.prm). Those won’t be supported, but that should be fine since I think the primary use of the loader is reading Tinker files for the AMOEBA FF.

peastman:
At some point we could extend the AMOEBA code to support other scale factors.  For now, the calculation code hardcodes those values, so we need to tell the user if their files ask for something different.  Energy would be calculated incorrectly.

peastman:
I have a PR at https://github.com/JMorado/openmm/pull/2 that adds some documentation, and fixes some issues I came across in the process.

The latest changes broke TestForceField.py.  I tried to get it working, but it wasn't obvious what to do.  AmoebaTorsionTorsionForceBuilder has two different methods both called `_getChiralAtomIndex()`, but taking different numbers of arguments.

JMorado:
I've fixed the implementation and all tests in `TestForceField.py` are now passing. 

Regarding the GK force, I've kept the option for `radiusType == "amoeba"` even though it is not currently used. The overlap factor `shct = 0.69` is also always constant despite a mapping being defined. Is this the intended behavior?

peastman:
It looks like that probably is some very old code.  I think we can get rid of `getAmoebaTypeRadius()` and `getObcShct()`.  As you say, they're never used.  I can't even find any options in the Tinker manual they would correspond to.  According to the manual, `shct` is the scale factors for HCT, which is a different method from OBC, so what is `getObcShct()` supposed to mean?

peastman:
My latest changes are in https://github.com/JMorado/openmm/pull/3, mostly fixing bugs and removing duplicated code.  Here are the main things I see that are still left to do.

- There are two forces still missing (stretch-torsion and angle-torsion).
- The tests for TinkerFiles still fail.
- Some of the comments above still need to be addressed, for example the incorrect residue and atom names.
- There's still a lot of duplicated code that could be cleaned up, but we don't need to hold up the beta for that.

JMorado:
Many thanks. I’m working on getting the tests to pass now, going force by force, adding the unit conversions you identified, and removing some duplicated code. I think there will still be some duplication in the end, especially in the post-processing parts of the generators in `ForceField`, but we can later think about the best way to tidy those up. I'll get back to the review comments after.

JMorado:
Here's a list of the forces that I've already tested and that currently match the outputs produced by Tinker:

- [x] `AmoebaBondForce`
- [x] `AmoebaAngleForce`
- [x] `AmoebaOutOfPlaneBendForce` (still getting a small mismatch for `test_Amoeba18PoltypePhenolWater`, viz. 2.11 (OpenMM) vs 2.0572 (Tinker) kcal/mol -- not sure if this is just numerical or not, I can't see anything obvious causing this)
- [x] `AmoebaInPlaneAngleForce`
- [x] `AmoebaStretchBendForceBuilder`
- [x] `AmoebaTorsionForce` (`PeriodicTorsionForce`)
- [x] `AmoebaPiTorsionForce`
- [x] `AmoebaUreyBradleyForce` (`HarmonicBondForce`)
- [x] `AmoebaWcaDispersionForce`
- [x] `AmoebaGeneralizedKirkwoodForce`
- [x] `AmoebaMultipoleForce`
- [x] `AmoebaVdwForce`

~~There are also forces not yet covered by tests, for which we should add a system containing them:~~

- [x] `AmoebaTorsionTorsionForce` (now covered -- still needs to be tested)

~~Finally, the forces that are still to be implemented:~~

- [x] `AmoebaStretchTorsionForce` (tested and working)
- [x] `AmoebaAngleBendForce` (tested and working)

peastman:
Do you want me to work on any of those things?

JMorado:
If you wouldn't mind having a go at implementing the missing forces, that would be a great help in speeding things up! I'll pick things up Monday and keep debugging/wrapping up.

peastman:
The latest changes are in https://github.com/JMorado/openmm/pull/4.  I implemented the stretch-torsion and angle-torsion forces and fixed a few other issues.  Here are the remaining problems.

- The out-of-plane-bend energy is still off.  The difference is small, but too big to just be numerical error.  Perhaps it's related to the ordering of atoms?
- `test_Amoeba13ForcesImplicit` fails with an IndexError.  The problem appears to be in the topology builder.  The xyz file lists 22 atoms, but the Topology that gets created only has 10.

JMorado:
Many thanks! I'll debug both issues today.

JMorado:
`test_Amoeba13ForcesImplicit` now passes; there was an issue with the identification of cap groups. I’m now adding more tests for different topologies.

Yes, you’re right, I think there’s still something wrong with the ~~out-of-plane bend energies~~... the nucleic acid test also fails for that term.

**EDIT**: Actually, I think there's something off with the angles in general. It's probably related to how they're identified because it seems to affect all angle-dependent energy terms (including the stretch-bend term).

peastman:
It looks like we're just about there!  Two tests are still failing, but really they're correct:

> FAILED TestTinkerFiles.py::TestTinkerFiles::test_Amoeba18Peptide - AssertionError: 1.9697 != 1.9696760582532609 to a relative tolerance of 1e-05
> FAILED TestTinkerFiles.py::TestTinkerFiles::test_Amoeba18PoltypePhenolWater - AssertionError: 2.0572 != 2.0571607723935754 to a relative tolerance of 1e-05

1.9696760582532609 rounds to 1.9697, and 2.0571607723935754 rounds to 2.0572, meaning they agree perfectly.  The tests only fail because of the limited number of decimal places in the Tinker output.  In those two assertions, we can pass 1e-4 for `rtol`.

Other than that, is there anything still left to do?

JMorado:
I've just commited the last few small fixes and I think (mostly) everything is working now! The only test that is failing is `test_Amoeba18Peptide`, specifically in this line: 

`self.assertEnergyEqual(-488.0403 - 110.9042,energies["AmoebaMultipoleForce"])`.

The error is:

> FAILED TestTinkerFiles.py::TestTinkerFiles::test_Amoeba18Peptide - AssertionError: -598.9445000000001 != -598.8658617916452 to a relative tolerance of 1e-05.

The difference is quite small, and the `ForceField` implementation produces the same results.

In terms of what's left to do, I'd say the following could be a priority:

1. Address remaining feedback from previous reviews.
2. Reduce the duplicated code between `TinkerReader` and `ForceField`. Parts of `TinkerReader` can be moved to the internal `amoebaforces` module, and `ForceField` can be adapted to reuse it.
3. Type hinting and code documentation.
4. Constraints are not currently supported -- should we add support for them?

Gathering feedback from the user community will also be essential for understanding the UX and identifying any potential edge cases the current implementation might be missing.

I’ll start by working on point 1., as it’s the most straightforward. Please let me know your thoughts on the remaining points.

peastman:
That's great!  Items 2 and 3 are good to do, but it's not critical they be finished before we release the beta.  I don't think we need constraints.  AMOEBA is never intended to be used with constraints, and that's the only force field TinkerFiles currently supports.

JMorado:
I'm starting to clean up the implementation to reduce the duplicated code between `ForceField` and `TinkerFiles` and will keep commiting to this PR over the next few days. For reference, here is the current progress checklist for the refactoring:

- [x] AmoebaBondForce
- [ ] AmoebaAngleForce
- [ ] AmoebaOutOfPlaneBendForce 
- [ ] AmoebaInPlaneAngleForce
- [x] AmoebaStretchBendForce
- [ ] AmoebaTorsionForce 
- [x] AmoebaPiTorsionForce
- [x] AmoebaUreyBradleyForce
- [x] AmoebaWcaDispersionForce
- [x] AmoebaGeneralizedKirkwoodForce
- [ ] AmoebaMultipoleForce
- [x] AmoebaVdwForce
- [ ] AmoebaTorsionTorsionForce
- [x] AmoebaAngleTorsionForce 
- [x] AmoebaStretchTorsionForce

peastman:
If everything other than the cleanup is done, shall we merge this and start a new PR?  That way we can move forward with the beta.

JMorado:
Sure, I've reverted the cleanup commits and will open a new PR so that the refactoring is fully contained.

peastman:
Tests are running now.  If everything passes I'll merge it.

peastman:
The Windows build had a hard crash in TestTinkerFiles:

```
Windows fatal exception: access violation

Thread 0x00000c94 (most recent call first):
  File "D:\conda-envs\build\Lib\site-packages\execnet\gateway_base.py", line 534 in read
  File "D:\conda-envs\build\Lib\site-packages\execnet\gateway_base.py", line 567 in from_io
  File "D:\conda-envs\build\Lib\site-packages\execnet\gateway_base.py", line 1160 in _thread_receiver
  File "D:\conda-envs\build\Lib\site-packages\execnet\gateway_base.py", line 341 in run
  File "D:\conda-envs\build\Lib\site-packages\execnet\gateway_base.py", line 411 in _perform_spawn

Current thread 0x000020a8 (most recent call first):
  File "D:\conda-envs\build\Lib\site-packages\openmm\openmm.py", line 5991 in getState
  File "D:\a\openmm\openmm\build\python\tests\TestTinkerFiles.py", line 30 in computeAmoebaEnergies
  File "D:\a\openmm\openmm\build\python\tests\TestTinkerFiles.py", line 91 in test_Amoeba18PoltypePhenolWater
```

Possibly that was just some weird intermittent problem.  I'll try rerunning it once the other builds finish.

JMorado:
It's strange that it only fails for `test_Amoeba18PoltypePhenolWater`. The particularities of this test are the Urey-Bradley force, which isn't covered by the other tests, and the `phenol.prm` file. Could this file be read incorrectly on Windows for some reason, thus leading to a corrupted system?

peastman:
It's hard to say.  I don't have an actual Windows machine, but I'll see if I can reproduce it in a VM.

peastman:
A fix is at https://github.com/JMorado/openmm/pull/5.  It also brings this up to date with the main branch again.

JMorado:
Many thanks -- I've just merged it.

peastman:
Great, thanks.  Test are running.  Cross your fingers.

peastman:
All tests passed, which means we can merge it!

JMorado:
Excellent! I'll open a PR for the refactoring.

PR #4770, opened 2025-01-15
Reduce LennardJonesGenerator memory consumption with many NBFixPair-matching pairs

epretti:
Currently, when an `<NBFixPair>` entry is present in a `<LennardJonesForce>`, `openmm.app.forcefield.LennardJonesGenerator` creates an entry in a dictionary for every _pair_ of matching atom types.  This can lead to significant memory consumption even when simply loading a force field XML file, _e.g._, loading [openmmforcefields' version of CHARMM36](https://github.com/openmm/openmmforcefields/blob/main/openmmforcefields/ffxml/charmm/charmm36_nowaters.xml) creates over 156 million entries in the `nbfixTypes` table and consumes over 31 GiB of memory.

This memory consumption can be cut significantly by storing the parameters for each `<NBFixPair>` entry and keeping track of which atom types can match which entries.  An entry for a particular pair of atom types can be found with a set intersection.

The test case for this is also updated to cover all combinations of (i) NBFix/non-NBFix type pairs, (ii) 1-4/non-1-4 interactions, and (iii) same/different types in a pair.  This test should pass with the old as well as this new implementation.

Previously, if multiple `<NBFixPair>` entries corresponded to the same atom type pair, the earlier set of parameters would be silently overwritten by the later set.  An error is now raised if this condition is detected when creating a system.  I checked all of the CHARMM36+water-ions combinations in openmmforcefields that contain such entries to ensure that this should never happen with any pair of atom types, but this behavior could be changed later if we ever encounter a force field where such overwriting is done intentionally.

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -6,9 +6,9 @@ Simbios, the NIH National Center for Physics-Based Simulation of
 Biological Structures at Stanford, funded under the NIH Roadmap for
 Medical Research, grant U54 GM072970. See https://simtk.org.
 
-Portions copyright (c) 2012-2024 Stanford University and the Authors.
+Portions copyright (c) 2012-2025 Stanford University and the Authors.
 Authors: Peter Eastman, Mark Friedrichs
-Contributors:
+Contributors: Evan Pretti
 
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the "Software"),
@@ -2553,20 +2553,38 @@ class LennardJonesGenerator(object):
 
     def __init__(self, forcefield, lj14scale, useDispersionCorrection):
         self.ff = forcefield
-        self.nbfixTypes = {}
+        self.nbfixParameters = []
+        self.nbfixTypes1 = defaultdict(set)
+        self.nbfixTypes2 = defaultdict(set)
         self.lj14scale = lj14scale
         self.useDispersionCorrection = useDispersionCorrection
         self.ljTypes = ForceField._AtomTypeParameters(forcefield, 'LennardJonesForce', 'Atom', ('sigma', 'epsilon'))
 
     def registerNBFIX(self, parameters):
         types = self.ff._findAtomTypes(parameters, 2)
+
         if None not in types:
+            sigma = _convertParameterToNumber(parameters['sigma'])
+            epsilon = _convertParameterToNumber(parameters['epsilon'])
+
+            # Retrieve the index of nbfixParameters into which this sigma and
+            # epsilon will be stored, then register this index with the atom
+            # types that should have this sigma and epsilon applied.
+            nbfixIndex = len(self.nbfixParameters)
+            self.nbfixParameters.append([sigma, epsilon])
             for type1 in types[0]:
-                for type2 in types[1]:
-                    epsilon = _convertParameterToNumber(parameters['epsilon'])
-                    sigma = _convertParameterToNumber(parameters['sigma'])
-                    self.nbfixTypes[(type1, type2)] = [sigma, epsilon]
-                    self.nbfixTypes[(type2, type1)] = [sigma, epsilon]
+                self.nbfixTypes1[type1].add(nbfixIndex)
+            for type2 in types[1]:
+                self.nbfixTypes2[type2].add(nbfixIndex)
+
+    def getNBFIX(self, type1, type2):
+        nbfixIndices = (self.nbfixTypes1[type1] & self.nbfixTypes2[type2]) | (self.nbfixTypes2[type1] & self.nbfixTypes1[type2])
+        if nbfixIndices:
+            if len(nbfixIndices) > 1:
+                raise ValueError('Multiple NBFixPair entries match atom types %s-%s.' % (type1, type2))
+            return self.nbfixParameters[nbfixIndices.pop()]
+        else:
+            return None
 
     def registerLennardJones(self, parameters):
         self.ljTypes.registerAtom(parameters)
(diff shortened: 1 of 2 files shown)


wrappers/python/openmm/app/forcefield.py
```diff
@@ -2553,20 +2553,38 @@ class LennardJonesGenerator(object):
 
     def __init__(self, forcefield, lj14scale, useDispersionCorrection):
         self.ff = forcefield
-        self.nbfixTypes = {}
+        self.nbfixParameters = []
+        self.nbfixTypes1 = defaultdict(set)
+        self.nbfixTypes2 = defaultdict(set)
```
peastman, line 2558: Do you actually need two different dicts?  It looks to me like you could use a single `nbfixTypes` field, and that would simplify the code that uses it.  Which type you call "type1" and "type2" is arbitrary, so they don't need to be kept separate.

wrappers/python/openmm/app/forcefield.py
```diff
@@ -2553,20 +2553,38 @@ class LennardJonesGenerator(object):
...
-                    epsilon = _convertParameterToNumber(parameters['epsilon'])
-                    sigma = _convertParameterToNumber(parameters['sigma'])
-                    self.nbfixTypes[(type1, type2)] = [sigma, epsilon]
-                    self.nbfixTypes[(type2, type1)] = [sigma, epsilon]
+                self.nbfixTypes1[type1].add(nbfixIndex)
+            for type2 in types[1]:
+                self.nbfixTypes2[type2].add(nbfixIndex)
+
+    def getNBFIX(self, type1, type2):
+        nbfixIndices = (self.nbfixTypes1[type1] & self.nbfixTypes2[type2]) | (self.nbfixTypes2[type1] & self.nbfixTypes1[type2])
```
peastman, line 2581: For example, this would simplify to just

```python
nbfixIndices = self.nbfixTypes[type1] & self.nbfixTypes[type2]
```

peastman (commented):
Very clever!  This looks like a good approach.

The one thing I want to make sure of is that all the set intersections aren't slowing it down too much.  Can you time how long it takes `createForce()` to run when using CHARMM36 for some moderate sized protein, such as ApoA1 (found in `examples/apoa1.pdb`)?  Probably it will be fine, but I just want to make sure.

wrappers/python/openmm/app/forcefield.py
epretti: I tried that at first and found it was incorrect.  Suppose you have a single NBFIX entry between types A and B.  `nbfixTypes` would look like `{"A": {0}, "B": {0}}`.  So `nbfixTypes["A"] & nbfixTypes["B"] = {0}`, which is correct, but `nbfixTypes["A"] & nbfixTypes["A"] = {0}` as well, which is wrong.  I suppose one could split into two dictionaries for same- and differently-typed cases instead, or use some other method for bookkeeping to avoid this problem, but thought this approach was the most obvious.  If you would prefer to handle this differently, I can change the implementation.

wrappers/python/tests/TestForceField.py
```diff
@@ -914,68 +914,81 @@ def test_LennardJonesGenerator(self):
...
     def test_NBFix(self):
         """Test using LennardJonesGenerator to implement NBFix terms."""
-        # Create a chain of five atoms.
+        # Create a chain of seven atoms.
 
         top = Topology()
         chain = top.addChain()
         res = top.addResidue('RES', chain)
-        top.addAtom('A', elem.oxygen, res)
+        top.addAtom('A', elem.carbon, res)
```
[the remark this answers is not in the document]
epretti: Thanks for catching this.  I didn't realize why that was being done.  I will restore the multiple elements.

wrappers/python/openmm/app/forcefield.py
peastman: That makes sense.

epretti:
> The one thing I want to make sure of is that all the set intersections aren't slowing it down too much. Can you time how long it takes `createForce()` to run when using CHARMM36 for some moderate sized protein, such as ApoA1 (found in `examples/apoa1.pdb`)? Probably it will be fine, but I just want to make sure.

Using the version of CHARMM that comes with OpenMM, full system creation (via `createSystem()`) averages 4.41 s with the old approach and 4.44 s with the new approach for `apoa1.pdb`.

Unfortunately, this built-in force field doesn't have very many NBFix entries, unlike the version of CHARMM that comes with openmmforcefields, and I believe I am now running into another open issue with that force field (openmm/openmmforcefields#170) that causes OpenMM to consume inordinate amounts of memory.  I will look into that next because otherwise openmmforcefields' CHARMM appears to just be unusable.

I would still like to confirm with the openmmforcefields CHARMM that system creation isn't slowed significantly, since this force field is the reason I made this fix in the first place, so I will try to get it working.  Force field XML _loading_, on the other hand, for this force field, is significantly sped up (from 167.1 s to 2.5 s) with reduced memory usage (from 33 GB to 971 MB peak).

epretti:
I was able to regenerate the force field from CHARMM data files and it no longer exhibits the aforementioned issue.  System creation time for the existing implementation is 32.8 s; for the new implementation, it is 29.5 s.  I'm not sure exactly why it's actually faster, but a separate run with the Python profiler turned on indicates about 0.29 s spent in `registerNBFIX()` during force field loading and 0.05 s spent in `getNBFIX()` during system creation, both small compared to residue template matching which takes up most of the creation time.  So the overhead of the set operations is relatively small.

peastman:
Ok, that's great!  Is this ready to merge?

epretti:
Yes, it should be ready.

peastman:
Ok, thanks!

PR #4776, opened 2025-01-20
Variable distance-based displacements for ATMForce

egallicc:
These changes allow the specification of particle displacements based on the vector distances between given particles. For example, this sets the displacement of particle 0 as the vector distance of particle 2 relative to particle 1:
```
atmforce->addParticle(2, 1);
atmforce->addParticle(Vec3(0, 0, 0));
atmfiorce->addParticle(Vec3(0, 0, 0));
```
After the coordinate transformation, the position of particle 0 is
```
pos[0] + (pos[2]-pos[1])
```

The extension serves two purposes:

- The displacements can be specified in internal system's coordinates rather than fixed lab-frame coordinates. For example, a ligand can be displaced in some direction relative to the receptor regardless of the orientation of the receptor in the simulation box. Also, the position of two molecules can be swapped even if their relative position changes with time. 
- It implements the coordinate swapping and variable displacements for relative binding free energies between macromolecular ligands as described in the work [Relative Binding Free Energy Estimation of Congeneric Ligands and Macromolecular Mutants with the Alchemical Transfer with Coordinate Swapping Method](https://arxiv.org/abs/2412.19971) 

The extension is backward compatible with the original fixed-lab displacements without API changes.

Questions, comments, and suggestions are welcome.


openmmapi/include/openmm/ATMForce.h
```diff
@@ -88,7 +88,23 @@ namespace OpenMM {
...
+ * the change in energy when the first particle is displaced by the vector distance from particle 1 to particle 2:
+ *
+ * \verbatim embed:rst:leading-asterisk
+ * .. code-block:: cpp
+ *
+ *    ATMForce *atmforce = new ATMForce("u0 + Lambda*(u1 - u0)");
+ *    atm->addGlobalParameter("Lambda", 0.5);
+ *    atm->addParticle(2, 1);
+ *    atm->addParticle(Vec3(0, 0, 0));
+ *    atm->addParticle(Vec3(0, 0, 0));
```
peastman, line 102: Is this intended to illustrate a typical use?  If so, I don't understand it.  I can see displacing every particle along the same direction, to move a ligand in a direction defined in internal coordinates.  But when would you move some particles in a relative direction and others in a fixed direction?  And when would you ever want to specify a displacement of (0, 0, 0)?

Do you also need a distance scale?  I would think you would often want to use two particles to specify a direction, but then specify the distance to move separately.

platforms/common/src/CommonKernels.cpp
```diff
@@ -8079,18 +8079,63 @@ void CommonCalcATMForceKernel::initialize(const System& system, const ATMForce&
...
-        displVector0[i] = mm_float4(displacement0[0], displacement0[1], displacement0[2], 0);
-    }
-    displ1.initialize<mm_float4>(cc, cc.getPaddedNumAtoms(), "displ1");
-    displ1.upload(displVector1);
-    displ0.initialize<mm_float4>(cc, cc.getPaddedNumAtoms(), "displ0");
-    displ0.upload(displVector0);
+    vector<int>  pj1Vector(cc.getPaddedNumAtoms());
+    vector<int>  pi1Vector(cc.getPaddedNumAtoms());
+    vector<int>  pj0Vector(cc.getPaddedNumAtoms());
+    vector<int>  pi0Vector(cc.getPaddedNumAtoms());
```
peastman, line 8085: Using four separate arrays like this will lead to inefficient memory access in the kernels.  If you pack them into a single int4 array, it will be faster.

openmmapi/include/openmm/ATMForce.h
egallicc: In this example, the second and third particles are stationary. That is, they are not displaced. This is the case for most of the particles in binding applications. For example,  the receptor and solvent atoms are not displaced and are entered in ATMForce with zero displacement.

True, fixed and particle distance displacements are unlikely to be used together in the same application. It is one or the other. Nevertheless, stationary particles have to be specified somehow. Maybe we should have an `addParticle()` method for that (no arguments).

platforms/common/src/CommonKernels.cpp
egallicc: Good point, I will do that.

openmmapi/include/openmm/ATMForce.h
peastman: What about a distance?  Will the displacement to apply always be exactly equal to the displacement between two other particles, or will it sometimes be a fixed distance in the direction defined by two particles?

I guess this depends on what the use cases are.  If the goal is to move it from one pocket to a different one, you would probably use the displacement between them.  If the goal is to move it into solvent, you would probably want a fixed distance, and use the two other particles to determine which direction is "outward".

What are the intended use cases?

egallicc: Right, as of now, we do not have a use case for a "direction + magnitude" displacement mode.

The immediate application of the new protocol is for the calculation of the relative binding free energy (RBFE) between two congeneric ligands that share a common core and differ in a peripheral side group attached to an anchor atom of the common core. These include macromolecular ligands that differ by the sidechain of one residue. This involves: 

- The displacement of the side group of one ligand to the position of the anchor atom of the other ligand and vice versa. For example, if the indexes of the anchor atoms of the two ligands are `a1` and `a2`, each atom of the side group of the first ligand is assigned a displacement `pos[a2] - pos[a1]`, which of course varies with time depending on the positions of the anchor atoms. 
- The swapping of the coordinates of the corresponding atoms of the common cores of the two ligands to preserve the structures of the bonded terms. So if atoms `i` and `j` are corresponding atoms, atom `i` is displaced by `pos[j]-pos[i]`, which results in atom `i` taking the coordinates of atom `j`.

Furthermore, we envision that the new variable distance-based displacements will eventually replace the current fixed lab-frame displacements for the other two main use cases of ATMForce for molecular binding:

- Absolute binding free energies (ABFE): where the ligand is displaced from a position in the solvent to the receptor binding site (or the other way around for dissociation). This is currently implemented as a fixed lab-frame displacement of the ligand, which could be replaced with a variable distance-based displacement between a reference atom of the ligand and a virtual particle placed in the solvent and attached to the receptor by restraints expressed in internal coordinates.
- Relative binding free energies between unrelated ligands (scaffold hopping). These are currently implemented using a fixed lab-frame displacement similar to ABFE, except that the two ligands are displaced in opposite directions. We intend to replace these fixed displacements with variable displacements based on the distance between two reference atoms of the two ligands. Here, the user will pick the two reference atoms rather than a fixed displacement. But we have not tried this yet.

So, I think we will eventually stop using fixed lab-frame displacements in our work.

peastman: Will this change be sufficient to address those and all other future use cases you can imagine?

I'd much rather err on the side of generality and change the API only once, rather than keep changing incrementally.  The latter is likely to produce an incoherent API.

egallicc: It is difficult to predict which coordinate transformations people will find useful in the future. But for now, I can't think of an existing or potential computational protocol that is not covered by the proposed API.

peastman: It isn't orthogonal, because `getParticleParameters()` returns both descriptions, the fixed displacements and also the atom indices.  Once a feature is released, we try very hard never to break source compatibility.  This PR does.  If someone has written

```python
displacement1, displacement0 = force.getParticleParameters(i)
```

it would no longer run since the number of return values has changed.  And if we then remove the fixed displacements in the future, it would change again and we would break source compatibility again.  That's why I want to give the API some more thought, first to see if there's a way we can avoid breaking compatibility now, and second to make absolutely certain we won't need to break it in the future.

egallicc: Right. Okay. Is there a remedy/workaround that has worked with other Forces?

We had anticipated that the fixed lab-frame displacements were temporary because binding free energies should not depend on the receptor position and orientation. Currently, the receptor must be restrained to a specific orientation dictated by the lab-frame displacement. We did not know at the time what form of internal coordinate displacements would be appropriate. I think this is it.

peastman: > We had anticipated that the fixed lab-frame displacements were temporary

I wish I'd know that when we first implemented it.  I certainly would have designed the API differently.  Once we release an interface and declare it to be supported, that means we have to maintain compatibility.  There's no workaround.

Can you describe all possible extensions that you might conceivably want in the foreseeable future?  Let's design something that's sufficiently general, we're confident we can support it and not have to break it.

egallicc: The most general coordinate transformation I can think of is one in which the coordinates of a particle are some arbitrary function of the positions of some particles:
```
newpos[i] = f_i( i_1, i_2, i_3, ... )
```

Where, for example,
```
f_i = pos[i] + h
```
would correspond to the existing fixed lab-frame displacements and
```
f_i = pos[i] + (pos[j] - pos[k])
```
to the variable displacement/coordinate swapping transformation proposed here.

egallicc: Shall we hold off on incorporating this extension in the mainline code for the time being as we gather more experience with variable displacements? A few months from now, we should have a stronger basis to formulate a solid proposal.

peastman: Here's what I suggest as an API.  This will be extensible, so we can add new types of displacements in the future without breaking compatibility.

- Create a class called `ATMForce::ParticleDisplacement` to describe how a particle is displaced.  This is an abstract class.  Subclasses describe particular types of displacements.
- Create subclasses for the particular types of displacements we want to implement.  `ATMForce::FixedDisplacement` will describe the current behavior of fixed lab frame displacements.  Other subclasses can describe other types of displacements.
- Add a new version of `addParticle()` that takes a `ParticleDisplacement` object, along with `getParticleDisplacement()` and `setParticleDisplacement()` methods.  Your example code would become

```c++
atm->addParticle(new ATMForce::FixedDisplacement(Vec3(0, 0, 0)));
```

- Mark the current `addParticle()`, `getParticleParameters()`, and `setParticleParameters()` methods as deprecated.  They'll still work, but people will be encouraged to use the new methods instead.  Internally they'll create a `FixedDisplacement`.  If you call `getParticleParameters()` on a particle that uses a different displacement type, it will throw an exception.
- The serialization code will need to be revised.  Currently it records a particle like this:

```c++
particles.createChildNode("Particle").setDoubleProperty("d1x", d1[0]).setDoubleProperty("d1y", d1[1]).setDoubleProperty("d1z", d1[2])
        .setDoubleProperty("d0x", d0[0]).setDoubleProperty("d0y", d0[1]).setDoubleProperty("d0z", d0[2]);
```

We can add a new `type` property to specify which displacement class to use, and all the other properties will be specific to the displacement class.  When reading old files, it will assume `type="fixed"`.

> A few months from now, we should have a stronger basis to formulate a solid proposal.

Sounds good.

egallicc: >>  A few months from now, we should have a stronger basis to formulate a solid proposal.

> Sounds good.

Agreed.

> Here's what I suggest as an API.

Yes, it looks like a good extensible solution. However, I am unsure how efficiently it could be implemented for GPU platforms. I guess particles can be sorted into groups depending on their transformations, and then kernels would transform the coordinates and gradients for each group. Using each transformation's `type` parameter should help with the grouping.

peastman: Efficiency shouldn't be a problem.  Within the kernel we'll still just be handling a fixed set of hardcoded operations.

egallicc:
A quick update on our progress. The results of tests of internal coordinate displacements look promising. We will do more to confirm the validity of the approach.

Concerning the proposed extensible API:

> Add a new version of addParticle() that takes a ParticleDisplacement object, along with getParticleDisplacement() and setParticleDisplacement() methods.

Would a Force object make sense in this context? For example, a CustomExternalForce with "x+h" as an expression could be used to translate an atom by some fixed amount or just "x" to assign the coordinate of one atom to another. More complicated transformations involving the coordinates of multiple atoms could be given in terms of a CustomCompoundBondForce or similar.

I don't know how custom Forces could be used to implement the actual coordinate transformation internally. (Is it even possible?) However, all the information about the transformation can be encoded and queried using the Force objects with existing APIs—just a thought.

peastman:
I was figuring on something much simpler than that.  There would still only be a small list of supported displacement types, initially perhaps only two.  And each of them would be hardcoded in the kernel.  The public API is more general so we can add other types in the future, but those too would probably be specialized, hardcoded types.

peastman:
It is, of course, conceivable that we could have a type that involves a custom expression for the position.  We would need to figure out what the expression could depend on.

egallicc:
I have completed a draft of the new ATMForce interface proposed above, which maintains the current API while deprecating it.

It defines transformation objects: `ATMTransformation` (a base virtual class), and `ATMFixedDisplacement` and `ATMVectordistanceDisplacement` (derived classes), which hold the parameters for each coordinate transformation. `ATMFixedDisplacement` refers to the fixed lab frame displacements that we supported so far, and `ATMVectordistanceDisplacement` refers to the variable particle distance displacements introduced above.

To create a stationary particle, one that is not displaced, do:
```
atmforce->addParticle();
```

A particle with fixed displacement is created with something like:
```
atmforce->addParticle(new ATMFixedDisplacement(d));
```
where `d` is the `Vec3` displacement vector.

Finally, a particle with variable particle distance displacement is created with, for example:
```
atmforce->addParticle(new ATMVectordistanceDisplacement(1, 0));
```
to set the displacement of the added particle as the vector distance between particles 1 and 0.

The new API also supports setting and getting the transformation objects of existing particles:
```
const ATMTransformation* transformation = getParticleTransformation(i); 
```
which can be queried for the corresponding parameters:
```
if ( transformation->getName() == "FixedDisplacement") {
   const Vec3 d1 = dynamic_cast<const ATMFixedDisplacement*>(transformation)->getFixedDisplacement1();
   const Vec3 d0 = dynamic_cast<const ATMFixedDisplacement*>(transformation)->getFixedDisplacement0();
}
```
and similarly for `ATMVectordistanceDisplacement`, which is labeled as `"VectordistanceDisplacement"`

Two maps, `ATMTransformationType` and `ATMTransformationName`, enable mapping from transformation names to the numerical transformation types used in serialization.

The current API:
```
atmforce->addParticle(d);
atmforce->getParticleParameters(i, d1, d0);
etc.
```
is supported but generates a deprecation warning when the first particle is created.

egallicc:
NOTES:

- Building of the C wrappers currently fails (hence the CI build failures), complaining about conflicting declarations of `addParticle`.  I don't know why. I assume the C wrappers support function overloading? Turning off `OPENMM_BUILD_C_AND_FORTRAN_WRAPPERS` lets the build proceed without problems (tested on Linux).
- The `addParticle(ATMFixedDisplacement(d))` and  `addParticle(ATMVectordistanceDisplacement(i,j))` Python wrappers fail at runtime, probably due to class ownership. I could not add `addParticle` to the `STEAL_OWNERSHIP` list in `swigInputConfig.py` because it does not appear to support overloaded functions. As a workaround, particles can be added from Python by first creating them and then adding their transformation, taking advantage that  `setParticleTransformation` is in the `STEAL_OWNERSHIP` list:
```
p = atmforce.addParticle()
atmforce.setParticleTransformation(p, ATMFixedDisplacement(d))
``` 
- I attempted but failed to add the `ATMTransformation` objects as members of the `ATMForce` class because I could not find a way to have SWIG assign units to the output of a member class. For example, I set
```
("ATMFixedDisplacement", "getFixedDisplacement1") : ("unit.nanometer", ()),
```
in `swigInputConfig.py` but something like the below did not work
```
("ATMForce::ATMFixedDisplacement", "getFixedDisplacement1") : ("unit.nanometer", ()),
```
- The SWIG interface would not build unless I set a "precedence" in the `Vec3` typemaps. In `typemaps.i` I did:
```
%typemap(typecheck, precedence=SWIG_TYPECHECK_DOUBLE_ARRAY, fragment="Py_SequenceToVec3") Vec3 { ...
```
I hope it is fine.

peastman:
I'm reviewing the code now.  I suggest a few changes to how classes are named.  Rather than prepending "ATM" to everything, let's make them contained classes.  `ATMTransformation` would become `ATMForce::Transformation`.  Or should it be `Displacement`, since the subclasses all use that word instead of transformation?  In [my proposal above](https://github.com/openmm/openmm/pull/4776#discussion_r1929586223) I suggested calling it `ATMForce::ParticleDisplacement`.  `ATMFixedDisplacement` would become `ATMForce::FixedDisplacement`.  `ATMVectordistanceDisplacement` would be `ATMForce::VectorDistanceDisplacement` (all words should be capitalized), except I'm not sure that name makes sense.  You don't specify either a vector or a distance.  Maybe `ParticleOffsetDisplacement` or something like that?

egallicc:
> I'm reviewing the code now.

Thanks!

> Rather than prepending "ATM" to everything, let's make them contained classes.

I tried to do that. But I got stuck at the SWIG unit assignments step. 
```
("ATMForce::ATMFixedDisplacement", "getFixedDisplacement1") : ("unit.nanometer", ()),
```
did not work. See the Notes above.

egallicc:
Yes, I agree, `ParticleOffsetDisplacement` is a better name.

peastman:
> I tried to do that. But I got stuck at the SWIG unit assignments step.

We'll figure out how to make it work.  It might take minor changes to the script, but it shouldn't be difficult.

egallicc:
> I think the implementation may be unnecessarily expensive. 

I will look into it. One thing to keep in mind is that as more transformations are added, the single kernel to support them all could become unmanageable.

openmmapi/include/openmm/ATMForce.h
```diff
@@ -400,6 +526,38 @@ class OPENMM_EXPORT ATMForce : public OpenMM::Force {
...
+     */
+    const ATMFixedDisplacement* getParticleFixedDisplacementTransformation(int index) const;
+
+    /**
+     * Returns the VectordistanceDisplacement transformation object associated with the particle or NULL
+     *
+     * @param index           the index of the particle
+     * @return                the pointer to the VectordistanceDisplacement transformation object
+     */
+    const ATMVectordistanceDisplacement* getParticleVectordistanceDisplacementTransformation(int index) const;
```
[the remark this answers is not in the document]
egallicc: I don't like them either. I had to add them to make the Python interface work. Python/SWIG refused to expose the methods of the derived classes, such as `ATMFixedDisplacement::getFixedDisplacement1()`, through the `ATMTransformation` object returned by `getParticleTransformation()`. Hence, the need to query the type of the transformation from the `ATMTransformation` object and call the corresponding specific get method to acquire the derived class object. Is there a trick to avoid this horrible workaround?

openmmapi/src/ATMForce.cpp
```diff
@@ -91,21 +93,103 @@ void ATMForce::setEnergyFunction(const std::string& energy) {
     energyExpression = energy;
 }
 
+int ATMForce::addParticle() {
+    particles.push_back(ParticleInfo(particles.size(), NULL));
```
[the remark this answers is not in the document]
egallicc: Sure. Would there still be a need for the caller of `getTransformation()` to check whether the pointer is NULL before dereferencing it?

egallicc:
> > I tried to do that. But I got stuck at the SWIG unit assignments step.
> 
> We'll figure out how to make it work. It might take minor changes to the script, but it shouldn't be difficult.

I looked at the SWIG configuration script. It is, unfortunately, too impenetrable for me. I would need some guidance.

The situation for the C/Fortran wrapper is probably similar.

Also, please see my code review replies above regarding the limitations I encountered with the Python interface. If there is a way to address them, it is well beyond my SWIG expertise.

I will start making the changes you suggested to the C++ interface as soon as there is some clarity about the implications for the Python interface.

Thanks!

openmmapi/include/openmm/ATMForce.h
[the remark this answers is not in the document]
peastman: SWIG handles this with factories, which we define [here](https://github.com/openmm/openmm/blob/643473cee91fbf8e232d485012f7b8cf13d68829/wrappers/python/src/swig_doxygen/swigInputBuilder.py#L248).

openmmapi/src/ATMForce.cpp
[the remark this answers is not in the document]
peastman: Since we're changing the return type to be a reference, NULL values aren't possible.

openmmapi/src/ATMForce.cpp
```diff
@@ -91,21 +93,103 @@ void ATMForce::setEnergyFunction(const std::string& energy) {
...
+	displacement1 = Vec3(0, 0, 0);
+	displacement0 = Vec3(0, 0, 0);
+    }
+    else if (particles[index].transformation->getName() == "FixedDisplacement") {
+	displacement1 = dynamic_cast<const ATMFixedDisplacement*>(transformation)->getFixedDisplacement1();
+	displacement0 = dynamic_cast<const ATMFixedDisplacement*>(transformation)->getFixedDisplacement0();
+    }
+    else {
+        throw OpenMMException("getParticleParameters(): particle lacks fixed displacements");
+    }
```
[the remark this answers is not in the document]
peastman: Defining a factory will cause SWIG to automatically return the correct subclass.

peastman:
Don't worry about the wrappers for the moment.  Let's get the C++ interface working the way we want.  Once we're happy with it, I can get the wrappers working.  SWIG sometimes requires a bit of black magic, but there's almost always a way to do what you want.

egallicc:
I revised the C++ API according to the suggestions. As agreed, I have not addressed the Python, C, and Fortran wrappers.

I opted for `CoordinateTransformation` as the name for the base class of the `FixedDisplacement` and `ParticleOffsetDisplacement` to cover potential future coordinate transformations other than displacements.

serialization/src/ATMForceProxy.cpp
```diff
@@ -41,6 +43,35 @@ using namespace std;
...
 
+void ATMForceProxy::storeParams(int numParticles, ATMForce& force, const SerializationNode& particles) const {
+    for (auto& p : particles.getChildren()){
+
+	//support older serialized ATMForces that did not store the transformation type
+	//or the particle offset displacement indexes
+        if (p.hasProperty("type")) {
+	    //normal behavior
+	    int transformation_type = p.getIntProperty("type");
+	    if (transformation_type == ATMForce::CoordinateTransformationType.at("FixedDisplacement")) {
```
peastman, line 54: As noted above, we should get rid of those fields.  I suggest something like

```c++
string type = p.getStringProperty("type");
if (type == "fixed") {
   ...
}
else if (type == "offset") {
   ...
}
```

peastman (commented):
The API looks much better now.  I'll get to work on the wrappers and see if I can get them working.

[My comment above](https://github.com/openmm/openmm/pull/4776#pullrequestreview-2955083448) about the kernels still applies.  Could they be merged so we don't get the overhead of launching so many additional kernels?

Also see [above](https://github.com/openmm/openmm/pull/4776#pullrequestreview-2565730004) about formatting.  There's still some incorrect formatting, and especially a lot of indentation being done with tabs instead of spaces.  Can you do a find/replace to turn them into spaces?

egallicc:
> The API looks much better now. 

Thank you for reviewing!

> I'll get to work on the wrappers and see if I can get them working.
> 

Thanks!

> [My comment above](https://github.com/openmm/openmm/pull/4776#pullrequestreview-2955083448) about the kernels still applies. Could they be merged so we don't get the overhead of launching so many additional kernels?
> 

Understood. Can we please postpone this optimization aspect to a future PR? There are some coordinate transformations that we are thinking about, and I am concerned that the kernel complexity might complicate the work. I promise we will get to the kernel optimization very soon if the current set of transformations is not expected to grow.

> Also see [above](https://github.com/openmm/openmm/pull/4776#pullrequestreview-2565730004) about formatting. There's still some incorrect formatting, and especially a lot of indentation being done with tabs instead of spaces. Can you do a find/replace to turn them into spaces?

Yes, I will fix the remaining formatting issues. (Calling out to any old emacser out there for a good `.emacs` file for OpenMM's C++ formatting!)

And I will take care of the other issues as soon as possible. 

Thank you for confirming the new documentation. I wanted to go through a "hybrid" version to highlight the changes.

peastman:
> Can we please postpone this optimization aspect to a future PR?

Ok.

egallicc:
> a good `.emacs` file for OpenMM's C++ formatting!)
> 

For reference, this appears to do the trick:
```
 '(c-default-style "stroustrup")
 '(indent-tabs-mode nil)
```

egallicc:
I addressed the issues with the C++ implementation. Let me know if there is anything I can do to smooth out the process of updating the wrappers.

peastman:
Fortran.  What a modern language!  It truncates all function names to 63 characters, because why would any name need to be longer than that?  Which means it can't distinguish between these functions, leading to a compile error:

```
OpenMM_ATMForce_ParticleOffsetDisplacement_getDestinationParticle1
OpenMM_ATMForce_ParticleOffsetDisplacement_getDestinationParticle0
```

I'll make it skip the second one, and very likely no one will ever try to call it from Fortran.  It's just annoying.

peastman:
I just created a pull request to your repo that fixes the wrappers.

egallicc:
I converted `TestAPIUnits.py` to the new ATMForce API. However, it fails due to missing units in the return values of `getFixedDisplacement1` even though the corresponding statements are in `swigInputConfig.py`:
```
("FixedDisplacement", "getFixedDisplacement1") : ("unit.nanometer", ()),
("FixedDisplacement", "getFixedDisplacement0") : ("unit.nanometer", ()),
```

I can't figure out the issue. Any ideas?

peastman:
This fixes it.

```
diff --git a/wrappers/python/src/swig_doxygen/swigInputBuilder.py b/wrappers/python/src/swig_doxygen/swigInputBuilder.py
index 5aa22f493..434a1cea3 100755
--- a/wrappers/python/src/swig_doxygen/swigInputBuilder.py
+++ b/wrappers/python/src/swig_doxygen/swigInputBuilder.py
@@ -439,6 +439,7 @@ class SwigInputBuilder:
 
     def writeMethods(self, classNode):
         methodList=getClassMethodList(classNode, self.skipMethods)
+        fullClassName = getText("compoundname", classNode)
 
         #write only Constructors
         for items in methodList:
@@ -627,7 +628,7 @@ class SwigInputBuilder:
 
                 if addText:
                     self.fOutPythonappend.write("%pythonappend")
-                    self.fOutPythonappend.write(" OpenMM::%s::%s(" % key)
+                    self.fOutPythonappend.write(" %s::%s(" % (fullClassName, methName))
                     sepChar=''
                     outputIndex=0
                     for pNode in paramList:
```

egallicc:
> This fixes it.

Thanks!

egallicc:
The API C++ documentation does not build. I have tried several experiments to no avail. The issue (duplicate definitions) appears specific to nested classes such as `CoordinateTransformation` and `FixedDisplacement`.

peastman:
This fixes it:

```
diff --git a/docs-source/api-c++/CMakeLists.txt b/docs-source/api-c++/CMakeLists.txt
index 33e1be43f..2c9fcf62d 100644
--- a/docs-source/api-c++/CMakeLists.txt
+++ b/docs-source/api-c++/CMakeLists.txt
@@ -56,6 +56,7 @@ add_custom_command(
 
 add_custom_command(
     OUTPUT "${CMAKE_BINARY_DIR}/api-c++/index.html"
+    COMMAND "${CMAKE_COMMAND}" -E rm "${CMAKE_CURRENT_BINARY_DIR}/generated/*::*"
     COMMAND "${PYTHON_EXECUTABLE}" -m sphinx . "${CMAKE_BINARY_DIR}/api-c++" -W --keep-going # Promote warnings to errors to catch undocumented classes
     WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
     DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/conf.py"
```
Breathe was creating separate files for ATMForce and for each of the inner classes.  But Sphinx includes documentation for inner classes on the page with the main class, so it produced a warning that a class was documented twice.  And we tell it to treat warnings as errors to catch if we miss documenting a class.

This change deletes the extra files for the inner classes so they only get documented once.

egallicc:
> This change deletes the extra files for the inner classes so they only get documented once.

Thanks!

peastman:
Sorry for taking a while to get back to this.  One last thing: atmforce.cc still has a lot of indentation done with tabs instead of spaces.  Can you do a find and replace to fix that?  Then we can merge this and do the additional optimization in another PR.

egallicc:
> One last thing: atmforce.cc still has a lot of indentation done with tabs instead of spaces. 

Sorry about that. It escaped the earlier cleaning. It's fixed now.

> do the additional optimization in another PR.

Yes. Thanks!

peastman:
Thanks!

Issue #4784, opened 2025-01-29
Calling unit.x in a frozen class causes memory leak like issue


JPMarquardt:
Title. A minimum reproducible snippet:

```
from dataclasses import dataclass
from openmm import unit
from tqdm import tqdm

@dataclass(frozen=True)
class test_unit:
    def __post_init__(self):
        unit.nano * unit.meter

for i in tqdm(range(round(10e10))):
    a = test_unit()
    del a
```

The number of iterations per second decreases precipitously.

Python v: 3.11.7
Openmm v: 8.1.1

peastman:
What leads you to think it's a memory leak?

Running your script with either `frozen=True` or `frozen=False` gives basically the same performance for me.

epretti:
I think the `dataclass` here shouldn't make a difference.  The slowdown in timing can be observed with
```
from openmm import unit
from tqdm import trange

for i in trange(10**6):
    unit.nano * unit.meter
```
What I think is happening is that `unit.meter_base_unit._conversion_factor_to` is filling up with identical entries that are then looped over in `unit.BaseUnit.define_conversion_factor_to`.  This gets called every time `unit.SIPrefix.__mul__` is invoked.  Possibly related is the note on line 67 of `wrappers/python/openmm/unit/prefix.py`:
```
# TODO - check for existing BaseUnit with same name, symbol, and factor
```
This isn't done in the current implementation; instead, a new object is created every time.

I can try to look into how to fix this.

peastman:
That would be great.

PR #4785, opened 2025-01-31
Compare and track openmm.unit.BaseUnit objects by name

epretti:
This addresses #4784.  Note that previously, the behavior of `BaseUnit` with respect to comparison of and interaction between different unit objects having the same name was ambiguous.  For instance, I would assume that after
```
import openmm.unit as unit
dim = unit.BaseDimension("test")
a = unit.BaseUnit(dim, "test_a", "test_a")
b = unit.BaseUnit(dim, "test_b", "test_b")
b.define_conversion_factor_to(a, 10)
b2 = unit.BaseUnit(dim, "test_b", "test_b")
```
either `b2` should behave equal to `b` and have conversion factor of `1` to `b` and `10` to `a`, or `b2` should not behave equal to `b` and refuse to be converted to `a` or `b`.  However,
```
for i in (a, b, b2):
    for j in (a, b, b2):
        try:
            print(f"{i.conversion_factor_to(j):4.1f}", end=" ")
        except Exception:
            print("????", end=" ")
    print()
```
prints
```
 1.0  0.1  0.1
10.0  1.0  1.0
????  1.0  1.0
```
so the behavior here is not transitive, which I think classifies as a bug.  In the new approach, the missing factor is reported as `10.0`.

In this new approach, if two `BaseUnit` objects have the same dimension and name, they compare equal and hash to the same value.  (I am assuming one is not supposed to modify `BaseUnit`s after they have been created.)  If we want different semantics than this, we can take a different approach.  The tables of conversion factors within `BaseUnit` objects has been changed to a global registry that keeps track of the conversion factors between them by their names.  If this change would break something that I missed, let me know so I can add a test case to catch it, as tests currently pass.

As for performance, in the old approach, `[unit.nano * unit.meter for _ in range(5000)]` takes 11 s, compared to the new approach in which it takes 0.047 s, and also doesn't eat up memory.

### DIFF
--- a/wrappers/python/openmm/unit/baseunit.py
+++ b/wrappers/python/openmm/unit/baseunit.py
@@ -68,7 +68,7 @@ class BaseUnit(object):
         self.dimension = base_dim
         self.name = name
         self.symbol = symbol
-        self._conversion_factors[self.dimension][self.name][self.name] = 1.0
+        BaseUnit._conversion_factors[self.dimension][self.name][self.name] = 1.0
 
     def __lt__(self, other):
         """
@@ -137,7 +137,7 @@ class BaseUnit(object):
             raise TypeError('Cannot define conversion for BaseUnits with different dimensions.')
         assert(factor != 0)
         assert(self != other)
-        conversion_factors = self._conversion_factors[self.dimension]
+        conversion_factors = BaseUnit._conversion_factors[self.dimension]
         conversion_factors_self = conversion_factors[self.name]
         conversion_factors_other = conversion_factors[other.name]
         # import all transitive conversions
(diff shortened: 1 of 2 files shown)


wrappers/python/openmm/unit/baseunit.py
```diff
@@ -47,23 +49,26 @@ class BaseUnit(object):
...
             Quantity string descriptions.
         """
         self.dimension = base_dim
         self.name = name
         self.symbol = symbol
-        self._conversion_factor_to = {}
-        self._conversion_factor_to[self] = 1.0
-        self._conversion_factor_to_by_name = {}
-        self._conversion_factor_to_by_name[self.name] = 1.0
+        self._conversion_factors[self.dimension][self.name][self.name] = 1.0
```
peastman, line 71: Did you mean that to be `BaseUnit._conversion_factors` rather than `self._conversion_factors`?  As written it's a per-object table, not a global one.

epretti: Right, thanks.  It worked because it was defined as a class attribute and never set from `self`, but that was confusing and inconsistent.  Changed.

peastman:
Looks good.  Thanks!

PR #4794, opened 2025-02-10
Fixes precision mismatch when appending to an XTC file with 9 atoms or less

### LINKED ISSUE #4793: "xtc_read(): precision mismatch" when creating a new reporter appending to an existing file
Hi,

I encountered an error of which I could find no information anywhere else. When running the following code with some dummy systems with the same atoms, I get an error as soon as I end up on the last step. I am running a Free Energy Perturbation, using a range of different systems forming one trajectory. This error seems to be caused by using a new xtc reporter on the same file. If I use the same xtc reporter, I get the correct results.

```
with open('system1.xml') as f:
    str = f.read()
system1 = mm.XmlSerializer.deserialize(str)
with open('system2.xml') as f:
    str = f.read()
system2 = mm.XmlSerializer.deserialize(str)
pdb = mmapp.PDBFile('topology.pdb')
topology = pdb.topology
positions = pdb.positions

integrator1 = mm.VerletIntegrator(1.0*mmunit.femtoseconds)
simulation1 = mmapp.Simulation(topology, system1, integrator1)
simulation1.context.setPositions(positions)
xtcrep1 = mmapp.XTCReporter('traj.xtc',1)
simulation1.reporters.append(xtcrep1)
simulation1.step(10)
positions1 = simulation1.context.getState(getPositions=True).getPositions()

integrator2 = mm.VerletIntegrator(1.0*mmunit.femtoseconds)
simulation2 = mmapp.Simulation(topology, system2, integrator2)
simulation2.context.setPositions(positions1)
xtcrep2 = mmapp.XTCReporter('traj.xtc',1,append=True)
simulation2.reporters.append(xtcrep2)
simulation2.step(10)
```

epretti:
Should fix #4793.  Note that this issue in the reading code is triggered by attempting to append to an XTC because the XTC writer reads before appending to check that the number of atoms is consistent.

### DIFF
--- a/wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
+++ b/wrappers/python/openmm/app/internal/xtc_utils/src/xtc.cpp
@@ -81,7 +81,9 @@ struct XTCFrame {
 
     // Read the next frame from the XTC file and store it in this object
     int readNextFrame(XDRFILE* xd) {
-        float in_prec;
+        // Preinitialize in_prec for the precision check below since it may not
+        // be modified by read_xtc if the coordinates to read are not compressed
+        float in_prec = prec;
         auto* p_ptr = reinterpret_cast<rvec*>(positions.data());
         int status = read_xtc(xd, natoms, &step, &time, box, p_ptr, &in_prec);
         if (status == exdrOK && prec != in_prec) {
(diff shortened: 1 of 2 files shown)


epretti:
> Can you add a test case that would have caught the problem?

Added.  This is just a copy of the general test that uses a smaller system.  Some of the other tests appear to be similar slightly modified versions, so I just went with that.  It is a bit repetitive, but I think it is worth testing the full write/read since the internal XTC code takes different paths and uses different formats depending on whether you are writing compressed or uncompressed coordinates.  I checked that this test fails on the master branch and passes with the change.

peastman:
Looks good.  Shall I merge it once the tests finish?

epretti:
Sure, it should be good to go.

PR #4795, opened 2025-02-11
Properly handle and test multiple registration of atom types without elements

epretti:
`ForceField` allows multiple atom type definitions with the same name if their parameters are identical and forbids them if they differ.  `registerAtomType()` allows atom types' elements to be `None` (by not specifying an `element` attribute in the `<Type>`): https://github.com/openmm/openmm/blob/39808f12cb7dc0465748cc3884b50594ef568523/wrappers/python/openmm/app/forcefield.py#L452-L456 but this is not handled correctly in the duplicate checking code, which assumes that the element is not `None` and that the attribute is always present:
https://github.com/openmm/openmm/blob/39808f12cb7dc0465748cc3884b50594ef568523/wrappers/python/openmm/app/forcefield.py#L447-L449 This came up in trying to split a Drude force field, which uses atom types without elements for the Drude particles, into multiple files.

The change here should fix this problem.  It also adds several tests for the duplicate checking functionality, which was not previously covered by the test suite.  Do the tests make sense (do they match the expected behavior)?

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -444,7 +444,7 @@ class ForceField(object):
         if name in self._atomTypes:
             #  allow multiple registrations of the same atom type provided the definitions are identical
             existing = self._atomTypes[name]
-            elementsMatch = ((existing.element is None and 'element' not in parameters) or ('element' in parameters and existing.element.symbol == parameters['element']))
+            elementsMatch = ((existing.element is None and 'element' not in parameters) or (existing.element is not None and 'element' in parameters and existing.element.symbol == parameters['element']))
             if existing.atomClass == parameters['class'] and existing.mass == float(parameters['mass']) and elementsMatch:
                 return
             raise ValueError('Found multiple definitions for atom type: '+name)
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -444,7 +444,8 @@ class ForceField(object):
         if name in self._atomTypes:
             #  allow multiple registrations of the same atom type provided the definitions are identical
             existing = self._atomTypes[name]
-            if existing.atomClass == parameters['class'] and existing.mass == float(parameters['mass']) and existing.element.symbol == parameters['element']:
+            elementsMatch = ((existing.element is None and 'element' not in parameters) or ('element' in parameters and existing.element.symbol == parameters['element']))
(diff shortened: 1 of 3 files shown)


wrappers/python/openmm/app/forcefield.py
```diff
@@ -444,7 +444,8 @@ def registerAtomType(self, parameters):
         if name in self._atomTypes:
             #  allow multiple registrations of the same atom type provided the definitions are identical
             existing = self._atomTypes[name]
-            if existing.atomClass == parameters['class'] and existing.mass == float(parameters['mass']) and existing.element.symbol == parameters['element']:
+            elementsMatch = ((existing.element is None and 'element' not in parameters) or ('element' in parameters and existing.element.symbol == parameters['element']))
```
peastman, line 447: This check could be slightly improved.  If the existing type does not have an element (`existing.element` is None) but the new type does (`'element'` is in `parameters`), you'll get an exception from trying to access `existing.element.symbol`.  That's not terrible, since it's supposed to throw an exception anyway, but it will be a confusing and less helpful exception.

epretti: Right, that case causes an `AttributeError`.  A check to ensure that the existing element is not `None` before retrieving its symbol is now added, and tests now ensure that a `ValueError` is raised regardless of the order of the types with and without the element specification.

epretti:
All the CI is now failing, apparently because of a deprecation: <https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down>

peastman:
This looks like a bug on their part.  The notice says they deprecated v1 and v2.  We specify v3.3.2, which is supposed to still be supported.  And the error message confirms that:

> Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v3.3.2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions.

Probably they put in a check for old versions, but wrote it wrong.  Try updating to v4 perhaps?

epretti:
> Probably they put in a check for old versions, but wrote it wrong. Try updating to v4 perhaps?

Looks like incorrect/incomplete documentation on GitHub's part.  Not only all of v1 and v2 but some v3.x.x and v4.x.x versions were also being deprecated, including the one we were using.  See https://github.com/actions/cache/discussions/1510 which has the correct information.  Seems to be running now with the latest version.

peastman:

epretti:
Yes, it should be ready to go.

PR #4812, opened 2025-02-27
Fix issues with virtual site handling

epretti:
This fixes two problems with virtual site handling that came up during the CHARMMconversion.

1. In `CharmmPsfFile`, code to add exceptions between lone pairs and Drude particles bonded to the same parent was only executing when Drude particles were present.  It failed to run if multiple lone pairs were present but there were no Drude particles in the system, as is the case for TIP5P water.  Lone pairs repelling each other shouldn't be a problem for sampled configurations in this case of rigid water molecules, but it adds a constant offset to the energy that shouldn't be there.

2. In `ForceField`, virtual sites in a residue would not have their `excludeWith` atom indices updated during patching.  This could cause exclusions for virtual sites in patched residues to get computed based on nonsensical neighbors, leading to incorrect energies and forces.


peastman:
Thanks!

PR #4819, opened 2025-03-04
Add PME optimizer and rigid body scripts

epretti:
Adds these two scripts from the old SimTK PyOpenMM repository, along with a README, to `/examples/extras/`.


peastman:
Looks good, thanks.

PR #4835, opened 2025-03-09
feature(forces): Use caching to speed up CustomCVForce and possibly the global variables too

lukeczapla:
The change proposed addresses the issue of speeding up simulations like aMD and Metadynamics by caching the global variables (of `CustomIntegrator`) and collective variables (of `CustomCVForce`).  It has a Python wrapper to add the new functions `getCVTimeSeries()` and `getGlobalVariableTimeSeries()` so that you can get this data and all of the Integrators consider the presence of the CVs and the `CustomIntegrator` will cache the global variables too.   The centralized `Context` provides these new functions.  
Everything seemed happy but HelloArgonInC.c, wasn't sure how to resolve this one but the explanation for why was very apparent.

### DIFF
--- a/openmmapi/include/openmm/Context.h
+++ b/openmmapi/include/openmm/Context.h
@@ -284,6 +284,14 @@ public:
      * belong to exactly one molecule.
      */
     const std::vector<std::vector<int> >& getMolecules() const;
+    /**
+     * If CustomCVForce is defined, it will return the time series of CollectiveVariables of size equal to the length of the last step (or size 0)
+     */
+    std::vector<std::vector<double>> getCVTimeSeries();
+    /**
+     * If CustomIntegrator is defined, it will return the time series of GlobalVariables of size equal to the length of the last step (or size 0)
+     */
+    std::vector<std::vector<double>> getGlobalVariableTimeSeries();
 private:
     friend class ContextImpl;
     friend class Force;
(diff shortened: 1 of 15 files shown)


peastman:
You're doing something complicated to achieve something simple.  There's no need for a new API, and as I've said repeatedly, there's nothing to optimize as far as global parameters are concerned.  They're already extremely fast, and this wouldn't benefit them.

There's room to optimize CustomCVForce by having it cache CV values, but that can be done entirely through internal changes to it.  The caching is easy.  The hard part is figuring out when the cache needs to be discarded.

lukeczapla:
Okay, updated the beginning comment there with something about HelloArgonInC **if you have any input there on fixing that**, let me copy it here, could address it perhaps next weekend: 

The change proposed addresses the issue of speeding up simulations like aMD and Metadynamics by caching the global variables (of `CustomIntegrator`) and collective variables (of `CustomCVForce`). It has a Python wrapper to add the new functions `getCVTimeSeries()` and `getGlobalVariableTimeSeries()` so that you can get this data and all of the Integrators consider the presence of the CVs and the CustomIntegrator will cache the global variables too. The centralized `Context` provides these new functions.
Everything seemed happy but HelloArgonInC.c, wasn't sure how to resolve this one but the explanation for why was very apparent.

lukeczapla:
Also, it **seems** that the cache doesn't need to be discarded because it's recomputed every time you call `calcForcesAndEnergy()` on the `CustomCVForce` and cleared in every Integrator before the steps.

lukeczapla:
I think it's the wrong one anyway, let me

PR #4852, opened 2025-03-14
Correctly assign CMAPs to torsions across multiple CMAPTorsionForce sections

### LINKED ISSUE #4850: Multiple CMAPTorsionForce sections result in incorrect map assignment
Currently, if two separate `<CMAPTorsionForce>` sections are present in a force field, as can happen if two force field files with CMAPs are being combined, OpenMM will create a single `CMAPTorsionForce` in resulting systems.  However, the indices of `map` attributes of any `<Torsion>` tags in `<CMAPTorsionForce>` sections other than the first will point to `<Map>` entries starting from the first section instead of the section that they are a part of.  For example, if
```
<CMAPTorsionForce>
    <Map>1 2 3 4</Map>
    <Torsion map="0" class1="A" class2="A" class3="A" class4="A" class5="B" />
</CMAPTorsionForce>
```
is followed by
```
<CMAPTorsionForce>
    <Map>5 6 7 8</Map>
    <Torsion map="0" class1="A" class2="A" class3="A" class4="B" class5="B" />
</CMAPTorsionForce>
```
then the `<Torsion>` in the second section points to the `<Map>` in the first section.  I would expect this `<Torsion>` to refer to the `<Map>` in *its own* section.  If that is the desired behavior, I will submit a PR with a fix (just an offset of the map index to point to the start of a section's own maps) and a new test case.  I wanted to check because CMAPs work differently from other bonded forces in that the parameters are defined "out-of-band" rather than with each entry, and I don't know if we want to provide any facility for a `<Torsion>` to refer to a `<Map>` in another section.

epretti:
Should fix #4850.  Adds a test that would have caught the issue.

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -2392,6 +2392,7 @@ class CMAPTorsionGenerator(object):
             ff.registerGenerator(generator)
         else:
             generator = existing[0]
+        mapOffset = len(generator.maps)
         for map in element.findall('Map'):
             values = [float(x) for x in map.text.split()]
             size = sqrt(len(values))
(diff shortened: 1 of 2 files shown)


peastman (approved):
Looks good.

PR #4867, opened 2025-03-27
Remove includes for Dart

peastman:
This removes some ancient code from the CMake scripts that hasn't been used in a long time.


peastman:
It looks like CMake 4.0 was just released and broke a bunch of things.  For the moment I've pinned CI to 3.x.

peastman:
Jenkins is also failing, but I don't think it's anything related to this PR.  It isn't able to check out the code from github.

PR #4871, opened 2025-03-30
Drude wrapper

### LINKED ISSUE #1293: What's up with plugins/amoeba/wrappers?
The C and FORTRAN wrappers don't include any of the plugins, but the amoeba plugin has a separate facility and separate make targets for generating C and FORTRAN wrappers in `plugins/amoeba/wrappers`. None of the other plugins have this. It seems kind of strange.

chen850512337:
I followed the openmm wrapper and amoeba wrapper to implement the Drude wrapper.
These changes can be compiled successfully to generate the `DrudeOpenMMCWrapper.h` and `DrudeOpenMMFortranModule.f90` in the install/include folder.
Resolves #1293


peastman:
Thanks!

PR #4875, opened 2025-04-01
Attempting to fix XTC/DCD restarting

### LINKED ISSUE #4873: Wrong step in DCD/XTCreporter when restarting simulations
Hi, I just noticed what I assume is a bug in the XTCreporter and DCDreporter class

https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/xtcfile.py#L146

If you look at how the step is calculated

```py
step = (self._modelCount - 1) * self._interval + self._firstStep
```

When appending, the `_modelCount` variable is overwritten here: https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/xtcfile.py#L57

And the `_firstStep` value is passed here: https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/xtcreporter.py#L110 directly from the `simulation.currentStep`

This leads to the following behaviour if I add a `print("STEP", step, "TIME", time, self._modelCount, self._interval, self._firstStep, self._dt)` right before writing the XTC frame
```
STEP 50 TIME 0.2 1 50 50 0.004
STEP 100 TIME 0.4 2 50 50 0.004
STEP 150 TIME 0.6 3 50 50 0.004
STEP 200 TIME 0.8 4 50 50 0.004
STEP 250 TIME 1.0 5 50 50 0.004
RESTART
STEP 550 TIME 2.2 6 50 300 0.004
STEP 600 TIME 2.4 7 50 300 0.004
STEP 650 TIME 2.6 8 50 300 0.004
STEP 700 TIME 2.8 9 50 300 0.004
STEP 750 TIME 3.0 10 50 300 0.004
```
As you can see, after the restart, the `_firstStep` is 300 since that indeed is the currentStep of the simulation, but also the `_modelCount` is 6. This means that it calculates wrongly the step as 550 instead of 300 which it should.

I took a look at the DCDreporter and it does nearly the same, but with an offset of +50 compared to XTCreporter, so after the restart it calculates the step as 600 instead of 300.

stefdoerr:
Fixes #4873 by writing directly the simulation.currentStep into the XTC and DCD files.

I'm making here some assumptions that the first step written in a file will be at `step=interval` but I think this is a valid assumption?

Also assuming that you would only append to an XTC/DCD file if you were resuming a previous simulation (i.e. via a checkpoint) and not creating a new Simulation from scratch which would set the currentStep to 0

### DIFF
--- a/wrappers/python/openmm/app/dcdfile.py
+++ b/wrappers/python/openmm/app/dcdfile.py
@@ -75,7 +75,6 @@ class DCDFile(object):
         self._topology = topology
         self._firstStep = firstStep
         self._interval = interval
-        self._modelCount = 0
         if is_quantity(dt):
             dt = dt.value_in_unit(picoseconds)
         dt /= 0.04888821
@@ -84,8 +83,6 @@ class DCDFile(object):
         if topology.getUnitCellDimensions() is not None:
             boxFlag = 1
         if append:
-            file.seek(8, os.SEEK_SET)
-            self._modelCount = struct.unpack('<i', file.read(4))[0]
             file.seek(268, os.SEEK_SET)
             numAtoms = struct.unpack('<i', file.read(4))[0]
             if numAtoms != topology.getNumAtoms():
@@ -98,7 +95,7 @@ class DCDFile(object):
             header += struct.pack('<4i', 164, 4, topology.getNumAtoms(), 4)
             file.write(header)
 
-    def writeModel(self, positions, unitCellDimensions=None, periodicBoxVectors=None):
+    def writeModel(self, positions, unitCellDimensions=None, periodicBoxVectors=None, currentStep=0):
         """Write out a model to the DCD file.
 
         The periodic box can be specified either by the unit cell dimensions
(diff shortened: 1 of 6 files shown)


peastman:
I see a few problems.  You changed a public API in an incompatible way by removing the `firstStep` argument.  We can't do that.  It will break existing code that specifies that argument.

Similarly, you added `currentStep` as a new required argument.  Technically it's optional since it has a default value, but you'll get an invalid file if you don't specify it, so really it's required.

Another problem is that you broke the handling of long trajectories:

https://github.com/openmm/openmm/blob/adf1843f8d2bdcc8bcb32ebefa337d03d471a0ac/wrappers/python/openmm/app/dcdfile.py#L133-L141

On modern hardware, it's easy to run a simulation with more than $2^{31}$ steps.  We need to make sure we never try to write a number to the file that exceeds the allowed range.  You changed it to write the current step index, which may be invalid:

https://github.com/openmm/openmm/blob/b835619922924ffb8e0b0f4236d856985b79db95/wrappers/python/openmm/app/dcdfile.py#L143-L145

stefdoerr:
Superseded by https://github.com/openmm/openmm/pull/4879 which doesn't modify the public API of the classes

PR #4879, opened 2025-04-02
Fixing XTC/DCD time and step writing

stefdoerr:
Fixes https://github.com/openmm/openmm/issues/4873

### DIFF
--- a/wrappers/python/openmm/app/dcdfile.py
+++ b/wrappers/python/openmm/app/dcdfile.py
@@ -130,7 +130,7 @@ class DCDFile(object):
         file = self._file
 
         self._modelCount += 1
-        if self._interval > 1 and self._modelCount*self._interval > 1<<31:
+        if self._interval > 1 and self._firstStep+self._modelCount*self._interval > 1<<31:
             # This will exceed the range of a 32 bit integer.  To avoid crashing or producing a corrupt file,
             # update the header to say the trajectory consisted of a smaller number of larger steps (so the
             # total trajectory length remains correct).
--- a/wrappers/python/openmm/app/dcdfile.py
+++ b/wrappers/python/openmm/app/dcdfile.py
@@ -64,7 +64,7 @@ class DCDFile(object):
         dt : time
             The time step used in the trajectory
         firstStep : int=0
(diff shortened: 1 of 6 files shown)


stefdoerr:
```
         self.assertEqual(5, modelCount)
>       self.assertEqual(10, currStep)
E       AssertionError: 10 != 12

TestDcdFile.py:110: AssertionError
____________________________ TestXtcFile.testAppend ____________________________

             integrator = mm.VerletIntegrator(0.001 * unit.picoseconds)
            simulation = app.Simulation(
                pdb.topology,
                system,
                integrator,
                mm.Platform.getPlatform("Reference"),
            )
            simulation.currentStep = 10
            xtc = app.XTCReporter(fname, 2, append=True)
            simulation.reporters.append(xtc)
            simulation.context.setPositions(pdb.positions)
            simulation.context.setVelocitiesToTemperature(300 * unit.kelvin)
            simulation.step(10)
            self.assertEqual(10, xtc._xtc._modelCount)
            self.assertEqual(10, xtc._xtc._getNumFrames())
            _, _, time, step = read_xtc(fname.encode("utf-8"))
>           self.assertTrue(np.allclose(np.arange(2, 21, 2) * 0.001, time))
E           AssertionError: False is not true
=========================== short test summary info ============================
FAILED TestDcdFile.py::TestDCDFile::testAppend - AssertionError: 10 != 12
FAILED TestDcdFile.py::TestDCDFile::testAtomSubset - AssertionError: 10 != 12
FAILED TestXtcFile.py::TestXtcFile::testAppend - AssertionError: False is not true
====== 3 failed, 341 passed, 1 skipped, 13 warnings in 186.04s (0:03:06) =======
```
Now we see that the tests don't pass, DCD writes a step larger than the actual step and the XTC writer doesn't write correctly when appending to the file

stefdoerr:
Ok this is the most conservative change I can make to fix the issue.
1. Modify the reporters to not pass `simulation.currentStep` as `firstStep` but `interval` instead since the reporters ought to write their first frame on the first interval
2. Fix the bug in DCD writer which was writing one step more than it actually contained

peastman:
Looks good.  Thanks!

PR #4881, opened 2025-04-04
Added computeCurrentPressure() to MonteCarloBarostat

### LINKED ISSUE #3647: Calculate the pressure tensor
Hi, I want to calculate the pressure tensor according to [https://docs.lammps.org/compute_pressure.html](https://docs.lammps.org/compute_pressure.html).
The problem seems to be mentioned already in #1989, however, it does not seems to be addressed.
According to the formula, the force needed ($f_{ij}$) is the force components from each atoms, but thte #1989 only gives the force on each atoms.
The documents in LAMMPS refer to a [paper](https://aip.scitation.org/doi/10.1063/1.3245303) that gives 3 method to calculate the virial.
So is there any existing code that use OpenMM to calculate pressure tensor (or stress tensor) I can refer to?

### LINKED ISSUE #3740: Time evolution of pressure
Dear Developers,
         How can I get the time evolution of  pressure from simulation from openmm. I wan to check the average pressure of the system. 

Thank you in advance.

Umesh Roy

peastman:
This adds a `computeCurrentPressure()` method to the barostat.  It allows you to compute instantaneous pressure, using a finite difference to approximate the molecular virial.

So far I've implemented it only for MonteCarloBarostat.  Adding similar methods to MonteCarloAnisotropicBarostat and MonteCarloMembraneBarostat should be straightforward.  They'll work similarly, but compute the pressure separately along each axis and return the result as a Vec3.

I'm less sure what to do for MonteCarloFlexibleBarostat.  Should it try to compute a complete tensor of pressures, since it allows the angles as well as the lengths of the box vectors to change?  Or, since it only allows you to specify a single scalar pressure to apply, should it just compute a single pressure by uniformly scaling the box?  If you specify `setScaleMoleculesAsRigid(false)`, clearly it should use an atomic virial rather than a molecular one.

Fixes #3740.  Fixes #3647.


z-gong:
Should pressure calculation be a method of `State` rather than `Barostat`?
It's common to query the pressure in an NVT simulation, e.g.
- evaluate surface tension of liquid-gas surface via $\frac{2P_z-P_x-P_y}{4L_z}$
- determine Young's modulus by apply a fixed strain to solid material

peastman:
The reason for putting it in the barostat is to make sure the pressure calculation is consistent with it.  Is the pressure a scalar or a vector?  Do you use a molecular or atomic virial?  Those are defined by the barostat.

If you're running a NVT simulation this still works.  Add an appropriate barostat defining how the pressure should be calculated, then call `setFrequency(0)` on it to disable pressure control.

z-gong:
So the `Barostat` class functions as both a barostat and a pressure calculator. This is a bit confusing from a user's perspective, but it will work as long as it's well-documented.

Another reason why we should compute pressure using the instantaneous kinetic energy rather than the predefined temperature: In an NPT simulation of an ideal gas, the virial contribution is zero. If we rely on the predefined temperature, the computed instantaneous pressure would be constant -- clearly inconsistent with the fluctuations expected in an NPT ensemble.

peastman:
The kinetic contribution is $NkT/V$.  That still fluctuates as the volume changes.

svandenhaute:
The kinetic contribution *averages to* $NkT/V$, but that doesn't mean that it is actually a correct estimator (i.e. with the correct pressure fluctuations). I'm not sure how much people care about fluctuations in pressure -- I suppose they are much less important than fluctuations in unit cell volume -- but why not use the actual estimator based on the sum of exterior products of atomic momenta?

As for `MonteCarloFlexibleBarostat`, I do believe it would make sense to have it estimate the full pressure tensor i.e. with six displacements rather than one, or at least support this use case with a flag `computeCurrentPressure(full=True)` for example. The cases where people would use this barostat is in situations where the unit cell shape (rather than its size) is expected to deviate substantially from its initial configuration, and in those cases it can be interesting to monitor the full pressure tensor to e.g. check whether its off-diagonal elements have a running average sufficiently close to zero.

peastman:
I have it implemented for MonteCarloBarostat, MonteCarloAnisotropicBarostat, and MonteCarloMembraneBarostat.  If you want to try it out or look over the code, it's ready for review.

I'm trying to figure out how to implement it for MonteCarloFlexibleBarostat.  In the other barostats, we use a finite difference to estimate $\partial E / \partial V$ for each of the diagonal elements of the box vectors.  We can't do the same for the off-diagonal elements.  Because of the restricted form OpenMM requires the box vectors to have, the off-diagonal elements don't affect the volume.  If I tried to use a finite difference to estimate the derivative with respect to those elements, the denominator would be zero.  Is it still possible to compute them using this approach?  Or does it mean the pressure tensor is always diagonal in the chosen coordinate system?

peastman:
I think this is pretty much complete and ready for review.  Please check my math as well as the code, including in the documentation!  I have some basic test cases for it, but it's hard to test robustly, especially for the flexible barostat.

peastman:
Right, I figured it out.  The formula is in eq. 5.68 of https://d-nb.info/1079218084/34.

svandenhaute:
Formula 5.68 seems wrong; it needs to be multiplied by the transpose of the unit cell matrix.

svandenhaute:
It's a long time ago, but I've once written a Python function to do this for OpenMM (though it's very convoluted): https://github.com/svandenhaute/openyaff/blob/fc8c605969a125d9ddd1119c3f19b30cc72aca3e/openyaff/wrappers.py#L82

peastman:
I think it's just a difference in how you define the scaling.  Do you add $\Delta h$ to the element, or do you multiply the element by $1+\Delta h$?  In one case you get that extra factor of $h$ and in the other you don't.

svandenhaute:
Sorry, yes, I judged too soon. That formula is correct because they are talking about strain derivatives (which are multiplicative and don't need the extra factor). I got confused because they use $h$ as finite difference variable for the strain, though I assumed it was a perturbation to an individual unit cell component.

PR #4886, opened 2025-04-07
Tidying up the tests

stefdoerr:
I changed all `systems/` paths in the tests to be relative to the test files, so that the tests can be executed from any directory, not just from inside the tests directory. It's useful for some python tools which execute pytest from the main directory of the project like `cibuildwheel`, but it's also generally good practice I believe.

### DIFF
--- a/wrappers/python/tests/TestAmberInpcrdFile.py
+++ b/wrappers/python/tests/TestAmberInpcrdFile.py
@@ -1,3 +1,4 @@
+import os
 import unittest
 from validateConstraints import *
 from openmm.app import *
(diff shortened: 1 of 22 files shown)


peastman:
The tests aren't intended to be invoked from the source directory.  You need to build the Python wrappers first.  They then get invoked from `build/python/tests`.

stefdoerr:
Yes I am aware. Still it forces you to run the tests from inside a specific directory and only works if you use cmake to build the wheels. Since I'm trying to build with cibuildwheel it doesn't like the fact that I change directory. 

Also some of the tests already used that method to detect the system files https://github.com/openmm/openmm/pull/4886/files#diff-0e4370f840b8f7b144931a2cbe4022250cc871ac7d70480dbd66fce369508c87L13 so there was already an inconsistency in how it was handled

PR #4907, opened 2025-04-23
Add correction for self energy of neutralizing plasma

### LINKED ISSUE #4905: Long-range electrostatics methods should include net charge correction
Methods for calculating long-range electrostatic interactions (standard Ewald summation and PME) should include a correction term $\Delta U_{\mathrm{bkgd}}=-(\sum_{i=1}^nq_i)^2/8\epsilon_0V\alpha^2$ to the total energy that accounts for the implicit presence of a uniform neutralizing background charge density.  This is discussed in:

- [Figueirido *et al.*, *J. Chem. Phys.* 103, 6133 (1995)](https://doi.org/10.1063/1.470721).
- [Hub *et al.*, *J. Chem. Theory Comput.* 10, 381 (2014)](https://doi.org/10.1021/ct400626b).  
- Allen and Tildesley, *Computer Simulation of Liquids*, 2nd ed. (2017)

Note that if the system is charge-neutral overall, this term is zero and has no effect.  If it is not charge-neutral, and the total charge is fixed, this term has no effect on forces, but it does affect the dependence of the energy on the volume.  For the constant potential method #4870 that we are developing, it will affect the charges we solve for if a total charge constraint is not in use.

The term appears important to include, as without it, the Ewald energy will have a dependence on the Ewald parameter $\alpha$ (that has no direct physical significance and is supposed to serve only as a numerical tuning parameter).  The term listed above should be the exact correction for this $\alpha$-dependence.  For example, here's a system of 2 positive unit charges separated by 1.5 nm along one axis of a 3-nm cube (net charge +2):
Image
Here's OpenMM's explicit solvent DHFR benchmark test system with ff99SB/TIP3P (net charge -11):
Image

peastman:
Fixes #4905.  When simulating a charged system with Ewald or PME, the self energy should include a term for the neutralizing plasma.  We were missing that term.  It's just a shift in the energy, so it doesn't affect forces.  It does depend on periodic box volume, though, which means it could affect the behavior of barostats.

This is only relevant to charged systems.  When a system has zero net charge the term vanishes.

### DIFF
--- a/platforms/common/src/kernels/nonbondedParameters.cc
+++ b/platforms/common/src/kernels/nonbondedParameters.cc
@@ -3,13 +3,15 @@
  */
 KERNEL void computeParameters(GLOBAL mixed* RESTRICT energyBuffer, int includeSelfEnergy, GLOBAL real* RESTRICT globalParams,
         int numAtoms, GLOBAL const float4* RESTRICT baseParticleParams, GLOBAL real4* RESTRICT posq, GLOBAL real* RESTRICT charge,
-        GLOBAL float2* RESTRICT sigmaEpsilon, GLOBAL float4* RESTRICT particleParamOffsets, GLOBAL int* RESTRICT particleOffsetIndices
+        GLOBAL float2* RESTRICT sigmaEpsilon, GLOBAL float4* RESTRICT particleParamOffsets, GLOBAL int* RESTRICT particleOffsetIndices,
+        GLOBAL real* RESTRICT chargeBuffer
 #ifdef HAS_EXCEPTIONS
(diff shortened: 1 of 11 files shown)


epretti:
Great - should this correction also be added to the other forces that do long-range electrostatics (I believe these are just `AmoebaMultipoleForce` and `HippoNonbondedForce`)?

peastman:
I think that's a question for @jayponder.  It does seem like AMOEBA should include this term, but I can't find that Tinker adds it.  Jay, do you know if Tinker includes the self-energy term for the uniform neutralizing plasma?  And if not, should it?

PR #4911, opened 2025-04-25
Uniform interface for FFTs

peastman:
This is the first piece of a project to unify more of the code for NonbondedForce between the GPU platforms.  I created a generic FFT3D interface, with each platform providing an implementation.  You create an instance by calling `createFFT()` on the ComputeContext.  This allows a lot of code to be made more consistent.


peastman:
@epretti one goal of this is to make it easier for you to implement constant potential, since the amount of code you'll need to duplicate will be greatly reduced.  I hope to finish it all within the next few days.

Issue #4915, opened 2025-04-30
Update CHARMM and Amber force field XML files


epretti:
openmm/openmmforcefields#356 and openmm/openmmforcefields#368 update openmmforcefields to include new versions of the CHARMM and Amber force field XML files reflecting the latest (CHARMM July 2024 / AmberTools 24) releases of these force fields.  We should update what we distribute with OpenMM once the Amber conversion is merged in to openmmforcefields.
- Do we want to include everything that is in both conversion updates, or just the "main" files?  For instance, the CHARMM conversion has some residues like D-amino acids, CGenFF model compounds, certain carbohydrate residue/patch sets, *etc.* split into separate force field files because they can cause problems when creating systems, requiring the use of the `residueTemplates` option to `createSystem()`.  The Amber conversion includes some new force fields, but also some older ones like `ff94`, `DNA.bsc0`, `RNA.ROC`, `RNA.YIL`, *etc.* that aren't currently distributed with OpenMM.
- Do we want to update existing XML files or just add new ones?  Replacing these files might break users' workflows or generate subtly different results.  CHARMM, unlike Amber, doesn't try to distribute all historical force field versions in a single release package; some force field files are just modified.  The current releases from the CHARMM developers include the "CHARMM36m" parameters for proteins but not the older non-"36m" versions.  So openmmforcefields' automated conversion doesn't include the old versions along with the new.  Presumably we don't want to just update the `charmm36.xml` that comes with OpenMM, but have a separate file with the new parameters.  And it looks like for Amber, some of the older force fields like `amber03.xml` were converted in a different way than openmmforcefields currently generates its `ff03.xml`, for example.  I don't know what the provenance of those files in the OpenMM repository is, and at least for that example, some atom types are different and they include different residues.

peastman:
I would tend to err on the side of including fewer files rather than more.  The reasons are

1. We don't want to include a force field if we would never recommend someone use it.  For example, there's no good reason to use an ancient force field like ff94.  Including it would just encourage people to make bad choices.
2. If for some reason someone really does want it, they can just install openmmforcefields.  Omitting it from OpenMM doesn't create a serious barrier for people who know what they're doing.
3. It's easier to add something later than to remove something we added and then regretted adding.

[...]

PR #4916, opened 2025-05-01
Add new CHARMM and Amber force fields

epretti:
For #4915.  Adds Amber ff19SB and ff19ipq protein force fields, and OL21 DNA force field.  Fixes some bugs in GLYCAM06j-1.  Adds 2024 CHARMM force field with CHARMM36m protein parameters, and 2023 CHARMM Drude force field.


peastman:
Looks good.  I suggest a few changes.  We should add an `amber19-all.xml` for including the complete ff19 force field.  We don't have the lipids yet, but we should still provide a single file for including everything we do have.  I'd also move the discussion of ff19 above ff14.  That's the recommended force field for anything it supports, which mostly means anything other than membrane simulations.  And we should explain that: ff19 is generally preferred, but we we're not yet marking ff14 as a legacy force field only because we don't yet have the lipids.

For the same reason, we should update the examples to use ff19.  That includes both in the manual and the corresponding `simulate*.py` files in the examples directory.

epretti:
Making these updates.  I notice we use TIP3P-FB a lot in the Amber examples.  However, the Amber manual recommends OPC for use with ff19SB:
> Our results [...] showed that ff19SB pairs best with the more accurate water model OPC [...] , and that the older TIP3P model has serious limitations when used with the QM-based ff19SB. As a result, we strongly recommend using ff19SB with OPC, and we recommend against use with TIP3P.

Should I switch our examples?  This would make them slightly more complicated since the virtual sites for the 4-site OPC have to be added to the topologies.  But we probably want users who are just copying the first example they see for simulating some PDB to use the recommended model.

peastman:
TIP3P-FB is a very good water model.  In fact, the parameters are extremely similar to OPC.  They were optimized differently but ended up at almost the same place.

TIP3P is not a good water model, and they're right that people shouldn't use it.

epretti:
I think I made all of the relevant updates for 14 -> 19, and checked to make sure the example `simulatePdb.py` works with ff19SB.  The Amber19 section is now first, with some updated notes.

I had to regenerate CHARMM due to a problem with some bad patches.  I am making a separate issue on openmmforcefields with more information.  But I have re-run the CHARMM validation suite on the version now included here to make sure that it passes.

docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -87,14 +87,14 @@ Make sure you include the single quotes around the file name.  OpenMM also can l
...
+    forcefield = ForceField('amber19-all.xml', 'amber14/tip3pfb.xml')
 
 This line specifies the force field to use for the simulation.  Force fields are
 defined by XML files.  OpenMM includes XML files defining lots of standard force fields (see Section :numref:`force-fields`).
 If you find you need to extend the repertoire of force fields available,
 you can find more information on how to create these XML files in Chapter :numref:`creating-force-fields`.
-In this case we load two of those files: :file:`amber14-all.xml`, which contains the
-Amber14 force field, and :file:`amber14/tip3pfb.xml`, which contains the TIP3P-FB water model.  The
+In this case we load two of those files: :file:`amber19-all.xml`, which contains the
+Amber19 force field, and :file:`amber14/tip3pfb.xml`, which contains the TIP3P-FB water model.  The
```
peastman, line 97: This feels weird, that you have to combine `amber19-all.xml` with a water model from the `amber14` directory.  I gather that's how Amber distributes them?  Still, I wonder if it would be better to duplicate the water models in the `amber19` directory.  They're tiny files.

epretti: I agree that this is confusing.  The trouble is that there isn't really one "Amber14" or "Amber19".  Amber distributes a whole lot of force fields for proteins, nucleic acids, *etc.* all together, and you have to look in the Amber manual to see what the "recommended" combinations are.  This can change from year to year as they release new force fields; at some point they switched from recommending ff14SB+OL15+OL3+Lipid17 (what OpenMM calls "Amber14") to ff19SB+OL21+OL3+Lipid21.  I described the latter under an "Amber19" section for consistency with the nomenclature already used by the manual, but I don't think either name is official, and we can change how we refer to it if we want.  Note that even if copying the water models, the "Amber19" combination still refers to the OL3 RNA model from the `amber14` directory, since Amber hasn't recommended a new RNA force field in a while.

peastman: Yeah, it's a bit awkward.  Let's copy the water models over though.  Since we do have separate directories with those names, I think it will be less confusing for users.  I'm less concerned about the RNA parameters, because most people will get them automatically from `amber19-all.xml` without needing to see what directory they come from.

epretti: Sounds good.  I've copied them over and updated the paths in the example script and the documentation.

peastman (approved):
Looks good.  Feel free to merge once the tests finish.

Issue #4096, opened 2023-06-03
Create issue templates


peastman:
It's been suggested that we create templates for new issues.  I'm not fond of the default github ones, which try to force everything into a rigid structure ("Expected Behavior", "Actual Behavior", etc.), but I do think there's value in providing a paragraph that summarizes how to create a good issue of each type.  I suggest offering the following issue types: Help Request, Bug Report, Feature Suggestion, and Other.  Here is the text for each type.

#### Help Request

To request help, please describe as precisely as possible what you are trying to accomplish, any steps you have taken so far to accomplish it, and what is not working.  The more background you can provide to help us understand the problem, the better.  Also include any of the following information that is relevant.

- The version of OpenMM you are using.
- The operating system and hardware (e.g. GPU model) you are using.
- The platform (CUDA, OpenCL, etc.) you are using.
- Any other software (Python libraries, setup tools, etc.) you are using along with OpenMM to implement your workflow.

#### Bug Report

To write a useful bug report, please include as much detail as possible.  Describe exactly what you are doing, what you expect to happen, and what is happening instead.  If possible, provide a complete test case that can be used to reproduce the problem including a complete, self-contained script and all necessary input files.  It is very difficult to fix a bug if we cannot reproduce it.  Also include all of the following information.

- The version of OpenMM you are using.
- The operating system and hardware (e.g. GPU model) you are using.
- The platform (CUDA, OpenCL, etc.) you are using.
- Any other software (Python libraries, setup tools, etc.) you are using along with OpenMM to implement your workflow.

#### Feature Suggestion

Feature suggestions are most useful when they include enough context to help us fully understand the suggested feature.  What problem is it meant to solve?  How would the feature contribute to solving the problem?  Have you considered alternative solutions that might also work?  If your suggestion is to implement a published algorithm, please include a link to the publication.

#### Other

No template.

sef43:
Looks good to me. For additions I would include a example command to show the openmm version they are actually using, e.g. `python -c "import openmm; print(openmm.version.version)"`

And a point saying that you should format code using markdown directives: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks

jchodera:
This looks useful!

For the information you ask to capture, I agree with @sef43 that a pointer to the FAQ with information on how to collect all the requested information would be useful.

Do we need to provide more guidance on how proposed features end up being accepted onto the roadmap?

peastman:
[...]

PR #4928, opened 2025-05-06
Update issue templates

peastman:


peastman:
Fixes #4096.  This creates a set of templates for creating issues.

Issue #4939, opened 2025-05-16
GPU memory leak


fiedorowicz1:
When running OpenFold's relaxation pipeline repeatedly in the same process, I observe that the GPU memory usage continues to grow looping over the same input. Repeating the relaxation in separate subprocesses, does not exhibit the same behavior, however.

I have identified that there is a memory leak affecting CUDA/HIP platforms in the interaction between the `Cuda/HipContext` destructor and the destructor of its member `Cuda/HipArray`s. Specifically, these lines in the array destructors check if the context is still vaild before freeing the memory: https://github.com/openmm/openmm/blob/bc8098795c716890aa6339431cddb753943c6ae4/platforms/hip/src/HipArray.cpp#L45
https://github.com/openmm/openmm/blob/bc8098795c716890aa6339431cddb753943c6ae4/platforms/cuda/src/CudaArray.cpp#L44
However, before the destructors of the member arrays are called, the context destructor marks the context as invalid leading to the underlying pointers of the member arrays being leaked:
https://github.com/openmm/openmm/blob/bc8098795c716890aa6339431cddb753943c6ae4/platforms/hip/src/HipContext.cpp#L379
https://github.com/openmm/openmm/blob/bc8098795c716890aa6339431cddb753943c6ae4/platforms/cuda/src/CudaContext.cpp#L390

A potential fix would be to skip this check. Doing so fixes the memory leak and works without issue in my limited testing. However, I am unaware of broader potential side-effects that could stem from skipping this check.

peastman:
Do you have a test case that reproduces the problem?  Immediately before setting `contextIsValid` it deletes the CUDA context:

https://github.com/openmm/openmm/blob/bc8098795c716890aa6339431cddb753943c6ae4/platforms/cuda/src/CudaContext.cpp#L388-L390

`cuCtxDestroy()` releases all memory from all objects associated with the context.  There's no need to free them individually, and trying to free them after the context has been destroyed will produce a crash.  That's why CudaArray has that check.

I suspect the cause of the growing memory usage may actually be something different.  If you have a reproducible test case, we can track down what it is.

fiedorowicz1:
Ah, so this may only affect the HIP platform in that case since `hipCtxDestroy` is deprecated and does not appear to be used by OpenMM for that reason.

I don't have a simple reproducer, but I can try to pull one together.

peastman:
That makes sense.  So possibly we should remove the check from HIP but not CUDA?

[...]

PR #4940, opened 2025-05-16
Fix GPU memory leak in context arrays

fiedorowicz1:
This PR provides a potential fix for a GPU-side memory leak on CUDA/HIP platforms in the Context/Array destructors (issue #4939).


peastman:
Thanks!

PR #4952, opened 2025-05-29
Allow empty loops to be parsed with `PdbxReader.py`

amorehead:
* Adds support for skipping mmCIF categories with empty loops. Otherwise, a `SyntaxError` may be unnecessarily raised.


peastman:
Can you provide an example file that doesn't load correctly without the change?

PdbxReader.py is part of the official [parser library](https://mmcif.wwpdb.org/docs/sw-examples/python/html/index.html#) from wwPDB, not something we wrote.  That makes me a little cautious about modifying it.  I don't know the code, and it's hard to be sure we won't break something.

amorehead:
Hi, @peastman. On second thought, I will close this PR, because the mmCIF files it addresses use invalid (unofficial) formatting. For example, some of them have empty loops (with field names defined but no rows present).

PR #4956, opened 2025-06-05
Handle relative include paths from force fields in data directories

epretti:
For `<Include>` in force field files, the user manual states that

> The file attribute gives the path of the file to include. It may be relative either to the directory containing the parent XML file (the one with the <Include> tag) or the OpenMM data directory (the one containing built in force fields).

However, this doesn't work if the including file is in a subdirectory of a data directory and points to the included file in that subdirectory using a relative path, *e.g.*, if `data/a/b.xml` includes `data/a/c.xml`, `<Include file="c.xml" />` doesn't work, so `<Include file="a/c.xml" />` is needed.  It seems like the test https://github.com/openmm/openmm/blob/2b8ad70394f38d296497215da757aa31ac36b87b/wrappers/python/openmm/app/forcefield.py#L283-L285 doesn't work here since `parentDir` is that of `file`, but if the including file is in a data directory, `f` ends up being its path instead.

This change ensures that `file` will contain a resolved path to the force field if it is in some data directory, so the relative path lookup will find files correctly.  This fixes an issue with some openmmforcefields files that rely on it working.  It also allows more cleanly reporting errors in loading force fields in data directories, since we don't actually always control such files (any package can register an entry point).

Does this behavior make sense?  If not, and includes from data directories aren't supposed to be able to do relative path lookups, I will make changes in openmmforcefields instead.  (This shouldn't affect any of the OpenMM-distributed force fields.)

### DIFF
--- a/wrappers/python/tests/TestForceField.py
+++ b/wrappers/python/tests/TestForceField.py
@@ -6,6 +6,8 @@ from openmm.unit import *
 import openmm.app.element as elem
 import openmm.app.forcefield as forcefield
 import math
+import shutil
+import tempfile
 import textwrap
 try:
     from cStringIO import StringIO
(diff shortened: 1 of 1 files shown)


peastman:
I don't think I follow.  If `file` is `data/a/b.xml`, then `parentDir = os.path.dirname(file)` is `data/a`, and `joined = os.path.join(parentDir, includeFile)` is `data/a/c.xml`.  Can you add a test case that demonstrates the problem and verifies the fix?

epretti:
If `/full/path/to/.../data` is a data directory, and you have `a/b.xml` and `a/c.xml` within it, where `a/b.xml` contains `<Include file="c.xml" />`, calling `ForceField("a/b.xml")` will have `file` equal to `a/b.xml` and `f` equal to `/full/path/to/.../data/a/b.xml`.  `f` is used to (correctly) load `a/b.xml` from the data directory.  But then `parentDir`, from `file`, ends up as `a`, and `joined` ends up as `a/c.xml`, but that doesn't exist (because the correct place to look would be `/full/path/to/.../data/a/c.xml`).  So `includeFile` ends up as `c.xml`, and assuming `/full/path/to/.../data/c.xml` doesn't exist, the import won't work.  I added a test case that should demonstrate this and works with the fix.

peastman:
Got it, thanks.  Looks good.

PR #4959, opened 2025-06-07
Fix computeNonbonded hang on the HIP platform

ex-rzr:
`computeNonbonded` hangs in some tests (without neighbor list)  on the HIP platform
Reproducible on ROCm 6.4 and 6.4.1 (maybe on older versions too) on various architectures (both CDNA and RDNA).

Affected tests:
* TestHipATMForce
* TestHipMonteCarloBarostat
* TestHipNonbondedForce
* TestHipVirtualSites

Disassembly shows that the compiler splits branches of `if (skipBase+tgx < NUM_TILES_WITH_EXCLUSIONS)` and does
`SHFL(skipTiles, TILE_SIZE-1) < pos` checks in them separately, even though `__builtin_amdgcn_ds_bpermute` is a convergent function. Apparently in this case not all lanes participate in each call.

Unlike other common or CUDA kernels, the HIP kernel uses shuffles instead of shared memory.
First I fixed this issue by adding `SYNC_WARPS` at the end of the loop to ensure that the check `SHFL(skipTiles, TILE_SIZE-1) < pos` stays convergent, this helped.
But then I simplified the code even more with ballot.

### DIFF
--- a/platforms/hip/src/kernels/nonbonded.hip
+++ b/platforms/hip/src/kernels/nonbonded.hip
@@ -72,7 +72,6 @@ extern "C" __launch_bounds__(THREAD_BLOCK_SIZE) __global__ void computeNonbonded
     const unsigned int totalWarps = (blockDim.x*gridDim.x)/TILE_SIZE;
     const unsigned int warp = blockIdx.x*(THREAD_BLOCK_SIZE/TILE_SIZE) + threadIdx.x/TILE_SIZE; // global warpIndex
     const unsigned int tgx = threadIdx.x & (TILE_SIZE-1); // index within the warp
-    const unsigned int tbx = threadIdx.x - tgx;           // block warpIndex
     mixed energy = 0;
     INIT_DERIVATIVES
 
@@ -250,9 +249,7 @@ extern "C" __launch_bounds__(THREAD_BLOCK_SIZE) __global__ void computeNonbonded
     int pos = (int) (startTileIndex+warp*numTileIndices/totalWarps);
     int end = (int) (startTileIndex+(warp+1)*numTileIndices/totalWarps);
     int skipBase = 0;
-    int currentSkipIndex = tbx;
-    int skipTiles;
-    skipTiles = -1;
+    int skipTiles = -1;
     for (; pos < end; pos++) {
 #endif
         real3 force = make_real3(0);
@@ -277,7 +274,7 @@ extern "C" __launch_bounds__(THREAD_BLOCK_SIZE) __global__ void computeNonbonded
 
         // Skip over tiles that have exclusions, since they were already processed.
 
-        while (SHFL(skipTiles, TILE_SIZE-1) < pos) {
+        while (BALLOT(skipTiles >= pos) == 0) {
             if (skipBase+tgx < NUM_TILES_WITH_EXCLUSIONS) {
                 int2 tile = exclusionTiles[skipBase+tgx];
                 skipTiles = tile.x + tile.y*NUM_BLOCKS - tile.y*(tile.y+1)/2;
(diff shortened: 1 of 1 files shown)


peastman:
All four of those tests are passing for me.  That's on a RX 5700 XT with ROCm 6.2.  I tried to update to 6.4.1, but I can't get it to install.  When I execute `sudo apt install ./amdgpu-install_6.4.60401-1_all.deb`, it fails with the message

```
Building module:
Cleaning build area...(bad exit status: 2)
. /tmp/amd.0tn8LmU2/.env && make -j4 KERNELRELEASE=6.11.0-24-generic TTM_NAME=amdttm SCHED_NAME=amd-sched -C /lib/modules/6.1
1.0-24-generic/build M=/tmp/amd.0tn8LmU2....(bad exit status: 2)
ERROR: Cannot create report: [Errno 17] File exists: '/var/crash/amdgpu-dkms.0.crash'
Error! Bad return status for module build on kernel: 6.11.0-24-generic (x86_64)
Consult /var/lib/dkms/amdgpu/6.8.5-2041575.24.04/build/make.log for more information.
```

Checking `make.log`, there are a lot of compilation errors.  Here is the first one.

```
/tmp/amd.0tn8LmU2/scheduler/./gpu_scheduler_trace.h:60:1: error: macro "__assign_str" passed 2 arguments, but takes just 1
   60 | );
      | ^~             
In file included from ./include/trace/trace_events.h:378:
./include/trace/stages/stage6_event_callback.h:34: note: macro "__assign_str" defined here
   34 | #define __assign_str(dst)                                               \
      | 
/tmp/amd.0tn8LmU2/scheduler/./gpu_scheduler_trace.h: In function ‘do_trace_event_raw_event_drm_sched_job’:
```

This is on Ubuntu 24.04.

ex-rzr:
> All four of those tests are passing for me. That's on a RX 5700 XT with ROCm 6.2.

Yes, I also had no issues in any tests on older versions (<= 6.2, if I remember correctly). But these tests hang on 6.4 and 6.4.1 on all GPUs I have tested (gfx908, 90a, 1030, 1100, 1201).

About the error during ROCm update, I found this thread and this comment: https://github.com/ROCm/ROCm/issues/3870#issuecomment-2655995422

> The solution was to uninstall amdgpu-dkms

Also: https://rocm.docs.amd.com/projects/radeon/en/latest/docs/install/native_linux/install-radeon.html#upgrade-to-newer-versions-of-radeon-software-for-linux

> The recommended method to upgrade is to uninstall, followed by an install.
> Radeon Software for Linux does not support in-place upgrades.

peastman:
Thanks!  I removed 6.2 and successfully installed 6.4.1.  But it no longer seems to include any files for OpenCL, even when I specify `--usecase=opencl`.  At least, `libOpenCL.so` is no longer present in `/opt/rocm/lib`.  Has it moved somewhere else?

ex-rzr:
What I found in 6.3 release notes:
https://rocm.docs.amd.com/en/docs-6.3.0/about/release-notes.html#opencl-icd-loader-separated-from-rocm

But I don't see anything about OpenCL on the linked pages about installation...

ex-rzr:
It looks like OpenCL can be installed with package manager: 
https://rocm.docs.amd.com/projects/install-on-linux/en/latest/reference/package-manager-integration.html

Additional info about the icd loader: https://foldingforum.org/viewtopic.php?t=42803&start=15

peastman:
Yeah, not very helpful instructions!  I managed to compile by [installing the Mesa version](https://kazanas.com/getting-opencl-1-2-to-work-with-amd-gpus-on-ubuntu/) with the command

```
sudo apt install mesa-opencl-icd ocl-icd-libopencl1 ocl-icd-opencl-dev clinfo opencl-headers -y
```

I confirmed the hang with 6.4.1.  I'll try your fix next.

peastman:
I confirmed the fix eliminates the hang.  I'm running the full test suite now.  If everything is ok I'll merge this and build a new release candidate for 8.3.

peastman:
All tests passed.  Thanks for the fix!

PR #4980, opened 2025-06-25
Fixed bug in computing pressure

peastman:
This fixes a bug that caused `computeCurrentPressure()` to return an incorrect value.  The tests were passing only because they happened to set all particle masses to 1.  See https://github.com/openmm/openmm/discussions/4926#discussioncomment-13503249.

### DIFF
--- a/platforms/common/src/kernels/monteCarloBarostat.cc
+++ b/platforms/common/src/kernels/monteCarloBarostat.cc
@@ -79,7 +79,7 @@ KERNEL void computeMolecularKineticEnergy(int numMolecules, GLOBAL mixed4* RESTR
             molVel += mass*trimTo3(v);
             molMass += mass;
         }
-        molVel *= RECIP((mixed) numAtoms);
+        molVel *= RECIP((mixed) molMass);
 #if COMPONENTS == 1
         ke[0] += 0.5f*molMass*dot(molVel, molVel);
 #else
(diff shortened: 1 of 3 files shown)


peastman:
@Park-Hyungshick can you test this and confirm that it fixes your problem?

hue0704:
Sorry for my late response.






(Left) OpenMM 8.3, (Right) MSTK function

Thank you for your modification! I think we almost done about this problem.
However, still, the kinetic contribution on pressure does not oscillate around the theoretical prediction (0.8).

peastman:
I can't reproduce that.  I ran the notebook you sent earlier using this fix.  At the end I print the average pressure for each method:

```python
print(np.mean(press_openmm[500:,:]))
print(np.mean(press_mstk[500:,:]))
```

The result is

```
414.04535284744276
413.9707275870031
```

How did you generate that plot?

hue0704:
Oh, I'm sorry. I've forgot that manually defined function (MSTK) used averaged contribution of kinetic term. Previously, I only plot the Pxx contribution. P=(Pxx+Pyy+Pzz)/3 properly fluctuate around the theoretical value 0.8.

Thank you for your contribution on OpenMM developing and also for your kind reply.

peastman:
Great, thanks!  And thanks so much for spotting this problem.

PR #4989, opened 2025-07-02
Update CHARMM36 2024 and add water models

### LINKED ISSUE #4976: Issues with the new charmm36_2024.xml and water
I am getting an error adding water using the new charm forcefield.  
I modified the modeller to print out the script that causes the issue of self._scripts (code attached), hope that helps to fix it.

```
pdb = PDBFile(pdb_file)
forcefield = ForceField('charmm36_2024.xml',"charmm36/water.xml")
modeller = Modeller(fixer.topology, fixer.positions)
modeller.addSolvent(forcefield, padding=5.0*unit.nanometers)
```

[script_causing_the_crash.txt](https://github.com/user-attachments/files/20847095/script_causing_the_crash.txt)

  File "/home/ubuntu/test.py", line 34, in <module>
    modeller.addSolvent(forcefield, padding=5.0*unit.nanometers)
  File "/venv/lib/python3.11/site-packages/openmm/app/modeller.py", line 662, in addSolvent
    self._addIons(forcefield, numTotalWaters, waterPos, positiveIon=positiveIon, negativeIon=negativeIon, ionicStrength=ionicStrength, neutralize=neutralize, residueTemplates=newResidueTemplates)
  File "/venv/lib/python3.11/site-packages/openmm/app/modeller.py", line 311, in _addIons
    system = forcefield.createSystem(self.topology, residueTemplates=residueTemplates)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/openmm/app/forcefield.py", line 1433, in createSystem
    exec(script, locals())
  File "<string>", line 49, in <module>
IndexError: list index out of range

epretti:
Include fixes and improvements from openmm/openmmforcefields#386, and add solvent XML files specifically intended for use with the 2024 release of CHARMM36 (differences include new NBFIXes and changed NBFIX parameters).  Fixes #4976.


peastman:
I looked into whether we can enforce using the correct water models for the force field.  It's definitely possible, though a little more complicated than I would like.  Inside a `<Script>` tag, `self` refers to the ForceField.  Unfortunately, it doesn't record a list of what files it was loaded from.  Perhaps there are some unique features (particular residue names, for example), that can be used to distinguish the two versions?  We also can have one file store an attribute into the ForceField that the other checks for.  The complication is that we can't guarantee what order the scripts will be run in, so each one would have to both store a value and check for a value.

epretti:
I think that storing an attribute is the least messy way to do it.  What do you think about something like this in all of the CHARMM XML files:
```xml
  <InitializationScript>
EXPECTED_VERSION = ... # "charmm36" or "charmm36_2024"
if hasattr(self, "_charmm_ff_version"):
    if self._charmm_ff_version != EXPECTED_VERSION:
        raise RuntimeError(f"CHARMM force field versions ({self._charmm_ff_version} and {EXPECTED_VERSION}) incompatible")
else:
    self._charmm_ff_version = EXPECTED_VERSION
  </InitializationScript>
```

peastman:
That looks reasonable.

peastman:
Looks good.

PR #5009, opened 2025-07-14
API: Move definitions of getNum...() member functions to implementation files

### LINKED ISSUE #5001: Clang/C++20: arithmetic on a pointer to an incomplete type
When including OpenMM in a project which is compiled with `clang++ -std=c++20` or later, the compilation fails with the error `arithmetic on a pointer to an incomplete type`.

Example to reproduce (installing via `uv`/`pip`, but the same happens when using the library compiled from source):
```cpp
// file test.cpp
#include <OpenMM.h>

int main() {
  return 0;
}
```
```shell-session
$ uv venv && uv pip install openmm
$ clang++ -std=c++20 -I.venv/lib/python3.13/site-packages/OpenMM.libs/include -L.venv/lib/python3.13/site-packages/OpenMM.libs/lib -lOpenMM test.cpp
In file included from test.cpp:1:
In file included from .venv/lib/python3.13/site-packages/OpenMM.libs/include/OpenMM.h:36:
In file included from .venv/lib/python3.13/site-packages/OpenMM.libs/include/openmm/BrownianIntegrator.h:35:
In file included from .venv/lib/python3.13/site-packages/OpenMM.libs/include/openmm/Integrator.h:35:
In file included from .venv/lib/python3.13/site-packages/OpenMM.libs/include/openmm/State.h:36:
In file included from .venv/lib/python3.13/site-packages/OpenMM.libs/include/openmm/serialization/SerializationNode.h:40:
In file included from /usr/bin/../lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/vector:68:
/usr/bin/../lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/bits/stl_vector.h:1119:44: error: arithmetic on a pointer to an incomplete type 'OpenMM::CMAPTorsionForce::MapInfo'
 1119 |         ptrdiff_t __dif = this->_M_impl._M_finish - this->_M_impl._M_start;
      |                           ~~~~~~~~~~~~~~~~~~~~~~~ ^
.venv/lib/python3.13/site-packages/OpenMM.libs/include/openmm/CMAPTorsionForce.h:65:21: note: in instantiation of member function 'std::vector<OpenMM::CMAPTorsionForce::MapInfo>::size' requested here
   65 |         return maps.size();
      |                     ^
.venv/lib/python3.13/site-packages/OpenMM.libs/include/openmm/CMAPTorsionForce.h:177:11: note: forward declaration of 'OpenMM::CMAPTorsionForce::MapInfo'
  177 |     class MapInfo;
      |           ^
...
many more errors like the one above
```
The above example uses `libstdc++`, but the same happens for `libc++`.

The root cause is that many OpenMM headers use forward declarations of internal classes like this:
```cpp
class Foo {
public:
    ...
    int getNumBars() const {
        return bars.size();
    }
    ...

private:
    ...
    class Bar;
    std::vector<Bar> bars;
    ...
};

class Foo::Bar {
    ...
};
```
I assume Clang fails to compile this in C++20 mode due to `std::vector<T>::size()` being made `constexpr`, as compiling with `-std=c++17` works without problems. Also, GCC has no issues regardless of the C++ standard used.

Here's a list of the files that are affected:
- `openmmapi/include/openmm/ATMForce.h`
- `openmmapi/include/openmm/CMAPTorsionForce.h`
- `openmmapi/include/openmm/CustomAngleForce.h`
- `openmmapi/include/openmm/CustomBondForce.h`
- `openmmapi/include/openmm/CustomCVForce.h`
- `openmmapi/include/openmm/CustomCentroidBondForce.h`
- `openmmapi/include/openmm/CustomCompoundBondForce.h`
- `openmmapi/include/openmm/CustomExternalForce.h`
- `openmmapi/include/openmm/CustomGBForce.h`
- `openmmapi/include/openmm/CustomHbondForce.h`
- `openmmapi/include/openmm/CustomIntegrator.h`
- `openmmapi/include/openmm/CustomManyParticleForce.h`
- `openmmapi/include/openmm/CustomNonbondedForce.h`
- `openmmapi/include/openmm/CustomTorsionForce.h`
- `openmmapi/include/openmm/CustomVolumeForce.h`
- `openmmapi/include/openmm/DPDIntegrator.h`
- `openmmapi/include/openmm/GBSAOBCForce.h`
- `openmmapi/include/openmm/GayBerneForce.h`
- `openmmapi/include/openmm/HarmonicAngleForce.h`
- `openmmapi/include/openmm/HarmonicBondForce.h`
- `openmmapi/include/openmm/NonbondedForce.h`
- `openmmapi/include/openmm/PeriodicTorsionForce.h`
- `openmmapi/include/openmm/RBTorsionForce.h`
- `openmmapi/include/openmm/System.h`

The simplest fix would be to move the inline definitions for functions of the type `Foo::getNumBars()` from the header to the relevant `.cpp` files in `openmmapi/src/` – I'd be happy to submit a pull request if this is fine for you.

Some extra info:
- OS: Fedora Linux 42
- Compiler version: Clang 20.1.7 (following some testing on [compiler explorer](https://godbolt.org/z/K81svTGbW), all Clang versions from 15 upward are affected)
- Output of `python -m openmm.testInstallation`:
    ```shell-session
    $ uv run python -m openmm.testInstallation

    OpenMM Version: 8.3
    Git Revision: 1ce5d91d9dedfdc273066fafa1a618bf05c25b85

    3 OpenCL - Successfully computed forces

    Reference vs. CPU: 6.29438e-06
    Reference vs. OpenCL: 6.75445e-06
    CPU vs. OpenCL: 7.5833e-07

    All differences are within tolerance.
    ```

pro7onmatrix:
Fixes #5001.

As mentioned in the issue thread, referencing `std::vector<T>::size()` for an incomplete (e.g., forward-declared) `T` is ill-formed and fails to compile with Clang in C++20 mode.

This PR moves all problematic function definitions of this type from the header to the implementation files.

Note: Defining the functions as `inline` in the header like this
```cpp
class ATMForce {
public:
    ...
    int getNumParticles() const;
    ...
private:
    ...
    class ParticleInfo;
    std::vector<ParticleInfo> particles;
    ...
};

class ATMForce::ParticleInfo { ... };

inline int ATMForce::getNumParticles() const { return particles.size(); }
```
also passes the tests, but fails to link when using the API in a third-party project.

### DIFF
--- a/openmmapi/include/openmm/ATMForce.h
+++ b/openmmapi/include/openmm/ATMForce.h
@@ -177,28 +177,20 @@ public:
      *
      * This should be the same number of particles as the System
      */
-    int getNumParticles() const {
-        return particles.size();
-    }
+    int getNumParticles() const;
     /**
      * Get the number of Forces included in the ATMForce.
      */
-    int getNumForces() const {
-        return forces.size();
-    }
+    int getNumForces() const;
     /**
      * Get the number of global parameters that the interaction depends on.
      */
-    int getNumGlobalParameters() const {
-        return globalParameters.size();
-    }
+    int getNumGlobalParameters() const;
     /**
      * Get the number of global parameters with respect to which the derivative of the energy
      * should be computed.
      */
-    int getNumEnergyParameterDerivatives() const {
-        return energyParameterDerivatives.size();
-    }
+    int getNumEnergyParameterDerivatives() const;
     /**
      * Get the algebraic expression that gives the energy of the system
      */
--- a/openmmapi/include/openmm/CMAPTorsionForce.h
+++ b/openmmapi/include/openmm/CMAPTorsionForce.h
@@ -61,15 +61,11 @@ public:
     /**
      * Get the number of maps that have been defined.
      */
-    int getNumMaps() const {
-        return maps.size();
-    }
+    int getNumMaps() const;
     /**
      * Get the number of CMAP torsion terms in the potential function
      */
-    int getNumTorsions() const {
-        return torsions.size();
-    }
+    int getNumTorsions() const;
     /**
      * Create a new map that can be used for torsion pairs.
      *
--- a/openmmapi/include/openmm/CustomAngleForce.h
+++ b/openmmapi/include/openmm/CustomAngleForce.h
@@ -94,28 +94,20 @@ public:
     /**
      * Get the number of angles for which force field parameters have been defined.
      */
-    int getNumAngles() const {
-        return angles.size();
-    }
+    int getNumAngles() const;
     /**
      * Get the number of per-angle parameters that the interaction depends on.
      */
-    int getNumPerAngleParameters() const {
-        return parameters.size();
-    }
+    int getNumPerAngleParameters() const;
     /**
      * Get the number of global parameters that the interaction depends on.
      */
-    int getNumGlobalParameters() const {
-        return globalParameters.size();
-    }
+    int getNumGlobalParameters() const;
     /**
      * Get the number of global parameters with respect to which the derivative of the energy
      * should be computed.
      */
-    int getNumEnergyParameterDerivatives() const {
-        return energyParameterDerivatives.size();
-    }
+    int getNumEnergyParameterDerivatives() const;
     /**
      * Get the algebraic expression that gives the interaction energy for each angle
      */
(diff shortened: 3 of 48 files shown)


peastman:
Looks good.  Thanks for the PR!

It's likely to be a long time before we really need this.  We're still only on C++14.  But we'll probably get to C++20 some day, and we'll be glad not to have this break when we do.

PR #5013, opened 2025-07-16
Don't reinitialize CustomIntegrator on every step when no computations are present

### LINKED ISSUE #5010: Memory leak when using getState(getEnergy=True) with CustomIntegrator (on CPU platform)
When running the following code, total memory usage steadily increases, which causes the code to fail at some point due to running out of memory. However, this memory leak only happens when CustomIntegrator() is used in conjunction with getState(getEnergy=True).

```python
from openmm.app import *
from openmm import *
from openmm.unit import *

inpcrd = AmberInpcrdFile('./1h.inpcrd')
prmtop = AmberPrmtopFile('./1h.prmtop', periodicBoxVectors=inpcrd.boxVectors)
system = prmtop.createSystem(nonbondedMethod=PME, nonbondedCutoff=0.9*nanometer, rigidWater=False)
# integrator = LangevinIntegrator(100*kelvin, 1/picosecond, 0.0005*picoseconds)
integrator = CustomIntegrator(0)
simulation = Simulation(prmtop.topology, system, integrator)
simulation.context.setPositions(inpcrd.positions)
while True:
        """
        When using LangevinIntegrator, both getState() calls work without memory leak.
        When using CustomIntegrator:
                - The first getState() call works without memory leak.
                - The second getState() call (with getEnergy=True) causes a memory leak.
        """
        # state = simulation.context.getState(getEnergy=False, getForces=True)
        state = simulation.context.getState(getEnergy=True, getForces=True)
       
```

[memory_test.zip](https://github.com/user-attachments/files/21229029/memory_test.zip)

epretti:
Fixes #5010.  I wouldn't expect this to come up in practice, but it should prevent a CustomIntegrator with no computation steps from leaking memory.

### DIFF
--- a/platforms/reference/include/ReferenceCustomDynamics.h
+++ b/platforms/reference/include/ReferenceCustomDynamics.h
@@ -43,6 +43,7 @@ class ReferenceCustomDynamics : public ReferenceDynamics {
 private:
 
     class DerivFunction;
+    bool initialized;
     const OpenMM::CustomIntegrator& integrator;
     std::vector<double> inverseMasses;
     std::vector<OpenMM::Vec3> sumBuffer, oldPos;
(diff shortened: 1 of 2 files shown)


peastman (approved):
Looks good.

PR #5018, opened 2025-07-18
Update AWS S3 bucket region

epretti:
Updates bucket region for documentation upload


epretti:
The deployment only runs on `master` so I can't verify that it's working here.  But this should be the only change necessary in the repository.  If it doesn't work, either there's a problem with the new secrets, or the permissions on AWS need to be fixed.

peastman:
Let's go ahead and try it then.

epretti:
It's failing, but I think know what setting in AWS I need to change to fix it.  When the other jobs finish and it lets me retry the Docs job, I will see if it works.

epretti:
Should be working now.

PR #5021, opened 2025-07-25
Always safely overwrite checkpoint and state files

epretti:
Addresses the first of two items from https://github.com/openmm/openmm/issues/1645#issuecomment-3080876997.  `Simulation.saveCheckpoint()` and `Simulation.saveState()` should now ensure that if OpenMM is interrupted while overwriting an existing file, the existing checkpoint data will not be lost.  Previously, only `CheckpointReporter` did this, and there was a possibility to overwrite another file if one happened to have been manually created with the same name that `CheckpointReporter` would choose for a temporary file.


peastman (approved):
Looks good.

PR #4995, opened 2025-07-03
Adaptive quantum thermal bath

peastman:
I'm implementing the adaptive quantum thermal bath (adQTB) algorithm.  This is a fast method of simulating nuclear quantum effects.  I've written the reference implementation so far.

Implements #4225.

PR #5030, opened 2025-08-05
Cleanup examples, documentation, and README files

epretti:
- Update some outdated links, including references to the SimTK forum
- Update some old license text and copyright years
- Remove some outdated information about validation
- Reorganize examples into subdirectories (I went with `python-examples` and `cpp-examples` instead of `Python Examples` and `C++ Examples` as previously suggested to avoid annoyances that come with spaces in names in makefiles, *etc.*)
- Remove a few references to SimTK separate from the headers on every source file

Let me know of any changes that aren't good and I can update the PR.

The remaining references to SimTK should all be in the source file headers.  I plan to update those in a later PR to make review easier, since all of those changes should look the same.


epretti:
Something's wrong with the link checking script since it's claiming <https://openmm.org/development> is a 404.

epretti:
This may be a limitation in how Vue.js and GitHub Pages work.  The routing of URLs to pages on openmm.org is handled on the client side by Javascript that executes on page load.  To make this work, URLs that don't match any resource on the server should serve the default index page and Vue.js can ignore/handle the error.  But GitHub Pages only supports making a 404.html which for openmm.org is symlinked to index.html, and whose contents get served with status 404 when something is not found.  So if you go to, *e.g.*, https://openmm.org/development in a browser, Vue.js will run and the correct content will be displayed, but the script just sees a 404 status code and gives up.

I can exclude openmm.org from the sites to check as an easy fix, although then CI will lose the ability to alert us if the website becomes completely unreachable for some reason.  I don't know if it's possible to change the way GitHub Pages works so as not to return 404 on valid pages.

peastman (commented):
Looks good.

We'll want to consider carefully the timing of updating the file headers.  It will affect every source file, so create conflicts for most open pull requests.  We should wait until the big PRs like #4870 and #4995 have been merged.

peastman:
> I can exclude openmm.org from the sites to check as an easy fix

If https://openmm.org/development is the only URL currently causing problems, we could exclude just that one.

epretti:
OK, updated.

Agreed that it will be good to wait for the big PRs to be done before changing all of the file headers.

peastman:
Looks good.  Go ahead and merge when it's ready.

PR #5031, opened 2025-08-06
Radius of gyration force

### LINKED ISSUE #5015: Radius of Gyration restraint force in OpenMM
Hello!

We are recently investigating moving from CHARMM to OpenMM for some coarse-grained MD runs for comparison with SAXS data. We are very excited with the potential performance of some initial runs. 

However, we are running into an issue with implementing a RGYR Restraint Force ala CHARMM in OpenMM. We followed your suggestion of creating one group per atom, and one group with all atoms, and a bond for each single-atom group with the group containing all atoms within a CustomCentroidBondForce wrapped in CustomCVForce that does the harmonic restraint. 
```   
num_atoms = len(atom_indices)
rg2_force = CustomCentroidBondForce(2, f"(distance(g1, g2)^2)/{num_atoms}")
for i in atom_indices:
    rg2_force.addGroup([i], [1.0])  # g1: single atom
rg2_force.addGroup(atom_indices, [1.0] * num_atoms
for i in range(num_atoms):
    rg2_force.addBond([i, num_atoms])  # last group is full group (index = num_atoms)
super().__init__("0.5 * k_rg * (sqrt(cv) - rg0)^2")
self.addGlobalParameter("k_rg", k_rg)
self.addGlobalParameter("rg0", rg0)
self.addCollectiveVariable("cv", rg2_force)
self.setForceGroup(force_group)
```

1. As you cautioned in the other issue - this works when the number of atoms involved in the calc of the collective variable is in 100s - time for setup is a few seconds. When the number of atoms which are involved in the calculation of Rg is closer to a 1000, then the setup time becomes slower but still few tens of seconds. However, when we go to ~8000 atoms, the code doesn't move past the Simulation creation step for ~30 mins (longest we waited to test it). Is this expected?
The attached test case uses ~8000 atoms for the Rg restraint - ```python scripts/md.py``` should repr it in the attached tar. There is an nsys profile attached in the tar in case its helpful. A profile for a successful run (around 1000 atoms) shows that most of this time, it is in the ```findMoleculeGroups()``` function. In the py code it is at the ```simulation = Simulation(modeller.topology, system, integrator)``` step. 
2. We see only one thread on the CPU being active when where it possibly hangs. Is this part expected to be multi-threaded or single threaded?  Is the setup part parallelizable? 
3. We are under the impression that the Rg / Rg-based force calculation itself is not time consuming, it is the setup part that is currently just hanging (or requiring time > 30 min to complete) - but the dynamics calculation after setup completes proceeds faster since its using the GPU/is optimized - is this right? 
4. So, if we implemented the restraint force via a plugin (i.e. without using the bond infrastructure), would that avoid the slow setup problem since we would not need to create a group of >> 1000 atoms involved in a bond - the possibly non-optimized single threaded part of the code? Is that a reasonable approach?

[openmm-test.tar.gz](https://github.com/user-attachments/files/21306632/openmm-test.tar.gz)

How you installed OpenMM (pip, conda, compile from source, etc.).
    - compiled from source 
- The operating system and hardware (e.g. GPU model) you are using.
    - Running on Perlmutter, A100
- The platform (CUDA, OpenCL, etc.) you are using.
   - CUDA 
- Any other software (Python libraries, setup tools, etc.) you are using along with OpenMM to implement your workflow.
   - None
- The output of the command `python -m openmm.testInstallation`. It reports details about your installation and checks for some basic problems.
```
OpenMM Version: 8.3.1
Git Revision: 6e13f1361dcc3ffb581c7c97e3adfcc61257b813

3 CUDA - Successfully computed forces

Reference vs. CPU: 6.30823e-06
Reference vs. CUDA: 6.74442e-06
CPU vs. CUDA: 7.12587e-07

All differences are within tolerance.
```
Thanks!

peastman:
This adds a RGForce class that computes the radius of gyration.  It is mainly intended for use with CustomCVForce so you can use Rg as part of restraint potentials, metadynamics, etc.

Fixes #5015.


dhruvakulkarni:
Hi Peter, Thank you for the addition, very helpful for our use case!  
Our preliminary tests show that the CPU bottleneck is no longer observed for the 10000 atom system we were previously stuck on in issue #5015 . We are in the process of testing larger systems and running more correctness tests.

peastman:
Thanks!  Let me know what you find.

PR #5032, opened 2025-08-07
fix for scaled 1-4 LJ interactions when NBFIX is present in Gromacs topology file

feiglab:
This fix correctly scales LJ 1-4 interaction when reading a Gromacs topology file where NBFIX is present, pair parameters are not given via pairtype, and scaling of LJ 1-4 interactions is not 1.0. This is needed to correctly read Amber-type force fields that use NBFIX, for example a99sbws-tip4p2005s.


peastman:
Thanks!  Was this problem introduced by #5026, or did it exist before then?

feiglab:
The problem was there before but didn't become clear until the other changes were implemented.

peastman:
Ok.  We should keep this in mind as a candidate fix to include in a future patch release.

PR #5036, opened 2025-08-11
Convert C++ types to Python types in API docs

### LINKED ISSUE #5033: Types in Python API docs
The Python API is automatically generated from the C++ API.  An unfortunate side effect is that the Python API docs are full of references to C++ types like `double` and `vector`.  We should look into whether we can improve that.  There's already [a function](https://github.com/openmm/openmm/blob/d2a5d7e40990be7c8b9764aa0366da05443badc9/wrappers/python/src/swig_doxygen/swigInputBuilder.py#L156-L167) to translate C++ types to Python types, but it does very little.  Expanding that function could probably lead to more useful docs.

peastman:
Fixes #5033.

The Python API is generated automatically from the C++ API.  An unfortunate consequence is that in the Python API docs, function signatures list the argument types and return types as the original C++ types, not the Python types they get translated into.  This adds code to translate them.

The result is still not perfect, but it's a big improvement.  Consider the argument to `context.setPositions()`.  In C++ its type is `const vector<Vec3>&`.  With this PR, it now gets translated to `Sequence[Vec3]` in the Python API docs.  That is indeed an accepted type for that argument, but it also accepts other types, such as a sequence of sequences, or a 2D array, or a `Quantity` that wraps any of those types and adds distance units to it.  So it still doesn't express the full range of accepted types, but it's a lot better than listing it as `vector<Vec3>` which isn't even a Python type.

The other big issue in the Python API docs is that all the code examples are still in C++.  Fixing that will be a separate project.

### DIFF
--- a/docs-source/api-python/process-docstring.py
+++ b/docs-source/api-python/process-docstring.py
@@ -86,6 +86,18 @@ substitutions = {'double':'float', 'long long':'int', 'string':'str',
                  'seti':'set[int]'
                 }
 
+def convert_type(type):
+    if type in substitutions:
+        type = substitutions[type]
+    if '<' in type:
+        match = re.match('vector<(.*?),.*>', type)
+        if match is not None:
+            type = f'tuple[{convert_type(match[1])}]'
+        match = re.match('map<(.*?),(.*?),.*>', type)
+        if match is not None:
+            type = f'Mapping[{convert_type(match[1])}, {convert_type(match[2])}]'
+    return type
+
 def process_signature(app, what, name, obj, options, signature, return_annotation):
     if return_annotation is not None:
         # Convert C++ types to Python types
@@ -93,8 +105,7 @@ def process_signature(app, what, name, obj, options, signature, return_annotatio
             return_annotation = return_annotation[5:]
         if return_annotation.endswith(' const &'):
             return_annotation = return_annotation[:-8]
-        if return_annotation in substitutions:
-            return_annotation = substitutions[return_annotation]
+        return_annotation = convert_type(return_annotation)
     return (signature, return_annotation)
 
 


epretti:
This is nice!
- It might be nice to strip off `*` as well as `&`, to avoid things like `System.addForce()` documented as taking a `Force *`, and others.
- Is there a way to do the replacements more uniformly?  I see `set[int]` and `tuple[int, int]` are in the substitution table, but then `CustomNonbondedForce.getInteractionGroupParameters()` still returns `set<int>` for example, and `CustomNonbondedForce.createExclusionsFromBonds()` takes the strange C++/Python hybrid `Sequence[pair<int, int>]`.  I'm sure handling everything properly is going to be really difficult though, so this is obviously better than not having any translation at all.
- Some strange-looking types remain in the documentation for `SerializationNode` and `SerializationProxy`.  Are these even intended to be usable from the Python API?

peastman:
Thanks, I'll take a look at those.  Let me know if you spot any others that look wrong.

Note that I had to use completely different mechanisms to translate arguments and return types.  The former happens while generating the interface files, before SWIG has been run.  The latter happens while building the API docs, after the SWIG wrapper has been generated.  Except there's also the complication of output arguments to C++ methods, which get turned into return values in the Python API.  Those get documented while building the interface files.

SerializationNode and SerializationProxy aren't really intended to be used from Python.  I guess there's no reason you couldn't, but I've never come across an application of them.  The only serialization related class that's normally used from Python is XmlSerializer.

peastman:
Those problems should now be fixed, except for the serialization ones.  Looking at those, I see that SerializationProxy has methods that take a `std::type_info`.  That's not going to be very usable from Python.

epretti:
I found a few more return type oddities, excluding the serialization ones:
- `QTBIntegrator.getTypeAdaptationRates()` has `map< int,double,std::less< int >,std::allocator< std::pair< int const,double > > >`.
- `RMSDForce.getReferencePositions()` has `vector< Vec3,std::allocator< Vec3 > >`.
- Enumeration types get the fully qualified C++ names even though they are just represented as integers in the Python API.  (Ideally they could be Python enums, but that's a different story.)  For instance, `NonbondedForce.getNonbondedMethod()` has `OpenMM::NonbondedForce::NonbondedMethod`.  Similarly for others.

I found those by logging and inspecting all of the unique outputs of `process_signature()` and `docstringTypemap()`, so hopefully that should actually be all of them.

peastman:
I fixed the `map` and `vector` return types.  I think the enumerations are ok, because that better expresses the intent.  Enums in C++ can also be interpreted as integers.  But really you should compare `getNonbondedMethod()` to the fields defined in NonbondedForce, not to simple integer values.

PR #5041, opened 2025-08-12
Fix alternate residue name for water

k-ujihara:
The `alt1` field of the water residue definition in `pdbNames.xml` is 'H20'. I believe this should be 'H2O', as 'H20' is assigned for quinolone derivative.

### DIFF
--- a/wrappers/python/openmm/app/data/pdbNames.xml
+++ b/wrappers/python/openmm/app/data/pdbNames.xml
@@ -274,7 +274,7 @@
   <Atom name="H72" alt1="2H5M"/>
   <Atom name="H73" alt1="3H5M"/>
  </Residue>
- <Residue name="HOH" alt1="H20" alt2="WAT" alt3="SOL" alt4="TIP3" alt5="TP3" alt6="TIP">
+ <Residue name="HOH" alt1="H2O" alt2="WAT" alt3="SOL" alt4="TIP3" alt5="TP3" alt6="TIP">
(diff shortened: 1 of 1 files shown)


peastman:
The alternate names are ones we've encountered in files found in the wild.  There's no logic to them.  They're just names other programs have used when writing PDB files.

k-ujihara:
Yes, though that might be a typo.

k-ujihara:
I fixed the number "0 (zero)" to the letter "O".

epretti:
If we changed or removed this name, we wouldn't be able to properly read PDB files that use it for water.  A search turns up a few other tools that also use `H20`, not `H2O`, as an alternate name for water, suggesting that such files do exist in the wild, so we shouldn't alter it.

k-ujihara:
I understand, but "H20" is already assigned to "6-chloro-4-(cyclohexyloxy)-3-isopropylquinolin-2(1H)-one". That could cause problems. I think just removing is one of the options.

epretti:
Do you have a specific PDB containing this molecule that OpenMM fails to read correctly?  This molecule is not a standard residue per the PDB specification, which means that its connectivity (`CONECT` records) should be specified explicitly in a PDB.  The predefined residues in OpenMM are just used so it can assign standard bonds to the standard residues for which `CONECT` records are not required to be present.

k-ujihara:
I understood the situation. Thank you for your explanation.

PR #5043, opened 2025-08-12
Replace C++ code examples in Python API docs

peastman:
The Python API is created automatically from the C++ API.  That includes the documentation for it.  Since the documentation in C++ header files has code examples in C++, that means the Python API documentation ends up with example code in C++ rather than Python.

I created a mechanism that lets us include separate C++ and Python versions of each example.  It looks like this:

```
 * \verbatim embed:rst:leading-asterisk
 * <c++>
 * CustomIntegrator integrator(0.001);
 * integrator.addComputePerDof("v", "v+0.5*dt*f/m");
 * integrator.addComputePerDof("x", "x+dt*v");
 * integrator.addComputePerDof("v", "v+0.5*dt*f/m");
 * </c++>
 * <python>
 * integrator = CustomIntegrator(0.001)
 * integrator.addComputePerDof("v", "v+0.5*dt*f/m")
 * integrator.addComputePerDof("x", "x+dt*v")
 * integrator.addComputePerDof("v", "v+0.5*dt*f/m")
 * </python>
 * \endverbatim
```

When building each version of the documentation, it includes the correct version of the code.

So far I've only converted CustomIntegrator to use this mechanism.  Before doing more, I want to get feedback on the approach and implementation.


epretti:
I think the syntax is fine.

The implementation with editing files using regular expressions and escaped XML entities feels like a bit of a hack, but I guess it's not the only place where we have to do ad-hoc processing to generate the documentation.

peastman:
Yeah, it's a little hacky, but it was the only thing I could come up with.  The Python docs are built with autodoc, which provides hooks to modify the results as they're being generated.  The C++ docs are built with Breathe, which doesn't provide anything like that.  That's why I had to modify the xml files before it reads them.

epretti:
I think there's one more `\verbatim` block in `Platform.h` that shows up in the [Python docs](https://docs.openmm.org/latest/api-python/generated/openmm.openmm.Platform.html#openmm.openmm.Platform).

peastman:
Thanks, fixed.

PR #5050, opened 2025-08-21
NBFIX in Gromacs for 1-4 pairs

feiglab:
Code fix to match Gromacs in applying NBFIXes to 1-4 pairs when there are no explicit (1-4) pair parameters given via 'pairtypes'.


peastman:
Thanks!  Can you give a little more information about the error?  In what situation does it happen, and what's the incorrect behavior?

feiglab:
When NBFIX pairs are present in the Gromacs topology file, the old code only applied them to non 1-4 pairs. That is fine for most scenarios but some force fields (amber99sb-disp) have NBFIX for atom types that (also) come up in 1-4 pairs. Unless alternate parameters are specific for the 1-4 pair (via the [ pairtypes ] section) GROMACS also applies the NBFIX to the atoms in the 1-4 pairs (with scaling factors if present). The openMM code didn't apply the NBFIXes and instead used standard combining rules for 1-4 interactions. The fix now applies NBFIXes if present to 1-4 pairs unless they are overwritten by explicit pairtype interactions from the topology file.

peastman:
Got it, thanks!

peastman:
Is this ready to merge once CI finishes?

feiglab:
I think so.

PR #5053, opened 2025-08-24
Implement setDrudeForce for the Drude plugin integrators

egallicc:
The Drude integrators use the DrudeForce stored in the System to identify Drude particles. However, this approach does not work when the DrudeForce is under the ATMForce. In this case, the System does not have a Drude Force even though it is used in inner contexts. This change lets the user optionally provide the DrudeForce directly.  For example:

```
integrator.setDrudeForce(XmlSerializer::clone<DrudeForce>(*drudeforce));
```

An alternative solution is to have the Drude integrators search the ATMForce for a DrudeForce in addition to the System. However, there might be other circumstances (CustomCVForce + Metadynamics, for example) when the DrudeForce is "hidden" under other Forces. The approach of giving the Drude integrator a copy of the DrudeForce directly might be more general.

This change does not modify the default API. It adds an optional 'setDrudeForce()' method and the corresponding `getDrudeForce()' to DrudeIntegrator. The change is duplicated for the DrudeNoseHooverIntegrator that does not use the DrudeIntegrator class.

The changes in the tests are placeholders. Complete tests will be implemented once this approach is approved.


peastman:
I worry this approach will be brittle and lead to mistakes.  The problem is that the DrudeForce you give to it isn't the same one added to the System.  There's nothing to ensure they're identical, opening the possibility of subtle errors.  And if you modify one, it doesn't modify the other.  Another mode of failure is that someone does try to pass the same DrudeForce object in both places, which will lead to a segfault.

I think the better approach is to create a mechanism for it to find DrudeForces contained inside other forces, like ATMForce or CustomCVForce.  I have an idea for how to do that.  Let me see what I can do.

egallicc:
Thanks! Yes, I agree, this solution is not as robust as using the DrudeForce used in the calculation, if there is a general way to find it.

peastman:
Superseded by #5054.

PR #5054, opened 2025-08-25
Drude integrators can find DrudeForce inside another force

peastman:
This is an alternative to #5053.  I added a new method to ForceImpl, `getContainedForces()`.  The default implementation returns an empty vector.  ATMForce and CustomCVForce override it to return the forces they contain.  All the Drude integrators now check it when searching for a DrudeForce.

cc @egallicc


egallicc:
Thanks!

PR #5065, opened 2025-09-05
Fix error message formatting for bonds to elements without atoms

epretti:
Fixes a minor bug in #5020 in which error messages for incorrect bonds to elements without atoms don't get formatted and an unrelated error gets shown instead.  Normally, there shouldn't be bonds to extra sites that would trigger this, but it can happen if one attempts to parameterize a topology with explicitly specified bonds to atoms for which OpenMM couldn't guess the elements.


peastman (approved):
Looks good.

PR #5073, opened 2025-09-15
Update constant potential entry in user guide

epretti:
Emphasizes that it is incorrect to set non-zero charges in a NonbondedForce when using constant potential, and adds a note about running constant potential simulations with polarizable models.


peastman:
Looks good.

PR #5074, opened 2025-09-16
Update file headers

epretti:
Replaces SimTK file headers with the updated header discussed previously.  This should remove the remaining references to the old <https://simtk.org> URL.


peastman:
Looks good.  Two more files have been added since you created this: tinkerfiles.py and amoebaforces.py.

PR #5080, opened 2025-09-19
Cleanup TinkerReader, ForceField, and amoebaforces

JMorado:
As discusses in #4769, this PR servers to reduce the duplicated code between `ForceField` and `TinkerFiles`. 
Current progress checklist for the refactoring:

- [x] AmoebaBondForce
- [ ] AmoebaAngleForce
- [ ] AmoebaOutOfPlaneBendForce 
- [ ] AmoebaInPlaneAngleForce
- [x] AmoebaStretchBendForce
- [ ] AmoebaTorsionForce 
- [x] AmoebaPiTorsionForce
- [x] AmoebaUreyBradleyForce
- [x] AmoebaWcaDispersionForce
- [x] AmoebaGeneralizedKirkwoodForce
- [ ] AmoebaMultipoleForce
- [x] AmoebaVdwForce
- [ ] AmoebaTorsionTorsionForce
- [x] AmoebaAngleTorsionForce 
- [x] AmoebaStretchTorsionForce 

We should also consider updating https://docs.openmm.org/latest/userguide/library/08_amoeba_plugin.html to reflect the current state of the AMOEBA code.

### DIFF
--- a/wrappers/python/openmm/app/internal/amoebaforces.py
+++ b/wrappers/python/openmm/app/internal/amoebaforces.py
@@ -67,12 +67,34 @@ class BaseAmoebaForceBuilder:
         ValueError
             If the atom does not have an associated element or atomic number.
         """
-        if hasattr(atom, 'element'):
-            return atom.element.atomic_number
-        elif hasattr(atom, 'atomicNumber'):
+        if hasattr(atom, 'atomicNumber'):
             return atom.atomicNumber
+        elif hasattr(atom, 'element'):
+            return atom.element.atomic_number
         else:
             raise ValueError(f"Atom {atom} does not have an associated element or atomic number.")
+
+    @staticmethod
+    def _getNeighbors(atom: Any) -> List[int]:
+        """
+        Get the indices of neighboring atoms bonded to the given atom.
+
+        Parameters
+        ----------
+        atom : Any
+            Atom object that has either a bonds attribute or a bondedAtoms attribute.
+
+        Returns
+        -------
+        List[int]
+            List of indices of neighboring atoms.
+        """
+        if hasattr(atom, 'bondedAtoms'):
+            return [a.index for a in atom.bondedAtoms]
+        elif hasattr(atom, 'bonds'):
+            return [a for a in atom.bonds]
+        else:
+            raise ValueError(f"Atom {atom} does not have bonded atoms information.")
 
     def _findExistingForce(self, sys: mm.System, forceType: type, energyFunction: Optional[str] = None, name: Optional[str] = None) -> Optional[Any]:
         """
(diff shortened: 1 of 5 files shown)


JMorado:
I've noticed that currently there are two (redundant, I think?) ways of generating the proper periodic torsions in `ForceField`: one uses [`PeriodicTorsionGenerator`](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/forcefield.py#L4023-L4122), and the other uses [`AmoebaTorsionGenerator`](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/forcefield.py#L4023-L4122). The latter seems to have been kept around to support [`amoeba2013.xml`](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/data/amoeba2013.xml), the only XML file that still uses the `AmoebaTorsionForce` tag (`amoeba2009.xml` and `amoeba2018.xml` already use `PeriodicTorsionForce`).

To avoid having to maintain both generators, I'd suggest we could regenerate the `amoeba2013.xml` file and remove `AmoebaTorsionGenerator`.

JMorado:
In terms of code, the refactoring is done from my side. Please let me know if you have any suggestions for further improvements to the current implementation.

I've also started updating the AMOEBA Plugin documentation, but  I'm not entirely sure whether references to the OpenMM-Tinker plugin are still valid. I can't find any of the files mentioned [here](https://docs.openmm.org/latest/userguide/library/06_integration_examples.html?highlight=tinker%20openmm#tinker-openmm). Should Section 15 of the documentation focus only on the current state of AMOEBA force field support in OpenMM?

wrappers/python/openmm/app/internal/amoebaforces.py
```diff
@@ -1091,245 +1493,147 @@ def createForce():
...
-
+        Get chiral atom index using consistent TINKER algorithm for both ForceField and TinkerFiles usage.
+        
         Parameters
         ----------
-        data : ForceFieldData or list
-            ForceFieldData object (for ForceField usage) or list of TinkerAtom objects (for TinkerFiles usage)
+        data : Union[List[List[int]], List[Any]]
+            Either a list of bond lists (ForceField usage) or list of TinkerAtom objects (TinkerFiles usage).
+            The method automatically detects the data type by checking for the 'bonds' attribute.
```
peastman, line 1587: Same here.

peastman (commented):
Sorry for the delay in reviewing this!  It mostly looks pretty good.  Just a few issues to address.

JMorado:
Many thanks again for the review. I’ve addressed all the feedback -- let me know if there’s anything else I should adjust or if it looks good as is.

peastman:
Looks good, thanks!

Issue #5081, opened 2025-09-19
AMOEBA force field XML files are out of date


JMorado:
Hi,  

Lately, I’ve noticed a few issues with the XML files describing the various AMOEBA force fields available out-of-the-box in OpenMM. The current XML files were created several years ago, and since then:  

- Some parameters have been introduced that are not included in the OpenMM force fields (see, e.g., the history of commits of the AMOEBA 2018 force field: [amoebabio18.prm](https://github.com/TinkerTools/tinker/commits/release/params/amoebabio18.prm)). This means that someone running what appears to be the same force field on Tinker and OpenMM may get slightly different outputs. I recognise that simply updating the XML files without notice could cause reproducibility issues. Some of these discrepancies are mitigated by the `TinkerReader`, which allows users to read the latest Tinker force fields, but we should still decide whether we want to update the force fields on the OpenMM side. If not, we could at least include a warning or note in the documentation stating that these force fields may be slightly diffrent from the Tinker ones.  
- There is the issue described in [this comment](https://github.com/openmm/openmm/pull/5080#issuecomment-3311853548), where old tags are still present in `amoeba2013.xml`.  
-  Importantly, and this one should be corrected asap, the `AmoebaAngleTorsionForce` parameters defined in `amoeba2018.xml` are incorrect. I had already mentioned my suspicion in #4769, but now I understand the root of the problem: the current `amoeba2018.xml` file was created on 2021-11-04, and just a couple of weeks later, [on the Tinker side](https://github.com/TinkerTools/tinker/commit/df789b7ec5cf8a201f3b8ceb2c246b0bde522f62) they apparently realised this force had some issues with the units. OpenMM retained the original parameters in `amoeba2018.xml`, which assume force constants given in kcal/mol/degrees, even though everything internally assumes radians. We should correct the units of these parameters. This issue only affected simulations with nucleic acids. The test in `TestForceField` still uses the [old results from Tinker](https://github.com/openmm/openmm/blob/master/wrappers/python/tests/TestForceField.py#L1982), which differ from what Tinker [currently produces](https://github.com/openmm/openmm/blob/master/wrappers/python/tests/TestTinkerFiles.py#L234). This explains why this discrepancy went undetected for so long.

Looking forward to hearing any thoughts on this. Many thanks!

peastman:
Thanks, we should definitely fix the incorrect parameters.

JMorado:
I'll submit a PR with the fixes.

JMorado:
[...]

PR #5086, opened 2025-09-19
Update AMOEBA XML, .prm files, and parsing of Tinker files

JMorado:
Regarding issue #5081, this turned out to be a bit more confusing than I initially expected. The OpenMM AMOEBA parameters are entirely correct, so I apologise for the earlier confusion. The real problem was that I had been using some outdated `amoebabio18.prm` files to calculate the Tinker energies. Once I updated those files, the tests showed no discrepancies. The new `.prm` files have a slightly different format (for example, in the `tortor` parameters), so I also had to update the `TinkerReader` to handle that.

This PR:

- Updates `processTinkerForceField.py` and `TinkerReader` to read `tortor` parameters in the format Tinker currently uses
- Replaces outdated `amoebabio18.prm` files across the codebase
- Updates the AMOEBA force field XML files
- Updates  `TestForceField.py::AmoebaTestForceField::test_Amoeba18BPTI`. I've replaced the `bpti.pdb` file with another one that has the exactly same protein structure but does not produce incorrect structures when using `pdbxyz`. I've also updated the atomic multipoles and polarization energies of the test, which are slightly different with the new parameters. All values match Tinker outputs, and all tests pass locally.


peastman:
What did you regenerate the XML files for?  What has actually changed in them?  I tried to look through the changes, but mostly they just seem to add lots of unnecessary digits to the values, for example

```xml
-  <WcaDispersion class="19" radius="0.17025" epsilon="0.46024" /> 
+  <WcaDispersion class="19" radius="0.17025" epsilon="0.46024000000000004" /> 
```

JMorado:
They bring the XML files up to date with the tinker .prm files, adding some new parameters and changing others, e.g.:

```xml
-  <WcaDispersion class="100" radius="0.134" epsilon="0.928848" /> 
-  <WcaDispersion class="101" radius="0.1715" epsilon="1.046" /> 
-  <WcaDispersion class="102" radius="0.20600000000000002" epsilon="1.4225600000000003" /> 
-  <WcaDispersion class="103" radius="0.21600000000000003" epsilon="1.79912" /> 
-  <WcaDispersion class="104" radius="0.23050000000000004" epsilon="2.1756800000000003" /> 
+  <WcaDispersion class="100" radius="0.1935" epsilon="1.61084" /> 
+  <WcaDispersion class="101" radius="0.21000000000000002" epsilon="1.75728" /> 
+  <WcaDispersion class="102" radius="0.134" epsilon="0.928848" /> 
+  <WcaDispersion class="103" radius="0.1715" epsilon="1.046" /> 
+  <WcaDispersion class="104" radius="0.20600000000000002" epsilon="1.4225600000000003" /> 
+  <WcaDispersion class="105" radius="0.21600000000000003" epsilon="1.79912" /> 
+  <WcaDispersion class="106" radius="0.23050000000000004" epsilon="2.1756800000000003" /> 
```
I can remove the trailing zeros but it will be difficult to get a meaningful diff because the reference XMLs also contain many of them. The order of some XML elements has changed for some forces, but that’s because how `processTinkerForceField.py` writes them. It's probably easier to see what's changed in the .prm files.

JMorado:
It should now be clearer what has changed in the XML files.

peastman:
In defining the angles, I see a lot of empty strings were changed to 0.  For example,

```xml
  <Angle class1="2" class2="1" class3="" class4="" k="0.053147454159115176"/>
```

changed to

```xml
  <Angle class1="2" class2="1" class3="0" class4="0" k="0.053147454159115176"/>
```

Is that intentional?

JMorado:
The builder for the out-of-plane bends [uses '0' as a wildcard](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/internal/amoebaforces.py#L471-L489) to identify some of the trigonal centres. This should be the only force for which wildcard atom classes are used in the Tinker .prm files.

The aim is actually for both `''` and `'0'` to be equivalent. In the refactoring in #5080 I add [code to handle both cases](https://github.com/openmm/openmm/pull/5080/files#diff-7e10bd00e3ea4dd2e148b51d75c4bedff6a777d116ef99cf7869f276e8cd5973R3776-R3780):

```python
if 'class3' in angle.attrib and 'class4' in angle.attrib:
    if angle.attrib['class3'] == '':
        angle.attrib['class3'] = '0'
    if angle.attrib['class4'] == '':
        angle.attrib['class4'] = '0'
```

I’ve wondered if it would be worth fixing the precision of the parameters in these XML files to a set number of decimal places, say 8 for example. That way the diffs would be more meaningful if more parameters are added in the future. I realise currently it’s painful to compare them on github... apologies for that! I found a side-by-side diff in my IDE to be much more helpful for this.

peastman:
`"0"` isn't a wildcard.  It means type 0, which is a valid atom type!  `""` is what we use for wildcards everywhere.

JMorado:
Thanks for clarifying. In the Tinker FFs, the atom classes/types are indexed starting at 1 rather than 0, which is why everything has been working fine so far. I’ll update this to keep it consistent with the other FFs.

peastman:
Just checking on the status of this.  We're about half way through the beta period, so all remaining changes need to be finished in the next couple of weeks.

JMorado:
Hi @peastman, I'll finish both PRs this week -- many thanks for reviewing #5080 already.

JMorado:
The XML files have been updated. I've also updated `processTinkerForceField.py` so that it can generate the 2009 and 2013 force fields, and I've added the necessary XML residue files. Please let me know if everything looks good or if any changes are needed. Thanks!

peastman:
Looks good.  Shall I merge this and #5080?

JMorado:
Yes, please. Since both PRs modify the test files, I would suggest merging #5080 first. If any conflicts arise in this PR afterward, I'll resolve them.

peastman:
I merged #5080, and this now has conflicts that prevent merging.

JMorado:
It should be mergeable now.

peastman:
Thanks!  I'll merge it when the builds finish.

Issue #5075, opened 2025-09-17
Minimizing glycosylations


e-pettersen:
I've written some code to get ChimeraX to use OpenMM to minimize structures.  It employs Antechamber to determine GAFF types and partial charges for non-standard residues.  For the most part, it works great.  But for some systems involving glycosylations, those residues wind up with long bonds (e.g. C-C bonds > 2 angstroms) and with much, much flatter ring systems than they should have.  Other systems with similar glycosylations work fine, and I haven't been able to discern the pattern of what causes the wayward behavior.

My question is: what information can I provide to you to make it easier to ascertain what is going wrong with the problematic systems?

--Eric

peastman:
I'm not sure what your question is.  Is the problem related to OpenMM or to Antechamber?

If you can provide a complete test case that can be used to reproduce the problem, including all necessary input files, that would make it clearer.

e-pettersen:
It's an OpenMM problem (AFAICT) -- the GAFF types and partial charges all seem reasonable.  I will try to work up a test script and example files that demonstrates the problem.  Should I attach those here?

peastman:
Yes, that would be great.

e-pettersen:
[minimize.tar.gz](https://github.com/user-attachments/files/22504166/minimize.tar.gz)

e-pettersen:
I didn't find a way to dump a Context to a file, so I had to make a script to create the Context from scratch that reproduces the problem.  I've attached a gzipped tar archive that has two files.  minimize.py is a Python script that reproduces the problem, and gaff-2.2.20.xml is a parameter file used by the script.  The script builds the problematic glycosylated structure from scratch, runs a minimization, and then prints out the C4-C5 bond length in the NAG residue -- which is nearly 2.5 angstroms, much too long for a C-C single bond.  That's just one example -- most of the bonds in the glycosylation residues are much too long.  Let me know if you need more info.

epretti:
Thanks for this test case.  There's a bug in the `__init__()` method of `_TemplateAtomData` (the type of the object you are assigning `parameters['charge']` on.  It causes all instances created to share the same `parameters` dictionary if one is not explicitly given when the object is created:
https://github.com/openmm/openmm/blob/05472c9a812927c863be67abbb3376e944b2c7ef/wrappers/python/openmm/app/forcefield.py#L745-L747
[...]

PR #5091, opened 2025-09-24
Create a new dictionary every time `_TemplateAtomData.__init__()` is called without `parameters`

epretti:
See #5075.


peastman (approved):
Thanks for figuring this out!

Issue #5094, opened 2025-09-30
Force computed from user code


peastman:
I'm looking into creating a new Force class that is computed by code provided by the user.  Most often that code would be written in Python, though it could also work from C++.  It would look something like this.

```python
class MyForceCalculation(UserForceCalculation):
    def calculate(self, positions, boxVectors):
        ...
        return energy, forces

force = UserForce(MyForceCalculation())
system.addForce(force)
```

This class would be useful for several things.

- It could be used to invoke ML models that can't be exported in supported formats like TorchScript and ONNX.
- OpenMM-ML could provide a generic MLPotential that wraps an ASE calculator, giving users immediate access to any potential function supported by ASE.
- It could be an alternative to custom forces for implementing user calculations.  The performance would be much slower than custom forces, but it could implement a much wider range of calculations.

peastman:
This is implemented in #5122.

PR #5122, opened 2025-10-30
Created PythonForce

peastman:
Implements #5094.  This is a force class that is implemented by user-supplied Python code.  Here is an example of what it looks like.

```python
def compute(state):
    pos = state.getPositions(asNumpy=True).value_in_unit(nanometer)
    k = state.getParameters()['k']
    energy = k*np.sum(pos*pos)
    force = -0.5*k*pos
    return energy*kilojoules_per_mole, force*kilojoules_per_mole/nanometer

force = PythonForce(compute, {'k':2.5})
```

Performance is obviously much slower than standard and custom forces, but it's very flexible.  It should be useful for supporting ML models that can only be executed with Python code.


peastman:
To see how bad the overhead is, I created a minimal force that doesn't compute anything:

```python
def compute(state):
    return 0.0, np.zeros((len(state.getPositions(asNumpy=True)), 3))
```

I ran some of the standard benchmarks to see how much it slows them down.  Here are the speeds on a RTX 4080.

|Test|Normal|With PythonForce|Ratio|
|---|---|---|---|
DHFR|2095|885|2.4
ApoA1|660|156|4.2
Cellulose|168|31|5.4

It's slow enough that you really don't want to use it with standard force fields.  It should be fine for most ML potentials though.

I'm hoping to find ways to speed it up a bit more.  That's where it is right now.

peastman:
I was able to speed it up a bit, though it doesn't change the overall conclusion: it's fine for ML, but you probably don't want to use it with conventional force fields.  Here are the new benchmarks.

|Test|Normal|With PythonForce|Ratio|
|---|---|---|---|
DHFR|2095|959|2.2
ApoA1|660|224|2.9
Cellulose|168|38|4.4

wenyan4work:
thank you for this addition and it will be very useful for both MLFF MD and ab-initio MD.

Actually we wrote our plugin for the same purpose https://github.com/bytedance/OpenMM-Python-Force and found that the overhead is actually lower than lammps when running even highly optimized MLFF models. So I think overhead should not be too-big an issue.

Issue #4746, opened 2024-12-05
SASA based force


peastman:
Implicit solvent models often include a term that is proportional to the solvent accessible surface area (SASA).  It's challenging to compute, especially since you need the derivatives as well.  GBSAOBCForce and AmoebaGeneralizedKirkwoodForce both use a fast but inaccurate approximation.

I want to add a force class that computes it accurately rather than just approximately.  The state of the art for computing it is based on alpha shape theory.  Here are some papers describing implementations of it on CPUs and GPUs, respectively.

https://doi.org/10.1021/acs.jcim.2c01346
https://doi.org/10.1021/acs.jctc.3c01366

Both those implementations are open source (LGPL for AlphaMol, MIT for dSASA), so we should be able to either use them directly or adapt them.

epretti:
We've discussed that there are challenges with an efficient implementation of the exact method.  I've taken a brief look into approximate methods that will hopefully be more accurate than what OpenMM currently supports.  It should definitely be possible to implement [LCPO](https://doi.org/10.1002/(SICI)1096-987X(19990130)20:2%3C217::AID-JCC4%3E3.0.CO;2-A), which is basically a particular form of three-body potential.  Since the method's parameters are assigned per atom based on elements and bond topology, the potential could be implemented in the library layer and the parameters assigned in the application layer (we'd have to think about the best way to expose this).

We could also consider [pwSASA](https://doi.org/10.1021/acs.jctc.8b00413), a newer method that Amber supports (`gbsa=3`).  In graduate school, a coworker and I actually implemented this for OpenMM; it's just a pairwise potential that we implemented with `CustomNonbondedForce`, and I don't know whether we'd need a more complicated implementation.  The most annoying part was hand-translating a hardcoded table from Fortran to Python for looking up the parameters, which Amber of course decided to do based on residue names.

epretti:
To follow up on this with a bit more information and thoughts about the API: at one point we had mentioned having something like a `SASAForce` for this.  However, it may be better to have separate classes for different methods, since the underlying functional forms as well as necessary parameters are different:

- The existing ACE model is probably well-suited for staying inside `GBSAOBCForce` since it is dependent on the computed Born radii.
- LCPO:
  - Requires specifying a radius and coefficients $P_1$, $P_2$, $P_3$, $P_4$ for each atom (type).
[...]

PR #5130, opened 2025-11-07
Add LCPO method

epretti:
Adds LCPO (linear combination of pairwise overlaps) as discussed in #4746.

- [X] Reference platform
- [X] CPU platform
- [x] Common platform
- [X] Parameters for Amber files
- [x] Parameters for Topology + XML
- [x] Optimize common platform

### DIFF
--- a/wrappers/python/openmm/app/internal/lcpo.py
+++ b/wrappers/python/openmm/app/internal/lcpo.py
@@ -244,14 +244,19 @@ def getLCPOParamsTopology(topology):
             if atomicNumbers[index2] > 1:
                 numHeavyBondsList[index1] += 1
 
-    # Identify carbonyl Os and their partner Cs, then identify carboxylate Cs.
-    carbonylO = {}
+    # Identify terminal O atoms (-O(-), =O) and their bonding partners, then
+    # identify partners that look like carboxylate C or phosphate P atoms.  Note
+    # that we follow Amber's implementation and assign phosphate O atoms the
+    # LCPO carboxylate O- type.  The original LCPO publication does not comment
+    # on the appropriate parameters for this case, and NAMD's implementation has
+    # inconsistent behavior between Amber and CHARMM, so we follow Amber here.
+    terminalO = {}
     for index1, (atomicNumber, bondedTo, numTotalBonds) in enumerate(zip(atomicNumbers, bondedToList, numTotalBondsList)):
         if atomicNumber == 8 and numTotalBonds == 1:
             index2, = bondedTo
-            if atomicNumbers[index2] == 6 and numTotalBondsList[index2] == 3:
-                carbonylO[index1] = index2
-    carboxylateC = {indexC for indexC, countO in collections.Counter(carbonylO.values()).items() if countO > 1}
+            if (atomicNumbers[index2], numTotalBondsList[index2]) in (6, 3) or (15, 4):
+                terminalO[index1] = index2
+    terminalOPartners = {indexPartner for indexPartner, countO in collections.Counter(terminalO.values()).items() if countO > 1}
 
     # Identify sp2-hybridized Ns with 3 bonding partners based on the
     # hybridization of surrounding C atoms.  This may fail in a few unusual
@@ -314,8 +319,9 @@ def getLCPOParamsTopology(topology):
                     _raiseLCPOException(atomicNumber, numTotalBonds, numHeavyBonds)
         elif atomicNumber == 8:
             if numTotalBonds == 1:
-                # sp2 O (check to see if it is a carboxylate O).
-                if index in carbonylO and carbonylO[index] in carboxylateC:
+                # sp2 O (check to see if it is a carboxylate or phosphate O: see
+                # above regarding assigning the carboxylate type on phosphates).
+                if index in terminalO and terminalO[index] in terminalOPartners:
                     params = LCPO_PARAMETERS['O_carboxylate']
                 else:
                     params = LCPO_PARAMETERS['O_sp2_1']
--- a/wrappers/python/tests/TestAmberPrmtopFile.py
+++ b/wrappers/python/tests/TestAmberPrmtopFile.py
@@ -350,7 +350,7 @@ class TestAmberPrmtopFile(unittest.TestCase):
         contextLCPO = Context(systemLCPO, VerletIntegrator(0.001), Platform.getPlatformByName("Reference"))
         contextNone.setPositions(pdb.positions)
         contextLCPO.setPositions(pdb.positions)
-        energyRef = 13.7350 * kilocalorie_per_mole
+        energyRef = 14.1908 * kilocalorie_per_mole
         energyLCPO = contextLCPO.getState(energy=True).getPotentialEnergy() - contextNone.getState(energy=True).getPotentialEnergy()
         self.assertAlmostEqual(energyLCPO.value_in_unit(kilocalorie_per_mole), energyRef.value_in_unit(kilocalorie_per_mole), 4)
 
--- a/wrappers/python/tests/systems/lcpo_test.prmtop
+++ b/wrappers/python/tests/systems/lcpo_test.prmtop
@@ -3772,17 +3772,17 @@ H   CX  H1  3C  H1  CT  HC  HC  HC  OH  HO  C   O   N   H   CX  H1  CT  HC  HC
 C*  CW  H4  NA  H   CN  CA  HA  CA  HA  CA  HA  CA  HA  CB  C   O   N   H   CX  
 H1  CT  HC  HC  CA  CA  HA  CA  HA  C   OH  HO  CA  HA  CA  HA  C   O   N   H   
 CT  H1  H1  H1  HO  OH  CI  H1  H1  CT  H1  OS  CT  H2  N*  CK  H5  NB  CB  CA  
-N2  H   H   NC  CQ  H5  NC  CB  CT  H1  CT  HC  HC  OS  P   O   O   OS  CI  H1  
+N2  H   H   NC  CQ  H5  NC  CB  CT  H1  CT  HC  HC  OS  P   O2  O2  OS  CI  H1  
 H1  CT  H1  OS  CT  H2  N*  CM  H4  CM  HA  CA  N2  H   H   NC  C   O   CT  H1  
-CT  HC  HC  OS  P   O   O   OS  CI  H1  H1  CT  H1  OS  CT  H2  N*  CK  H5  NB  
-CB  C   O   NA  H   CA  N2  H   H   NC  CB  CT  H1  CT  HC  HC  OS  P   O   O   
+CT  HC  HC  OS  P   O2  O2  OS  CI  H1  H1  CT  H1  OS  CT  H2  N*  CK  H5  NB  
+CB  C   O   NA  H   CA  N2  H   H   NC  CB  CT  H1  CT  HC  HC  OS  P   O2  O2  
 OS  CI  H1  H1  CT  H1  OS  CT  H2  N*  CM  H4  CM  CT  HC  HC  HC  C   O   NA  
 H   C   O   CT  H1  CT  HC  HC  OH  HO  HO  OH  CI  H1  H1  CT  H1  OS  CT  H2  
 N*  C4  H4  C4  HA  CA  N2  H   H   NC  C   O   CT  H1  CT  H1  OH  HO  OS  P   
-O   O   OS  CI  H1  H1  CT  H1  OS  CT  H2  N*  C5  H5  NB  CB  CA  N2  H   H   
-NC  CQ  H5  NC  CB  CT  H1  CT  H1  OH  HO  OS  P   O   O   OS  CI  H1  H1  CT  
+O2  O2  OS  CI  H1  H1  CT  H1  OS  CT  H2  N*  C5  H5  NB  CB  CA  N2  H   H   
+NC  CQ  H5  NC  CB  CT  H1  CT  H1  OH  HO  OS  P   O2  O2  OS  CI  H1  H1  CT  
 H1  OS  CT  H2  N*  CS  H4  CS  HA  C   O   NA  H   C   O   CT  H1  CT  H1  OH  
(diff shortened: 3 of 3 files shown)


docs-source/usersguide/application/02_running_sims.rst
```diff
@@ -983,6 +983,16 @@ Debye-Huckel screening parameter\ :cite:`Srinivasan1999`:
 
     system = prmtop.createSystem(implicitSolvent=OBC2, implicitSolventKappa=1.0/nanometer)
 
+By default, OpenMM uses the ACE approximation :cite:`Schaefer1998`\ :cite:`Ponder`
+to the solvent-accessible surface area, but you can also use the LCPO
+approximation :cite:`Weiser1999`:
+::
+
+    system = prmtop.createSystem(implicitSolvent=OBC2, gbsaModel='LCPO')
```
peastman: How about calling the argument `sasaMethod`?  The name `gbsaModel` suggests something much more general.  This option doesn't effect the GB part of the calculation, and also doesn't affect how energies depend on the SASA.  It's just the method used to  estimate the SASA.

docs-source/usersguide/theory/02_standard_forces.rst
```diff
@@ -462,6 +462,42 @@ where :math:`r_i` is the atomic radius of particle *i*\ , :math:`r_i` is
...
+   S_i=4\pi r_i^2
+
+where :math:`r_i` is the radius of atom :math:`i`, and
+
+.. math::
+   A_{ij}=2\pi r_i\left(r_i-\frac{d_{ij}}{2}-\frac{r_i^2-r_j^2}{2d_{ij}}\right)
+
+Note that the coefficients used in LCPOForce should be set to the product of the
+coefficients in the LCPO manuscript :cite:`Weiser1999` with the desired surface
+energy.
```
peastman: This sentence is a little confusing.  I'm not quite sure what it means.

openmmapi/include/openmm/LCPOForce.h
```diff
@@ -0,0 +1,144 @@
...
+    int getNumParticles() const;
+    /**
+     * Add parameters for a particle.  This should be called once for each particle in the System.
+     * When it is called for the i'th time, it specifies the parameters for the i'th particle.
+     *
+     * @param radius  the radius of the particle for the LCPO method, including the solvent probe radius if applicable
+     * @param p1      the product of the surface energy with the value of the LCPO parameter P1 for the particle
+     * @param p2      the product of the surface energy with the value of the LCPO parameter P2 for the particle
+     * @param p3      the product of the surface energy with the value of the LCPO parameter P3 for the particle
+     * @param p4      the product of the surface energy with the value of the LCPO parameter P4 for the particle
```
peastman: Would it make sense to pass the surface area scale as a separate parameter rather than asking the user to premultiply by it?  That better matches how it's described in the paper and specified in Amber.

docs-source/usersguide/application/02_running_sims.rst
epretti: I can change it if you think it is OK to do so; I agree that `sasaMethod` makes more sense but I didn't want to break the existing API so I kept the existing name.

openmmapi/include/openmm/LCPOForce.h
epretti: I can change this.  Should it be a per-particle parameter?  For Amber, the surface energy is a single parameter for the whole system, although there is an undocumented and possibly incomplete feature in the source code that lets you define different energies per amino acid.

epretti:
> Did I understand correctly, and each triplet gets processed six times?

That's correct; all of the permutations are processed for each triplet.  The calculation is asymmetric so that was the most natural way for me to write it.

> If so, is it necessary to do that (because the calculation is different each time)?  Or could you process each one only once and multiply the result by 6?

I'd have to do separate calculations for each permutation in the inner loops, but it's possible that some time can be saved building the neighbor list like this, and perhaps there would be an opportunity to reuse a few values between calculations for the different permutation.  I've already found that recomputing the distance between i and k to check if they are neighbors is faster than scanning the neighbor list.  I will test this out as well, because I think it would then be possible to do half as many of those distance checks in the inner loop.

openmmapi/include/openmm/LCPOForce.h
peastman: If it's usually specified as a single number, that's how I'd do it.  There's no loss of generality, since you can always absorb variation between particles into the other parameters.  It's just to make the interface match what people expect.

We also could make it a context parameter, allowing you to inexpensively change it.  I don't know whether there are cases where someone would want to do that.

epretti:
@peastman You can take a look at the common platform implementation now if you want.

I have been spending some time trying to optimize this.  I have a few test systems of various sizes, and the current performance I am getting is as follows (CUDA single precision, ns/day, median of 5 runs 10s each) compared to ACE and the slow CustomManyParticleForce implementation of LCPO for reference:

| System | Atoms | ACE | LCPO (custom) | *LCPO (this)* |
| --- | --- | --- | --- | --- |
| Chignolin | 138 | 1823 | 882 | 1235 |
| DHFR | 2489 | 1170 | 225 | 759 |
| Nucleosome | 25095 | 157 | 30 | 118 |

Right now it seems to be about 25-35% slower than just using ACE depending on system size (for the nonbonded options selected), but a lot faster than the CustomManyParticleForce implementation.  Let me know if you see any obvious opportunities for further optimization.  In case you want to profile it yourself, the benchmark script and inputs I have been using are here:

[benchmark.tar.gz](https://github.com/user-attachments/files/23781835/benchmark.tar.gz)

docs-source/usersguide/application/02_running_sims.rst
peastman: > Another option is to rename the AmberPrmtopFile argument to sasaMethod and make gbsaModel a deprecated alias for it.

That sounds like a good choice.

peastman:
Did you try ordering `condensedPos` in the other way (the x coordinates of all atoms, followed by the y coordinates of all atoms, etc.)?  It might lead to better coalescing of memory access in `findNeighbors` and `computeInteraction`.

peastman:
I wonder if there's a way to split up the work more evenly in `computeInteraction`?  Each thread takes an (i, j) pair and loops over all the neighbors of j.  For DHFR, the number of neighbors varies from 0 to 50, leading to some threads doing much more work than others.

epretti:
> Did you try ordering `condensedPos` in the other way (the x coordinates of all atoms, followed by the y coordinates of all atoms, etc.)? It might lead to better coalescing of memory access in `findNeighbors` and `computeInteraction`.

I had already tried that and (was a bit surprised to find) it didn't help.  But I think I did a few other optimizations since then, so I will double-check to see if it has any effect on the current version.

> I wonder if there's a way to split up the work more evenly in computeInteraction? Each thread takes an (i, j) pair and loops over all the neighbors of j. For DHFR, the number of neighbors varies from 0 to 50, leading to some threads doing much more work than others.

Maybe `computeNeighborStartIndices` could do a prefix sum on `numNeighbors * (numNeighbors - 1) / 2` as well as just `numNeighbors` so each thread could figure out where to start to evaluate a single `(i, j, k)` pair.  The test for inclusion of a triple would have to check if `j` and `k` are neighbors.  I don't know how well this would work and if it would be faster or not, but I can try it.

peastman:
Here is another possible optimization.  In the inner loop of `computeInteraction`, you first check the distance between i and k, which is fairly quick to do.  If it's beyond the cutoff, you skip the rest of the calculation, which is the much more expensive part.  On DHFR, less than 50% of k atoms pass the check, so you can usually skip the expensive calculation.

Except of course on a GPU, you can't.  If even one thread passes the check, every thread in the warp pays the cost of the expensive calculation.  Thread divergence leads to a lot of wasted calculation.

You might be able to improve the performance by restructuring the loop like this.

1. Split the loop into smaller blocks, maybe around 10 atoms.
2. For each block, first do the cutoff check on every atom.  For any atom that passes the check, store `k`, `kParams`, and `ikDelta` into a buffer.
3. Loop over the atoms you stored in the buffer and do the expensive calculation on them.

This will hopefully produce less thread divergence and better efficiency.

You can use shared memory for the buffer.  It's very fast to access and you don't use it for anything else.  The larger the blocks are, the less variation you'll get between threads in how many atoms are stored into the buffer.  But if you make them too large, you'll run out of shared memory and force it to reduce occupancy.  You can figure out a reasonable block size based on those considerations.

peastman:
You also could make `condensedPos` a `real4`.  Put the radius into the fourth element.  That way you only need one load for the whole thing, which really should be more efficient when you aren't getting good coalescing.  In `findNeighbors` you won't need to access `parameters` at all, since the only thing you use it for is the radius.  And in `computeInteraction` you'll only need to load it for the k atoms that pass the distance check.

You also could split up `parameters` into a pair of real2 arrays, one with `radius` and `p2`, and the other with `p3` and `p4`.  The reason is that in the inner loop of `computeInteraction`, you never use `p2`, and `radius` would already have been loaded from `condensedPos`.  So you wouldn't have to load as much data for each k atom.

epretti:
Here's where I'm at with this feature:

*Parameter assignment:*

I've added a heuristic approach for assigning the LCPO parameters from a Topology.  The two non-trivial cases are:

1. Determining whether an N with 3 bonds is sp3 (trigonal pyramidal) or sp2 (trigonal planar, either positively charged with a double bond and two single bonds, part of an aromatic ring, or most commonly, part of an amide).
2. Determining whether an O with 1 bond is part of an ordinary carbonyl group (and should have sp2 hybridization), or part of a carboxylate ion (such that it gets special LCPO parameters).

(1) is handled by checking the hybridization of C atoms bonded to an N atom in question.  I think this may fail for some cases like azides where N is bonded to itself.  However, the new test case I added covering all of the standard amino acids and nucleotides correctly assigns all of these N atoms.

(2) is handled by specifically identifying the topology of a carboxylate.  This correctly identifies all of the carboxylate O atoms, but it doesn't actually match Amber!  This is because Amber reuses the carboxylate `O2` atom type it checks for in this case for oxygens attached to phosphate groups, so it ends up assigning the carboxylate type to them.

I'm not sure what behavior we should implement here.  The LCPO paper doesn't mention doing this, such that I think assigning the carboxylate type to phosphate O atoms is wrong.  I looked at what NAMD does; NAMD reading Amber inputs emulates the Amber behavior, while when it reads CHARMM inputs, it will end up assigning an sp3 atom type to these O atoms (whether this is intended behavior is also unclear since it is the result of a default/fallthrough case).

*Trying to optimize further:*

> > Did you try ordering `condensedPos` in the other way (the x coordinates of all atoms, followed by the y coordinates of all atoms, etc.)? It might lead to better coalescing of memory access in `findNeighbors` and `computeInteraction`.
> 
> I had already tried that and (was a bit surprised to find) it didn't help. But I think I did a few other optimizations since then, so I will double-check to see if it has any effect on the current version.

Confirmed that this doesn't help the current version.

> You might be able to improve the performance by restructuring the loop like this.
> 
> 1. Split the loop into smaller blocks, maybe around 10 atoms.
> 
> 2. For each block, first do the cutoff check on every atom.  For any atom that passes the check, store `k`, `kParams`, and `ikDelta` into a buffer.
> 
> 3. Loop over the atoms you stored in the buffer and do the expensive calculation on them.

I already tried something like this before with the old neighbor checking approach I was using and it didn't help.  I've tested it again with the current cutoff-based approach.  I've now tried it with different neighbor block sizes, thread block sizes, and shared memory access patterns.  I couldn't find any combination of parameters that wasn't slower than not doing this.

> You also could make `condensedPos` a `real4`. Put the radius into the fourth element.

Inexplicably this slows down `findNeighbors` by 50% for me; I can't understand why!  This offsets any benefit to `computeInteraction` in the test cases I have.

peastman:
> Inexplicably this slows down findNeighbors by 50% for me; I can't understand why!

That's very strange.  Can you attach your modified version of lcpo.cc?  I'd like to try profiling it.

epretti:
Here's a patch.  The version I had stashed also has the suggested splitting of `parameters` into separate arrays, which I recall didn't appear to do much either way.

[condensedPosReal4.patch](https://github.com/user-attachments/files/23940546/condensedPosReal4.patch)

peastman:
I don't see the slowdown you describe, though I also don't see any significant benefit.  According to the compute profiler, the median duration of `findNeighbors` decreases from 39.9 to 39.0.  The median duration of `computeInteraction` decreases from 59.9 to 59.3.  Those differences are small enough that I'm not convinced they aren't just noise.

epretti:
That's odd; I wonder if it's something different with the hardware, or how the kernel is being compiled.  What version of the CUDA toolkit are you using?

peastman:
12.6 toolkit and 580.105.08 driver.

Try running it through the compute profiler.  With that big a slowdown, you ought to see something dramatically different.

For `findNeighbors`, the only really big change I see is "L1TEX Global Load Access Pattern" (under Memory Workload Analysis),  which increases from 10.6 to 31.7 bytes used.  Other than that, the differences are quite small.  Slight decreases in memory throughput and the number of instructions executed.  The warp state statistics barely change at all.

epretti:
The only thing that seems different that I suppose might be meaningful is that apparently there are more warp stalls due to "Stall Wait" and "Stall Short Scoreboard":

Screenshot 2025-12-04 at 14 34 57

(green = baseline, blue = patched)

Apparently according to the inline help, "Stall Short Scoreboard" can be due to "Warp was stalled waiting for a scoreboard dependency on a MIO (memory input/output) operation (not to L1TEX). The primary reason for a high number of stalls due to short scoreboards is typically memory operations to shared memory."  It is odd since the change between the two versions of the kernel shouldn't affect its use of shared memory, just how and what it reads from global memory.

I am using toolkit 13.0 and 580.105.08 driver.  Installing toolkit 12.6 and recompiling both versions linked against that version of the toolkit, I cannot reproduce the slowdown, and in fact the overall simulation speed for DHFR goes up about 1.8% with the `real4` `condensedPos` patch, whereas with 13.0, it goes down about 2.0%.

I can try to dig further into this, but based on these results, it seems at least plausible to me that there is an optimizer regression in some version of the toolkit > 12.6 and <= 13.0.

peastman:
I made a few attempts at optimizations, but none of them show signs of significant potential speedups.  I think we can merge this as it is.

PR #5137, opened 2025-11-10
Workaround for incorrect FFT results on Intel CPU OpenCL 2025.3

### LINKED ISSUE #5134: Intel CPU OpenCL constant potential tests are failing
I'm noticing some constant potential tests now failing in CI and it appears to be happening only on the Intel CPU OpenCL platform and (maybe?) only in mixed precision.  I'm looking into this now.

epretti:
Fixes #5134.

I didn't add a check to try to test for the buggy versions.  There is theoretically a way to distinguish between them by calling `clGetDeviceInfo` with `CL_DRIVER_VERSION`, but the result doesn't correspond directly with the installed version of Intel OpenCL:

- Release `2025.2.1-7` returns `2025.20.8.0.06_160000`
- Release `2025.3.0-639` returns `2025.20.10.0.10_160000`
- Release `2025.3.1-760` returns `2025.20.10.0.23_160000`

Even if we assume that version `2025.20.10.0.*_*` and up are bad, we'd have to parse this string, which would break if Intel ever changes the format of it in the future.


peastman (approved):
Looks good.

PR #5148, opened 2025-11-25
Add pythonforce.i to SWIG input files.

9527567:
Added pythonforce.i to SWIG input files for Python. The lack of this file resulted in a failure when using single core compilation, as `pythonforce.i` could not be found.


peastman:
Thanks!

PR #5151, opened 2025-11-26
Minor optimization to PME

peastman:
This avoids a few unnecessary atomic operations in charge spreading.  Running the standard benchmarks on RTX 4080, I get an overall speedup in the range of 1%.  Not a big deal, but still consistent improvement.

### DIFF
--- a/platforms/common/src/kernels/pme.cc
+++ b/platforms/common/src/kernels/pme.cc
@@ -102,7 +102,7 @@ KERNEL void gridSpreadCharge(GLOBAL const real4* RESTRICT posq,
                 ybase = ybase*GRID_SIZE_Z;
                 int index = xbase + ybase + zindex;
                 real add = dzdx*data[iy].y;
-                if (add > 2.3e-10f || add < -2.3e-10f) { // Smallest value representable in 64 bit fixed point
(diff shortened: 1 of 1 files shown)


platforms/common/src/kernels/pme.cc
```diff
@@ -102,17 +102,19 @@ KERNEL void gridSpreadCharge(GLOBAL const real4* RESTRICT posq,
                 ybase = ybase*GRID_SIZE_Z;
                 int index = xbase + ybase + zindex;
                 real add = dzdx*data[iy].y;
+                if (add > 2.3e-10f || add < -2.3e-10f) { // Smallest value representable in 64 bit fixed point
```
ex-rzr: I suggest to replace it with `fabs(add) > 2.3e-10f`. The reason is that absolute value can be free because (on AMD GPUs), many instructions can take abs or negative value of source registers. However, surprisingly clang already does this optimization: https://godbolt.org/z/75TrdYWbj

It turns out that Nvidia also can encode abs as a part of instruction, and the compiler doesn't see the opportunity for an optimization: https://godbolt.org/z/vh5oMW1e7

So on CUDA it saves 1 instructions (but it's likely that it is not noticeable in the overall performance)

PR #5153, opened 2025-11-30
Support opc3 water model in Modeller.addSolvent

jgullingsrud:


peastman:
If you specify `opc3` it will throw an exception.  It will expect there to be a file called opc3.pdb containing a pre-equilibrated box of water.

Of course, it really doesn't matter.  All water models are similar enough that once you run an energy minimization and a short equilibration, it makes no difference which water model was used to build the original box.  The only thing the `model` argument really matters for is four and five site water models, so it can add the correct set of extra particles.  Even then it's just saving you a single call to `addExtraParticles()`.

I've been vaguely meaning to update the documentation to discuss that.

jgullingsrud:
No problem, feel free to close.

peastman:
I took a look at the existing documentation for `addSolvent()`, and I think it does a pretty good job of describing the situation.  Here's what the API documentation says:

> There exist many different water models, many of which are very similar to each other. This method creates preequilibrated water boxes for a limited set of water models. In most cases, they work equally well for other models that involve the same number of particles. For example, to simulate a box of TIP3P-FB water, use this method to create a box of TIP3P water, construct a System using TIP3P-FB parameters, and perform a local energy minimization to correct for the small differences between the models. Likewise, a box of TIP4P-Ew water can be used for most four site water models.

Do you think that's sufficiently clear, or should I change anything about it?

The discussion in the user guide doesn't say anything about this, so I should add similar text there.

PR #5172, opened 2025-12-28
Fix Python SyntaxErrors in processAmberForceField.py

cclauss:
`print()` is a function in Python 3.


peastman:
That's an obsolete file that hasn't been used in years.  It was used to convert the early Amber force fields.  More recent force fields are converted with the infrastructure at https://github.com/openmm/openmmforcefields.  We keep it around just to preserve a record of how the earlier force fields were converted.

PR #5174, opened 2025-12-28
ruff-rules-for-comprehensions

cclauss:
% `ruff check --select=C4,PERF --statistics ` # https://docs.astral.sh/ruff
* https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

### DIFF
--- a/wrappers/python/openmm/app/internal/amber_file_parser.py
+++ b/wrappers/python/openmm/app/internal/amber_file_parser.py
@@ -558,7 +558,7 @@ class PrmtopLoader(object):
         self._cmapList=[]
         forceConstConversionFactor = (units.kilocalorie_per_mole).conversion_factor_to(units.kilojoule_per_mole)
         for ii in range(0,len(cmapPointers),6):
-            if any([int(cmapPointers[ii+jj])<0 for jj in range(5)]):
+            if any(int(cmapPointers[ii+jj])<0 for jj in range(5)):
                 raise ValueError("Found negative cmap atom pointers %s"
                                 % ((cmapPointers[ii],
                                    cmapPointers[ii+1],
@@ -1062,7 +1062,7 @@ def readAmberSystem(topology, prmtop_filename=None, prmtop_loader=None, shake=No
     epNames = ['EP', 'LP']
     ep = [i for i in range(prmtop.getNumAtoms()) if isWater[i] and prmtop.getAtomName(i)[:2] in epNames]
     if len(ep) > 0:
-        epRes = set((prmtop.getResidueNumber(i) for i in ep))
+        epRes = {prmtop.getResidueNumber(i) for i in ep}
         numRes = max(epRes)+1
         # For each residue that contains an "extra point", find the oxygen, hydrogens, and points.
         waterO = []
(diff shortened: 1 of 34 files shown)


peastman:
Most of these changes look arbitrary.  For example, `{}` is no better or worse than `dict()`.  This seems to be an automated style checker trying to enforce a particular set of coding conventions that we don't follow.

I also prefer not to change code if there's no good reason, since it complicates the version history and can make it harder to track down when a particular change was made.

cclauss:
This PR is about performance, not style.
> `{}` is no better or worse than `dict()`.

The former has no functional call overhead while the latter does.

Ask ChatGPT:
> What is Python function call overhead?

tl;dr: ~80–150 nanoseconds for every call.

peastman:
That seems like a pretty extreme case of premature optimization.  Did you identify that any of these changes was in a performance critical routine, or measure that it produced a meaningful benefit?

cclauss:
I and others have measured the performance differences in microbenchmarks and by examining the generated Python bytecodes.  See the `Why is this bad?` section of the command:
% [`ruff rule C408`](https://docs.astral.sh/ruff/rules/unnecessary-collection-call)

---
% `python3.14 -m timeit "dict()"`
```
5000000 loops, best of 5: 52.9 nsec per loop
```
% `python3.14 -m timeit "{}"`
```
10000000 loops, best of 5: 27.9 nsec per loop
```
---
% `python3.14`
```
>>> def func_0():
...     d = dict()
...
>>> def func_1():
...     d = {}
...
>>> from dis import dis
>>> dis(func_0)
  1           RESUME                   0

  2           LOAD_GLOBAL              1 (dict + NULL)
              CALL                     0
              STORE_FAST               0 (d)
              LOAD_CONST               0 (None)
              RETURN_VALUE
>>> dis(func_1)
  1           RESUME                   0

  2           BUILD_MAP                0
              STORE_FAST               0 (d)
              LOAD_CONST               0 (None)
              RETURN_VALUE
>>> from timeit import timeit
>>> timeit(func_0)
0.09176879108417779
>>> timeit(func_1)
0.06369720806833357
```

wrappers/python/openmm/app/internal/amber_file_parser.py
```diff
@@ -558,7 +558,7 @@ def getCMAPDihedrals(self):
         self._cmapList=[]
         forceConstConversionFactor = (units.kilocalorie_per_mole).conversion_factor_to(units.kilojoule_per_mole)
         for ii in range(0,len(cmapPointers),6):
-            if any([int(cmapPointers[ii+jj])<0 for jj in range(5)]):
+            if any(int(cmapPointers[ii+jj])<0 for jj in range(5)):
```
cclauss: % `python3.14 -m timeit "any([x for x in range(1000)])"`
```
10000 loops, best of 5: 24.1 usec per loop
```
% `python3.14 -m timeit "any(x for x in range(1000))"`
```
1000000 loops, best of 5: 256 nsec per loop
```
% `python3.14 -m timeit "all([x for x in range(1000)])"`
```
10000 loops, best of 5: 24.2 usec per loop
```
% `python3.14 -m timeit "all(x for x in range(1000))"`
```
1000000 loops, best of 5: 237 nsec per loop
```

peastman:
Microbenchmarks are not relevant to real world performance.  If a function accounts for 0.1% of the time for your real world operation, and your optimization speeds up that function by 0.1%, the speedup to the real world operation is 0.0001%.

You highlighted a line in `getCMAPDihedrals()`.  If you profile loading an Amber file, you'll find that function takes a completely negligible fraction of the time, so any speedup to it has no meaningful benefit.  If you then profile just that function, you'll find the speedup is quite small, saving only a fraction of an already insignificant amount of time.

PR #5175, opened 2025-12-30
Rename file that contains Python 2 code

cclauss:
> devtools/forcefield-scripts/processAmberForceField.py -->
> devtools/forcefield-scripts/processAmberForceField.py2

This helps linters and humans understand that this code is for Python 2, which died 6 years ago on January 1, 2020.
* #5172


peastman:
I decided not to merge this, since it's another PR that doesn't fix any real problem.  We don't use any linter that gets confused, and if we did we certainly wouldn't run it on the scripts in that directory, which are all obsolete.  And by updating the modification date it would make that script look like it was still actively developed, when in fact it hasn't been modified in years.

Issue #5157, opened 2025-12-04
device querying


chrisiacovella:
This is somewhat related to this issue https://github.com/openmm/openmm/issues/52 

Basically, it would be nice if openMM had some functionality built-in that would allow us to "check" a node before we actually try to start a simulation.   While this of course could be accomplished by writing some simple code (as in the other issue, or querying nvidia-smi for CUDA devices), as was noted in the last meeting, these functions already exist in OpenMM and could just be easily wrapped into some utilities.   Part of my thought here is that I want to be able to better control how a script fails if we can't find the expected device so that return value from a failed run could be used to direct a future run (e.g., to aggregate a list of nodes where the GPU can't be addressed so I can avoid spawning jobs on those nodes that are just going to fail). 

I think there could be two pieces to this:

First, would be create a simple wrapper that returns a list of the available platforms, e.g., `Platform.getAvailablePlatforms()`, rather than having to loop through each platform and then get the name. 
```
[Platform.getPlatform(i).getName() for i in range (Platform.getNumPlatforms())]
```
Openmmtools utils(https://github.com/choderalab/openmmtools/blob/7b6e86cd6431a5c4f459d44b904d9542578e161a/openmmtools/utils/utils.py#L561)  has a bunch of functions that already wrap these functions, that look at the precision which could be useful to have that more easily accessed directly via OpenMM. 

Secondarily, and more to the heart of the issue, would be good to have a function that would return the device(s) for a given platform (or all available platforms found)   I might be going about it the wrong way, but the way I've done this in the is to just create a dummy system to get this info quickly, putting this is a try/except to catch if there are weird cases where it finds the platform but not the device.  
```
system = System()
system.addParticle(1.0)
properties = {'DeviceIndex': '0'}
platform = Platform.getPlatform('CUDA')
context = Context(system, integrator, platform, properties)
device_name = context.getPlatform().getPropertyValue(context, "DeviceName")
```    
So it could be useful to have a  function that would return the device associated with a platform name, (or None if not found), without the user having to setup the real system or a dummy system.

peastman:
What about a `getDevices()` method that returns a list of dicts, one for each available device?  Each dict would contain the values of the platform-specific properties for that device.  For the CUDA platform, it would include "DeviceIndex" and "DeviceName".  For OpenCL, it would also include "OpenCLPlatformIndex" and "OpenCLPlatformName".  For Reference or CPU it would return an empty list, since they don't have selectable devices.

[...]

PR #5192, opened 2026-01-26
API for querying devices

peastman:
Implements #5157.  This adds a `getDevices()` method to Platform.  It returns a list of maps, one for each available device.  They contain the values for platform-specific properties.  You can pass them directly as the `properties` argument when creating a Context to select the corresponding device.

There's an optional `filters` argument which takes property values.  It will only return devices that are compatible with those values.  For example, you might pass `{'Precision':'mixed'}` to only get devices compatible with mixed precision, or `{'OpenCLPlatformIndex':'1'}` to specify a particular OpenCL implementation.

cc @chrisiacovella

### DIFF
--- a/platforms/cuda/src/CudaPlatform.cpp
+++ b/platforms/cuda/src/CudaPlatform.cpp
@@ -164,6 +164,15 @@ void CudaPlatform::setPropertyValue(Context& context, const string& property, co
 }
 
 vector<map<string, string> > CudaPlatform::getDevices(const map<string, string>& filters) const {
+    try {
+        CudaContext::ensureCudaInitialized();
+    }
+    catch (...) {
+        // CUDA couldn't be initialized, so report no devices.
+
+        return {};
+    }
+
     // Check for properties that might act as filters.
 
     int deviceIndex = -1;
@@ -173,7 +182,6 @@ vector<map<string, string> > CudaPlatform::getDevices(const map<string, string>&
 
     // Loop over devices.
 
-    CudaContext::ensureCudaInitialized();
     vector<map<string, string> > results;
     int numDevices;
     if (cuDeviceGetCount(&numDevices) != CUDA_SUCCESS)


epretti:
I just tried this out on my system.  Running
```python
import openmm
platform = openmm.Platform.getPlatformByName("CUDA")
print(len(platform.getDevices()))
system = openmm.System()
system.addParticle(1)
openmm.Context(system, openmm.VerletIntegrator(1), platform)
print(len(platform.getDevices()))
```
prints
```
0
1
```
*i.e.*, if a `Context` using the CUDA platform hasn't been created yet, no devices will be returned.  My guess is maybe `cuDeviceGetCount` fails if `cuInit` hasn't been called by `CudaContext::CudaContext` yet?  (OpenCL doesn't have this problem.)

epretti:
Also, a thought about the API: would it make sense for the Reference and CPU platforms to return a list with a single empty dictionary inside (instead of an empty list)?  I'm just imagining writing, *e.g.*, a benchmarking script that wants to iterate over all platforms and all devices for each platform.  For the Reference and CPU platforms, there would be exactly one possibility for what to pass as the `properties` argument when creating a Context, which would be an empty dictionary representing the only possible "device".  I can also see the argument for returning an empty list for these platforms since the idea of separate selectable devices isn't meaningful for them.  So feel free to ignore this idea if you don't like the sound of it.

peastman:
I can see arguments either way for the Reference and CPU platforms.  My original idea was that an empty list would indicate the platform didn't use selectable devices.  But it also can indicate the platform does use selectable devices, but none are available.  So perhaps a single empty dict is a better way to signal it.

peastman:
@chrisiacovella can you review this and make sure it meets your requirements?

chrisiacovella (approved):
This looks great!  I think it will be very useful.

peastman:
Thanks!

PR #5197, opened 2026-01-30
Residue templates can specify constraints

peastman:
This is another change to support future OpenFF force fields.  A template can now include `<Constraint>` tags which specify constraints explicitly required by the force field.

cc @j-wags @IAlibay

### DIFF
--- a/wrappers/python/openmm/app/forcefield.py
+++ b/wrappers/python/openmm/app/forcefield.py
@@ -928,6 +928,14 @@ class ForceField(object):
                 for atom in self.addedExternalBonds:
                     newTemplate.addExternalBondByName(atom.name)
 
+                # Build the list of constraints.
+
+                for atom1, atom2, distance in template.constraints:
+                    a1 = template.atoms[atom1]
+                    a2 = template.atoms[atom2]
+                    if a1 in atomMap and a2 in atomMap:
+                        newTemplate.addConstraint(atomMap[a1], atomMap[a2], distance)
+
                 # Add new virtual sites.
 
                 indexMap = dict((i, newAtomIndex[atom.name]) for i, atom in enumerate(self.addedAtoms[index]+self.changedAtoms[index]))


IAlibay (commented):
Just the one question from me re: HMR, otherwise I only had a quick look through but this looks good to me.

peastman:
The special case handling for HMR is specific to water.  If you've told it to make water rigid and a residue is named `HOH`, it omits HMR.  In all other cases it does not.

IAlibay:
> The special case handling for HMR is specific to water.  If you've told it to make water rigid and a residue is named `HOH`, it omits HMR.  In all other cases it does not.

Thanks for the clarification. Am I correct in understanding that now, with the OpenFF water models, this could cause a case where you could have HMR + constraints on a water and that this is a case that wouldn't happen in OpenMM normally? 

If so, my question is purely "could this be a problem in terms of system stability"? Or better put, is the fact that OpenMM doesn't apply HMR to constrained waters purely because that's what Hopkins et al. did or if there's a deeper stability issue that needs it that way?

peastman:
By default, any residue named `HOH` will be rigid and will not get HMR.  The user can override it by specifying `rigidWater=False` when calling `createSystem()`, and the force field can also override it by adding `rigidWater="false"` as an attribute of the template.

The only problem with HMR on rigid water is that it increases the moment of inertia, which slows down rotational motions.  That's true of all uses of HMR.  We treat rigid water as a special case because we know it's unnecessary in that case, and water makes up the majority of most systems.

j-wags (approved):
This looks good to me. It might be worth documenting/testing what happens if both a bond parameter and a constraint (each with different distances) are defined for the same bond. I'd imagine that either an error is raised or the constraint distance would take precedence, but it could be good to make sure.

peastman:
The constraint takes precedence.  This is tested in `testTemplateConstraints()`.  For the OG-HG bond the constraint specifies 1.123 and the bond specifies 1.1.  It checks that the correct value was used.

PR #5198, opened 2026-01-31
GPU implementation of L-BFGS

### LINKED ISSUE #1237: GPU LBFGS Minimizer

epretti:
Resolves #1237.  I'd like to do some more testing of this on various systems and machines, and profiling on various system sizes.  But it should at least be ready to look at.

### DIFF
--- a/platforms/common/src/CommonMinimizeKernel.cpp
+++ b/platforms/common/src/CommonMinimizeKernel.cpp
@@ -565,7 +565,7 @@ void CommonMinimizeKernel::evaluateGpu(ContextImpl& context) {
     }
 
     // Convert the forces from fixed to floating point format.  If they are too
-    // large, the energy in returnValue will be set to NaN to signal that the
+    // large, the energy in returnValue will be set to FLT_MAX to signal that the
     // results are invalid and we must fall back to computing forces on the CPU.
 
     convertForcesKernel->execute(numParticles);


platforms/common/src/CommonMinimizeKernel.cpp
```diff
@@ -0,0 +1,650 @@
...
+
+    map<string, string> defines;
+    defines["THREAD_BLOCK_SIZE"] = cc.intToString(threadBlockSize);
+    defines["NUM_CONSTRAINT_BLOCKS"] = cc.intToString(numConstraintBlocks);
+    defines["NUM_CONSTRAINTS"] = cc.intToString(numConstraints);
+    defines["NUM_PADDED"] = cc.intToString(cc.getPaddedNumAtoms());
+    defines["NUM_PARTICLES"] = cc.intToString(numParticles);
+    defines["NUM_VARIABLE_BLOCKS"] = cc.intToString(numVariableBlocks);
+    defines["NUM_VARIABLES"] = cc.intToString(numVariables);
+    defines["NUM_VECTORS"] = cc.intToString(numVectors);
```
peastman: Could these be made into kernel arguments rather than compile time constants?  Making them constants is unlikely to have any significant speed benefit for these kernels.  And then the cached kernels wouldn't be specific to a particular system, so they wouldn't need to be recompiled for every new system.

platforms/common/src/CommonMinimizeKernel.cpp
```diff
@@ -0,0 +1,754 @@
...
+
+    if (!isSetup) {
+        // Load system and integrator information.
+
+        mixedIsDouble = cc.getUseMixedPrecision() || cc.getUseDoublePrecision();
+        elementSize = mixedIsDouble ? sizeof(double) : sizeof(float);
+        threadBlockSize = cc.getMaxThreadBlockSize();
+        numVariableBlocks = (numVariables + threadBlockSize - 1) / threadBlockSize;
+        numConstraintBlocks = (numConstraints + threadBlockSize - 1) / threadBlockSize;
+        pinnedMemory = cc.getPinnedBuffer();
```
peastman: Make sure this is safe.  It's a shared buffer that should only be used briefly while you're sure nothing else will try to use it too.  For example, calling `getPositions()` or `setPositions()` will overwrite it.

epretti: In between the asynchronous download into this buffer being enqueued and waited on, only minimizer-related kernels should be running.

platforms/cuda/src/kernels/common.cu
```diff
@@ -18,6 +18,7 @@
 #define SYNC_THREADS __syncthreads();
 #define MEM_FENCE __threadfence_block();
 #define ATOMIC_ADD(dest, value) atomicAdd(dest, value)
+#define NAN nanf("")
```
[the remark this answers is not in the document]
epretti: OK, I think I have something that works.  It sets `FLT_MAX` to mark overflow, and uses `!(fabs(energy) < FLT_MAX)` to check for it; I think this should avoid introducing NaNs, but also catch any NaNs and infinities that do show up.  As for the fallback paths for computing dot products, I just run them automatically if evaluating on the CPU, since that's the only case when the gradients could be large enough that it matters.

epretti:
Some benchmarks I ran to get a sense of the performance of this (CUDA, mixed precision, bonds to hydrogen constrained) are as follows.  Times are median of 7 runs.  Coordinates are perturbed by ± 0.5 Å from a minimized starting configuration to generate the initial conditions.  (The optimizer works fine when perturbing by more than this, but then both algorithms just spend a lot of their time evaluating large forces on the CPU.  As is, initial forces are generally betweeen $10^6$ and $10^8$ kJ/mol/nm.)

For water boxes of various sizes:

| Atoms  | Old (s) | New (s) |
| ------ | --- | --- |
| 2685   | 0.30 | 0.19 |
| 21480  | 1.42 | 0.38 |
| 72495  | 5.48 | 1.12 |
| 171840 | 14.97 | 1.98 |
| 335625 | 31.90 | 4.91 |
| 579960 | 63.16 | 8.23 |
| 920955 | 126.44 | 18.42 |

And, for tiling a box with solvated DHFR:

| Atoms  | Old (s) | New (s) |
| ------ | --- | --- |
| 23558  | 1.61 | 0.40 |
| 188464 | 16.69 | 3.80 |
| 636066 | 104.79 | 23.58 |

peastman:
Excellent!  That will be a really nice speedup.  Let me do one more review, and then it should be ready to merge soon.

epretti:
I'm seeing new failures on the CUDA CI, and something's now wrong with single precision OpenCL on my Mac :(  I need to figure out what's causing those.

epretti:
The test that's failing on Mac for me is actually failing on the master branch in the same place for what appears to be the same reason.  It's `testLargeForces`, and it seems like when the forces are overflowing the fixed-point buffers, they're ending up equaling 0, so the gradient comes to 0 and the minimizer stops, thinking it has succeeded.  On OpenCL and CUDA on my machine with the NVIDIA card, overflow seems to cause forces to go to $\pm 2^{31}$, but they're getting zeroed on the Mac.  I'm not entirely sure why this difference is happening and might want to investigate further, but it seems like it's not due to new changes here.

peastman:
The GPUs must be handling overflow differently.  I can't find anything in the OpenCL spec about how it should be handled, so it's likely implementation dependent.  The CUDA programming guide is also mostly silent about it, but I did find [one brief mention](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=atomicadd#floating-point-standard): "The behavior of integer division by zero and integer overflow is left undefined by IEEE-754."  That's all it says about it.

I also found the following in the [Wikipedia article on integer overflow](https://en.wikipedia.org/wiki/Integer_overflow):

> Most commonly, the resulting bit-pattern is the same as if the operation was performed modulo $2^W$, where W is the word size in bits. The operation also sets or unsets one or more flags that indicate whether overflow has occurred. On some processors like [graphics processing units](https://en.wikipedia.org/wiki/Graphics_processing_unit) (GPUs) and [digital signal processors](https://en.wikipedia.org/wiki/Digital_signal_processor) (DSPs) which support [saturation arithmetic](https://en.wikipedia.org/wiki/Saturation_arithmetic), overflowed results may be clamped, i.e. set to the minimum value in the representable range if the result is below the minimum and set to the maximum value in the representable range if the result is above the maximum.

My guess is that Nvidia GPUs saturate and Apple GPUs wrap.  We should look into whether we can improve this behavior in the future, but since it's not a regression, it shouldn't block this PR.

The Wikipedia article also notes that the C spec treats signed and unsigned ints differently.  Unsigned ints are required to wrap, but the behavior of signed ints is undefined.  But there's no guarantee that GPUs follow that.

peastman:
The GPUs must be handling overflow differently.  I can't find anything in the OpenCL spec about how it should be handled, so it's likely implementation dependent.  The CUDA programming guide is also mostly silent about it, but I did find [one brief mention](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=atomicadd#floating-point-standard): "The behavior of integer division by zero and integer overflow is left undefined by IEEE-754."  That's all it says about it.

I also found the following in the [Wikipedia article on integer overflow](https://en.wikipedia.org/wiki/Integer_overflow):

> Most commonly, the resulting bit-pattern is the same as if the operation was performed modulo $2^W$, where W is the word size in bits. The operation also sets or unsets one or more flags that indicate whether overflow has occurred. On some processors like [graphics processing units](https://en.wikipedia.org/wiki/Graphics_processing_unit) (GPUs) and [digital signal processors](https://en.wikipedia.org/wiki/Digital_signal_processor) (DSPs) which support [saturation arithmetic](https://en.wikipedia.org/wiki/Saturation_arithmetic), overflowed results may be clamped, i.e. set to the minimum value in the representable range if the result is below the minimum and set to the maximum value in the representable range if the result is above the maximum.

My guess is that Nvidia GPUs saturate and Apple GPUs wrap.  We should look into whether we can improve this behavior in the future, but since it's not a regression, it shouldn't block this PR.

The Wikipedia article also notes that the C spec treats signed and unsigned ints differently.  Unsigned ints are required to wrap, but the behavior of signed ints is undefined.  But there's no guarantee that GPUs follow that.

peastman:
The GPUs must be handling overflow differently.  I can't find anything in the OpenCL spec about how it should be handled, so it's likely implementation dependent.  The CUDA programming guide is also mostly silent about it, but I did find [one brief mention](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=atomicadd#floating-point-standard): "The behavior of integer division by zero and integer overflow is left undefined by IEEE-754."  That's all it says about it.

I also found the following in the [Wikipedia article on integer overflow](https://en.wikipedia.org/wiki/Integer_overflow):

> Most commonly, the resulting bit-pattern is the same as if the operation was performed modulo $2^W$, where W is the word size in bits. The operation also sets or unsets one or more flags that indicate whether overflow has occurred. On some processors like [graphics processing units](https://en.wikipedia.org/wiki/Graphics_processing_unit) (GPUs) and [digital signal processors](https://en.wikipedia.org/wiki/Digital_signal_processor) (DSPs) which support [saturation arithmetic](https://en.wikipedia.org/wiki/Saturation_arithmetic), overflowed results may be clamped, i.e. set to the minimum value in the representable range if the result is below the minimum and set to the maximum value in the representable range if the result is above the maximum.

My guess is that Nvidia GPUs saturate and Apple GPUs wrap.  We should look into whether we can improve this behavior in the future, but since it's not a regression, it shouldn't block this PR.

The Wikipedia article also notes that the C spec treats signed and unsigned ints differently.  Unsigned ints are required to wrap, but the behavior of signed ints is undefined.  But there's no guarantee that GPUs follow that.

epretti:
I did some simple testing of `float` to `long` conversion, and it appears that indeed my NVIDIA GPU clamps, with positive overflow giving `0x7fffffffffffffff` and negative overflow or NaN giving `0x8000000000000000`.  On the other hand, my Apple appears to wrap, so for large enough floating-point values, the significand gets shifted off to the left and leaves only 0.  (Infinities and NaN also turn into 0.)  For Intel OpenCL, overflow is always giving `0x8000000000000000` (although this may be the underlying behavior of the AMD CPU).

In any case, this would explain the observed behavior, and I agree that this is not a regression.  I am as happy as I can be with this PR until we can get more people to try the new implementation, so I think it can be merged once you are OK with it.

platforms/common/src/CommonMinimizeKernel.cpp
```diff
@@ -0,0 +1,757 @@
...
+            getConstraintEnergyForcesKernel->setArg(8, kRestraint);
+        }
+        else {
+            getConstraintEnergyForcesKernel->setArg(8, (float) kRestraint);
+        }
+        getConstraintEnergyForcesKernel->execute(numConstraints);
+    }
+
+    // Convert the forces from fixed to floating point format.  If they are too
+    // large, the energy in returnValue will be set to NaN to signal that the
```
peastman: FLT_MAX not NaN I think?

philipturner:
Yes! This is a big W. Have yet to try it out, it should be faster.

PR #5202, opened 2026-02-04
Add keepSegids parameter for keeping segment (chain) ids in PDB file

GorAleks:
Previously, PDB files saved with "keepIDs=True" omitted segids, complicating PSF generation and VMD manipulation of PSF-based structures. This change ensures segids are correctly written, enabling PSF creation from PDB files.

### DIFF
--- a/wrappers/python/openmm/app/pdbfile.py
+++ b/wrappers/python/openmm/app/pdbfile.py
@@ -266,7 +266,7 @@ class PDBFile(object):
                 map[atom.attrib[id]] = name
 
     @staticmethod
-    def writeFile(topology, positions, file=sys.stdout, keepIds=False, extraParticleIdentifier='EP'):
+    def writeFile(topology, positions, file=sys.stdout, keepIds=False, extraParticleIdentifier='EP', keepSegids=False):
         """Write a PDB file containing a single model.
 
         Parameters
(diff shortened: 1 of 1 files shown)


peastman:
Thanks.  Can you explain what you mean by segids?  According to [the PDB spec](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM), those columns are unused and should be left blank.  And `chain.id` already gets stored in column 22, which is where the chain identifier is supposed to go.

If your goal is to store chain IDs that are more than one character, I recommend saving to PDBx/mmCIF format instead.  The PDB format is deprecated.

GorAleks:
Segids aren't strictly required by the PDB spec, but they're essential for PSF generation and VMD workflows. Without them, VMD's psfgen assigns everything to a generic "SYS" segment, which breaks multi-component selection, modifications, and system partitioning.
chainID (column 22) and segid (73-76) serve different purposes in VMD. You'll always see both in CHARMM-GUI or VMD-generated PDBs. Psfgen reads segids from column 73-76 specifically.

peastman:
What is a segid?  Can you point to the documentation where they're described?

GorAleks:
here (segname ~ segid)
https://www.ks.uiuc.edu/Research/vmd/plugins/doxygen/readpdb_8h.html#a19a9:~:text=%C2%A0%20chain%2C-,const%20char%20*%C2%A0,%C2%A0%20segname,-)%C2%A0

and here (segid properly)

https://www.ks.uiuc.edu/Research/vmd/plugins/doxygen/readpdb_8h-source.html#l00352:~:text=00263%20/*%20ATOM%20field,be%20null%2Dterminated.

peastman:
Those comments are incorrect.  The actual specification is at https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM.  Columns 73-76 are supposed to be left blank, and there is no such field as "segID".  Someone made up their own nonstandard change to the format.

epretti:
From what I recall when working with CHARMM files, this is a nonstandard feature used by CHARMM.  From the point of view of adherence to the official PDB specification, it should neither generate nor process such malformed PDBs, but unfortunately, it does, and so CHARMM and related tools end up handling PDBs with these fields.

Issue #5203, opened 2026-02-05
uv cannot install openmm[cuda12]


stefdoerr:
I was trying to use openmm in `uv` and I was getting this warning:

```
uv sync

Using CPython 3.11.13
Creating virtual environment at: .venv
warning: The package `openmm==8.4.0.post2` does not have an extra named `cuda12`
```
Indeed when I try to run I see that OpenMM is installed without the extra cuda12 libs.

I was scratching my head because when I looked at the METADATA file in the package it looks like this
```
Requires-Dist: numpy
Provides-Extra: cuda12
Requires-Dist: OpenMM-CUDA-12==8.4.0.post2; extra == "cuda12"
Provides-Extra: cuda13
Requires-Dist: OpenMM-CUDA-13==8.4.0.post2; extra == "cuda13"
Provides-Extra: hip6
Requires-Dist: OpenMM-HIP-6==8.4.0.post2; extra == "hip6"
Provides-Extra: hip7
Requires-Dist: OpenMM-HIP-7==8.4.0.post2; extra == "hip7"
Dynamic: author
Dynamic: description
Dynamic: download-url
Dynamic: home-page
Dynamic: license
Dynamic: platform
Dynamic: provides-extra
Dynamic: requires-dist
Dynamic: summary
```
which indeed defines the extras. 

But according to Gemini:
> Under PEP 643, if a field is marked as Dynamic, it means the static metadata (the METADATA file in the .dist-info directory) is not considered the "source of truth" for that field.
> Instead, the installer is told: "Don't trust what's written here; you have to actually run the build backend (like setuptools or flit) to determine what the extras/dependencies really are."

So uv is stricter than pip in applying the PEP and it fails. Could the extras possibly be defined without "dynamic"?

peastman:
I don't understand what it's saying.  Here is [the documentation](https://packaging.python.org/en/latest/specifications/core-metadata/#dynamic-multiple-use) on dynamic fields.  It only applies to sdist packages, not bdist ones.  It's for fields whose value gets set at build time.

Also, I think the whole concept of dynamic fields is only relevant to packages defined by a pyproject.toml file.    OpenMM uses a setup.py file, which is inherently dynamic.

stefdoerr:
https://packaging.python.org/en/latest/specifications/core-metadata/#dynamic-multiple-use
Maybe this is more to the point

[...]

PR #5207, opened 2026-02-10
Specify extras_require on all platforms

peastman:
This is an attempt at fixing #5203.


peastman:
I'll merge this and then try building wheels to see whether it works on all platforms.

Issue #5201, opened 2026-02-04
Patches not applied as expected - Charge not updated


m-julian:
I need to modify the charges on certain residues and have been experimenting around with doing that  through patches, however patches never get applied because existing templates already match to the residues. I would expect that if a patch is loaded in a force field, it should always be applied where matches occur. I defined a test force field for a hydrogen molecule below:

Hydrogen ffxml:

```
<?xml version="1.0"?>
<ForceField>
  <AtomTypes>
    <Type name="H2H" class="H6" element="H" mass="1.008"/>
  </AtomTypes>
  <Residues>
    <Residue name="HH">
      <Atom name="H1" type="H2H" charge="0.0"/>
      <Atom name="H2" type="H2H" charge="0.0"/>
      <Bond atomName1="H1" atomName2="H2"/>
    </Residue>
  </Residues>
  <HarmonicBondForce>
    <Bond class1="H6" class2="H6" length="0.074" k="450000"/>
  </HarmonicBondForce>
  <NonbondedForce coulomb14scale="1.0" lj14scale="1.0">
    <UseAttributeFromResidue name="charge" />
    <Atom type="H2H" sigma="0.296" epsilon="0.065"/>
  </NonbondedForce>
</ForceField>
```

Hydrogen patch:

```
<?xml version='1.0' encoding='utf-8'?>
<ForceField>
  <Patches>
   <Patch name="TEST">
   <ApplyToResidue name="HH"/>
   <ChangeAtom name="H1" type="H2H" charge="3.5" />
   </Patch>
  </Patches>
</ForceField>
```

Hydrogen PDB:
```
ATOM      1  H1  HH  A   1       0.000   0.000   0.000  1.00  0.00           H
ATOM      2  H2  HH  A   1       0.740   0.000   0.000  1.00  0.00           H
```

Python code:

```
from openmm.app import ForceField, Modeller, PDBFile, NoCutoff, Simulation, Topology
from openmm import Platform, LangevinIntegrator, System, NonbondedForce
from openmm import unit
from typing import Iterable

def print_atom_class_assignments(
    topology: Topology,
    force_field: ForceField,
    resnames_to_ignore: Iterable[str] | None = None,
):
    if resnames_to_ignore:
        resnames_to_ignore = set(resnames_to_ignore)
    else:
        resnames_to_ignore = {}

    data = ForceField._SystemData(topology)
    # this sets attributes for system data, so needs to get called
[...]

PR #5208, opened 2026-02-10
Update patch documentation

m-julian:
Clarified the definition of patches to make it clear that patches are used to create new residue templates, which match different residues. Patches do not modify existing residue templates. Related to discussion in https://github.com/openmm/openmm/issues/5201

### DIFF
--- a/docs-source/usersguide/application/06_creating_ffs.rst
+++ b/docs-source/usersguide/application/06_creating_ffs.rst
@@ -203,7 +203,7 @@ of the site position in the local coordinate system), and :code:`wo1`\ ,
 =========
 
 A "patch" is a set of rules for creating new residue templates from existing ones.
-These new templates should match a different set of residues contained in the topology,
+These new templates should match a different set of residues contained in the topology;
 therefore a patch must specify a change in atoms and bonds.
 For example a terminal amino acid contains slightly different atoms than
 one in the middle of a chain.  A force field could of course define multiple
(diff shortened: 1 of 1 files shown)


peastman:
Thanks!  I'm not sure about these changes.  The new wording in the first sentence is fine ("rules for creating new residue templates from existing ones"), but the changes after that are potentially confusing.  For example,

> Note that parameters of existing templates cannot be changed through patches.  Any parameters (such as :code:`charge`\ ) specified in a patch apply only to the new templates.

That only makes sense to someone who has the misconception that patches modify existing templates rather than creating new ones.  To someone who doesn't think that, it's unclear why you're talking about parameters, and it sounds like you're saying contradictory things about what a patch can do.

And then you changed this text:

> A patch can alter multiple templates at once.  This is useful for creating bonds between molecules, and allows the atom types in one residue to depend on the identity of the other residue it is bonded to.

to this:

> A multi-residue patch can be used to create new templates that define bonds between specific residues, and allows the atom types in one residue to depend on the identity of the other residue it is bonded to.

The original version states what a multi-residue patch is (one that's applied to multiple templates at once), and then gives a couple of examples of things they can be used for.  The revised version makes it sound like the examples are the definition, which is confusing.  Can we make that clearer?  How about, "A patch can be applied to multiple templates at once.  This is useful for..."

m-julian:
I've removed the parameter/`charge` sentences, and update the multi-residue patch section with the text you suggested. Essentially, the biggest change now is in the first few sentences that say a patch is used to create new residue templates that match against other residues in the topology.

peastman:
Thanks.  I made a few more edits.  Let me know if they look ok.

m-julian:
Yep, looks good to me. Thanks again for helping out.

peastman:
Thanks!

PR #5213, opened 2026-02-16
Added two more synonyms for HOH

### LINKED ISSUE #5212: Add additional synonyms for `HOH` to the `pdbNames.xml` file
We found that maestro can output PDBs with water residue names `SPC` and `T4P`. Would it be possible to add those as synonyms for HOH to the `pdbNames.xml` file or would that lead to confusions?

peastman:
Fixes #5212.

### DIFF
--- a/wrappers/python/openmm/app/data/pdbNames.xml
+++ b/wrappers/python/openmm/app/data/pdbNames.xml
@@ -274,7 +274,7 @@
   <Atom name="H72" alt1="2H5M"/>
   <Atom name="H73" alt1="3H5M"/>
  </Residue>
- <Residue name="HOH" alt1="H20" alt2="WAT" alt3="SOL" alt4="TIP3" alt5="TP3" alt6="TIP" alt7="T4P" alt8="SPC">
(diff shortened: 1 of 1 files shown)


stefdoerr:
If you want a few more synonyms, I think I picked these out of the VMD source code some time ago:

```json
    "water_resnames": [
        "H2O",
        "HH0",
        "OHH",
        "HOH",
        "OH2",
        "SOL",
        "WAT",
        "TIP",
        "TIP2",
        "TIP3",
        "TIP4",
        "SPC"
    ],
```
Also I see you have `H20` in your file, should probably be `H2O` instead, no? (zero vs letter O)

hannahbaumann:
This works great @peastman , thank you so much!

peastman:
That does make more sense.  All of these are names that I came across in actual files at one point.  On the other hand, anyone who uses `H20` instead of `H2O` kind of deserves what they get!  We make reasonable efforts to support nonstandard names, but that seems like pushing it a little too far.

PR #5253, opened 2026-04-02
Merge fixes for 8.5.1

peastman:
We're preparing to do an 8.5.1 patch release.  This merges the following changes from the main branch: #5236, #5242, #5243, #5251.

### DIFF
--- a/olla/include/openmm/kernels.h
+++ b/olla/include/openmm/kernels.h
@@ -1994,10 +1994,10 @@ public:
     /**
      * Initialize the kernel.
      *
-     * @param system     the System this kernel will be applied to
+     * @param context    the ContextImpl this kernel will be applied to
      * @param force      the CustomCPPForceImpl this kernel will be used for
      */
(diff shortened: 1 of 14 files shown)


epretti (approved):
The changeset looks OK to me.  But for some reason, I don't see all of the CI jobs running.

epretti:
Maybe they are not all getting started because the GitHub CI [is only supposed to run](https://github.com/openmm/openmm/blob/1f2203f60b383794d4ff813b8e413eb2942da77b/.github/workflows/CI.yml#L4-L9) on `master` (as we'd normally want).  Is there a way to get them to run on the 8.5 branch manually?  Would be good for peace of mind before the release.

peastman:
It looks like [it's possible](https://stackoverflow.com/questions/63362126/github-actions-how-can-i-run-a-workflow-created-on-a-non-master-branch-from-t) with the github command line tool.  Which I've never managed to get working...

peastman:
I'll merge once Jenkins finishes, so we'll at least have that test.  There will also be another PR to update the version number.

PR #5254, opened 2026-04-03
Update version number to 8.5.1

peastman:
In preparation for releasing 8.5.1.


peastman:
@epretti I'm manually running the tests on a few computers, since CI isn't running here.  On my Mac I consistently get a failure in TestOpenCLLocalEnergyMinimizer:

> exception: Assertion failure at TestLocalEnergyMinimizer.h:229

epretti:
Do you get an identical failure on Mac on 8.4 and/or earlier OpenMM versions?  If you comment out that one assertion and recompile the test, does it pass?  If so, it is likely due to [this issue](https://github.com/openmm/openmm/pull/5198#issuecomment-3863298583) that we recognized was not a regression and decided not to change for now (OpenCL on Apple hardware wraps instead of saturating `float`-to-`long` conversion, so large forces could get recognized as zero and the minimizer believes it is converged).

peastman:
That makes sense.  Just wanted to check.

All Jenkins builds passed.  In addition to the Mac, I also tested under Linux with an Nvidia GPU.  All tests passed.  I think that means we can go ahead with the release candidate.

I haven't been able to test HIP unfortunately.  (It doesn't get tested by CI either.)  My machine with an AMD GPU died.  A new one is on order and should come in a few weeks.

PR #5257, opened 2026-04-06
Created ReplicaExchangeSampler

peastman:
This is a first draft of replica exchange.  To use it, you first create your Simulation as usual.  Then create a sampler:

```python
sampler = ReplicaExchangeSampler(states, simulation, steps)
```

where `states` is the list of states to sample in [the format of MultistateSampler](https://github.com/openmm/openmm/blob/1f2203f60b383794d4ff813b8e413eb2942da77b/wrappers/python/openmm/app/internal/multistatesampler.py#L33-L45) and `steps` is the number of steps to integrate for each replica on each iteration.  Then you can simulate it by calling

```python
sampler.simulate(iterations)
```

Each iteration involves integrating each replica for `steps` time steps, then attempting exchanges.  Output and analysis are done by adding reporters, much like Simulation:

```python
sampler.reporters.append(reporter)
```

But in this case, a reporter just means any function that takes a ReplicaExchangeSampler as its argument.  It gets called at the end of every iteration.  We'll need to think about what built in reporters we want to provide.  Many users will likely want to write their own as well.

Users can customize this class to use different algorithms but subclassing it and overriding `simulateReplica()` and/or `exchangeReplicas()`.

### DIFF
--- a/wrappers/python/openmm/app/replicaexchangesampler.py
+++ b/wrappers/python/openmm/app/replicaexchangesampler.py
@@ -34,11 +34,122 @@ import math
 import random
 
 class ReplicaExchangeSampler(object):
-    def __init__(self, states: list, simulation: "mm.app.Simulation", stepsPerIteration: int):
+    """
+    ReplicaExchangeSampler uses replica exchange to simulate a system in a collection of different states.  It supports
+    both temperature replica exchange, in which the states correspond to different temperatures, and Hamiltonian replica
+    exchange, in which they correspond to different potential functions.  It also can combine them to vary temperature
+    and potential function at the same time.
+
+    You provide a Simulation describing the system to simulate and a list of states.  It creates multiple replicas of the
+    system, one for each state.  It then simulates all the replicas at once while periodically swapping states between
+    them in a way that ensures correct sampling.
+
+    Replica exchange is used for a variety of purposes.  Here are some of the more common ones.
+
+    - Temperature replica exchange is most often used to accelerate sampling of rare events.  A particular transition might
+      only happen rarely at physiological temperature.  You therefore simulate both the temperature you care about and also
+      higher temperatures at which the transition happens more easily.  Allowing each replica to move between temperatures
+      produces accurate sampling at low temperature of the entire phase space.
+
+    - Hamiltonian replica exchange may be used to sample an alchemical transition between two endpoints.  It produces
+      efficient sampling at every point along the transition pathway from which a free energy difference can be computed.
+
+    Each state (not to be confused with a State object) is represented by a dict containing property values.  All states
+    must specify values for the same properties.  They describe the ways in which the states differ from each other.  The
+    following properties are supported.
+
+    - 'temperature': the simulation temperature
+    - 'groups': a set containing the force groups to include when computing the energy, for example {0, 2}.  It also may
+      be a weighted sum specified as a dict.  For example, {0:1.0, 2:0.5} means to compute the energy of group 0 plus
+      half the energy of group 2.
+    - Context parameters, specified by the parameter name
+    - Global variables defined by a CustomIntegrator, specified by the variable name
+
+    For example, a typical use of temperature replica exchange might define the states as
+
+    >>> states = [{'temperature':t*kelvin} for t in np.geomspace(300.0, 500.0, 10)]
+
+    A typical use of Hamiltonian replica exchange might include forces that depend on a parameter `lambda`.  The states
+    could be defined as
+
+    >>> states = [{'lambda':x} for x in np.linspace(0.0, 1.0, 11)]
+
+    States may also specify multiple properties.  This example includes all combinations of two different parameters.
+
+    >>> states = [{'lambda1':x, 'lambda2':y} for x in np.linspace(0.0, 1.0, 6) for y in np.linspace(0.0, 1.0, 6)]
+
+    To run a replica exchange simulation, first create a Simulation object in the normal way.  Then create a
+    ReplicaExchangeSampler:
+
+    >>> sampler = ReplicaExchangeSampler(states, simulation, stepsPerIteration)
+
+    The third argument is the number of time steps to integrate during each iteration.  Then perform a specified number
+    of sampling iterations by calling
+
+    >>> sampler.simulate(iterations)
+
+    Because replica exchange involves simulating many distinct replicas, the standard reporters used to generate output
+    from a simulation are often not appropriate.  ReplicaExchangeSampler instead provides its own reporting mechanism.
+    Define a function that takes a ReplicaExchangeSampler as its only argument:
+
+    >>> def report(sampler):
+    >>>     # Generate whatever output you want here
+
+    and then add it to the sampler:
+
+    >>> sampler.reporters.append(report)
+
+    The function will be called after every iteration.
+
+    Attributes
+    ----------
+    states: list[dict]
+        The states to sample
+    simulation: openmm.app.Simulation
+        The Simulation defining the System, Context, and Integrator to use
+    stepsPerIteration: int
+        The number of time steps to integrate for each replica on each iteration
+    reinitializeVelocities: bool
+        If true, a replica's velocities are reinitialized from a Boltzmann distribution every time its state changes.
+        This may sometimes improve stability, but also decreases efficiency.
+    exchangesPerIteration: int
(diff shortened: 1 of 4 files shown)


wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,111 @@
...
+from openmm.app.internal.multistatesampler import MultistateSampler
+import math
+import random
+
+class ReplicaExchangeSampler(object):
+    def __init__(self, states: list, simulation: "mm.app.Simulation", stepsPerIteration: int):
+        self.states = states
+        self.simulation = simulation
+        self.stepsPerIteration = stepsPerIteration
+        self.exchangesPerIteration = len(states)**3
```
IAlibay: It might be good to optionally let users pass in a value for this?

wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,111 @@
...
+            self.replicaStateEnergy = energies
+            self.exchangeReplicas()
+            self.currentIteration += 1
+            for reporter in self.reporters:
+                reporter(self)
+
+    def simulateReplica(self, index: int, steps: int | None = None):
+        self.simulation.context.setState(self.replicaConformation[index])
+        if self._kT is not None:
+            # Rescale velocities for replicas whose temperatures have changed.
```
IAlibay, line 221: As part of the discussion on rescale vs reinitialize, it would be great to let users choose to always reinitialize.

wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,111 @@
...
+                    break
+            si = self.replicaStateIndex[i]
+            sj = self.replicaStateIndex[j]
+            if self._kT is None:
+                kT = unit.MOLAR_GAS_CONSTANT_R*self.simulation.integrator.getTemperature()
+                exponent = (self.replicaStateEnergy[i][si]-self.replicaStateEnergy[i][sj]+self.replicaStateEnergy[j][sj]-self.replicaStateEnergy[j][si])/kT
+            else:
+                exponent = (self.replicaStateEnergy[i][si]-self.replicaStateEnergy[j][si])/self._kT[si] + (self.replicaStateEnergy[j][sj]-self.replicaStateEnergy[i][sj])/self._kT[sj]
+            prob = min(1.0, math.exp(exponent))
+            if random.random() <= prob:
```
IAlibay, line 265: It might be good to accumulate the acceptance matrix so that it can be reported to users.

IAlibay (commented):
I only had a brief look through (cc @jthorton who might have time for a more detailed look through). The overall API looks great to me!

wrappers/python/openmm/app/replicaexchangesampler.py
peastman: In what format?  What exactly should be recorded?

peastman:
I marked this as ready for review.  In addition to adding API documentation, I made two functional changes.  First, I added an option to reinitialize velocities after state changes.  Second, I reduced the default number of exchange attempts from $N^3$ to $N^2$.

Here are some things I intentionally did not include in this PR:

- Any built in reporters.  We still need to decide what we want.  (@IAlibay remember I could use examples of replica exchange H5MD files from other programs.)
- Documentation other than API docs.  Once we're done with all the multistate samplers I'll add a section to the manual discussing all of them.  Probably a tutorial too.

wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,269 @@
...
+                j = random.randrange(n)
+                if i != j:
+                    break
+            si = self.replicaStateIndex[i]
+            sj = self.replicaStateIndex[j]
+
+            # Compute the probability of exchanging their states.
+
+            if self._kT is None:
+                kT = unit.MOLAR_GAS_CONSTANT_R*self.simulation.integrator.getTemperature()
```
jthorton, line 260: can we avoid doing this at every exchange attempt if the temperature is constant accross states? Maybe in the init it would make sense to have self._kT be formed to be `self._kT = [unit.MOLAR_GAS_CONSTANT_R*self.simulation.integrator.getTemperature()] * len(states)`

wrappers/python/openmm/app/replicaexchangesampler.py
jthorton: It could be an `len(states) * len(states)` matrix one for the swap attempts and one for the swap acceptance, see my suggestions.  We could also add a property which returns the exchange acceptance rate?

wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,269 @@
...
+        self.reinitializeVelocities = reinitializeVelocities
+        self.exchangesPerIteration = len(states)**2
+        self.currentIteration = 0
+        self.reporters = []
+        self.replicaStateIndex = list(range(len(states)))
+        self._previousReplicaStateIndex = self.replicaStateIndex[:]
+        self.replicaConformation = [simulation.context.getState(positions=True, velocities=True, parameters=True, integratorParameters=True)]*len(states)
+        self.replicaStateEnergy = None
+        self._sampler = MultistateSampler(states, simulation.context)
+        if 'temperature' in states[0]:
```
jthorton, line 160:
```suggestion
        n_states = len(states)
        self.exchangeAttempts = np.zeros((n, n), dtype=int)
        self.exchangeAcceptances = np.zeros((n,n), dtype=int)
        if 'temperature' in states[0]:
```

wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,269 @@
...
+            sj = self.replicaStateIndex[j]
+
+            # Compute the probability of exchanging their states.
+
+            if self._kT is None:
+                kT = unit.MOLAR_GAS_CONSTANT_R*self.simulation.integrator.getTemperature()
+                exponent = (self.replicaStateEnergy[i][si]-self.replicaStateEnergy[i][sj]+self.replicaStateEnergy[j][sj]-self.replicaStateEnergy[j][si])/kT
+            else:
+                exponent = (self.replicaStateEnergy[i][si]-self.replicaStateEnergy[j][si])/self._kT[si] + (self.replicaStateEnergy[j][sj]-self.replicaStateEnergy[i][sj])/self._kT[sj]
+            prob = min(1.0, math.exp(exponent))
```
jthorton, line 264:
```suggestion
            # record the attempt
            self.exchangeAttempts[si, sj] += 1
            self.exchangeAttempts[sj, si] += 1
            prob = min(1.0, math.exp(exponent))
```

wrappers/python/openmm/app/replicaexchangesampler.py
```diff
@@ -0,0 +1,269 @@
...
+            if random.random() <= prob:
+                # Accept the exchange.
+
+                self.replicaStateIndex[i] = sj
+                self.replicaStateIndex[j] = si
```
jthorton, line 269:
```suggestion
                self.replicaStateIndex[j] = si
                # record the acceptance
                self.exchangeAcceptances[si, sj] += 1
                self.exchangeAcceptances[sj, si] += 1
```

wrappers/python/openmm/app/replicaexchangesampler.py
peastman: I did it this way so it would work if the user changes the temperature after the ReplicaExchangeSampler is created.  For example, they might change the temperature during equilibration.

wrappers/python/openmm/app/replicaexchangesampler.py
peastman: This probably won't give what you're expecting, because one or both states may have already been exchanged once.  I think simple counting like this is only valid in programs that only attempt exchanges between nearest neighbors, and alternate even and odd pairs so each state is guaranteed to only get one exchange attempt on each iteration.

It's better to think of the entire sequence of $N^2$ pairwise exchange attempts as a single permutation of the states.  Then you can simply look at the states and record what has changed.

peastman:
I'm going to merge this so I can add the reporter in a separate PR.  Feel free to keep making comments.  They can be addressed in other PRs.

PR #5261, opened 2026-04-10
Created ReplicaExchangeReporter

peastman:
This is the standard reporter for writing output from a replica exchange simulation.

### DIFF
--- a/wrappers/python/openmm/app/replicaexchangereporter.py
+++ b/wrappers/python/openmm/app/replicaexchangereporter.py
@@ -1,0 +1,144 @@
+"""
+replicaexchangereporter.py: Writes output for replica exchange simulations
+
+This is part of the OpenMM molecular simulation toolkit.
+See https://openmm.org/development.
+
+Portions copyright (c) 2026 Stanford University and the Authors.
+Authors: Peter Eastman
+Contributors:
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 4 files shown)


wrappers/python/openmm/app/replicaexchangereporter.py
```diff
@@ -0,0 +1,203 @@
...
+            else:
+                energy /= sampler._kT
+            print(','.join([str(x) for x in energy.flatten()]), file=self._energy)
+            self._energy.flush()
+        for i, conf in enumerate(sampler.replicaConformation):
+            if len(self._stateReporters) > 0:
+                self._stateReporters[sampler.replicaStateIndex[i]].report(sampler.simulation, conf)
+            if len(self._replicaReporters) > 0:
+                self._replicaReporters[i].report(sampler.simulation, conf)
+            safesave.save(mm.XmlSerializer.serialize(conf), os.path.join(self.directory, f'checkpoint_{i}.xml'))
```
epretti, line 203: If the process gets interrupted here, I think it's possible to be left with a mix of old and new checkpoint files, as well as trajectories/logs with different numbers of frames.  It's probably unlikely but not impossible.

peastman: I thought about that, but I don't have a good solution.  The only completely safe way to do it would be to create a new copy of the entire folder, update all the files in the copy at once, and then delete the other one.

epretti: For the checkpoint files, presumably one could atomically write one file that contained the data for all of the states.  That would at least make sure the checkpoints were always valid.

For files that get appended to, the way [GROMACS handles it](https://manual.gromacs.org/current/user-guide/managing-simulations.html#appending-to-output-files) is to either (1) truncate the files on restart from a checkpoint (after verification with checksums also recorded in the checkpoint), or (2) write files in multiple parts that presumably could be concatenated later.

peastman: I think we can call that out of scope for right now.  In addition to being complicated to implement, it would probably end up being more difficult for the user.  For example, instead of having a single trajectory file, they would have multiple files that they would have to concatenate.

peastman:
I'm going to merge this so I can go on to the next PR in the series.

PR #5271, opened 2026-04-25
Avoid host wait for PME post-computation synchronization

alephpiece:
This replaces the host-side wait in `SyncQueuePostComputation` with a queue-side wait.

When PME runs on a separate queue, the post-computation step only needs to ensure that later work on the default queue does not run until the PME queue has completed. Enqueuing a wait on the current queue preserves that dependency, while avoiding a CPU-side stall. The `addEnergy` kernel is still launched after the wait, so the PME energy buffer is not consumed until the reciprocal-space work has finished.

Before the common NonbondedForce implementation in #4922, the CUDA, HIP, and OpenCL implementations used queue/stream-side synchronization for this step (`cuStreamWaitEvent`, `hipStreamWaitEvent`, and an OpenCL barrier wait list). This change restores the same synchronization pattern through the common `ComputeEvent::queueWait()` API.

### DIFF
--- a/platforms/common/src/CommonCalcNonbondedForce.cpp
+++ b/platforms/common/src/CommonCalcNonbondedForce.cpp
@@ -194,7 +194,9 @@ public:
     }
     double computeForceAndEnergy(bool includeForces, bool includeEnergy, int groups) {
         if ((groups&(1<<forceGroup)) != 0) {
-            event->wait();
+            // Keep the dependency on reciprocal-space work, but avoid stalling the host
+            // when the default queue can wait for the PME queue asynchronously.
+            event->queueWait(cc.getCurrentQueue());
             if (includeEnergy)
                 addEnergyKernel->execute(pmeEnergyBuffer.getSize());
         }
--- a/platforms/common/src/CommonCalcNonbondedForce.cpp
+++ b/platforms/common/src/CommonCalcNonbondedForce.cpp
@@ -194,8 +194,6 @@ public:
     }
     double computeForceAndEnergy(bool includeForces, bool includeEnergy, int groups) {
         if ((groups&(1<<forceGroup)) != 0) {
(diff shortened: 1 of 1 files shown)


platforms/common/src/CommonCalcNonbondedForce.cpp
```diff
@@ -194,7 +194,9 @@ class CommonCalcNonbondedForceKernel::SyncQueuePostComputation : public ComputeC
     }
     double computeForceAndEnergy(bool includeForces, bool includeEnergy, int groups) {
         if ((groups&(1<<forceGroup)) != 0) {
-            event->wait();
+            // Keep the dependency on reciprocal-space work, but avoid stalling the host
+            // when the default queue can wait for the PME queue asynchronously.
```
peastman: Can you remove this comment?  It describes what you changed, not what the code does.  For someone reading the code in the future, it would be confusing.

peastman (commented):
Thanks!  This looks like a reasonable change in principle, though in practice I'm not observing any benefit in the benchmarks.  Have you identified any cases where it produces a measurable speedup?

alephpiece:
I've been tuning the PME path on HIP platform for a while and observed a slight performance gain on pme, apoa1pme, and apoa1ljpme with this change.

For example, the pme case improved from ~950 ns/day to ~990 ns/day.

> These numbers are from a tuned build on a CDNA-like GPU; the official benchmark has not been released yet.

peastman:
I do see a small speedup on HIP.  These are on a Radeon RX 9070 XT:

|Test|Current|This PR|
|---|---|---|
pme|1564|1594
apoa1pme|593|597
apoa1ljpme|432|438

The change is just barely above the noise level, but it is measurable.

I've started the tests running.  Assuming everything passes, I'll merge once it finishes.

PR #5276, opened 2026-04-27
Allow preservation of LRC coefficients when updating force parameters

lohedges:
The LRC coefficient cache introduced in #5239 keys by the current values of the force's global parameters, avoiding the cost of recomputing the coefficients when switching between discrete states. In our free-energy perturbation engine, [SOMD2](https://github.com/OpenBioSim/somd2), we move between discrete states by changing the particle parameters in the force, rather than using global parameters. Since the LRC cache is cleared whenever `updateParametersInContext` is called, we have no way to take advantage of the caching, despite the LRC being uniquely defined for each set of atom parameters.

This PR adds an optional `preserveLongRangeCorrection` kwarg to the `updateParametersInContext` method of `CustomNonbondedForce` that stops the cache being cleared. Combined with a dedicated global parameter used as an explicit cache key (which need not appear in the energy expression at all), this enables per-state LRC caching in multi-state simulations where particle parameters change. The same kwarg is added to the regular `NonbondedForce`, allowing a user to skip recomputation of the LRC correction when LJ parameters are unchanged, e.g. when the force is only being used for electrostatics.

Testing with [Sire](https://github.com/OpenBioSim/sire/tree/feature_preserve_lrc), we obtain the following (per-window) results for the bound leg of a typical ABFE perturbation. This is performing an energy scan over 24 lambda windows using the CUDA platform on a 5070 Ti.

| | No LRC | LRC, `preserveLongRangeCorrection=True` | LRC, `preserveLongRangeCorrection=False` |
|---|---|---|---|
| `set_lambda` | ~2.3 ms | ~5.4 ms | ~63 ms |
| `get_potential_energy` | ~0.35 ms | ~0.38 ms | ~1.6 ms |
| **Total** | **~2.7 ms** | **~5.8 ms** | **~65 ms** |
| Speedup vs `False` | 24x | **11x** | baseline |
| Speedup vs no LRC | baseline | 2.2x slower | 24x slower |

Energies were also found to agree within numerical precision across the lambda scan using `preserveLongRangeCorrection=True` or `False`.

While debugging this implementation I came across an issue with the LRC calculation in the `CustomNonbondedForce`, which caused convergence errors for our custom forces. While `prepareLongRangeCorrection` correctly counts pair interactions per particle class pair, `calcLongRangeCorrection` called `integrateInteraction` for _all_ class pairs before multiplying by the count. If a zero-count pair happens to not converge, this would crash the entire LRC computation. As such, a zero count guard was placed before the calculation.


lohedges:
Just to say that you obviously don't need a correction in the regular force for electrostatics only. However, ours does include LJ terms that are unperturbed.

peastman:
Thanks for the contribution.  I'm not sure this is the right solution.  It adds a flag that effectively means, "Compute the wrong answer."  I try to avoid that whenever possible!

Let's see if we can come up with a better solution.  Can you open an issue to discuss it?  Provide a self-contained test case for it.  Create a system that's typical of the ones you really work with, and repeatedly modify it in ways that are typical of the real changes you make.  The goal is to have a realistic code we can profile and optimize.

lohedges:
Many thanks for the feedback. I'll try to write something up later in the week and open an issue thread to discuss.

I've reworked the existing code so that changes to LJ parameters are autodetected in the regular `NonbondedForce`, so there is no need for the `preserveLongRangeCorrection` kwarg there. This means that the issue reduces to figuring out a way to cache the LRC for the `CustomNonbondedForce` when using discrete states described via particle parameter changes, not changes to global parameters.

I agree that my existing approach is more of a hack than anything and it would be good to avoid giving users a way to shoot themselves in the foot. I just couldn't think of any way to extend the caching approach implemented in #5239 without keying on the particle and global parameters, which would be painful. We could probably workaround this issue ourselves by changing the way we use custom forces for standard FEP schedules, i.e. use global parameters where it makes sense and parameter updates only when doing something more exotic.

lohedges:
I guess what we should really do is calculate and cache `CustomNonbondedForce` LRC ourselves for each state, then use `CustomVolumeForce` to implement it.

lohedges:
That worked perfectly. Apologies for not thinking about this harder in the first place :-)

The reduced scope of this PR is:

1) NonbondedForce::updateParametersInContext previously recomputed the dispersion correction on every call, even when only charges were updated. This adds a sigma/epsilon snapshot that is compared against the updated range [firstParticle, lastParticle] on each call; the dispersion coefficient is only recomputed when a change is detected. The snapshot is initialised during initialize() and updated whenever a change triggers a recompute.

2) CustomNonbondedForceImpl::calcLongRangeCorrection iterated over all particle class pairs when computing the long-range correction coefficient, including pairs where the count had zero particles. A guard now skips any pair where the particle count for that pair class is zero.

peastman:
Just a reminder to open an issue about this and provide a test case we can optimize for.  There's no hurry.  I just want to be sure it doesn't get forgotten.

lohedges:
Thanks, and apologies for the lack of an update. I've managed to work around this:

https://github.com/OpenBioSim/sire/pull/440 

Here I compute and cache things outside of OpenMM, then apply via several `CustomVolumeForce`s depending on what type of simulation is being performed. This seems to work well and performance is good. I'm still in the process of testing it more thoroughly, though. 

The changes in this PR aren't needed for the above, so please feel free to close if they are unlikely to be useful in general. (I think our original use case was quite atypical.)

peastman:
Good to know.  An issue and test case would still be useful, if you feel inclined.  You may not be the only person who wants to do something like this.

lohedges:
Note that the discussion has moved to #5410. A new prototype implementation that supersedes the one originally implemented here is discussed there, with the PR branch used here still used as the reference.

PR #5279, opened 2026-05-02
Make sure contexts are deselected before evaluation

### LINKED ISSUE #5277: Minimizing with PythonForce using PyTorch on CUDA fails with `CUDA_ERROR_INVALID_HANDLE`
With Python 3.14.4, OpenMM 8.5.1 and PyTorch 2.10.0, the following:
```python
import openmm
import openmm.unit
import torch

system = openmm.System()
system.addParticle(1)

def evaluate(state):
    positions = torch.from_numpy(state.getPositions(asNumpy=True).value_in_unit(openmm.unit.nanometer)).to(device="cuda")
    energy = positions.square().sum()
    forces = -2 * positions
    return energy.item(), forces.cpu().numpy()
force = openmm.PythonForce(evaluate)
system.addForce(force)

context = openmm.Context(system, openmm.VerletIntegrator(1), openmm.Platform.getPlatformByName("CUDA"))
context.setPositions([[10, 0, 0]])
# context.getState(energy=True)
openmm.LocalEnergyMinimizer.minimize(context)
```
fails with
```
  File "/home/.../test.py", line 19, in <module>
    openmm.LocalEnergyMinimizer.minimize(context)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/.../lib/python3.14/site-packages/openmm/openmm.py", line 16692, in minimize
    return _openmm.LocalEnergyMinimizer_minimize(context, tolerance, maxIterations, reporter)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
openmm.OpenMMException: Error invoking kernel: CUDA_ERROR_INVALID_HANDLE (400)
```
This does not occur if the `context.getState(energy=True)` line is uncommented, the OpenMM platform is specified as something other than `CUDA`, or the PyTorch device is set to `cpu`.

I am currently working on finding the cause.

I ran into another issue (with a more complicated reproduction) where memory in PyTorch buffers seemed to get corrupted but I haven't ruled out a problem in PyTorch or with my code.  If it is actually related to this, I will update this issue.

epretti:
Fixes #5277.  Ensures that the (CUDA) context is deselected before `LocalEnergyMinimizer` calls `calcForcesAndEnergy`.  There are several other places in the code where this occurs (and which can cause a `PythonForce` callback to be invoked with OpenMM's CUDA context on the stack), so I created a `ContextDeselector` to pop a context and push it again.  If there is a better way to accomplish this, I could use it instead of adding this class, but trying to use only `ContextSelector` made some methods very messy.

The checks in `PythonForce` and `CustomCPPForce` for other instances of these forces should also now not be triggered when they find themselves.

### DIFF
--- a/platforms/common/include/openmm/common/ContextSelector.h
+++ b/platforms/common/include/openmm/common/ContextSelector.h
@@ -49,6 +49,25 @@ private:
     ComputeContext& context;
 };
 
+/**
+ * This class deselects a ComputeContext by calling popAsCurrent() on the
+ * context when it is created and pushAsCurrent() when it goes out of scope.
+ * This can be useful to temporarily undo the effect of a ContextSelector and
+ * must only be used when the context is already selected.
+ */
+
+class OPENMM_EXPORT_COMMON ContextDeselector {
+public:
+    ContextDeselector(ComputeContext& context) : context(context) {
+        context.popAsCurrent();
+    }
+    ~ContextDeselector() {
+        context.pushAsCurrent();
+    }
+private:
+    ComputeContext& context;
+};
+
 } // namespace OpenMM
 
 #endif /*OPENMM_CONTEXTSELECTOR_H_*/
--- a/platforms/common/src/CommonIntegrateCustomStepKernel.cpp
+++ b/platforms/common/src/CommonIntegrateCustomStepKernel.cpp
@@ -683,7 +683,10 @@ void CommonIntegrateCustomStepKernel::execute(ContextImpl& context, CustomIntegr
             }
             else {
                 recordChangedParameters(context);
-                energy = context.calcForcesAndEnergy(computeForce, computeEnergy, forceGroups);
+                {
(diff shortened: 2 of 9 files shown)


peastman:
Looks good.  Would it make sense to make `popAsCurrent()` throw an exception if it isn't actually current?  Something like this.

```c++
void CudaContext::popAsCurrent() {
    CUcontext popped;
    if (contextIsValid) {
        cuCtxPopCurrent(&popped);
        if (popped != context)
            throw OpenMMException("Called popAsCurrent() on a context that is not current");
    }
}
```

That will catch code paths where a ContextDeselector isn't inside a ContextSelector.

peastman (approved):

PR #5294, opened 2026-05-25
refactor CMake

e-kwsm:
- fix: call cmake_minimum_required firstly

  > CMake Warning (dev) at CMakeLists.txt:21 (PROJECT):
  >   cmake_minimum_required() should be called prior to this top-level project()
  >   call.  Please see the cmake-commands(7) manual for usage documentation of
  >   both commands.
  > This warning is for project developers.  Use -Wno-dev to suppress it.

- refactor: replace deprecated PythonInterp with Python3
  https://cmake.org/cmake/help/v3.17/module/FindPythonInterp.html


peastman:
Changing `PYTHON_EXECUTABLE` to `Python3_EXECUTABLE` is a breaking change that will impact everyone's existing build directories.  And it's fixing an obsolete problem that doesn't exist anymore: all 2.x versions of Python have been out of support for years now, and there's no longer any difference between `python` and `python3`.

e-kwsm:
FindPythonInterp is deprecated in CMake 3.12, and FindPython or FindPython3 should be used instead; the resulting variables are renamed as well.

PR #5302, opened 2026-05-29
Fix `AmoebaMultipoleForce::ForceInfo` particle groups used for molecule detection and atom reordering

### LINKED ISSUE #5300: Incorrect `AmoebaMultipoleForce::ForceInfo` can break `reorderAtoms()` and produce wrong CUDA results
# Summary

`AmoebaMultipoleForce::ForceInfo` appears to provide incomplete connectivity information to `ComputeContext::findMoleculeGroups()`.

As a result, when `AmoebaMultipoleForce` is used on CUDA, atoms that are covalently related may fail to be assigned to the same molecule. Then `reorderAtoms()` can reorder atoms incorrectly across molecule boundaries, which breaks the assumptions behind precomputed exclusion tiles / covalent scaling and leads to incorrect results.

This does not affect CPU, since atom reordering is a CUDA optimization. It also does not affect systems where some other force already supplies sufficient connectivity information (for example, a `CustomNonbondedForce` with exclusions/exceptions that connect the full molecule graph).

# Root Cause

`ComputeContext::findMoleculeGroups()` builds molecular connectivity by taking the union of all particle groups reported by every `ComputeForceInfo`:

```c++
for (OpenMM::ComputeForceInfo* force : forces) {
    vector<int> particles;
    for (int j = 0; j < force->getNumParticleGroups(); j++) {
        force->getParticlesInGroup(j, particles);
        for (int k = 1; k < (int) particles.size(); k++) {
            atomBondSets[particles[k]].insert(particles[k-1]);
            atomBondSets[particles[k-1]].insert(particles[k]);
        }
    }
}
```

So `getParticlesInGroup()` is effectively expected to return a connected set of particles for each “group”.

However, the current `AmoebaMultipoleForce::ForceInfo` implementation does:

```c++
int getNumParticleGroups() {
    return 7*force.getNumMultipoles();
}

void getParticlesInGroup(int index, vector<int>& particles) {
    int particle = index/7;
    int type = index-7*particle;
    force.getCovalentMap(particle, AmoebaMultipoleForce::CovalentType(type), particles);
}
```

There are two problems:
1. `getParticlesInGroup()` does **not** include the central particle itself.
2. The hard-coded value 7 is inconsistent with the actual enum: `AmoebaMultipoleForce::CovalentEnd` == 8.

Because `findMoleculeGroups()` only adds edges between consecutive entries inside particles, if the central particle is omitted, the returned list may fail to connect the full covalent neighborhood to that particle. 

> For Example, if particle 0's only Covalent12 is particle 1, then `getParticlesInGroup` only return `[1]`, then the loop `for (int k = 1; k < (int) particles.size(); k++)` will not be entered

This can produce incomplete connected components and therefore incorrect molecule grouping.

# Why this causes wrong CUDA results

`NonbondedUtilities::initialize()` builds exclusion tiles based on the original atom ordering.

Later, on CUDA, `ComputeContext::reorderAtoms()` may reorder atoms by molecule for better memory locality.

This is safe only if all atoms that belong to one molecule are correctly identified and reordered together.

If `findMoleculeGroups()` incorrectly splits one physical molecule into multiple connected components, `reorderAtoms()` can interleave atoms from different molecules into what it thinks are “equivalent molecules”. After that, precomputed exclusion/covalent scaling information no longer matches the reordered internal atom layout.

In the Amoeba CUDA kernels, tiles with special covalent scaling are handled through `exclusionTiles`, while the normal neighbor path assumes scale factors (`dScale=1, pScale=1`). If a tile that should be treated as an exclusion tile is no longer recognized after reordering, interactions can go through the wrong path and use incorrect scaling factors.

This was observed as a mismatch between CPU and CUDA results in induced field calculations.

cmd23333:
This PR fixes `AmoebaMultipoleForce::ForceInfo` so that molecule detection in `ComputeContext::findMoleculeGroups()` sees complete connectivity.

#Problem

`AmoebaMultipoleForce::ForceInfo::getParticlesInGroup()` currently returns only the covalent map entries, but omits the central particle. `findMoleculeGroups()` interprets each returned particle group as a connected chain and builds molecular connectivity from consecutive entries in that vector. Omitting the central particle can therefore split a real molecule into multiple connected components.

This becomes problematic on CUDA because `reorderAtoms()` reorders atoms by molecule. If molecule detection is wrong, atoms can be reordered across molecule boundaries, while precomputed exclusion tiles and covalent scaling tables still assume the original internal molecule structure.

Fixes #5300 

#Changes
- Replace hard-coded `7` with `AmoebaMultipoleForce::CovalentEnd`
- Include the central particle in `getParticlesInGroup()`

#Result
This restores correct molecule grouping and fixes the observed CPU/CUDA mismatch in our test case.

### DIFF
--- a/plugins/amoeba/platforms/common/src/AmoebaCommonKernels.cpp
+++ b/plugins/amoeba/platforms/common/src/AmoebaCommonKernels.cpp
@@ -198,7 +198,7 @@ public:
         int particle = index / AmoebaMultipoleForce::CovalentEnd;
         int type = index - AmoebaMultipoleForce::CovalentEnd * particle;
         force.getCovalentMap(particle, AmoebaMultipoleForce::CovalentType(type), particles);
-        particles.insert(particles.begin(), particle);
+        particles.push_back(particle);
     }
     bool areGroupsIdentical(int group1, int group2) {
         return ((group1%AmoebaMultipoleForce::CovalentEnd) == (group2%AmoebaMultipoleForce::CovalentEnd));
--- a/plugins/amoeba/platforms/common/src/AmoebaCommonKernels.cpp
+++ b/plugins/amoeba/platforms/common/src/AmoebaCommonKernels.cpp
@@ -192,15 +192,16 @@ public:
         return true;
     }
     int getNumParticleGroups() {
-        return 7*force.getNumMultipoles();
+        return AmoebaMultipoleForce::CovalentEnd * force.getNumMultipoles();
     }
     void getParticlesInGroup(int index, vector<int>& particles) {
-        int particle = index/7;
-        int type = index-7*particle;
+        int particle = index / AmoebaMultipoleForce::CovalentEnd;
+        int type = index - AmoebaMultipoleForce::CovalentEnd * particle;
         force.getCovalentMap(particle, AmoebaMultipoleForce::CovalentType(type), particles);
+        particles.insert(particles.begin(), particle);
     }
     bool areGroupsIdentical(int group1, int group2) {
-        return ((group1%7) == (group2%7));
+        return ((group1%AmoebaMultipoleForce::CovalentEnd) == (group2%AmoebaMultipoleForce::CovalentEnd));
     }
 private:
     const AmoebaMultipoleForce& force;


peastman:
Thanks!  This looks good except for one minor suggestion:

```c++
particles.insert(particles.begin(), particle);
```

Inserting a value at the start of a vector is expensive because it requires rewriting the entire content of the vector.  It will be faster if you change that to `particles.push_back(particle)`.

cmd23333:
Thanks for the suggestion! That makes sense. I’ve updated it to use `particles.push_back(particle)` instead of inserting at the beginning in the latest commit.

peastman:
Looks good.  Thanks!

PR #5306, opened 2026-05-31
Enable split PME streams for HIP LJPME

alephpiece:
This PR splits HIP LJPME reciprocal-space PME work across two PME streams: one for Coulomb PME and one for dispersion PME.

OpenMM already supports running PME work on a separate stream from direct-space work.  For LJPME, the Coulomb and dispersion reciprocal branches are independent, but they previously executed serially on the same PME stream.

This change gives the dispersion branch its own PME queue, grid, atom-grid indices, sorter, energy buffer, and completion event so the two branches can overlap without sharing mutable scratch state.

The change is enabled only for HIP LJPME when Coulomb interactions are present and PME streams are in use.  Non-HIP platforms, non-LJPME systems, and existing `DisablePmeStream` behavior are unchanged.

Performance was measured as the average of three 60 second benchmark runs on CDNA architecture.

| Case | Before | After | Change|
|------|--------|-------|-------|
|apoa1ljpme| 277.741 ns/day | 294.423 ns/day | +6.01% |

The trace below shows the intended overlap between Coulomb PME and dispersion PME work in an LJPME step.

image

### DIFF
--- a/platforms/common/include/openmm/common/CommonCalcNonbondedForce.h
+++ b/platforms/common/include/openmm/common/CommonCalcNonbondedForce.h
@@ -45,7 +45,8 @@ namespace OpenMM {
 class CommonCalcNonbondedForceKernel : public CalcNonbondedForceKernel {
 public:
     CommonCalcNonbondedForceKernel(std::string name, const Platform& platform, ComputeContext& cc, const System& system) : CalcNonbondedForceKernel(name, platform),
-            hasInitializedKernel(false), cc(cc), pmeio(NULL), stepsToSort(0) {
+            hasInitializedKernel(false), cc(cc), pmeio(NULL), stepsToSort(0), useSplitLJPMEStream(false),
+            dispersionSyncQueue(NULL) {
     }
     ~CommonCalcNonbondedForceKernel();
     /**
@@ -133,6 +134,8 @@ private:
     ComputeArray cosSinSums;
     ComputeArray pmeGrid1;
     ComputeArray pmeGrid2;
+    ComputeArray pmeDispersionGrid1;
+    ComputeArray pmeDispersionGrid2;
     ComputeArray pmeBsplineModuliX;
     ComputeArray pmeBsplineModuliY;
     ComputeArray pmeBsplineModuliZ;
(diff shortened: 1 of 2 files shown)


peastman:
I see you limit this to only cases when `cc.getSIMDWidth() == 64`.  I assume the intent is to only use it on CDNA, not RDNA?  Is that because you tested on both and only saw a benefit on CDNA?  I don't see any reason that the SIMD width in itself should matter.

For the same reason, did you only implement it for HIP because there's no benefit for CUDA?

peastman:
I don't have access to a CDNA GPU.  I tested on a Radeon RX 9070 XT (after removing the check for SIMD width).  I tried different systems ranging from 23K to 1M atoms.  The effect is a huge slowdown (>50%) for small systems, switching over to a very small speedup (~2%) on the largest ones.

|System|Original|This PR|
|---|---|---|
DHFR|1151|542
ApoA1|459|398
Cellulose|114|116
STMV|41.1|41.7

alephpiece:
Thanks for testing on RDNA, that's really helpful data!

I don't have access to RDNA hardware, so I couldn't measure it directly. I did test the same split-stream scheme on RTX 4090/CUDA, and it also regresses there:

| Case | Before | After | Change |
| --- | ---: | ---: | ---: |
| ljpme (DHFR) | 1949.230 | 1898.827 | -2.6% |
| apoa1ljpme | 692.254 | 668.166 | -3.5% |

In contrast, both LJPME cases improved on the CDNA HIP backend:

| Case | Before | After | Change |
| --- | ---: | ---: | ---: |
| ljpme (DHFR) | 753.047 | 818.354 | +8.67% |
| apoa1ljpme | 277.741 | 294.659 | +6.09% |

So the current evidence suggests this optimization is backend/hardware dependent: it helps in my CDNA/HIP measurements, but regresses on RTX 4090/CUDA, and your Radeon RX 9070 XT/HIP result points in the same direction.

My current read of the Nsight Systems trace is that the PME overlap itself is real, but the key question is whether that local overlap shortens the whole simulation step. For `apoa1ljpme` on RTX 4090/CUDA:

| Metric | split - baseline | Definition / reading |
| --- | ---: | --- |
| PME union saved | +119.243 us/step | Reduction in the merged PME-kernel coverage within a step; the two PME branches do overlap and the local PME coverage is shorter |
| Kernel union saved | -1.697 us/step | Reduction in the merged coverage of all GPU kernels within a step; this does not get shorter |
| Gap penalty | +13.683 us/step | Increase in step time not covered by GPU kernels; split adds non-kernel / scheduling gap |
| Step saved | -15.380 us/step | Reduction in the profiled step interval; the step is slower, consistent with the formal regression |

By this accounting, `step_saved = kernel_union_saved - gap_penalty`, so `-1.697 - 13.683 = -15.380 us/step`.

I think that, for this OpenMM stream graph on RTX 4090, the local PME saving is mostly consumed by concurrent-kernel interference and extra GPU-side scheduling/queue gap before it can reduce the step time.

The detailed timeline points in the same direction: main/list/sort work and aggregate PME work become longer, and the added gap is mostly post-launch GPU-side wait rather than host `cuLaunchKernel` API time. Nsight Compute did not show a matching isolated kernel resource-shape or occupancy regression, so this looks more like a runtime scheduling/contention issue than poorer CUDA code generation for the affected kernels.

peastman:
I think this change probably isn't worth making.  The reasons are:

- It's only useful on a very small set of GPUs.  It seems to always be harmful on NVIDIA GPUs, and also on some AMD GPUs.
- At best it only produces a very small speedup, whereas in the bad cases it can produce a huge slowdown.
- We don't really understand the reasons it helps on the GPU you tested but hurts everywhere else, which means we can't predict the effect on other GPUs we haven't tested, including ones not yet released.

So I think the risks of adding it outweigh any potential benefit.

PR #5307, opened 2026-06-02
Fix deepcopy on Topology bonds to reference newly copied atoms

patelvyom:
Currently, when deepcopying an `openmm.app.Topology` object, the `Bond` objects in the new topology retain references to the original `Atom` objects rather than pointing to the newly copied Atom objects. This results in an invalid copied topology state where its bonds point to external atoms.

This is caused by `Bond.__deepcopy__` ignoring the memo dictionary and directly copying the `self[0]` and `self[1]` original references.

Fix: Updated `Bond.__deepcopy__` in `openmm/app/topology.py` to properly use deepcopy on its atom references along with the memo dictionary. This ensures that the bond correctly looks up and assigns the newly created copied atoms.

### DIFF
--- a/wrappers/python/openmm/app/topology.py
+++ b/wrappers/python/openmm/app/topology.py
@@ -509,7 +509,7 @@ class Bond(namedtuple('Bond', ['atom1', 'atom2'])):
         return self.__dict__
 
     def __deepcopy__(self, memo):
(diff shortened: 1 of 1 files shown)


peastman:
Looks good.  Thanks for the fix!

PR #5317, opened 2026-06-13
Add GPU ESP kernel for NonbondedForce PME

lu1and10:
## Description

This adds an opt-in ESP reciprocal-space backend for charge-charge Coulomb interactions in `NonbondedForce` PME on GPU platforms.

The default remains the existing Gaussian Ewald/B-spline PME path.  Users select the new path with:

```cpp
force.setReciprocalSpaceKernelType(NonbondedForce::ESPKernel);
```

The implementation is intentionally scoped to the existing common GPU PME machinery:

- ESP real-space split tables replace the direct-space `erfc()` Coulomb split when ESP is selected.
- ESP/PSWF spreading weights replace B-spline spreading/interpolation in `pme.cc`.
- ESP reciprocal convolution uses the prolate split transform and PSWF inverse moduli.
- Standard PME remains unchanged for the default `PMEKernel`.

### DIFF
--- a/openmmapi/include/openmm/NonbondedForce.h
+++ b/openmmapi/include/openmm/NonbondedForce.h
@@ -145,6 +145,22 @@ public:
         LJPME = 5
     };
     /**
+     * This is an enumeration of the reciprocal space kernels that may be used
+     * with PME for charge-charge Coulomb interactions.
+     */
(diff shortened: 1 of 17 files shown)


lu1and10:
Placeholder pending cleanup.

PR #5322, opened 2026-06-20
Add computeStressTensor() to MonteCarloFlexibleBarostat.

### LINKED ISSUE #5315: Incorrect pressure tensor from MonteCarloFlexibleBarostat.computeCurrentPressure() for triclinic boxes (inconsistent finite-difference deformation)
# MonteCarloFlexibleBarostat.computeCurrentPressure() is wrong for triclinic boxes

## Summary

`MonteCarloFlexibleBarostat::computeCurrentPressure()` returns an incorrect pressure tensor when the periodic box is **non-orthorhombic**.

This is separate from the volume/sign fix in #5248 / #5251; this is about the deformation used in `computePressureComponent`, which #5251 did not touch.

## Cause

In `openmmapi/src/MonteCarloFlexibleBarostatImpl.cpp`, `computePressureComponent` computes a normal component (`component < 3`) like this:

```cpp
Vec3 scale(1, 1, 1);
if (component < 3) {
    scale[component] = 1.0+delta;
    kernel.getAs<ApplyMonteCarloBarostatKernel>().scaleCoordinates(context, scale[0], scale[1], scale[2]);
}
newBox[scaleVec][scaleComponent] *= 1.0+delta;   // only ONE box element
```

For the XX component it scales every atom's x coordinate by `(1+delta)` but only scales the single box element `box[0][0] = a_x`. A consistent uniaxial x-strain must scale the x component of all box vectors (`a_x`, `b_x`, `c_x`). When `b_x` or `c_x` is nonzero (tilted box), the box deformation no longer matches the coordinate scaling, so the finite difference is not the
stress.

The shear components (`component >= 3`) are also affected: they perturb a single off-diagonal box element and *apply no coordinate scaling* at all, so they are not a strain either.

## Reproducing

A plain Lennard-Jones system, velocities set to zero so only the potential part remains. The reference is a textbook uniaxial strain finite difference (deform atoms and every box vector by the same gradient), using the same `delta = 1e-3` OpenMM uses.

```python
# reproducer.py 
```

Output:

```
  c_x/a      xx omm      xx ref  xx err %   yy diff   zz diff
   0.00       355.9       355.9      -0.0      0.00      0.00
   0.15       328.2      2243.2     -85.4     -0.00      0.00
   0.30        85.5      2579.5     -96.7      0.00     -0.00
   0.45     -1357.2      -585.7     131.7     -0.00      0.00
```

Orthorhombic (`c_x = 0`) is exact. As the tilt grows, `xx` diverges by ~85-130%, while `yy`/`zz` remain exact to rounding.

## Proposed fix

For the normal components, scale the matching Cartesian component of every box vector, not just the diagonal element:

```cpp
if (component < 3) {
    scale[component] = 1.0+delta;
    kernel.getAs<ApplyMonteCarloBarostatKernel>().scaleCoordinates(context, scale[0], scale[1], scale[2]);
    for (int j = component; j < 3; j++)
        newBox[j][component] *= 1.0+delta;          // a_x, b_x, c_x for XX
}
else
    newBox[scaleVec][scaleComponent] *= 1.0+delta;
```

(and the analogous change in the second-energy block). With this change the `xx`/`yy`/`zz` components match the strain-FD reference exactly.

The shear components need a true shear of the coordinates (e.g. `x += g*z`), which the current `scaleCoordinates(sx, sy, sz)` kernel cannot express, so a complete fix for the off-diagonal terms is a larger change.

## Environment

- OpenMM 8.5.1 (also present on `master`)
- Reference platform (platform independent; the FD logic is in the API layer)

[reproducer.py](https://github.com/user-attachments/files/28875832/reproducer.py)

Vivek-R-S:
This is a new finite-difference stress-tensor API that applies a consistent strain to both the atomic coordinates and box vectors, while leaving computeCurrentPressure() unchanged. It also adds a triclinic regression test, Python unit metadata, and user-guide/API documentation.

Closes #5315

### DIFF
--- a/docs-source/usersguide/theory/02_standard_forces.rst
+++ b/docs-source/usersguide/theory/02_standard_forces.rst
@@ -909,22 +909,26 @@ The six non-zero elements are as follows.
 * The elements below the diagonal reflect the pressure acting to change the box
   angles.  In equilibrium they should average to zero.
 
-computeStressTensor()
-^^^^^^^^^^^^^^^^^^^^^
+MonteCarloFlexibleBarostat can also compute the stress tensor, which measures the
+mechanical stress in a material.\ :cite:`Thompson2009`  It is defined as the
+derivative of the energy with respect to a uniform strain of the system: if a
+small strain :math:`\epsilon_{jk}` is applied to both the atomic coordinates and
+the periodic box, the stress is
 
-MonteCarloFlexibleBarostat also provides computeStressTensor(), which estimates
-the stress tensor from a consistent strain finite difference.  For each of the
-six independent tensor components, a small deformation gradient is applied to
-both the atomic coordinates and all three box vectors.  The potential energy
-is evaluated at positive and negative strain, and the corresponding stress
-component is obtained from the central difference.
+.. math::
 
-The method returns the six independent elements in the order
-(XX, YY, ZZ, XY, XZ, YZ), in bar.  Tension is positive.  By default only the
-configurational (potential energy) contribution is included.  If includeKinetic
-is set to true, the kinetic contribution is added as well.  When
-getScaleMoleculesAsRigid() is true, coordinates are deformed by moving each
-molecule as a rigid unit.
+   \sigma_{jk} = -\frac{1}{V} \frac{\partial E}{\partial \epsilon_{jk}}
+
+where :math:`V` is the volume.  The stress tensor is symmetric, so it has six
+independent elements: three normal components that describe stretching or
+compression along each axis, and three shear components that describe the
+response to a shape change.  A positive value corresponds to tension.
+
+The stress tensor is the quantity needed to study the mechanical response of a
+material.  For example, applying a known deformation to a crystal and measuring
+the resulting stress gives the elastic constants, and monitoring the stress
+during a simulation indicates whether a structure is under residual tension or
+compression.
 
 CMMotionRemover
 ***************
--- a/docs-source/usersguide/theory/02_standard_forces.rst
+++ b/docs-source/usersguide/theory/02_standard_forces.rst
@@ -909,6 +909,23 @@ The six non-zero elements are as follows.
 * The elements below the diagonal reflect the pressure acting to change the box
   angles.  In equilibrium they should average to zero.
 
+computeStressTensor()
+^^^^^^^^^^^^^^^^^^^^^
+
+MonteCarloFlexibleBarostat also provides computeStressTensor(), which estimates
+the stress tensor from a consistent strain finite difference.  For each of the
+six independent tensor components, a small deformation gradient is applied to
+both the atomic coordinates and all three box vectors.  The potential energy
+is evaluated at positive and negative strain, and the corresponding stress
+component is obtained from the central difference.
+
+The method returns the six independent elements in the order
+(XX, YY, ZZ, XY, XZ, YZ), in bar.  Tension is positive.  By default only the
+configurational (potential energy) contribution is included.  If includeKinetic
+is set to true, the kinetic contribution is added as well.  When
+getScaleMoleculesAsRigid() is true, coordinates are deformed by moving each
+molecule as a rigid unit.
+
 CMMotionRemover
 ***************
 
--- a/openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
+++ b/openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
@@ -190,6 +190,23 @@ public:
      *                   pressure tensor in the order (XX, YY, ZZ, XY, XZ, YZ)
      */
     void computeCurrentPressure(Context& context, std::vector<double>& pressure) const;
+    /**
+     * Compute the instantaneous stress tensor of a system to which this barostat is applied.
+     * The stress is estimated by a consistent strain finite difference: for each tensor
+     * component, a small deformation gradient is applied to both the atomic coordinates
+     * and all box vectors, and the potential energy is differentiated with respect to
+     * the strain.
+     *
+     * The returned values are in bar, with tension positive.  When includeKinetic is
+     * true, the kinetic contribution is included as well.  When getScaleMoleculesAsRigid()
+     * is true, coordinates are deformed by moving each molecule as a rigid unit.
+     *
+     * @param context         the Context for which to compute the stress tensor
+     * @param stress          on exit, the six independent elements in the order
+     *                        (XX, YY, ZZ, XY, XZ, YZ), in bar
+     * @param includeKinetic  if true, include the kinetic contribution to the stress tensor
+     */
+    void computeStressTensor(Context& context, std::vector<double>& stress, bool includeKinetic=false) const;
 protected:
     ForceImpl* createImpl() const;
 private:
--- a/openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
+++ b/openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
@@ -198,7 +198,7 @@ public:
      * the strain.
      *
      * The returned values are in bar, with tension positive.  When includeKinetic is
-     * true, the kinetic contribution is included as well.  When getScaleMoleculesAsRigid()
+     * true, the kinetic energy contribution is included as well.  When getScaleMoleculesAsRigid()
      * is true, coordinates are deformed by moving each molecule as a rigid unit.
      *
      * @param context         the Context for which to compute the stress tensor
--- a/platforms/common/include/openmm/common/CommonKernels.h
+++ b/platforms/common/include/openmm/common/CommonKernels.h
@@ -1359,7 +1359,8 @@ public:
      * @param scaleY     the scale factor by which to multiply particle y-coordinate
      * @param scaleZ     the scale factor by which to multiply particle z-coordinate
      */
-    void scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ);
+    void scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ,
+                          double scaleXY=0, double scaleXZ=0, double scaleYZ=0);
     /**
      * Reject the most recent Monte Carlo step, restoring the particle positions to where they were when
      * saveCoordinates() was last called.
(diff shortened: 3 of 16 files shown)


docs-source/usersguide/theory/02_standard_forces.rst
```diff
@@ -909,6 +909,23 @@ The six non-zero elements are as follows.
 * The elements below the diagonal reflect the pressure acting to change the box
   angles.  In equilibrium they should average to zero.
 
+computeStressTensor()
+^^^^^^^^^^^^^^^^^^^^^
```
peastman: No need for an extra header here.

What you've written is API documentation, not theory documentation.  We don't need to list method or argument names here.  Instead we want to give information about what it's for.  Something like this:

"MonteCarloFlexibleBarostat can also compute the stress tensor, defined as..."

then give the formula that defines the stress tensor and say what its significance is and what it's used for.

openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
```diff
@@ -190,6 +190,23 @@ class OPENMM_EXPORT MonteCarloFlexibleBarostat : public Force {
...
     void computeCurrentPressure(Context& context, std::vector<double>& pressure) const;
+    /**
+     * Compute the instantaneous stress tensor of a system to which this barostat is applied.
+     * The stress is estimated by a consistent strain finite difference: for each tensor
+     * component, a small deformation gradient is applied to both the atomic coordinates
+     * and all box vectors, and the potential energy is differentiated with respect to
+     * the strain.
+     *
+     * The returned values are in bar, with tension positive.  When includeKinetic is
+     * true, the kinetic contribution is included as well.  When getScaleMoleculesAsRigid()
```
peastman: I think you mean "kinetic energy contribution"?

peastman (commented):
Thanks!  This looks like a good start.  I have a few suggestions for how it could be improved.

The way you've implemented it is very slow: download the coordinates from the GPU to the host, scale them on the CPU in a single thread, and upload them back to the GPU.  Could you restructure it in terms of calls to `scaleCoordinates()`?  That will do it much more efficiently.

Your test case just reproduces the same computation you implemented in `computeStressComponent()`.  It tells us if the two implementations produce the same result, but it doesn't tell us if you implemented the correct calculation.  Are there any simple cases for which the correct answer is known analytically?  That would give an independent check.

Vivek-R-S:
> The way you've implemented it is very slow: download the coordinates from the GPU to the host, scale them on the CPU in a single thread, and upload them back to the GPU. Could you restructure it in terms of calls to `scaleCoordinates()`? That will do it much more efficiently.

Thank you for your comments! If I understand correctly, `scaleCoordinates` kernel applies independent per-axis scaling (`x' = scaleX·x`, etc.), so it can handle the three normal components (XX, YY, ZZ). However, the shear components require mixing coordinates e.g. XY strain needs `x += δ·y`, which can't be expressed as a combination of `(scaleX, scaleY, scaleZ)`. So the `setPositions()` path can't be fully avoided without a new kernel.

I can refactor XX/YY/ZZ to use `scaleCoordinates()` and keep `setPositions()` only for the three shear components. I'll also add a note to the user guide warning that `computeStressTensor()` is not cheap and should not be called frequently inside a tight MD loop. What do you think?

Vivek-R-S:
> Your test case just reproduces the same computation you implemented in `computeStressComponent()`. It tells us if the two implementations produce the same result, but it doesn't tell us if you implemented the correct calculation. Are there any simple cases for which the correct answer is known analytically? That would give an independent check.

I can add a test with two particles connected by a harmonic bond, where the analytic stress tensor is known:

$$\sigma_{ij} = \frac{k(|\mathbf{r}| - r_0)\, r_i\, r_j}{|\mathbf{r}|\, V}$$

I've already verified locally that `computeStressTensor()` matches this for a bond tilted in all three directions (so all six components are nonzero). I can add it as a test function alongside the existing triclinic test (or should I remove it?).

openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
Vivek-R-S: > I think you mean "kinetic energy contribution"?

Yes! I will make the correction :)

docs-source/usersguide/theory/02_standard_forces.rst
Vivek-R-S: > No need for an extra header here.
> 
> What you've written is API documentation, not theory documentation. We don't need to list method or argument names here. Instead we want to give information about what it's for. Something like this:
> 
> "MonteCarloFlexibleBarostat can also compute the stress tensor, defined as..."
> 
> then give the formula that defines the stress tensor and say what its significance is and what it's used for.

Makes sense, I will make the changes in the next commit :)

peastman:
> So the setPositions() path can't be fully avoided without a new kernel.

Let's add a new kernel then.  `scaleCoordinates()` is a very simple kernel.  Adding another version that applies shear as well should be easy.  Or we could just add a few more arguments to `scaleCoordinates()` for specifying the shear.  Let me know if you have questions about how to do it.

> I can add it as a test function alongside the existing triclinic test

That sounds perfect.

peastman:
One other test that would be useful is to create identical systems in OpenMM and another program that computes the stress tensor such as LAMMPS, and check that they produce identical results.  The test case doesn't need to invoke the other program.  It can just hardcode the expected values, and include a comment describing where they came from.

Vivek-R-S:
> Or we could just add a few more arguments to `scaleCoordinates()` for specifying the shear.

Done. extended scaleCoordinates() with optional shear factors (defaulting to zero, so existing barostats are unchanged) and computeStressComponent() now uses it for all six components.

> One other test that would be useful is to create identical systems in OpenMM and another program that computes the stress tensor such as LAMMPS, and check that they produce identical results.

Added testStressTensorAgainstLammps(), a 6-atom LJ system checked against LAMMPS virial stress (hardcoded values). [Here](https://github.com/user-attachments/files/29265902/in.dummy_lj_stress.txt) is the lammps input script for reference:  Also added the analytic two-particle harmonic bond test.

docs-source/usersguide/theory/02_standard_forces.rst
Vivek-R-S: Rewritten in the latest commit. Added the defining formula and reference.

openmmapi/include/openmm/MonteCarloFlexibleBarostat.h
Vivek-R-S: Fixed in the latest commit.

Vivek-R-S:
Pretty sure the CUDA/OpenCL CI failure in TestMonteCarloFlexibleBarostatX (the triclinic stress test) in the [689559d](https://github.com/openmm/openmm/pull/5322/commits/689559dec5bbacc218358dfa4ac820f74b948560) commit was a GPU atom reordering issue, the same underlying bug that @epretti just fixed at the kernel level in #5326.

The latest commit sidesteps this by deforming the coordinates continuously instead of saving/restoring around each evaluation, mirroring computePressureComponent(), and works independently of #5326

platforms/common/include/openmm/common/CommonKernels.h
```diff
@@ -1359,7 +1359,8 @@ class CommonApplyMonteCarloBarostatKernel : public ApplyMonteCarloBarostatKernel
      * @param scaleY     the scale factor by which to multiply particle y-coordinate
      * @param scaleZ     the scale factor by which to multiply particle z-coordinate
      */
-    void scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ);
+    void scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ,
+                          double scaleXY=0, double scaleXZ=0, double scaleYZ=0);
```
peastman, line 1366: It's missing `@param` tags for the new arguments.

platforms/reference/include/ReferenceKernels.h
```diff
@@ -1881,7 +1881,8 @@ class ReferenceApplyMonteCarloBarostatKernel : public ApplyMonteCarloBarostatKer
      * @param scaleY     the scale factor by which to multiply particle y-coordinate
      * @param scaleZ     the scale factor by which to multiply particle z-coordinate
      */
-    void scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ);
+    void scaleCoordinates(ContextImpl& context, double scaleX, double scaleY, double scaleZ,
+                          double scaleXY=0, double scaleXZ=0, double scaleYZ=0);
```
peastman, line 1888: It's missing `@param` tags for the new arguments.

peastman (commented):
This looks really good!

I notice that your test cases only ever call it with `includeKinetic=false`.  When would you pass `true` for it?  We should either include a test for that option, or if it isn't generally used in practice, remove it.

peastman:
Any thoughts on the `includeKinetic` option?

peastman:
We're aiming to release the beta of OpenMM 8.6 next week.  It would be great if we could include this.  After that we'll be in feature freeze, so it would need to wait for 8.7.

If you don't expect to have time to do anything more, what about just removing the `includeKinetic` option?  We can always add it back in a future release once we have tests for it.

Vivek-R-S:
> We're aiming to release the beta of OpenMM 8.6 next week. It would be great if we could include this. After that we'll be in feature freeze, so it would need to wait for 8.7.

Thanks for letting me know! I had a thesis submission this Wednesday, so I couldn't get to this until now.

I've added a test for the includeKinetic option (testKineticStressTensorAgainstLammps). I think it's worth keeping. LAMMPS reports the pressure as the full virial + kinetic tensor. For gases or dilute fluids the kinetic part dominates so it's worth keeping it.

For my own use (configurational stress of solids for elastic constants) I only need includeKinetic=false, since at a fixed temperature the kinetic part is essentially a constant offset.

The test assigns nonzero velocities to the same 6-atom LJ system and checks computeStressTensor(..., includeKinetic=true) against the full LAMMPS pressure tensor (I scaled the velocities up so the kinetic term is comparable to the virial, otherwise it's a ~0.01% perturbation the tolerance wouldn't catch). 
[Heres's](https://github.com/user-attachments/files/29630774/in.lj_stress_kinetic.txt) the LAMMPS input used to generate the hardcoded values, for reference.

platforms/common/include/openmm/common/CommonKernels.h
Vivek-R-S: Fixed in the latest commit

platforms/reference/include/ReferenceKernels.h
Vivek-R-S: Fixed in the latest commit.

peastman:
That looks great.  Thanks for the new feature!

PR #4106, opened 2023-06-08
Attempt at fixing errors with barostat

peastman:
This is an attempt at fixing #4071.  It splits `saveCoordinates()` and `scaleCoordinates()` into two separate methods.

PR #5326, opened 2026-06-24
Don't overwrite atomsWereReordered in scaleCoordinates()

epretti:
Fixes an issue reported at https://github.com/openmm/openmm/discussions/5185#discussioncomment-17244825.  Barostats' pressure computations involve multiple calls to `scaleCoordinates()`, so it was possible for `restoreCoordinates()` to miss a reordering.

If you think it is worth adding test cases for the barostats that would have identified this issue, let me know.


peastman:
That code has been the source of a lot of bugs, so we should test it very carefully.  See in particular #4049, #4106, #4119, #4364, and #4954.  All of them are related to atom reordering happening in the middle of a barostat move.  I don't see any problems with your fix, but it wouldn't be the first time a fix for this situation looked correct but still had subtle errors!

epretti:
I've added tests that check that particle coordinates don't jump too far between steps.  For me, the GPU platform tests fail without the fix and pass with it.

Why does `MonteCarloMembraneBarostat` not have GPU platform tests?  Should it?

peastman:
All the barostats are implemented with the same kernel, ApplyMonteCarloBarostatKernel.  It doesn't need to be retested for every type of barostat.  If we check any one barostat on every platform, that's enough to test all the platform-specific code.

Some of the others are tested on multiple platforms just because some tests are fairly expensive, so we only want to run them if a GPU is available,

epretti:
I had added the test to three barostats because `computeCurrentPressure()` in `MonteCarloBarostatImpl`, `MonteCarloFlexibleBarostatImpl`, and `MonteCarloMembraneBarostatImpl` all call `saveCoordinates`, `scaleCoordinates`, and `restoreCoordinates` different numbers of times or in different orders.  But I'll add the test only for `MonteCarloBarostat` if that's preferable.

peastman:
Looks good.

I don't know what happened with the Windows test failure, but I don't think it's related to this.

epretti:
OK, I'll merge.

Looking at CI, our Windows build has been broken for a little while.  Are you working on it or planning to, or should I take a look?

peastman:
Take a look if you want.

PR #5330, opened 2026-06-26
Fix Windows CI

epretti:
CI for Windows has been broken since the [Visual Studio version was updated](https://github.com/actions/runner-images/issues/14017), so this updates the path to `vcvarsall.bat`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -382,7 +382,7 @@ jobs:
         env:
           CMAKE_FLAGS: ${{ matrix.CMAKE_FLAGS }}
         run: |
-          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
+          call "C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
           mkdir build
           cd build
           cmake -G "NMake Makefiles JOM" ^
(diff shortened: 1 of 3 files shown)


epretti:
The Windows run failed but it looked like a stochastic failure in one of the Python tests; the build was successful.  I'm running it again to see if it is reproducible.

Incidentally, after looking at a month of scheduled daily CI runs between May 9 to June 9 (before the Windows build broke completely), 25 of those runs failed.  15 of those failures included the Windows run failing.  8 of those Windows runs failed on the same test `TestExpandedEnsembleSampler::testTemperature`.  That is the same failure I saw the first time here.  That might just be by chance, but it is a lot of failed runs.

peastman:
I've observed that stochastic failures are more common on Windows than other platforms.  It's not just that test.  Other ones are also more likely to fail.  I would suspect that it's using a less accurate floating point mode, but it really shouldn't be.  If you don't specify a mode, and as far as I can tell we don't, the default is supposed to be `/fp:precise`.  I'm not sure what else could account for it.

epretti:
I think there was an actual bug here.  I found that the standard deviation of temperatures coming out after calling `context.setVelocitiesToTemperature()` was $\sqrt{3}$ too high on Windows.  It turned out that with MSVC, a line like this:
https://github.com/openmm/openmm/blob/1da3cbb0bfebe8447f1b7773352c26cfde56144a/openmmapi/src/Integrator.cpp#L107
causes the vector to get populated with three copies of the same value.  This is undefined or unspecified behavior, but GCC seems to compile it with the intended behavior.

PR #5331, opened 2026-06-26
Use standard OpenCL C++ header

### LINKED ISSUE #5327: Updated OpenCL C++ headers incompatible with some versions of OpenCL C headers
#5296 appears to have updated the OpenCL C++ headers distributed with OpenMM.  When trying to do a fresh rebuild of OpenMM on Linux, I encountered many errors of the form:

```
In file included from /.../openmm/platforms/opencl/./include/OpenCLArray.h:34,
                 from /.../openmm/platforms/opencl/src/OpenCLArray.cpp:25:
/.../openmm/platforms/opencl/src/opencl.hpp:1870:62: error: missing binary operator before token "("
 1870 | #if CL_KHR_COMMAND_BUFFER_EXTENSION_VERSION > CL_MAKE_VERSION(0, 9, 5)
      |                                                              ^
```

It appears version 2024.10.24 and later of the C++ headers assumes the existence of the `CL_MAKE_VERSION` macro.  However, this macro is an OpenCL 3.0 feature.  In version 2024.10.24 and later of the OpenCL *C* headers, it is always available regardless of the `CL_TARGET_OPENCL_VERSION`, but in version 2024.05.08 and earlier, it is gated behind `CL_TARGET_OPENCL_VERSION >= 300` or `CL_VERSION_3_0` being defined.  If the OpenCL C system headers are older (as they were in my Linux distribution), OpenMM will fail to build since it defines `CL_HPP_TARGET_OPENCL_VERSION` as `120`, causing the `CL_MAKE_VERSION` macro to be missing.

This doesn't happen to me on macOS; I think the OpenCL headers are so old that other feature macros that would activate the sections of the C++ headers containing `CL_MAKE_VERSION` just don't exist.

We could treat this as:
- Not a bug: users compiling from source should have to install the latest OpenCL C headers, [from source](https://github.com/KhronosGroup/OpenCL-Headers) if not available from their OS.
- OpenMM problem: we could distribute an older version of the C++ headers or otherwise work around this.
- OpenCL problem: we could report this to Khronos as a bug, although they only seem to say that the C++ headers are meant to work with all earlier versions of the OpenCL API, not necessarily the C headers specifically.

peastman:
Fixes #5327.  `opencl.hpp` is a standard header on all modern OpenCL distributions.  Apple still uses an ancient version that doesn't include it.  That's why we bundle a copy.  On all other platforms, we should use the standard one instead of the bundled one to avoid compatibility issues.

### DIFF
--- a/platforms/opencl/CMakeLists.txt
+++ b/platforms/opencl/CMakeLists.txt
@@ -17,6 +17,12 @@ if(BUILD_TESTING AND OPENMM_BUILD_OPENCL_TESTS)
     SUBDIRS (tests)
 endif(BUILD_TESTING AND OPENMM_BUILD_OPENCL_TESTS)
 
+find_file(OPENCL_HPP_PATH opencl.hpp "${OPENCL_INCLUDE_DIR}/CL")
+mark_as_advanced(OPENCL_HPP_PATH)
+if(OPENCL_HPP_PATH)
+    ADD_DEFINITIONS(-DHAS_OPENCL_HPP)
+endif()
+
 # The source is organized into subdirectories, but we handle them all from
 # this CMakeLists file rather than letting CMake visit them as SUBDIRS.
 SET(OPENMM_SOURCE_SUBDIRS . ../common)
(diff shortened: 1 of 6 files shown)


peastman:
Because of course nothing ever works...

NVIDIA's OpenCL claims to be 3.0.  But they still include an ancient `cl.hpp` header designed for 1.x instead of the modern `opencl.hpp` header that's supposed to be included with 3.0.

PR #5340, opened 2026-07-08
Use more threads for CI

peastman:
The CI scripts specify two threads/processes for compiling and running tests.  At some point, Github updated their [configurations](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) so most free runners have four cores instead of two.  This updates the scripts accordingly.  Hopefully it should make builds run faster.


peastman:
Using more processes for the Python tests made it slower, so I removed that part.  This ended up only making a small difference, but most builds are at least a little bit faster.

PR #5341, opened 2026-07-08
Fix replicaexchangereporter reporter format

jthorton:
Fixes a possible typo when storing the trajectory format during the construction of the reporter.


peastman:
Thanks!

PR #5344, opened 2026-07-09
Implemented fix for new residue-aware cmaps added in GMX2026

mattiabondanza:
In new version of Gromax (2026.1) with some new forcefields (amber19sb) contact maps are defined by atomtypes+residue :
```filename="share/gromacs/top/amber19sb.ff/cmap.itp"
[ cmaptypes ]
C-* N-GLY XC-GLY C-GLY N-* 1 24 24\
...
```
With the original code, this causes missing parameters error when such topology files are used. With this patch, during parameter assignment, we search for dashes in cmap types definitions. If they are present, then a new mode to perform residue name aware comparison between the required parameter and the available parameters list is enabled. Since the residue names used here are the original ones (so `"HIP"` is different from `"HID"` etc.), original residue name are stored in residues objects and used in this context only.

### DIFF
--- a/wrappers/python/openmm/app/gromacstopfile.py
+++ b/wrappers/python/openmm/app/gromacstopfile.py
@@ -613,6 +613,7 @@ class GromacsTopFile(object):
                         if resName in PDBFile._residueNameReplacements:
                             resName = PDBFile._residueNameReplacements[resName]
                         r = top.addResidue(resName, c)
+                        r._original_name = fields[3]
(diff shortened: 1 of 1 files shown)


peastman:
Thanks!  Is there a spec for the new syntax describing how it should be interpreted?

Adding new fields to a standard class feels wrong to me.  I think it would be better for the GromacsTopFile to store that information in its own fields.

We'll need a test case to verify the new feature works correctly.

PR #5346, opened 2026-07-10
Fixed a GPU thread synchronization bug

### LINKED ISSUE #5328: Non-reproducible forces and energies when using PME and CustomCVForce on GPU platforms (CUDA and OpenCL)
When calculating energies and/or forces, they are non-reproducible, with large random deviations when CUDA is used in combination with PME and a CustomCVForce. This affects at least openMM versions 8.2, 8.4 and 8.5.2 (tested).

To re-create the problem, I re-calculated the energies of a single snapshot of a simulation. All files necessary to reproduce the bug are attached. Note that the bug happens rather sporadically, so when the values seem fine, re-run the script a few times, or use a higher trial number. In my tests, even a trial number of 250 did not lead to reproducible results, sometimes triggering the large energy spread and force RMS deviation, sometimes not. Although it did consistently trigger only on simulations with a CustomCVForce. OpenCL simulations also seem to be affected, at least in version 8.4 and 8.5.2 (I did not observe this in 8.2, but as the issue is happening sporadically, I might just have missed it).

Regarding CUDA, double precision seems to be affected the most.

I suspect that the underlying reason has something to do with a CUDA race condition, but can not confirm this due to a lack of expertise in CUDA programming. However, setting `DisablePmeStream` to `True` will lead to reproducible forces and energies and mitigate the bug.

I think there is also a related issue in #5008, but maybe that has nothing to do with it.

Image
Image
Image
Image
Image
Image

[results_8.2.csv](https://github.com/user-attachments/files/29327191/results_8.2.csv)
[results_8.4.csv](https://github.com/user-attachments/files/29327190/results_8.4.csv)
[results_8.5.2.csv](https://github.com/user-attachments/files/29327188/results_8.5.2.csv)

Below is an AI generated description of my findings. Take the conclusions with a grain of salt, but maybe it can help with diagnosing:

# AI SUMMARY

# Nondeterministic all-groups energy/forces on GPU when a `CustomCVForce` is combined with a PME `NonbondedForce`

## Summary

On the **CUDA** and **OpenCL** platforms, `Context.getState(getEnergy=True, getForces=True)`
(all force groups) returns **different results on every call** for a fixed Context — same
positions, same box, same parameters, no integration in between — when the `System` contains
both a `CustomCVForce` and a PME `NonbondedForce`.

The same single-point energy varies run-to-run by **thousands to ~15,000 kJ/mol**, and per-atom
forces by **several kJ/mol/nm RMS** (with individual atoms far higher). Removing the
`CustomCVForce` makes the result bit-reproducible. The `CPU` and `Reference` platforms are
always reproducible. This looks like a race between the PME reciprocal-space stream and the
nested inner `Context` that a `CustomCVForce` evaluates mid-computation.

This is a correctness bug: it silently corrupts any all-groups energy/force evaluation, and
because `Integrator.step()` computes all-groups forces, it perturbs the integrated dynamics of
any PME simulation that uses a `CustomCVForce` (RMSD restraints, ATM-style setups, etc.) on a
GPU platform.

## Reproducer

`reproduce.py` (attached) builds **two systems from the same PDB** (amber14/ff14SB + tip3p, PME,
1 nm cutoff, HBonds constraints, ~62k atoms):

- **`plain`** — the amber14 system as-is (bonded + PME `NonbondedForce`).
- **`with_cv`** — *identical*, plus 4 `RMSDForce`-backed `CustomCVForce` restraints
  (`0.5*k*rmsd^2`, 80 atoms each, reference = a different snapshot), each in its own force group.

The `CustomCVForce` is the **only** difference between the two systems — there is deliberately
no barostat or any other stochastic element. For each platform/precision the all-groups energy
and forces of one fixed snapshot are evaluated **250 times** and the spread (max − min for
energy; max per-repeat RMS per-atom deviation for forces) is reported.

```
python reproduce.py --n-repeats 250 --out results.csv     # run once per OpenMM build/env
```

## Results

Energy spread = max − min over 250 repeats [kJ/mol]; Force = max per-repeat RMS per-atom
deviation [kJ/mol/nm]. `nopme` = CUDA property `DisablePmeStream='true'`.

### OpenMM 8.2

| platform | precision   | plain Eptp | **+CV Eptp** | plain F | **+CV F** |
|----------|-------------|-----------:|-------------:|--------:|----------:|
| CUDA     | single      |   3.9e-03  |     1.9e-03  | 4.2e-05 |  4.1e-05  |
| CUDA     | mixed       |   3.3e-04  | **2.9e+03**  | 4.1e-05 | **1.1e+00** |
| CUDA     | mixed+nopme |   2.8e-04  |     2.7e-04  | 3.8e-05 |  3.8e-05  |
| CUDA     | double      |   0.0      | **9.1e+03**  | ~0      | **1.1e+01** |
| CUDA     | double+nopme|   0.0      |     0.0      | ~0      |  ~0       |
| OpenCL   | single/mixed/double | 0.0 |     0.0      | ~0      |  ~0       |
| CPU      | double      |   2.4e-04  |     2.4e-04  | 1.9e-04 |  1.7e-04  |
| Reference| double      |   0.0      |     0.0      | ~0      |  ~0       |

### OpenMM 8.4

| platform | precision   | plain Eptp | **+CV Eptp** | plain F | **+CV F** |
|----------|-------------|-----------:|-------------:|--------:|----------:|
| CUDA     | single      |   9.8e-03  |     7.8e-03  | 4.0e-05 |  4.1e-05  |
| CUDA     | mixed       |   2.8e-04  |     3.2e-04  | 4.1e-05 |  4.2e-05  |
| CUDA     | double      |   0.0      | **1.3e+04**  | ~0      | **2.7e+00** |
| CUDA     | double+nopme|   0.0      |     0.0      | ~0      |  ~0       |
| OpenCL   | single      |   0.0      | **2.1e+03**  | 0.0     | **9.1e-01** |
| OpenCL   | mixed       |   0.0      | **4.5e+03**  | 0.0     | **1.3e+00** |
| OpenCL   | double      |   0.0      | **1.3e+04**  | ~0      | **1.2e+01** |
| CPU      | double      |   2.4e-04  |     2.4e-04  | 1.9e-04 |  1.9e-04  |
| Reference| double      |   0.0      |     0.0      | ~0      |  ~0       |

### OpenMM 8.5.2

| platform | precision   | plain Eptp | **+CV Eptp** | plain F | **+CV F** |
|----------|-------------|-----------:|-------------:|--------:|----------:|
| CUDA     | single      |   7.8e-03  | **2.8e+00**  | 4.1e-05 | **3.7e-03** |
| CUDA     | mixed       |   2.6e-04  |     3.1e-04  | 4.2e-05 |  3.9e-05  |
| CUDA     | double      |   0.0      | **1.5e+04**  | ~0      | **1.8e+00** |
| CUDA     | double+nopme|   0.0      |     0.0      | ~0      |  ~0       |
| OpenCL   | single      |   0.0      | **3.3e+03**  | 0.0     | **1.2e+00** |
| OpenCL   | mixed       |   0.0      | **4.9e+00**  | 0.0     | **1.8e-02** |
| OpenCL   | double      |   0.0      | **1.2e+04**  | ~0      | **1.2e+01** |
| CPU      | double      |   2.4e-04  |     2.4e-04  | 1.9e-04 |  1.9e-04  |
| Reference| double      |   0.0      |     0.0      | ~0      |  ~0       |

(Bold = nondeterministic, i.e. `+CV` spread ≫ the ~1e-4 floating-point floor while `plain` is
at the floor.)

## What the results show

1. **The `CustomCVForce` is necessary.** In every row the `plain` system is reproducible to the
   floating-point floor (≤ ~1e-2 at single precision, exactly 0.0 at double on GPU). Only the
   `with_cv` system breaks. The two systems are otherwise bit-identical, so nothing physical
   distinguishes them — the variation is a pure software bug.

2. **It is not a rare outlier.** For CUDA double + CV, essentially *all* 250 repeats deviate
   (247–250 of 250 land > 1 kJ/mol from the median), spanning ~9,000–15,000 kJ/mol depending on
   version. The "correct" value is the floor reached by CPU/Reference/`nopme`; the GPU all-groups
   path lands somewhere random above it almost every call.

3. **It affects both GPU platforms.** CUDA is affected on all three versions. **OpenCL is also
   affected on 8.4 and 8.5.2** (all precisions), though it was clean on 8.2.

4. **The triggering precision is version-dependent** (consistent with a timing race whose outcome
   depends on kernel scheduling): CUDA *double* breaks on all three versions; CUDA *mixed* breaks
   only on 8.2; CUDA *single* breaks only on 8.5.2. The double-precision CUDA path is the most
   reliably broken across versions.

5. **`DisablePmeStream='true'` removes it on CUDA.** Every CUDA case that breaks under default
   properties becomes reproducible (back to the floor) with `DisablePmeStream='true'`. This
   strongly localizes the cause to the PME reciprocal-space stream. (There is no OpenCL
   equivalent property to test the same way.)

## Suspected mechanism

PME reciprocal-space work runs on a separate, non-blocking GPU stream and is merged back into the
main stream after the rest of the nonbonded computation. A `CustomCVForce` is unusual in that its
`execute()` runs a *whole separate nested `Context`* in the middle of the outer force evaluation,
with its own kernel launches and host-side synchronizations. That re-entrancy appears to
interleave with the PME stream merge such that the all-groups energy/force reduction can read the
PME contribution before it has been written, or mix it inconsistently — producing a different
result each call. `DisablePmeStream='true'` (which keeps PME on the main stream) avoids the race,
which is consistent with this explanation. Plain PME systems and per-force-group queries (which
never evaluate PME and the `CustomCVForce` together in one all-groups reduction) are unaffected,
which is why this is not seen everywhere.

## Impact

- **Single-point energies/forces:** any `getState` over all groups is silently wrong run-to-run.
- **Dynamics:** `Integrator.step()` computes all-groups forces every step, so trajectories of a
  PME + `CustomCVForce` system on a GPU are integrated on intermittently corrupted forces (RMS
  ~kJ/mol/nm per step, individual atoms much larger). The thermostat keeps the run stable, but
  the sampled ensemble is perturbed by an uncontrolled stochastic force with no
  fluctuation–dissipation relationship.
- **Workaround for users today:** set the CUDA platform property `DisablePmeStream='true'`
  (confirmed above), or compute the total energy as the sum of per-force-group `getState` calls,
  which avoids evaluating PME and the `CustomCVForce` in the same all-groups reduction.

## Environment

- OpenMM 8.2, 8.4, 8.5.2 (conda-forge builds), reproduced on the same machine/GPU.
- System: ~62k atoms, amber14/ff14SB + tip3p, PME (1 nm cutoff), HBonds constraints.
- Attached: `reproduce.py` (self-contained, needs `final.pdb` + `solvated.pdb`), `plot.py`,
  `results/results_*.csv`.

## How to reproduce

```bash
# one run per OpenMM build (e.g. one conda env per version)
python reproduce.py --n-repeats 250 --out results/results_<version>.csv
python plot.py        # energy scatter + spread bar charts per version
```

Expected: `plain` reproducible everywhere; `with_cv` nondeterministic on CUDA (and OpenCL on
8.4/8.5.2); `DisablePmeStream='true'` restores reproducibility on CUDA.

# Reproducer script and files

[plot.py](https://github.com/user-attachments/files/29332749/plot.py)
[reproduce.py](https://github.com/user-attachments/files/29332748/reproduce.py)
[pdb_files.zip](https://github.com/user-attachments/files/29332800/pdb_files.zip)

Note that reproduce.py also contains additional systems to dissect whether it's actually the CustomCVForce, or the RMSDForce. These are not included in the data presented above, as I did not run them yet.

All energy evaluations were done on a NVIDIA L40S GPU.

peastman:
Fixes #5328.  This was a subtle threading issue that could cause incorrect results in RMSDForce, RGForce, and OrientationRestraintForce.  But usually it didn't.  It only seemed to occur when the GPU was trying to run multiple sets of kernels on different streams at the same time.  I really don't know why!

### DIFF
--- a/platforms/common/src/kernels/orientationRestraintForce.cc
+++ b/platforms/common/src/kernels/orientationRestraintForce.cc
@@ -11,6 +11,7 @@ DEVICE real reduceValue(real value, LOCAL_ARG volatile real* temp) {
             temp[thread] = temp[thread] + temp[thread+step];
         SYNC_WARPS;
     }
+    SYNC_THREADS;
     for (int step = 32; step < LOCAL_SIZE; step *= 2) {
         if (thread+step < LOCAL_SIZE && thread%(2*step) == 0)
             temp[thread] = temp[thread] + temp[thread+step];
--- a/platforms/common/src/kernels/rg.cc
+++ b/platforms/common/src/kernels/rg.cc
@@ -11,6 +11,7 @@ DEVICE real reduceValue(real value, LOCAL_ARG volatile real* temp) {
             temp[thread] = temp[thread] + temp[thread+step];
         SYNC_WARPS;
     }
+    SYNC_THREADS;
     for (int step = 32; step < LOCAL_SIZE; step *= 2) {
         if (thread+step < LOCAL_SIZE && thread%(2*step) == 0)
             temp[thread] = temp[thread] + temp[thread+step];
--- a/platforms/common/src/kernels/rmsd.cc
+++ b/platforms/common/src/kernels/rmsd.cc
@@ -14,6 +14,7 @@ DEVICE real reduceValue(real value, LOCAL_ARG volatile real* temp) {
             temp[thread] = temp[thread] + temp[thread+step];
         SYNC_WARPS;
     }
+    SYNC_THREADS;
     for (int step = 32; step < LOCAL_SIZE; step *= 2) {
         if (thread+step < LOCAL_SIZE && thread%(2*step) == 0)
             temp[thread] = temp[thread] + temp[thread+step];


ex-rzr:
I may have a plausible explanation but it's just my guess.

I use AMD terminology but it's similar on Nvidia: CU - Compute Unit (SM - Streaming Multiprocessor on Nvidia), EU - Execution Unit or SIMD.
The vertical axes is time.

The left side is when there is only one stream so only one kernel `K0` like `computeCorrelationMatrix` is executed on each CU. The kernel uses a small amount of resources (register, shared memory) so each CU can process multiple thread blocks.
Even without the synchronization you added all warps (`W0..W3`) of different thread blocks of this kernel (`K0 B0` and `K0 B1`) **usually** work with the same pace, i.e. each operation take a similar number of cycles. So the warps act like they are synchronized even though they are not (by a lucky chance).

The right side is what may happen when there is another stream that can launch a different kernel (`K1`) on the same CU. Let's assume that some warps (`K1 B0 W2..W3`) do more work than others (e.g. load more data from global memory). This breaks "synchronicity" of the kernel `K0`'s warps so they execute different instructions at next cycles and see different states of shared memory.

peastman:
Sounds plausible?  I always assumed that each CU would only execute one kernel at a time.  Multiple streams would allow different CUs to execute different kernels.  But perhaps that assumption is wrong?

ex-rzr:
As far as I know, there is no detailed official information about scheduling for both Nvidia and AMD.
But some conclusions can be done from experiments in these papers:
 * Exploring AMD GPU Scheduling Details by Experimenting With “Worst Practices”: https://www.cs.unc.edu/~anderson/papers/rtns21b.pdf  
see page, for example, page 8
 * GPU Scheduling on the NVIDIA TX2: Hidden Details Revealed: https://www.cs.unc.edu/~anderson/papers/rtss17c.pdf  
see page 5

In general, they are similar, i.e. SM/CU can execute thread blocks of different kernels. The main difference is "cutting ahead" on AMD described in the first paper: when one kernel requires fewer resources (threads, registers, shared memory) it will not free enough resources on CU when finished for launching another kernel requiring more resources.

PR #5348, opened 2026-07-13
Fix undefined behavior from operator[]

vedranmiletic:
Fedora builds packages with [libstdc++'s _GLIBCXX_ASSERTIONS](https://gcc.gnu.org/wiki/LibstdcxxDebugMode) turned on. In order for OpenMM tests to pass on Fedora after building, we have to avoid the undefined behavior.

### DIFF
--- a/platforms/cpu/src/CpuConstantPotentialForce.cpp
+++ b/platforms/cpu/src/CpuConstantPotentialForce.cpp
@@ -246,7 +246,7 @@ CpuConstantPotentialCGSolver::CpuConstantPotentialCGSolver(int numParticles,int 
 
 void CpuConstantPotentialCGSolver::solveImpl(CpuConstantPotentialForce& conp, ThreadPool& threads, Kernel& pmeKernel) {
     ensureValid(conp, threads, pmeKernel);
-    
+
     double offset;
     float error, paramScale, alpha, beta;
     const float errorTarget = conp.cgErrorTol * conp.cgErrorTol * numElectrodeParticles;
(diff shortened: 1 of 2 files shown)


platforms/cpu/src/CpuConstantPotentialForce.cpp
```diff
@@ -274,8 +274,20 @@ void CpuConstantPotentialCGSolver::solveImpl(CpuConstantPotentialForce& conp, Th
...
+    // this aborts under -D_GLIBCXX_ASSERTIONS with numElectrodeParticles ==
+    // chargeDerivatives.size() == 2, i.e. a perfectly ordinary, non-empty
+    // case). conp.chargeDerivatives can also be legitimately empty
+    // (numElectrodeParticles == 0), which breaks &vec[0] the same way. Use
+    // .data(), which is well-defined for pointer arithmetic up to and
+    // including size(), instead of operator[]/&vec[0]/&vec[n]. The same
+    // pattern recurs a few more times below for the same reason.
     conp.getDerivatives(threads, pmeKernel);
-    grad.assign(&conp.chargeDerivatives[0], &conp.chargeDerivatives[numElectrodeParticles]);
+    grad.assign(conp.chargeDerivatives.data(), conp.chargeDerivatives.data() + numElectrodeParticles);
```
peastman: I think this could be written more cleanly as

```c++
grad.assign(conp.chargeDerivatives.begin(), conp.chargeDerivatives.end());
```

And similarly in the other places.

peastman (commented):
Thanks!

Could you remove the comments?  They don't explain what the code does, just why you're modifying it in particular ways.  They're only relevant to this PR.  To someone reading the code in the future, they would be confusing and not helpful.

platforms/cpu/src/CpuConstantPotentialForce.cpp
vedranmiletic: Did that and added the assert that size is the same as the offset used originally.

peastman:
Looks good.  Thanks!

vedranmiletic:
> Looks good. Thanks!

Thank you!

I have several pull requests more from trying to make a Fedora package and will submit them for review in the coming days.

PR #5353, opened 2026-07-15
Link librt for shm_open and shm_unlink in virtmem.cpp

prlw1:
Currently librt is only linked for Linux, and no other unices.


prlw1:
librt is needed for shm_open on more OSes than just Linux

It may be that one can link to librt for all of  UNIX AND NOT APPLE, but just in case, I test for librt's existence in the same way as currently done for libdl.

With the change, on NetBSD which has librt but not libdl, I correctly have
```
/tmp/openmm/lib/libOpenMM.so:
        -lrt.1 => /usr/lib/librt.so.1
        -lc.12 => /usr/lib/libc.so.12
        -lstdc++.9 => /usr/lib/libstdc++.so.9
        -lm.0 => /usr/lib/libm.so.0
        -lgcc_s.1 => /usr/lib/libgcc_s.so.1
        -lpthread.1 => /usr/lib/libpthread.so.1
```

peastman:
Looks good.  Thank you!

PR #5355, opened 2026-07-16
Use GNUInstallDirs instead of hardcoding "lib" for install paths

vedranmiletic:
Every CMakeLists.txt that installs a shared library or plugin (top- level, all platforms/*/sharedTarget and staticTarget dirs, and every plugins/*/CMakeLists.txt and plugins/*/platforms/*/CMakeLists.txt) hardcoded "/lib" or "/lib/plugins" as the install destination, as did wrappers/python/pysetup.cmake.in for OPENMM_LIB_PATH. On any system using the lib64 convention for 64-bit libraries (Fedora/RHEL and most other mainstream distro packaging, which is what CMake's own GNUInstallDirs module is for for exactly this purpose) that silently installs everything into the wrong, non-multilib-safe directory.

Add include(GNUInstallDirs) once at the top level and reference ${CMAKE_INSTALL_LIBDIR} everywhere "lib" was hardcoded, instead of a literal string. This changes nothing on platforms where CMAKE_INSTALL_LIBDIR defaults to "lib" (the vast majority), and correctly switches to "lib64" wherever GNUInstallDirs determines that convention applies -- confirmed with a from-scratch build/install on this exact kind of system, where CMAKE_INSTALL_LIBDIR is auto-detected as "lib64" with no extra configuration needed: all four top-level shared libraries and every dlopen()'d platform plugin correctly land under lib64/ (and lib64/plugins/), OPENMM_LIB_PATH is substituted correctly for the Python wrapper build, and the resulting openmm._openmm extension links against and gets the correct lib64 RUNPATH.


peastman:
In most of these places, `lib` does not refer to a standard system directory.  It's the name of a directory inside OpenMM's installation area, which is usually inside a Python environment.  This change would likely break our pip and conda packages.

vedranmiletic:
Ouch.

Well, this could work as a fallback:

```
if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
  set(CMAKE_INSTALL_LIBDIR lib CACHE PATH "Object code libraries")
endif()
```

And then `include(GNUInstallDirs)` could be made optional. Would that be acceptable?

peastman:
It's not about how you set the value.  It's that you're using it in places that aren't appropriate.  What name an OS gives to its standard directories has no connection to what name we give to subdirectories inside the OpenMM installation.

PR #5356, opened 2026-07-16
Remove stray shebangs and normalize CRLF line endings in unit/ modules

vedranmiletic:
These openmm.unit and amber_file_parser modules are library code, never meant to be run directly, but eleven of them carried a "#!/bin/env python" shebang despite shipping without the executable bit set. doctests.py additionally had Windows-style CRLF line endings throughout the file, unlike the rest of the codebase. Neither affects Python import/execution, but both are packaging-lint failures for any downstream (e.g. rpmlint's non-executable-script and wrong-script-end-of-line-encoding checks) and are simple to fix at the source.


peastman:
Looks good.  Thanks.

Issue #3796, opened 2022-09-30
Changes to support PyPI packages


peastman:
To be able to create PyPI packages, there are some issues we need to resolve which will probably involve changes to code or build scripts.  This issue is to discuss possible ways of solving them.

### CUDA

The biggest issue we need to consider is how to deal with CUDA.  Different CUDA releases are not binary compatible with each other (or only binary compatible over a narrow range of releases), so at runtime OpenMM needs the libraries from the particular toolkit version it was built against.  Each toolkit in turn requires particular driver versions, and users often can't control what driver version they have (for example on a cluster they don't administer).  That means it's helpful to build OpenMM against multiple CUDA versions.  We need a way for each user to get an OpenMM build and CUDA libraries that are compatible with each other and with their driver.

Conda handles this with two mechanisms: first by providing all the libraries in a `cudatoolkit` package, and second by allowing a package to have multiple builds that differ only in which version of that package they require.  At install time, the conda client tries (not always successfully) to automatically select a version of `cudatoolkit` that is compatible with the user's driver, and then installs the corresponding versions of other packages.

For installing with pip, NVIDIA provides their own repository with all the redistributable libraries.  You can install packages from it either by adding an extra URL to `requirements.txt`, or alternatively by first installing the `nvidia-pyindex` package.  See https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#pip-wheels-installation-linux for more information.

On the other hand, pip does not support having multiple builds that differ only in what CUDA version they're built against.  A given OpenMM version has to be built against a single CUDA version, which would presumably be the oldest (and therefore most compatible) one we support.  And that means any other package in the same environment that also uses CUDA (e.g. PyTorch)  also has to be built against the same version.

PyTorch works around this issue by providing their own repository with additional versions.  For example, if you type `pip install torch` you get one that's built against CUDA 10.2.  If you instead want a package built against CUDA 11.6 you type `pip install torch --extra-index-url https://download.pytorch.org/whl/cu116`.  We could consider doing something similar, though it's probably more trouble than it's worth.

Another possibility discussed in #3196 is that a single OpenMM package could contain multiple versions of the plugins built against different CUDA versions.  Within the `plugins` directory you would have `libOpenMMCUDA10.2.so`, `libOpenMMCUDA11.0.so`, etc.  At runtime, the one matching the available CUDA libraries would be loaded, and all the others would fail to load since the libraries they depended on couldn't be found.

### Plugins

[...]

PR #5357, opened 2026-07-16
Set VERSION/SOVERSION on the four installed shared libraries

vedranmiletic:
libOpenMM.so, libOpenMMAmoeba.so, libOpenMMDrude.so, and libOpenMMRPMD.so are installed directly into the library directory (unlike the CPU/PME/platform plugins, which are dlopen()'d from a private plugins/ directory and never linked against directly) but never had VERSION/SOVERSION set, so they shipped as bare, unversioned .so files -- not something a shared library meant for general linking should do on any ELF platform, since it prevents running multiple ABI- incompatible versions side by side and misses the usual mechanism for detecting an ABI break at link/load time.

Set VERSION to the full OPENMM_MAJOR_VERSION.MINOR_VERSION.BUILD_VERSION and SOVERSION to just the major version, matching the versioning scheme OpenMM already uses elsewhere (e.g. the OPENMM_VERSION string). CMake then creates the usual real libFoo.so.MAJOR.MINOR.PATCH file plus libFoo.so.MAJOR and libFoo.so symlinks automatically; verified with a from-scratch build that all four libraries get correct SONAME entries and the expected symlink chain.


peastman:
We used to provide versioned libraries, but removed it because it's incompatible with PyPI packaging, tends to cause problems, and doesn't really serve any useful purpose for the way OpenMM is intended to be used.  See #3796 for background.

vedranmiletic:
I see. I also noticed that Debian package still exists, but I didn't see that they patch this behavior. It could be that their packaged version (8.1.2) is old enough not to include the change.

Would having a compile-time option (e.g. `OPENMM_VERSION_SHARED_LIBS`, default OFF) that produces versioned libraries be acceptable?

peastman:
I'm really sorry to hear that the Debian package exists.  8.1.2 is an ancient version that no one should be using.  We specifically asked them not to create a package to avoid that situation.  Apparently they did it anyway, and now there's an ancient package that no one is updating or maintaining, but probably some people end up installing. :(

PR #5358, opened 2026-07-16
Support building the Python extension against a DESTDIR-staged install

vedranmiletic:
setup.py links openmm._openmm against OPENMM_LIB_PATH, which is always the final, real installation directory -- correct for the runtime rpath, but wrong as the linker's search path in any packaging workflow that stages a build into a temporary root (DESTDIR) before the real system install, since the just-built C++ libraries only exist under DESTDIR + OPENMM_LIB_PATH at that point, not OPENMM_LIB_PATH itself. This is the normal pattern used by most Linux distro packaging (RPM, deb, etc.) via "make install DESTDIR=...", and currently fails there with "cannot find -lOpenMM" (confirmed with a from-scratch build).

Search DESTDIR + OPENMM_LIB_PATH (empty DESTDIR is a no-op, so this is unchanged when not building via a staged install) while leaving the extension's runtime_library_dirs -- and the existing Darwin -rpath, which already did this correctly -- pointing at the clean OPENMM_LIB_PATH, since DESTDIR never exists on the system the extension actually runs on.

Verified with a full two-pass DESTDIR build (matching the pattern OpenMM's own docs describe: build+install the C++ libraries, then build the Python wrapper referencing the now-installed libraries): fails with "cannot find -lOpenMM" without this change, builds and installs cleanly with it, and the resulting extension's RUNPATH correctly points at the real (non-DESTDIR) OPENMM_LIB_PATH either way.

### DIFF
--- a/wrappers/python/setup.py
+++ b/wrappers/python/setup.py
@@ -194,6 +194,17 @@ def buildKeywordDictionary(major_version_num=MAJOR_VERSION_NUM,
     openmm_lib_path = os.getenv('OPENMM_LIB_PATH')
     if not openmm_lib_path:
         reportError("Set OPENMM_LIB_PATH to point to the lib directory for OpenMM")
+    # Packaging workflows that stage a build into DESTDIR before the final
+    # system install (e.g. "make install DESTDIR=...", used by most Linux
+    # distro packagers) build this extension after the C++ libraries have
+    # only been installed to DESTDIR + OPENMM_LIB_PATH, not the real
+    # OPENMM_LIB_PATH yet, so the linker needs to search there instead.
+    # The extension's runtime rpath must still be the real OPENMM_LIB_PATH,
+    # since DESTDIR never exists on the system the extension actually runs
+    # on -- see the runtime_library_dirs assignment below, and the Darwin
+    # -rpath in extra_link_args below, which already uses openmm_lib_path
+    # (not this DESTDIR-aware variant) for the same reason.
+    openmm_lib_search_path = os.environ.get('DESTDIR', '') + openmm_lib_path
 
     extra_compile_args=['-std=c++11']
     extra_link_args=[]
@@ -218,7 +229,7 @@ def buildKeywordDictionary(major_version_num=MAJOR_VERSION_NUM,
                 os.environ['CC'] = 'clang'
                 os.environ['CXX'] = 'clang++'
 
-    library_dirs=[openmm_lib_path]
+    library_dirs=[openmm_lib_search_path]
     include_dirs=openmm_include_path.split(';')
     include_dirs.append(numpy.get_include())
 
@@ -231,7 +242,7 @@ def buildKeywordDictionary(major_version_num=MAJOR_VERSION_NUM,
                     "extra_compile_args": extra_compile_args,
                     "extra_link_args": extra_link_args}
     if platform.system() != "Windows":
-        extensionArgs["runtime_library_dirs"] = library_dirs
+        extensionArgs["runtime_library_dirs"] = [openmm_lib_path]
     setupKeywords["ext_modules"] = [Extension(**extensionArgs)]
     setupKeywords["ext_modules"] += cythonize('openmm/app/internal/*.pyx')
 


peastman:
This looks ok, but it should be used very cautiously.  The installation scripts assume you do a complete `make install` before building the Python module.  If the workflow you describe happens to work, it's by chance, not design.  We make no promises it won't break in the future!

What's the reason for adding this?  Your code comment mentions Linux distro packagers, but we don't use any of those.  The only packaging mechanisms we support are pip and conda, and we have no plans to add any others in the foreseeable future.

vedranmiletic:
> What's the reason for adding this? Your code comment mentions Linux distro packagers, but we don't use any of those. The only packaging mechanisms we support are pip and conda, and we have no plans to add any others in the foreseeable future.

I am a co-maintainer of OpenMM in Arch User Repository (AUR). We have 8.5.1 there so, while not perfect, we are a bit more up to date compared to Debian. AUR is a user repository for community sharing of package build instructions and it has some recommendations the packages should follow, but not really a set of hard requirements.

I would also like to submit the RPM package of OpenMM to Fedora. In this case, instead of using user repositories, I plan to submit the package to the main repository, which requires me to follow relatively detailed [packaging guidelines, in part checked by rpmlint](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_use_rpmlint). Hence the pull requests.

I am grateful for #5356 and #5348 that you merged already.

Regarding this one, #5355 and #5357, I would appreciate if OpenMM could support the behaviors required for the distro packaging available in the upstream version. I am more than willing to rework these pull requests so the Fedora-friendly behaviors aren't used by default so nothing changes for your packaging environments. Furthermore, I don't expect any promises that this method of installation will work indefinitely, I can adapt the RPM specification file if OpenMM changes in the future for whatever reason.

It doesn't matter to me whether getting the Fedora packaging guidelines-compatible version requires me to set several CMake configuration switches, this is far preferred to carrying these patches downstream and rebasing them for every new release.

peastman:
I really wish you wouldn't do that.  You're just going to hurt people.  The only packaging mechanisms we support are pip and conda, and we have no plans to change that.

OpenMM is a difficult library to package.  It needs to work with multiple CUDA versions, multiple HIP versions, multiple Python versions, etc.  It also is part of a large ecosystem of interacting packages: OpenMM, OpenMM-Torch, OpenMM-ML, PDBFixer, OpenMM-Setup, etc.  All of them need to be packaged and distributed in a way that makes them compatible with each other.

The pip and conda packages went through an extended period of testing and validation before we were confident of the build process and considered them ready to use.  Your packages won't have gone through that process, which means they'll be very likely to have problems.  Encouraging users to install unvalidated packages is just going to hurt them.

And then without a firm commitment for anyone to maintain them, it's inevitable they'll often be out of date.  Encouraging users to install old versions that don't have the latest features and bug fixes will also hurt them.

And then when people run into problems, they won't go to you.  They'll come here and expect us to help them with problems caused by buggy, out of date packages we didn't create.  You're creating a support burden that will fall entirely on us.

So please, don't create any alternate distribution channels.  I appreciate that you want to make OpenMM easy to install, but it would do more harm than good.

vedranmiletic:
Thanks for the feedback, I understand where you are coming from. Fedora has the [Science and Technology Special Interest Group](https://fedoraproject.org/wiki/Category:SciTech_SIG), but it's volunteer-driven and it's hard to make any, let alone firm, commitment on package maintenance long into the future.

Therefore, I will respect your wish and cease my effort to submit this package to Fedora.

I will continue to (co-)maintain the Arch User Repository build instructions for OpenMM because that's completely unofficial (the landing page explicitly says something like "user produced, use at your own risk"). While there were packages that eventually moved from AUR to the official distribution, such as RDKit, there are no such plans for OpenMM as far as I know.

The changes that I proposed could also be useful for users that build from source on Linux distributions in general. If you would be willing to accept them as opt-in (i.e. off by default), please let me know so I can rework them. If not, that's fine too.

peastman:
Thanks, I really appreciate it!

I'll merge this one, since it's small and doesn't affect anyone who doesn't use it.  I'm going to close #5355 and #5357, which are much more likely to cause problems.

vedranmiletic:
Thank you!

PR #5359, opened 2026-07-20
Call reporters before exchanging replicas

### LINKED ISSUE #5354: `ReplicaExchangeReporter` logs post-exchange `replicaStateIndex`
`ReplicaExchangeSampler.simulate()` currently does (per iteration):

- propagate each replica (simulateReplica(i))
- compute energies for each replica conformation in all states (computeAllEnergies())
- exchange state assignments (exchangeReplicas())
- call reporters

Because reporters are called after `exchangeReplicas()`, `ReplicaExchangeReporter` records `replicaStateIndex` values that are post-exchange, while the conformations and energy matrix rows correspond to samples generated during propagation under the pre-exchange assignment.

This is fine for tracking current assignment, but it is ambiguous/misaligned for downstream MBAR-style analysis when users interpret reported state index as “state that generated this sample” (used for N_k / provenance).

The easiest fix might be to just change the order in simulate and call the reports before performing the exchange?

This is using openmm=8.6.0beta from conda-forge

peastman:
Fixes #5354.  This changes the order of steps in replica exchange so that reporters are invoked before exchanging states.

### DIFF
--- a/wrappers/python/openmm/app/replicaexchangesampler.py
+++ b/wrappers/python/openmm/app/replicaexchangesampler.py
@@ -195,10 +195,10 @@ class ReplicaExchangeSampler(object):
                 self.simulateReplica(i)
                 energies.append(self._sampler.computeAllEnergies())
             self.replicaStateEnergy = energies
-            self.exchangeReplicas()
             self.currentIteration += 1
             for reporter in self.reporters:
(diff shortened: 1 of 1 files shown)


jthorton (approved):
Thanks @peastman this works great!

PR #5363, opened 2026-07-21
Update `ReplicaExchangeReporter` energy output

### LINKED ISSUE #5362: `ReplicaExchangeReporter` reports units in the energy.csv
I am not sure if it was intended but we report units in the energy file but not the volumes file and it makes it slightly more awkward to access the data than the [doc string suggests](https://github.com/openmm/openmm/blob/e91a4d9e5a21bcefdbd9468df7a6d296788108b5/wrappers/python/openmm/app/replicaexchangereporter.py#L60), my current method to load the file is:

```python
raw_energy = np.loadtxt(old_solvent_path / "energy.csv", delimiter=',', dtype=str)
flat_energy = np.array([float(x.split()[0]) for x in raw_energy.flatten()])
energy_data = flat_energy.reshape(-1, n_states, n_states)
```
It would probably be fine to strip the units and match the output of the `volumes.csv` file or we should update the doc string to show how to load the data.

jthorton:
Fixes #5362.

### DIFF
--- a/wrappers/python/openmm/app/replicaexchangereporter.py
+++ b/wrappers/python/openmm/app/replicaexchangereporter.py
@@ -200,11 +200,8 @@ class ReplicaExchangeReporter(object):
         self._log.flush()
         if self._energy is not None:
             import numpy as np
-            energy = unit.Quantity(np.array(sampler.replicaStateEnergy))
-            if sampler._kT is None:
-                energy /= unit.MOLAR_GAS_CONSTANT_R*sampler.simulation.integrator.getTemperature()
-            else:
-                energy /= sampler._kT
+            energy = np.array(sampler.replicaStateEnergy)
+            energy /= np.array(sampler._kT)
             print(','.join([str(x) for x in energy.flatten()]), file=self._energy)
             self._energy.flush()
         if self._volume is not None:
--- a/wrappers/python/openmm/app/replicaexchangereporter.py
+++ b/wrappers/python/openmm/app/replicaexchangereporter.py
@@ -200,7 +200,7 @@ class ReplicaExchangeReporter(object):
         self._log.flush()
         if self._energy is not None:
             import numpy as np
-            energy = np.array(sampler.replicaStateEnergy)
+            energy = unit.Quantity(np.array(sampler.replicaStateEnergy))
             if sampler._kT is None:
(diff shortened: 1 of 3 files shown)


peastman:
Can you provide a test demonstrating the problem this is meant to fix?

See [this test case](https://github.com/openmm/openmm/blob/e91a4d9e5a21bcefdbd9468df7a6d296788108b5/wrappers/python/tests/TestReplicaExchangeSampler.py#L153-L157), which uses a ReplicaExchangeReporter to write out energies, and they all come out correctly with no units.

jthorton:
Yes I was confused why that test worked, I think its due to the way `_kT` is constructed when the `temperature` is defined in the states, it also becomes a lists of unit objects in that case https://github.com/openmm/openmm/blob/e91a4d9e5a21bcefdbd9468df7a6d296788108b5/wrappers/python/openmm/app/replicaexchangesampler.py#L169-L174. But yes I'll add a test for this case.

wrappers/python/openmm/app/replicaexchangereporter.py
```diff
@@ -200,7 +200,7 @@ def __call__(self, sampler: "app.ReplicaExchangeSampler"):
         self._log.flush()
         if self._energy is not None:
             import numpy as np
-            energy = np.array(sampler.replicaStateEnergy)
+            energy = unit.Quantity(np.array(sampler.replicaStateEnergy))
             if sampler._kT is None:
                 energy /= unit.MOLAR_GAS_CONSTANT_R*sampler.simulation.integrator.getTemperature()
```
jthorton:
```suggestion
                temp = sampler.simulation.integrator.getTemperature()
                energy /= [unit.MOLAR_GAS_CONSTANT_R*temp for i in range(len(energy))]
```
This also seems to fix the issue.

jthorton:
@peastman test added I also confirmed the test fails before the changes in the PR.

peastman:
This version will give incorrect results if the user changes the integrator's temperature after creating the sampler.

I have the feeling it's a mistake to convert it to a numpy array, and we're making things unnecessarily complicated.  Let me experiment with it a bit.

peastman:
I have an alternate fix in #5366.  It ensures that whenever we have a list of energies, the units are attached to the list, not to the individual elements.  Take a look and see if it looks ok to you.

peastman:
Superseded by #5366.

PR #5366, opened 2026-07-23
More consistent units handling in replica exchange

### LINKED ISSUE #5362: `ReplicaExchangeReporter` reports units in the energy.csv

peastman:
Fixes #5362.

### DIFF
--- a/wrappers/python/openmm/app/replicaexchangereporter.py
+++ b/wrappers/python/openmm/app/replicaexchangereporter.py
@@ -200,11 +200,11 @@ class ReplicaExchangeReporter(object):
         self._log.flush()
         if self._energy is not None:
             import numpy as np
-            energy = np.array(sampler.replicaStateEnergy)
+            energy = np.array(sampler.replicaStateEnergy.value_in_unit(unit.kilojoules_per_mole))
             if sampler._kT is None:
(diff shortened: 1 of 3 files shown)


jthorton (approved):
Thanks @peastman this looks great!

jthorton:
Should we align how this is done in the expanded ensemble sampler as well https://github.com/openmm/openmm/blob/e91a4d9e5a21bcefdbd9468df7a6d296788108b5/wrappers/python/openmm/app/expandedensemblesampler.py#L297-L302 that didn't have this problem and does it slightly differently.

Issue #3603, opened 2022-05-17
Constraints blowing up


jlmaccal:
We're working on finalizing our implementation of the Martini CG force field in OpenMM, but are running into problems with the topology for cholesterol.

In Martini, four atoms in cholesterol are constrained in a diamond pattern with five constraints. I have constructed a minimum failing example below. LINCS can handle this construction fine, but OpenMM's constraint algorithm blows up.

```python
import openmm as mm
import openmm.unit as u
import numpy as np


system = mm.System()
system.addParticle(50)
system.addParticle(50)
system.addParticle(50)
system.addParticle(50)

# Add constraints
system.addConstraint(0, 1, 0.4904)
system.addConstraint(0, 2, 0.6019)
system.addConstraint(1, 2, 0.2719)
system.addConstraint(3, 1, 0.7237)
system.addConstraint(3, 2, 0.5376)

platform = mm.Platform.getPlatformByName("Reference")
integrator = mm.LangevinIntegrator(
    300 * u.kelvin, 1 / u.picosecond, 0.002 * u.picoseconds
)
context = mm.Context(system, integrator, platform)

positions = np.array(
    [
        [5.566, 7.728, 3.213],
        [5.494, 7.973, 3.631],
        [5.354, 7.788, 3.773],
        [5.295, 7.811, 4.307],
    ],
    dtype=float,
)
context.setPositions(positions)
context.applyConstraints(1e-4)
new_positions = context.getState(getPositions=True).getPositions(asNumpy=True)
print(new_positions)
```

The OpenMM algorithm appears to behave well up to a tolerance of `1e-2`, but blows up for tighter tolerances.

peastman:
This arrangement of constraints seems to be ill-conditioned, though it isn't immediately obvious to me why.  If you remove any one constraint, it works correctly.  I can sort of see why it might be: each of the triangles 0-1-2 and 1-2-3 on its own would uniquely determine the positions of atoms 1 and 2.  Enforcing both triangles at once can therefore cause them to interfere with each other.  That's not to say the constraints are fundamentally in conflict: there certainly should exist a solution.  But reaching it may be numerically difficult.

If I recall correctly, LINCS cheats and only ever does two iterations?  It doesn't try to iterate constraints to convergence.  Watching the constraint solver in OpenMM, it isn't until the third iteration that it starts to show serious signs of blowing up.

jlmaccal:
To be honest, the details of constraint algorithms are something I've never really dug into.

[...]

PR #5367, opened 2026-07-24
Improve CCMA robustness with delayed position damping

lyfar:
# Draft PR: Improve CCMA robustness with delayed position damping

## Summary

This draft adds a fallback for CCMA position constraints:

- Reference uses full corrections for iterations 0 through 11, then half
  corrections.
- Common preserves its existing half corrections at iterations 0 and 1, uses
  full corrections at iterations 2 through 11, then returns to half
  corrections.
- Velocity constraints are unchanged.

The threshold of 12 is empirical. This is a robustness fallback, not a general
convergence guarantee.

## Motivation

The four-particle reproducer in #3603 diverges under the current Reference
position iteration. For one anchored linearized mode, the update has a
certified eigenvalue in `(-1.8, -1.75)`. Applying half corrections maps that
mode to approximately `(-0.4, -0.375)`.

This explains one failure mode of the Reference recurrence. It does not prove
global CCMA convergence or cover every constraint graph.

## Changes

- Apply delayed half-step damping to Reference and Common position
  corrections.
- Pass an explicit position-or-velocity flag to the shared Common kernel so
  velocity behavior stays unchanged.
- Add the public four-particle reproducer to the shared SETTLE test suite.

## Tests

Before this patch, the new four-particle regression fails its squared-distance
assertion on Reference and CPU.

With this patch:

- Reference passes.
- CPU passes.
- OpenCL single precision passes.
- Existing Reference, CPU, and OpenCL SETTLE and Verlet tests pass.
- A manual OpenCL test with more than 1,024 constraints passes position and
  velocity checks, exercising the split-kernel path.

The companion audit also models the exact current and patched Common schedules
on 80 bounded float64 inputs. The patched schedule loses none of the 65
current-schedule successes and makes the original issue case converge at
`1e-4` and `1e-6`. This is distance-only numerical evidence. It does not
replace compiled backend tests.

## Known limits

The later eight-particle mixed-angle fixture from #3603 passes on Reference and
CPU. OpenCL single precision still misses the strict squared-distance tolerance
on four of five constraints. This draft therefore does not claim to resolve
every configuration reported in #3603.

Still outstanding:

- CUDA runtime testing
- HIP runtime testing
- OpenCL mixed and double precision
- full CI
- performance measurements
- broader slow-convergence systems

## Maintainer question

Is delayed position damping an acceptable fallback direction for CCMA? Should
the threshold remain a fixed iteration count, become platform-specific, or be
triggered by measured residual behavior before this moves beyond Draft?

### DIFF
--- a/platforms/common/src/IntegrationUtilities.cpp
+++ b/platforms/common/src/IntegrationUtilities.cpp
@@ -765,6 +765,7 @@ IntegrationUtilities::IntegrationUtilities(ComputeContext& context, const System
         ccmaUpdateKernel->addArg(ccmaDelta2);
         ccmaUpdateKernel->addArg(ccmaConverged);
         ccmaUpdateKernel->addArg();
+        ccmaUpdateKernel->addArg();
         ccmaFullKernel->addArg();
         ccmaFullKernel->addArg(ccmaAtoms);
         ccmaFullKernel->addArg(ccmaNumAtomConstraints);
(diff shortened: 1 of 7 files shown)


peastman:
I don't think this is an appropriate change.  #3603 is a pathological case involving a poorly conditioned constraint matrix.  You're modifying the standard behavior to make that pathological case work better, but the result will be slower performance for normal cases.

PR #5368, opened 2026-07-24
Validate complete constraint triangles before solver initialization

lyfar:
## Summary

- Reject non-finite or non-positive constraint distances when OpenMM creates a `Context`.
- Reject complete constraint triangles that are degenerate or violate the strict triangle inequality.
- Reject isolated SETTLE triangles whose float-stored geometry is not numerically safe for the solver.

## Motivation

OpenMM accepts a system with three unit-mass particles and constraint distances `1, 1, 2`. It selects SETTLE after storing those distances as `float`. Position constraints then compute a zero `rb` and `ra` before dividing by `ra`, producing NaN positions. The corresponding velocity constraint system is singular.

A strictly valid double-precision triangle can reach the same failure: the double immediately below `2` rounds to `2.0f`, so `1, 1, nextafter(2, 0)` becomes exactly collinear in SETTLE's stored geometry.

The new validation raises an `OpenMMException` before a platform initializes its constraint solver.

## Numerical policy

For every complete triangle, the validator sorts the double-precision side lengths and checks:

`smallest <= largest - middle`

The subtraction form avoids overflow and enforces a strict, nondegenerate triangle without a tolerance band.

The additional float check applies only to an isolated potential SETTLE cluster: all three particles have exactly two constrained neighbors, all have nonzero mass, and at least two float-stored sides compare equal under SETTLE's selection rules.

For that cluster, validation rejects:

- zero or non-finite float-stored sides;
- non-strict float-stored side lengths;
- a SETTLE radicand that is not both positive and normal.

The radicand is evaluated with the same float arithmetic used by the common SETTLE path:

`arm*arm - (0.5f*base)*(0.5f*base)`

Strictly valid near-degenerate triangles that are not isolated SETTLE clusters remain available to CCMA. Incomplete constraint graphs remain accepted because their missing distances may admit a realization.

## Tests

The shared SETTLE tests now cover:

- exact `1, 1, 2`;
- adjacent double values below and above `2`;
- the first valid double value that rounds to a collinear SETTLE geometry;
- the last value that remains nondegenerate after float storage, including finite position and velocity constraint execution;
- float-square overflow (`1e20`), a positive subnormal radicand (`1e-19`), and float-square underflow (`1e-30`);
- an embedded near-degenerate triangle routed through CCMA, with finite position and velocity results;
- existing valid, inconsistent, extreme-scale, embedded, scalar-distance, and partial-graph cases.

`TestReferenceSettle` and `TestCpuSettle` pass after the final guard.

The configured 124-test serial suite was also run twice before the final narrow normal-radicand guard. Each run passed 123 tests and had one different stochastic test fail; each failed test passed immediately on isolated rerun. Both SETTLE tests passed in both full runs.

CUDA was unavailable locally. The common float storage and kernel arithmetic were inspected, and the unsafe arithmetic boundaries are covered by the shared tests for upstream platform CI.

## Scope

This change validates constraint input at `Context` creation. SETTLE and CCMA keep their existing algorithms, damping, convergence criteria, and tolerances. Issue #3603 remains a separate solver-convergence problem.

### DIFF
--- a/openmmapi/src/ContextImpl.cpp
+++ b/openmmapi/src/ContextImpl.cpp
@@ -50,6 +50,49 @@ using namespace OpenMM;
 using namespace std;
 const static char CHECKPOINT_MAGIC_BYTES[] = "OpenMM Binary Checkpoint\n";
 
+namespace {
+
+bool violatesStrictTriangleInequality(double distance1, double distance2, double distance3) {
+    double distances[] = {distance1, distance2, distance3};
+    sort(distances, distances+3);
+    // Subtraction avoids overflow.  Equality is unsafe: constraint solvers
+    // require a nondegenerate triangle.
+    return distances[0] <= distances[2]-distances[1];
+}
+
+bool isUnsafeAtSettlePrecision(double distance1, double distance2, double distance3) {
+    float distance12 = (float) distance1;
+    float distance13 = (float) distance2;
+    float distance23 = (float) distance3;
+    float arm, base;
+    if (distance12 == distance13) {
+        arm = distance12;
+        base = distance23;
+    }
+    else if (distance12 == distance23) {
+        arm = distance12;
+        base = distance13;
+    }
+    else if (distance13 == distance23) {
+        arm = distance13;
+        base = distance12;
+    }
+    else
+        return false;
+    float stored[] = {distance12, distance13, distance23};
+    for (int i = 0; i < 3; i++)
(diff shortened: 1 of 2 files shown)


peastman:
Can you explain the motivation for this change?  It adds a lot of cost to Context creation, but only checks for one specific type of singular constraint matrix while leaving lots of others unchecked.  What is the concrete problem it's meant to solve?

I assume the code is AI generated?  It has a lot of the issues that tend to indicate that.

PR #5373, opened 2026-07-24
ExpandedEnsembleSampler records the state index before exchange

### LINKED ISSUE #5370: `ExpandedEnsembleSampler` issues
Following up from the changes in the replica exchange sampler it would be great to align the methods so that we can use them with MBAR:
-  volumes are not reported which are required for the full reduced potentials 
- The state reported is the post exchange state not the state that generated the sample conformation https://github.com/openmm/openmm/blob/e91a4d9e5a21bcefdbd9468df7a6d296788108b5/wrappers/python/openmm/app/expandedensemblesampler.py#L258-L260

peastman:
Fixes #5370.

### DIFF
--- a/wrappers/python/openmm/app/expandedensemblesampler.py
+++ b/wrappers/python/openmm/app/expandedensemblesampler.py
@@ -255,9 +255,10 @@ class ExpandedEnsembleSampler(object):
             def report(self, simulation, state):
                 sampler = self.sampler
                 if simulation.currentStep%sampler.stepsPerIteration == 0:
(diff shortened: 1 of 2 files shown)


jthorton (approved):
LGTM!

wrappers/python/openmm/app/expandedensemblesampler.py
jthorton, line 121: We should add info here that the last exchange will be lost when restarting and that it could cause issues if the run has frequent restarts.

PR #5375, opened 2026-07-25
Move serialization subdirectory before plugins/amoeba and plugins/drude

vedranmiletic:
plugins/amoeba and plugins/drude each check OPENMM_BUILD_SERIALIZATION_TESTS to decide whether to add their own serialization/tests subdirectories. That variable is declared by serialization/CMakeLists.txt, but on a fresh (single-pass) configure it was previously being added_subdirectory'd after both plugins, so the check evaluated as undefined/false and their tests were silently skipped (147 tests instead of 138 on a clean configure). Moving the serialization subdirectory earlier, right after the CPU platform and before the plugins, ensures the variable is defined before it's needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -372,6 +372,16 @@ IF(OPENMM_BUILD_CPU_LIB)
    ADD_SUBDIRECTORY(platforms/cpu)
 ENDIF(OPENMM_BUILD_CPU_LIB)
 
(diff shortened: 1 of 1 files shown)


peastman:
Looks good.  Thanks!

vedranmiletic:
Thank you!

PR #5377, opened 2026-07-28
Give kernel interface classes a key function

vedranmiletic:
On Linux, an OpenMM built against libc++ could not create a Context on any plugin platform from Python as Kernel::getAs() threw std::bad_cast. The abstract kernel classes declared only pure virtual members and inline constructors, so they had no key function, and every library that included the headers emitted its own weak copy of the typeinfo. Python loads the extension module with RTLD_LOCAL, so those copies were never unified. libc++abi compares typeinfo by address and the cast failed; libstdc++ compares mangled names and survived.

Declare a virtual destructor for each of the 56 kernel interface classes and for KernelImpl, and define it out of line. Each destructor is then the key function for its class, so the vtable and typeinfo are emitted once with strong linkage and every plugin refers to that single definition.

Three separate cast boundaries were affected, each only visible once the previous one was fixed: libOpenMM to plugin, plugin to plugin, and API library to plugin.

The destructors must stay out of line. Defaulting one on its declaration in a header makes it inline, which leaves the class without a key function and silently reintroduces the problem: nothing is diagnosed at build time with either standard library and it is invisible at run time on libstdc++.

### DIFF
--- a/olla/include/openmm/kernels.h
+++ b/olla/include/openmm/kernels.h
@@ -92,6 +92,7 @@ public:
     }
     CalcForcesAndEnergyKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~CalcForcesAndEnergyKernel();
     /**
      * Initialize the kernel.
      * 
@@ -136,6 +137,7 @@ public:
     }
     UpdateStateDataKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~UpdateStateDataKernel();
     /**
      * Initialize the kernel.
      *
@@ -255,6 +257,7 @@ public:
     }
     ApplyConstraintsKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~ApplyConstraintsKernel();
     /**
      * Initialize the kernel.
      *
@@ -287,6 +290,7 @@ public:
     }
     VirtualSitesKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~VirtualSitesKernel();
     /**
      * Initialize the kernel.
      *
@@ -311,6 +315,7 @@ public:
     }
     MinimizeKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~MinimizeKernel();
     /**
      * Initialize the kernel.
      *
@@ -338,6 +343,7 @@ public:
     }
     CalcHarmonicBondForceKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~CalcHarmonicBondForceKernel();
     /**
      * Initialize the kernel.
      * 
@@ -375,6 +381,7 @@ public:
     }
     CalcCustomBondForceKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~CalcCustomBondForceKernel();
     /**
      * Initialize the kernel.
      *
@@ -412,6 +419,7 @@ public:
     }
     CalcHarmonicAngleForceKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~CalcHarmonicAngleForceKernel();
     /**
      * Initialize the kernel.
      * 
@@ -449,6 +457,7 @@ public:
     }
     CalcCustomAngleForceKernel(std::string name, const Platform& platform) : KernelImpl(name, platform) {
     }
+    virtual ~CalcCustomAngleForceKernel();
     /**
      * Initialize the kernel.
      *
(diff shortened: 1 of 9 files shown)


peastman:
We've built OpenMM on many different Linux distributions over many years and never seen that problem.  How exactly do you reproduce it?

vedranmiletic:
I'm seeing this on Fedora, but the instructions below should work on a typical Linux distro.

Using the following Python script:

``` python
#!/usr/bin/env python

import openmm

s = openmm.System()
s.addParticle(1.0)
openmm.Context(s, openmm.VerletIntegrator(0.001),
               openmm.Platform.getPlatformByName("Reference"))
print("Reference context created succesfully")

s = openmm.System()
s.addParticle(1.0)
openmm.Context(s, openmm.VerletIntegrator(0.001),
               openmm.Platform.getPlatformByName("CPU"))
print("CPU context created succesfully")
```

Let's first build and install with GCC or Clang/libstdc++:

```
cmake -S . -B gccbuild -G Ninja -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_INSTALL_PREFIX=~/software/openmm-gcc -DOPENMM_PYTHON_USER_INSTALL=ON -DOPENMM_BUILD_COMMON=OFF -DOPENMM_BUILD_CUDA_LIB=OFF -DOPENMM_BUILD_OPENCL_LIB=OFF -DOPENMM_BUILD_HIP_LIB=OFF
cmake --build gccbuild --target install
cmake --build gccbuild --target PythonInstall
```

or

```
cmake -S . -B clangbuild -G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=~/software/openmm-clang -DOPENMM_PYTHON_USER_INSTALL=ON -DOPENMM_BUILD_COMMON=OFF -DOPENMM_BUILD_CUDA_LIB=OFF -DOPENMM_BUILD_OPENCL_LIB=OFF -DOPENMM_BUILD_HIP_LIB=OFF
cmake --build clangbuild --target install
export CC=clang CXX=clang++
cmake --build clangbuild --target PythonInstall
```

Running the script prints:

```
Reference context created succesfully
CPU context created succesfully
```

Now let's build with Clang/libc++:

```
cmake -S . -B clangcxxbuild -G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS='-stdlib=libc++' -DCMAKE_INSTALL_PREFIX=~/software/openmm-clangcxx -DOPENMM_PYTHON_USER_INSTALL=ON -DOPENMM_BUILD_COMMON=OFF -DOPENMM_BUILD_CUDA_LIB=OFF -DOPENMM_BUILD_OPENCL_LIB=OFF -DOPENMM_BUILD_HIP_LIB=OFF
cmake --build clangcxxbuild --target install
export CC=clang CXX=clang++ CXXFLAGS="-stdlib=libc++"
cmake --build clangcxxbuild --target PythonInstall
```

After running the script:

```
Reference context created succesfully
  File "/home/vmiletic/workspace/libcxx-rtti-repro/repro-script.py", line 13, in <module>
    openmm.Context(s, openmm.VerletIntegrator(0.001),
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                   openmm.Platform.getPlatformByName("CPU"))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vmiletic/.local/lib/python3.14/site-packages/openmm/openmm.py", line 6217, in __init__
    _openmm.Context_swiginit(self, _openmm.new_Context(*args))
                                   ~~~~~~~~~~~~~~~~~~~^^^^^^^
openmm.OpenMMException: std::bad_cast
```

As noted, accessing the plugin platform raises the exception. Let me know if I missed something. 

FWIW, I was skeptical whether my Fedora installation is special in some way. In that regard, I should note that I'm also able to observe the issue on my FreeBSD installation (where libc++ happens to be the default C++ library).

peastman:
I'm not sure what to make of that.  I do much of my development on a Mac, which uses clang and libc++ for everything.  In fact, we enforce that:

https://github.com/openmm/openmm/blob/398801cf1aadf1e0f82f09046ff1ed510ad00388/CMakeLists.txt#L121

I've never seen anything like the problem you describe.

vedranmiletic:
> I'm not sure what to make of that. I do much of my development on a Mac, which uses clang and libc++ for everything. In fact, we enforce that:
> 
> https://github.com/openmm/openmm/blob/398801cf1aadf1e0f82f09046ff1ed510ad00388/CMakeLists.txt#L121
> 
> I've never seen anything like the problem you describe.

I observed this with ELF, since this is what both Linux and FreeBSD use, while macOS use Mach-O. I don't think it will necessary be reproducible on that platform. Do you want me to prepare a Docker container where this is reproducible?

PR #5381, opened 2026-07-29
Minor performance improvements to Modeller

badisa:
The following changes are some minor performance improvements to reduce the time it takes to build systems, primarily looking at solvent systems. In some cases I was able to reduce the time to constructing a system by ~5 seconds (a large membrane system).

## Changes
* Use list comprehensions where appropriate (IE where the size of the list isn't unbounded and thus would have memory implications) in the Modeller. See [this](https://switowski.com/blog/for-loop-vs-list-comprehension/) for how list comprehensions end up being faster than generators.
* Exit early from _addIons if there aren't any ions to be added, avoiding a createSystem call that is redundant.
* Re-order a set of nested loops to avoid repeatedly looping through the same residue multiple times to get an oxygen index.

The changes were benchmarked using [Pyspy](https://github.com/benfred/py-spy) and this should produce identical systems as the current changes.

### DIFF
--- a/wrappers/python/openmm/app/modeller.py
+++ b/wrappers/python/openmm/app/modeller.py
@@ -1727,8 +1727,8 @@ class _CellList(object):
                 vectors = (Vec3(width[0], 0, 0), Vec3(0, width[1], 0), Vec3(0, 0, width[2]))
         self.positions = deepcopy(positions)
         self.cells = {}
-        self.numCells = tuple((max(1, int(floor(vectors[i][i]/maxCutoff))) for i in range(3)))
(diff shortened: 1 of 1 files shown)


peastman:
Thanks!  I wouldn't have expected just replacing `()` with `[]` to have a significant difference, but my tests confirm it does:

```
$ python -m timeit "tuple((i+1 for i in range(3)))"
2000000 loops, best of 5: 167 nsec per loop
$ python -m timeit "tuple([i+1 for i in range(3)])"
5000000 loops, best of 5: 65.6 nsec per loop
```

Tests are running now.

PR #5388, opened 2026-08-06
Fixed implicit solvent forces on CPU platform with triclinic boxes

### LINKED ISSUE #5387: Implicit solvent computes incorrect distances for triclinic box on CPU platform
GBSAOBCForce and CustomGBForce incorrectly wrap distances for triclinic boxes (i.e. they assume an orthorhombic box). See the relevant code below:

https://github.com/openmm/openmm/blob/398801cf1aadf1e0f82f09046ff1ed510ad00388/platforms/cpu/src/CpuGBSAOBCForce.cpp#L395-L405

https://github.com/openmm/openmm/blob/398801cf1aadf1e0f82f09046ff1ed510ad00388/platforms/cpu/src/CpuCustomGBForce.cpp#L668-L673

pujaltes:
Fixes #5387

PBC Implicit solvent simulations with triclinic boxes were incorrectly calculating distances along each cartesian axis independently. This PR includes a corrected calculation and test for implicit solvent triclinic distances on the CPU platform.

### DIFF
--- a/platforms/cpu/include/CpuCustomGBForce.h
+++ b/platforms/cpu/include/CpuCustomGBForce.h
@@ -44,8 +44,10 @@ private:
 
     bool cutoff;
     bool periodic;
+    bool triclinic;
     const CpuNeighborList* neighborList;
     float periodicBoxSize[3];
+    AlignedArray<fvec4> periodicBoxVec4;
     float cutoffDistance, cutoffDistance2;
     int numValues, numParams;
     const std::vector<std::set<int> > exclusions;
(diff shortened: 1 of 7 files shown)


peastman:
Thanks!

I guess no one ever hit this before, because no one ever tried to use triclinic boxes with implicit solvent.  The main uses of triclinic boxes are to simulate crystals and to  reduce the amount of explicit solvent that needs to be simulated.

peastman:
Github Actions is currently having problems, and it keeps cancelling jobs.  I'll wait until they report the problems are resolved before trying to resubmit them again.

peastman:
Looks good.

pujaltes:
Great, thanks!

Yeah it's admittedly a very weird use case (hopefully useful for some work we're doing). I really appreciate that OpenMM allows for an maintains all of these different and niche simulation settings.

PR #5393, opened 2026-08-20
Add CI on self-hosted runners

epretti:
Adds CI for OpenCL and CUDA on self-hosted runners.

Note that like the existing Jenkins setup, these runs don't do the Python tests, but this should be possible to add if we want to.  Unlike the existing Jenkins setup, this doesn't add any CPU tests, since we already have many such configurations on GitHub runners, including ARM on Linux.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -675,3 +675,30 @@ jobs:
           ((exitcode+=$?))
           echo "::endgroup::"
           exit $exitcode
+
+  gpu:
+    runs-on: self-hosted
+    name: ${{ matrix.name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Linux GPU
+            CMAKE_FLAGS: |
+              -DOPENMM_BUILD_REFERENCE_TESTS=OFF \
+              -DOPENMM_BUILD_CPU_TESTS=OFF \
+              -DOPENMM_BUILD_EXAMPLES=OFF \
+              -DOPENMM_BUILD_C_AND_FORTRAN_WRAPPERS=OFF \
+              -DOPENMM_BUILD_PYTHON_WRAPPERS=OFF \
+        steps:
+          - uses: actions/checkout@v7
+
+          - name: "Build"
+            run: |
+              mkdir build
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -675,3 +675,71 @@ jobs:
...
+    name: ${{ matrix.name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Linux GPU OpenCL
+            python-version: "3.13"
+            env: ubuntu-arm
+            CMAKE_FLAGS: |
+              -DOPENMM_BUILD_CUDA_TESTS=OFF \
```
peastman: This should probably be `OPENMM_BUILD_CUDA_LIB`, and similar for OpenCL below.  With `OPENMM_BUILD_CUDA_TESTS` it still compiles the CUDA platform, just not the test cases.

.github/workflows/CI.yml
```diff
@@ -675,3 +675,71 @@ jobs:
...
+        shell: bash -l {0}
+        run: |
+          cd build
+          make -j8 install
+
+      - name: "Test OpenMM"
+        shell: bash -l {0}
+        run: |
+          cd build
+          python ../devtools/run-ctest.py --parallel 2 --timeout 600 --job-duration 900 --attempts 3
```
peastman: Have you tried a larger value for `--parallel`?  It's a 20 core processor, and most of the tests are dominated by CPU time.

epretti: My experience from testing this with Jenkins had been that `--parallel 2` was best and `--parallel 4` was slower.  It seemed that with enough tests running simultaneously, the GPU did become a bottleneck.  I've just verified that.  I never tried `--parallel 3`, so I'll check it to be sure.

peastman: It looks like that saves a few minutes.

epretti: Indeed, it seems to be the fastest option.

.github/workflows/CI.yml
epretti: Should be updated now.

PR #5405, opened 2026-08-28
Make `PythonForce.getPickledFunction()` return `bytes`

### LINKED ISSUE #5404: `PythonForce.getPickledFunction()` returns opaque object
[Documentation](https://docs.openmm.org/latest/api-python/generated/openmm.openmm.PythonForce.html#openmm.openmm.PythonForce.getPickledFunction) claims that `PythonForce.getPickledFunction()` returns `tuple[char]`.  However,
```py
import openmm
def compute(state):
    pass
force = openmm.PythonForce(compute)
print(force.getPickledFunction())
```
gives `<Swig Object of type 'std::vector< char,std::allocator< char > > *' at 0x...>`.  This object doesn't provide a way to get the bytes.  Ideally this function would just return a `bytes`.

epretti:
Should fix #5404.


peastman:
This looks fine, though really I never intended anyone to call `getPickledFunction()`.  It exists mainly for use in serialization.  The documentation claiming the return type is `tuple[char]` was automatically generated, not something I wrote.

epretti:
Would you rather it just be excluded from the Python API/documentation altogether?

I did run into this problem because I was trying to use this function; specifically, when testing the link-atom feature for OpenMM-ML, I wanted to easily inspect some objects inside the callable for the `PythonForce` after it had been created.  I imagine this isn't something a lot of people would want to do, but it was useful for me.

peastman:
That's fine.  There's no reason to block it if it's useful.

Issue #5397, opened 2026-08-24
Proposal: Explore a native Metal Platform for Apple silicon Macs


NORPG:
## Background

OpenMM’s macOS GPU support has gone through several stages.

After Apple deprecated OpenCL at WWDC18, the long-term direction of Mac GPU support was discussed in [#2093](https://github.com/openmm/openmm/issues/2093) and [#2489](https://github.com/openmm/openmm/issues/2489). OpenMM subsequently added native arm64 Mac support through [#2939](https://github.com/openmm/openmm/issues/2939) and [#2978](https://github.com/openmm/openmm/pull/2978), while continuing to use OpenCL for Apple GPUs.

Apple silicon GPU performance was later investigated in [#3847](https://github.com/openmm/openmm/issues/3847). Several improvements were made to the existing OpenCL Platform, including [#3917](https://github.com/openmm/openmm/pull/3917), [#3960](https://github.com/openmm/openmm/pull/3960), and [#3978](https://github.com/openmm/openmm/pull/3978). Metal-specific experiments in [#3924](https://github.com/openmm/openmm/issues/3924) and [#3959](https://github.com/openmm/openmm/pull/3959) also showed that some Apple GPU features, such as SIMD-group operations, are not fully exposed through Apple’s OpenCL implementation.

A native Metal Platform was previously discussed in [#3639](https://github.com/openmm/openmm/issues/3639). The main concerns were the implementation and maintenance cost, uncertain end-to-end performance benefits, and the lack of FP64 support on Apple GPUs.

The external [`openmm-metal`](https://github.com/philipturner/openmm-metal) plugin provides useful prior experience. It demonstrates substantial OpenMM Platform integration and supports many Apple FP32 tests, including PME. However, it translates OpenCL kernels through `cl2Metal` rather than using native MSL throughout. Its recent compatibility problems—[openmm-metal#3](https://github.com/philipturner/openmm-metal/issues/3), [openmm-metal#4](https://github.com/philipturner/openmm-metal/issues/4), and [openmm-metal#5](https://github.com/philipturner/openmm-metal/pull/5)—also show the risks of compiler dependencies, private AIR interfaces, and OpenMM API drift.

OpenCL still works on current Apple silicon Macs. This proposal is not based on an assumption that it will disappear soon, and it does not propose removing the existing OpenCL Platform.

## Proposal

Would the OpenMM project be open to an experimental, out-of-tree native Metal Platform for Apple silicon?

“Native Metal” here means:

[...]

PR #5416, opened 2026-09-07
[WIP] Add an experimental native Metal Platform for Apple silicon

NORPG:
Related discussion: #5397
Current commit: [8f6a7332f](https://github.com/NORPG/openmm/commit/8f6a7332f4bca326cd43366f2916da396db661ae)

### Current technical choices

- **Target:** Apple silicon, macOS 13+, one GPU, single precision.
- **Structure:** Follow the CUDA/HIP runtime interfaces and reuse existing Common context, array-wrapper, and force-info sources.
- **Host implementation:** Private Objective-C++ with ordinary C++11-facing interfaces.
- **Build:** Optional Metal build, disabled by default.

Following the [review feedback](https://github.com/openmm/openmm/pull/5416#issuecomment-5575475013), I restarted the branch from `master` and replaced the earlier prototype with a smaller foundation. The priority is simple, readable code that follows existing platforms, with eventual Common Compute kernel reuse and minimal Metal-specific code.

### Implementation status

- Basic runtime and standard state buffers through `ComputeContext`, `ArrayInterface`, `ComputeQueue`, `ComputeEvent`, `ComputeProgram`, and `ComputeKernel`.
- Buffer transfers and clearing, argument rebinding, native MSL 3.0 compilation, threadgroup dispatch, and queue/event synchronization.
- Doxygen documentation for ownership, asynchronous transfers, bindings, and unsupported interfaces.

This is **not yet a registered simulation Platform**, and it does not execute existing Common kernel sources. The earlier simulation paths and custom fixed-point implementation are no longer included. Forces, integrators, sorting, neighbor lists, and FFT/PME remain unimplemented.

### Host API and language choice

- **Objective-C++:** implemented in this branch.
- **metal-cpp:** alternative using a private C++17 implementation.
- **Swift:** alternative behind a C/C++ bridge.

Objective-C++ requires the least porting work because the runtime already exists behind C++11 interfaces. Retaining it still needs maintainer agreement, given the concern raised in #5397.

[Apple's metal-cpp](https://developer.apple.com/metal/cpp/) is a header-only interface requiring C++17. We could apply that requirement only to the optional Metal implementation and keep `MTL::`/`NS::` types private. I would pin a release and rebuild/test header updates without assuming a separate stable wrapper ABI.

Swift would add build/runtime integration and ownership/error-handling work. [Direct C++ interoperability requires Swift 5.9+ and C++14+](https://www.swift.org/documentation/cxx-interop/status/) at the bridge boundary; an isolated adapter or C bridge could preserve C++11-facing interfaces. Swift's [stable Apple-platform ABI](https://www.swift.org/blog/library-evolution/) is distinct from its [evolving C++ interop compatibility policy](https://www.swift.org/documentation/cxx-interop/#source-stability-guarantees-for-mixed-language-codebases).

All three options concern the host layer. MSL kernels and Common Compute reuse remain shared goals. Which option would maintainers prefer to evaluate?

### Validation

- `TestMetalComputeContext` passes with Metal API Validation and GPU Validation enabled.
- Coverage includes transfers, resizing/rebinding, complete threadgroups, grid-stride execution, clearing, pinned-memory readback, queue/event ordering, and invalid inputs.
- Doxygen generation completes without warnings.

These tests validate runtime interfaces only—not simulation correctness, Common shader compatibility, or performance. Build instructions are in the [Metal README](https://github.com/NORPG/openmm/blob/Objective-C/platforms/metal/README.md).

Feedback on whether this smaller foundation follows the intended CUDA/HIP porting approach would be appreciated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -357,6 +357,13 @@ IF(OPENMM_BUILD_HIP_LIB)
     ADD_SUBDIRECTORY(platforms/hip)
 ENDIF(OPENMM_BUILD_HIP_LIB)
 
+# Metal Platform
+
+SET(OPENMM_BUILD_METAL_LIB OFF CACHE BOOL "Build the experimental Metal Platform")
+IF(OPENMM_BUILD_METAL_LIB)
+    ADD_SUBDIRECTORY(platforms/metal)
+ENDIF(OPENMM_BUILD_METAL_LIB)
+
 # Common compute files
 
 SET(OPENMM_BUILD_COMMON OFF CACHE BOOL "Build common files even if CUDA or OpenCL platforms are not built")
--- a/platforms/metal/CMakeLists.txt
+++ b/platforms/metal/CMakeLists.txt
@@ -1,0 +1,88 @@
+# --------------------------------------------------------------------------
+# OpenMM
+#
+# This is part of the OpenMM molecular simulation toolkit.
+# See https://openmm.org/development.
+#
+# Metal Platform code:
+# Portions copyright (c) 2026 Chun-Chi Hung.
+# Authors: Chun-Chi Hung
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+# --------------------------------------------------------------------------
+
+# Metal Platform: minimal implementation of the Common compute interfaces.
+# This target does not yet register an OpenMM simulation Platform.
+
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    message(FATAL_ERROR "The Metal Platform is only available on macOS")
+endif()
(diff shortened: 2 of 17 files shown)


peastman:
It looks to me like you're writing this from scratch instead of basing it on one of the existing platforms.  I would definitely not do it that way.  You'll end up with a lot of code that has no similarity to the other platforms and that behaves differently (and likely incorrectly) in a lot of ways.

Instead, I would start from either the CUDA or HIP platform (whichever you think is more similar to Metal, but they're very similar to each other so it doesn't matter that much) and convert it.

This code looks really complicated, and a lot of it I have no idea what it's doing.  For example, MetalQueue.mm is 235 lines.  CudaQueue.cpp is only 44, and most of that is just the license header at the top.  What is different about Metal that requires so much more code?  Similarly, MetalArray.mm is 301 lines while CudaArray.cpp is only 134.  MetalKernels.cpp is 675 lines while CudaKernels.cpp is only 103.

Then there's a whole lot of code related to fixed point values.  I can't figure out why that's needed.  In the other platforms we just use a 64 bit int when we need a fixed point value.  The conversion is done with a single trivial function:

https://github.com/openmm/openmm/blob/0da03998df892bbb0a954ad3767c30a0cc53a11c/platforms/opencl/src/kernels/common.cl#L80-L82

NORPG:
Thanks for the feedback. I've replaced the earlier prototype in [8f6a7332f](https://github.com/NORPG/openmm/commit/8f6a7332f4bca326cd43366f2916da396db661ae) and updated the PR description.

- Reworked the branch from `master` around the CUDA/HIP Common runtime interfaces.
- Removed the standalone force/integrator paths, checkpoint handling, offline metallibs, and custom fixed-point/probe code.
- Kept a minimal runtime, standard state buffers, focused tests, and Doxygen documentation.

This version is not yet a simulation Platform and does not execute existing Common kernels. Objective-C++ remains provisional. The aim is a small, reviewable foundation for Common Compute reuse. The runtime test passes with Metal API and GPU validation enabled.

PR #5418, opened 2026-09-08
add strict element detection for GROMACS topology parsing

wenyan4work:
Add a switch to allow users to choose a stricter mode

### DIFF
--- a/docs-source/usersguide/application/02_running_sims.rst
+++ b/docs-source/usersguide/application/02_running_sims.rst
@@ -325,6 +325,15 @@ Unix-like operating systems.  So in :autonumref:`Example,Gromacs example` we act
 this parameter, but if the Gromacs files were installed in any other location,
 we would need to include it.
 
+By default, OpenMM uses the atomic numbers in the :code:`[ atomtypes ]` section
+to identify elements, and guesses from atom names when atomic numbers are missing.
+Names can be ambiguous: for example, :code:`CA` can refer to a carbon atom or a
+calcium ion. To require explicit atomic numbers, pass
+:code:`allowElementGuessing=False` to :class:`GromacsTopFile`. This raises a
+:class:`ValueError` if an atom in the topology uses a type without an atomic
+number. An explicit atomic number of zero represents a particle with no element,
+such as a dummy particle, and is accepted in either mode.
+
 .. _using-charmm-files:
 
 Using CHARMM Files
--- a/wrappers/python/openmm/app/gromacstopfile.py
+++ b/wrappers/python/openmm/app/gromacstopfile.py
@@ -160,12 +160,13 @@ class GromacsTopFile(object):
 
     def _processFile(self, file):
         append = ''
-        for line in open(file):
-            if line.strip().endswith('\\'):
-                append = '%s %s' % (append, line[:line.rfind('\\')])
-            else:
-                self._processLine(append+' '+line, file)
-                append = ''
+        with open(file) as inputFile:
+            for line in inputFile:
+                if line.strip().endswith('\\'):
+                    append = '%s %s' % (append, line[:line.rfind('\\')])
+                else:
+                    self._processLine(append+' '+line, file)
+                    append = ''
 
     def _processLine(self, line, file):
         """Process one line from a file."""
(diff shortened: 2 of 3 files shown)


peastman:
This PR makes a lot of different changes, most of which seem to be unrelated to the feature you described.  It's not clear to me what all of them are for.

How does Gromacs handle this situation?  The [documentation](https://manual.gromacs.org/current/reference-manual/topologies/topology-file-formats.html) says that atomic number is optional.  What happens when it's missing?

wenyan4work:
This is a comparison generated by GPT-6. `allowElementGuessing=False` is a new mode which is not supposed to be identical as original gromacs behavior (stricter, such that later stages like HMR, constraints with HBond/HAgnles are more reliable).


</html>Reviewed `codex/gromacs-parser` (`620f6ab7`) against local `master` (`af1260bb`). **No confirmed introduced regressions found.** The default preserves the guessing heuristic, but parser fixes can change results.

1. **With `allowElementGuessing=True` versus original OpenMM**

   The heuristic is unchanged: explicit atomic numbers take precedence; otherwise `CL`, `NA`, and `MG` receive special handling, followed by first-letter lookup. Thus `CA` still guesses carbon; an unrecognized name produces `element=None`.

   | Input or condition | Original OpenMM | This branch |
   |---|---|---|
   | Correctly parsed, ordinary atom-type records | Uses explicit number or guesses | Same |
   | Seven-column record with single-character charge, e.g. `T 20 40.08 0 A 0.3 0.2` | Misaligns columns, guesses carbon for `CA`, then fails building the system | Correctly assigns calcium and builds the system |
   | Eight-column record with single-character mass | Can misalign columns | Correctly parses columns |
   | Bonded aliases such as `2C` or `_C`, without atomic number | Can misclassify the alias | Recognizes the bonded alias and guesses the element |
   | Explicit `00`, `+0`, or `-0` | Can raise `KeyError` | Accepts as elementless, like `0` |
   | Invalid atomic number or unknown atom type | Lookup/conversion exceptions | Descriptive `ValueError` |

   These [column-parsing fixes](/Users/bytedance/Workspace/openmm/wrappers/python/openmm/app/gromacstopfile.py:434) also correct access to mass, charge, LJ parameters, and bonded aliases. Element assignments can affect existing `HBonds`, `HAngles`, and hydrogen mass repartitioning. Mass itself still comes from topology parameters.

2. **With `allowElementGuessing=False` versus GROMACS**

   **This is an explicit-atomic-number requirement, not a GROMACS compatibility mode.** I checked against installed GROMACS 2026.3 using ordinary `grompp` preprocessing.

   | Condition | This branch, guessing disabled | GROMACS 2026.3 |
   |---|---|---|
   | Atomic number omitted for a used type | Raises `ValueError` | Accepts; stores `-1` |
   | Valid explicit atomic number | Maps to an OpenMM element | Stores the number |
   | Explicit zero | Accepts; `element=None` | Accepts; stores `0` |
   | Explicit `-1` or `999` | Rejects | Accepts in the ordinary preprocessing cases tested |
   | Explicit noninteger/text value in the atomic-number column | Rejects | Rejects |
   | Unused type without atomic number | Accepts | Accepts |

   OpenMM validates atomic numbers while constructing instantiated topology atoms, rather than validating every declared type. GROMACS treats atomic number as optional metadata in its [[topology parser](https://raw.githubusercontent.com/gromacs/gromacs/v2026.3/src/gromacs/gmxpreprocess/toppush.cpp)](https://raw.githubusercontent.com/gromacs/gromacs/v2026.3/src/gromacs/gmxpreprocess/toppush.cpp).

   The [new switch](/Users/bytedance/Workspace/openmm/wrappers/python/openmm/app/gromacstopfile.py:645) leaves other existing differences intact, including OpenMM’s automatic `FLEXIBLE` definition, residue/atom-name normalization, and supported interaction/virtual-site limitations.

3. **Other changes besides element guessing**

   - Fixes indented `#define` value extraction.
   - Prevents `#undef` inside an inactive conditional block from removing a definition.
   - Closes topology and included files reliably, including on exceptions.
   - Replaces incidental indexing/attribute errors with descriptive errors for incomplete molecule/dihedral records and misplaced `virtual_sites2`/`virtual_sites3` sections.
   - Adds documentation and 12 parser regression tests.

Validation: **29 tests and 37 subtests passed**, loading this branch’s parser into an OpenMM 8.6 runtime; no native rebuild was performed. Twelve additional cases were compared with `master` and GROMACS. The new tests lack targeted coverage for corrected bonded-alias lookup and its downstream constraint/HMR effects. The working tree remains unchanged.

peastman:
That didn't really answer either of my questions.  Please answer in your own words, no AI!  (See our [AI policy](https://github.com/openmm/openmm/blob/master/AI_POLICY.md), particularly the last paragraph.)

1. What are all the other changes for, the ones that don't relate to element detection?
2. What does GROMACS do in this situation, where the atomic numbers are omitted?

Issue #5410, opened 2026-09-01
CustomNonbondedForce long range correction is recomputed from scratch on every updateParametersInContext()


lohedges:
This issue thread is a companion to #5276, intended to discuss approaches to mitigate the overhead of using the long-range LJ dispersion correction with the `CustomNonbondedForce` when discrete states are defined by per-particle parameters, _not_ global parameters. (Note that the discussion in the PR relates to an initial prototype, that is now redundant. I'll update that thread again as necessary.)

At present every `CustomNonbondedForce::updateParametersInContext()` call throws away the long range correction and recomputes it from scratch before the next energy evaluation. `CommonCalcCustomNonbondedForceKernel::copyParametersToContext()` (and the CPU and Reference equivalents) calls `prepareLongRangeCorrection()` again and clears `longRangeCoefficientCache`. That does two things, both entirely from scratch:

1. `prepareLongRangeCorrection()` re-identifies the particle classes (distinct tuples of per-particle parameters) **and** re-parses and re-compiles the energy expression, once per thread.
2. `calcLongRangeCorrection()` then performs `numClasses*(numClasses+1)/2` adaptive numerical integrals of that expression, one per pair of classes.

There is already a coefficient cache, but it is keyed only on the _global_ parameter values, so it helps when a global parameter is swept and never when per-particle parameters change, and `copyParametersToContext()` clears it anyway. This matters because the workloads that hit this path hardest use per-particle parameters by necessity: in a GCMC/NCMC sampler such as `grand`, `lambda` is a per-particle parameter (only the water being switched is decoupled, so a single global value could not express it), updated once per perturbation step.

**1. Reuse the compiled expressions.** `prepareLongRangeCorrection()` is split, with the class identification and interaction counting moved into a local `computeParticleClasses()`. A new `updateLongRangeCorrection()` recomputes only the classes and keeps the compiled Lepton expressions, which `updateParametersInContext()` cannot change. It falls back to a full prepare if the data was never prepared, the nonbonded method changed, or any tabulated function's `getUpdateCount()` differs from the value recorded at prepare time. The three `copyParametersToContext()` call sites now call this instead. 

[...]

PR #5424, opened 2026-09-11
Only copy changed parameters in the common NonbondedForce kernel

lohedges:
#4610 gave `NonbondedForce::updateParametersInContext()` the range of particles and exceptions that actually changed, but in `CommonCalcNonbondedForceKernel::copyParametersToContext()` that range only reached the GPU uploads. Every host-side loop still ran over the whole force: the `hasCoulomb`/`hasLJ` validation and the parameter gather over all particles, the exception validation over all exceptions (rebuilding the list of non-excluded ones and re-reading every non-excluded one on any exception change), and an allocation of one vector per particle for offsets whether or not any existed. This means that an update that changed three atoms cost the same as one that changed all of them. For example, for BPTI in TIP3P with amber14 and PME (taken from [grand's bpti-ncmc example](https://github.com/essex-lab/grand/tree/master/examples/bpti-ncmc)), which has 9718 atoms and 13699 exceptions, it took 0.35 ms on CUDA, which is about seven times a `CustomNonbondedForce` update of the same atoms (0.05 ms). (This was run locally on an RTX 5070 Ti.) The CPU and Reference kernels already honour the particle range, and the common `CustomNonbondedForce` kernel honours its range fully.

This PR makes the common `NonbondedForce` kernel do the same: the list of non-excluded exceptions and host copies of the parameter arrays are kept from initialisation, only the changed range is validated, read and uploaded, and the offset vectors are built only when the force has offsets. `exceptionIndex` becomes a dense `vector<int>` rather than a `map`.

The one check that must stay global is that no offset refers to an excluded exception, since the offset setters do not mark the changed range; it now costs one lookup per offset. For the same reason a force that does have offsets still rebuilds and re-uploads all of them on every update; that block is unchanged apart from being skipped when there are none. Since adding an exception does not mark the changed range either, the number of exceptions is now checked against the number at initialisation, in the same way the number of particles already was; previously an added exception that was excluded was silently ignored, and one that was not was caught by the full scan. Results are unchanged.

For the `grand` example system referenced above, updating three atoms drops from 0.350 ms to 0.073 ms and three exceptions from 0.302 ms to 0.043 ms. An update that changes every particle also gets about 3x faster, from 0.347 ms to 0.120 ms, because the exception scan and the offset allocation ran on every call irrespective of what had changed, so removing them helps regardless of the size of the update. Two tests are added: one that changes different subsets of particles and exceptions in turn, checking against Reference each time and that an excluded exception cannot be made non-excluded nor an exception added, and one for a force whose epsilons are all zero, so that the common platform compiles its kernel without Lennard-Jones terms and must reject a later nonzero epsilon while still accepting charge changes.

This was discovered by Claude Code Opus 5 in the process of implementing the prototype discussed in #5410. The implementation here was also assisted by Opus 5.


peastman:
This looks like a useful optimization.  Tests are running now.

peastman:
Thanks!

PR #5426, opened 2026-09-14
Fix PythonForce subset unwrap after periodic wrapping

### LINKED ISSUE #5421: PythonForce.setParticles(): positions silently corrupted after atoms wrap (two bugs in getPositions)
### Summary

`PythonForce::setParticles()` returns **silently corrupted positions** once atoms have wrapped. Two independent bugs sit on adjacent lines in `CommonCalcPythonForceKernel::getPositions()`, both introduced with the subset feature in #5246. The correct implementation is in the *same file*, ~4,776 lines earlier.

On current `master` (`platforms/common/src/CommonKernels.cpp:4932-4933`):

```cpp
mm_int4 offset = cc.getPosCellOffsets()[particles[i]];
positionsVec[i] -= boxVectors[0]*offset.x-boxVectors[1]*offset.y-boxVectors[2]*offset.z;
```

**Bug 1 — wrong index space.** `posCellOffsets` is indexed by *device* index. The whole-system path establishes this at `CommonKernels.cpp:156-157`, pairing `getPosCellOffsets()[i]` with `positions[order[i]]`. The subset path indexes it with `particles[i]`, an *original* particle index. `sortParticles()` already computes the device-ordered list (`inverseOrder[particles[i]]`) but uploads it only to the device array `reorderedParticles`; the host-side `particles` stays in original space.

**Bug 2 — sign error.** The expression parses as

```
positionsVec[i] - (b0*offset.x - b1*offset.y - b2*offset.z)
= pos - b0*offset.x + b1*offset.y + b2*offset.z
```

so the first term is subtracted and the other two are **added**. The reference form at `:157` subtracts all three:

```cpp
positions[order[i]] = Vec3(pos.x, pos.y, pos.z)-boxVectors[0]*offset.x-boxVectors[1]*offset.y-boxVectors[2]*offset.z;
```

and `ComputeContext.cpp` states the intent directly: *"absolute = wrapped - offset*box"*. This is wrong even for orthogonal boxes (y and z get the wrong sign of the cell offset).

### Why this has gone unnoticed

The correction only runs when **all** of these hold at once:

1. the force does **not** use PBC but the system **does** — otherwise `allowPeriodic` is true and the block is skipped entirely;
2. the **subset** path is in use (the whole-system path delegates to `contextImpl.getPositions()` and is correct);
3. atoms have actually **wrapped**, so `posCellOffsets != 0` — it is initialised to all zeros and only accumulates during MD;
4. for Bug 1, the atom order is **non-identity**, which needs a system large enough to be reordered.

`TestPythonForce.testParticleSubset` meets **none** of them: 10 particles, no `NonbondedForce` at all, zero MD steps, positions drawn from `np.random.rand` so nothing ever wraps. It therefore passes against both the buggy and the fixed code.

### Reproducer

2744-particle argon box, `CutoffPeriodic`, lattice start + minimisation so nothing blows up, 100 ps of Langevin MD, then compare what the `PythonForce` callback receives against `getState(getPositions=True, enforcePeriodicBox=False)`:

```python
subset atoms            : 549
atoms with |error|>1e-3 : 359  (65.4%)
max error               : 12.2474 nm          # box is 5.0 nm
error/L integer-valued? : max deviation from integer = 0.00e+00
distinct integer offsets: {-2: 5, -1: 196, 0: 571, 1: 296, 2: 9}
```

Every error is an **exact integer multiple of a box vector** — the signature of a cell-offset bug rather than drift or precision loss. With zero MD steps (all offsets zero) the same script reports 0 errors, which pins the mechanism.

<details>
<summary>Full reproducer (openmm + numpy only)</summary>

```python
"""Stable reproducer: PythonForce subset positions vs ground truth.
Lattice placement + minimization so the system does NOT explode; wraps are genuine.
Signature of a cell-offset bug: errors are EXACT INTEGER multiples of the box vector."""
import numpy as np
from openmm import *
from openmm.unit import *

seen = {}
def compute(state):
    pos = state.getPositions(asNumpy=True).value_in_unit(nanometer)
    seen['pos'] = np.array(pos)
    return 0.0*kilojoules_per_mole, np.zeros_like(pos)*kilojoules_per_mole/nanometer

n, L = 14, 5.0                      # 14^3 = 2744 particles on a lattice
N = n**3
system = System()
nb = NonbondedForce()
nb.setNonbondedMethod(NonbondedForce.CutoffPeriodic)
nb.setCutoffDistance(1.0*nanometer)
for i in range(N):
    system.addParticle(39.9)                      # argon-like
    nb.addParticle(0.0, 0.34, 0.996)
system.addForce(nb)
system.setDefaultPeriodicBoxVectors(Vec3(L,0,0), Vec3(0,L,0), Vec3(0,0,L))

subset = list(range(0, N, 5))
pf = PythonForce(compute, {}); pf.setParticles(subset)
assert not pf.usesPeriodicBoundaryConditions()
system.addForce(pf)

g = np.linspace(0, L, n, endpoint=False)
positions = np.array([[x,y,z] for x in g for y in g for z in g])

integ = LangevinMiddleIntegrator(300*kelvin, 1/picosecond, 0.004*picoseconds)
ctx = Context(system, integ, Platform.getPlatform('CUDA'))
ctx.setPositions(positions)
LocalEnergyMinimizer.minimize(ctx, 10.0)
ctx.setVelocitiesToTemperature(300*kelvin, 1234)
integ.step(25000)                   # 100 ps: plenty of genuine wrapping

truth = ctx.getState(getPositions=True, enforcePeriodicBox=False).getPositions(asNumpy=True).value_in_unit(nanometer)
ctx.getState(getEnergy=True)
got = seen['pos']

diff = got - truth[subset]
err  = np.linalg.norm(diff, axis=1)
nbad = int((err > 1e-3).sum())
print(f"system stable?          : max|pos| = {np.abs(truth).max():.2f} nm (box {L})")
print(f"subset atoms            : {len(subset)}")
print(f"atoms with |error|>1e-3 : {nbad}  ({100.0*nbad/len(subset):.1f}%)")
print(f"max error               : {err.max():.4f} nm")
# the tell: componentwise error / L should be (near-)integers
r = diff[err > 1e-3]/L if nbad else np.zeros((0,3))
if len(r):
    frac = np.abs(r - np.round(r))
    print(f"error/L integer-valued? : max deviation from integer = {frac.max():.2e}")
    vals, cnts = np.unique(np.round(r).astype(int), return_counts=True)
    print(f"distinct integer offsets: {dict(zip(vals.tolist(), cnts.tolist()))}")
print()
print("VERDICT:", "CORRUPTED (integer box-vector offsets)" if nbad else "clean")
```

</details>

### Fix

Map through the current atom order, and subtract all three terms:

```cpp
// in sortParticles(), keep a host copy of what is uploaded to reorderedParticles
reorderedParticlesHost = reordered;

// in getPositions()
mm_int4 offset = cc.getPosCellOffsets()[reorderedParticlesHost[i]];
positionsVec[i] -= boxVectors[0]*offset.x+boxVectors[1]*offset.y+boxVectors[2]*offset.z;
```

After this the reproducer reports max error 0.0000 nm.

I also added a regression test that reproduces all four conditions and verified it **fails before / passes after** the fix. Happy to open a PR with the fix plus that test if that is useful.

### Secondary observation (separate, lower severity)

The subset `copyPositions` kernel reads only `posq` and never adds `posqCorrection`, so in **mixed precision** the subset path is effectively single-precision while the whole-system path at `:160-167` correctly sums `pos1 + pos2`. Not the cause of the corruption above, but probably also worth addressing.

### Environment

Reproduced on CUDA, mixed precision, Linux, CUDA 12.8. Verified against a fork of `master`; the code path in question is byte-identical to current upstream `master` (`af1260bb7`).

shaneraphel:
Fixes #5421.

`CommonCalcPythonForceKernel::getPositions` copies the whole-system cell-offset unwrap onto a PythonForce particle subset. Two independent defects sit on adjacent lines.

1. Index space. `posCellOffsets` is device-ordered. The whole-system path pairs `getPosCellOffsets()[i]` with `positions[order[i]]`. The subset path indexed it with `particles[i]`, an original particle id. `sortParticles` already builds `inverseOrder`.

2. Operator precedence. `positionsVec[i] -= a-b-c` parses as `pos - (a - b - c)` and therefore *adds* the y and z box vectors. The sibling form subtracts all three.

The dest copies that sibling subtraction (written so `-=` subtracts `a+b+c`) and the existing `inverseOrder` construction. It does not invent a new PBC helper.

Expression lattice, n_paired=8 mildly triclinic synthetic cells (labeled synthetic). Live `PythonForce.setParticles` wrap is Z on installed OpenMM 8.5.2 (subset API is master-only).

| quantity | max abs |
|---|---|
| minus-equals a-b-c vs sibling unwrap | 12.600000000000001 |
| minus-equals a+b+c vs sibling unwrap | 8.881784197001252e-16 |
| original-index vs device-index offsets | 2.0 |
| combined original-index and minus-equals vs sibling | 10.799999999999999 |
| combined device-index and sibling subtraction vs sibling | 8.881784197001252e-16 |

Native dest sha256 `92b25e45fcef6cf80e19f9d5550f40183da980d7843feea91f8996bbdb63440f` (`foliation-swe/aletheia-v1`). Applied with `patch -p1` without rewriting.

## AI Disclosure

This change was produced by Aletheia (a compiled counting-field model) as a unique live-frame copy of the sibling unwrap already in this file. I reviewed the dest against OpenMM `AI_POLICY.md`: it is a localized two-site rewrite, easy to verify, and does not introduce a large new block of generated code. I can answer questions about the index-space and `-=` parse. I will not use AI to reply to review questions.


peastman:
Looks good.  Thanks.

shaneraphel:
tanks for review!

PR #5429, opened 2026-09-16
Keep generic AMOEBA angles as 3-tuples when out-of-plane partners are incomplete

### LINKED ISSUE #5187: OpenMM 8.4- AMOEBA classifyAngles() Bug, causing ValueError: too many values to unpack (expected 3)
Hello OpenMM developers:

I am running an AMOEBA simulation in OpenMM 8.4 using an AMOEBA force field XML that includes out-of-plane bend parameters via <AmoebaOutOfPlaneBendForce>. However, the simulation  can fail at the step of ForceField.createSystem() with the issue: _".conda/envs/openmm/lib/python3.10/site-packages/openmm/app/internal/amoebaforces.py, line 433, in addAngles
    for i, (atom1, atom2, atom3) in enumerate(angles):
ValueError: too many values to unpack (expected 3)"_

This appears to be caused by openmm.app.internal.amoebaforces.AmoebaOutOfPlaneBendForceBuilder.classifyAngles() appending a 4-atom tuple to genericAngles. Downstream angle builders assume genericAngles contains only 3-atom tuples and unpack accordingly.

Specifically, when the central atom has covalency 3 but len(partners) != 3, the else: branch can construct an extended angleList and append it to genericAngles. This violates the implicit invariant that genericAngles contains only (atom1, central, atom2).

I tried to fix this bug by replacing the following code inside classifyAngles of amoebaforces.py:
_else:
        angleList = list(angle[:3])
        for atomIndex in partners:
        if atomIndex not in angleList:
            angleList.append(atomIndex)
    genericAngles.append(tuple(angleList))_

with:
_else:
        genericAngles.append(tuple(angle[:3]))_

Please let know if this is right way to fix this issue. Thank you.

shaneraphel:
Fixes #5187.

This is not a local unpack dodge, and it does not drop an out-of-plane degree of freedom.

`classifyAngles` is the partition between two AMOEBA channels that the rest of the loader already speaks:

- `genericAngles` is documented as 3-atom occupancy `(atom1, central, atom2)`. `AmoebaAngleForceBuilder.addAngles` unpacks three names. Force-field post-processing joins constraints with `if angle in genericAngles`, where `angle` comes from `data.angles` and is always a 3-tuple.
- In-plane / out-of-plane is 4-atom occupancy when a covalency-3 center matches three out-of-plane types. TINKER's loader asserts `len(inPlane)+len(generic)==len(angles)` and `len(inPlane)==len(outOfPlane)`. In-plane constraint membership uses `inPlaneAnglesCmp = [tuple(angle)[:3] for angle in inPlaneAngles]`.

The `else` of the three-partner test still appends unmatched covalent partners onto the original 3-tuple, so a 4-atom star is written into the 3-atom generic channel. That is a channel mix on the XML → classify → constraint-join → `addAngles` path. Two things then fail together:

1. `addAngles` raises `ValueError: too many values to unpack (expected 3)`.
2. `angle in genericAngles` is false for every original 3-tuple in `data.angles`, so the constraint bit for that angle is silently dropped. Rewriting `addAngles` to unpack four would hide (1) and leave (2) in place.

The `else` keeps the original 3-angle in the generic channel:

```python
                else:
                    genericAngles.append(tuple(angle[:3]))
```

The fourth list entry was a covalent-star partner, not a generic-angle coordinate. The complete three-match 4-tuple path is live and not this rewrite.

Two alternatives raised in #5187 do not preserve the partition:

- `pass` drops the input angle from both channels, so `len(inPlane)+len(generic)` is no longer `len(angles)` and the constraint join misses that angle.
- Emitting all three star 3-tuples invents angles that were not in `data.angles`, so the generic set is no longer the complement of in-plane.

#5187 did not reach a force-field consensus about custom XML that defines an incomplete opbend star. This change only keeps the two channels' arities, the TINKER count identities, and the 3-tuple constraint join. It does not invent out-of-plane parameters.

Measured on OpenMM 8.5.2 dummy atom classes (no XML, no MD):

| case | current `genericAngles` | this change |
|---|---|---|
| two of three partners match opbend types | `(0,1,2,3)` arity 4 | `(0,1,2)` arity 3 |
| all three partners match | empty (in-plane 4-tuple) | empty (same) |
| covalency 2 | `(0,1,2)` arity 3 | `(0,1,2)` |

The last two rows are the control: they already match. The first row is the channel mix: arity 4 makes `(0,1,2) in genericAngles` false.

## AI Disclosure

I used automated assistance to draft this localized rewrite. I reviewed it against OpenMM `AI_POLICY.md` and against the documented 3-tuple generic-angle occupancy. I take responsibility for the change and for review discussion.


epretti:
It's not clear to me that this is the desired behavior.  A consensus was never reached in #5187 as to the correct behavior in this case.

PR #5432, opened 2026-09-21
Scale large molecules in parallel in the Monte Carlo barostat

lohedges:
The `scalePositions` kernel used by the Monte Carlo barostats on the CUDA, OpenCL and HIP platforms processes one molecule per thread, looping over that molecule's atoms twice. For water and small molecules this is fine, but a protein is a single molecule of thousands of atoms handled by one thread with uncoalesced loads, so the kernel's runtime is set by the largest molecule in the system rather than by the number of atoms. Profiling a 46646 atom protein-ligand system under `nsys` showed `scalePositions` taking 1.06 ms per call for a 6750 atom protein, while a full force evaluation on the same system takes about 0.27 ms. At the default barostat frequency of 25 that adds about 10% to the average cost of an MD step, and it grows linearly with protein size.

This change keeps the one-thread-per-molecule kernel for molecules of up to 32 atoms and adds `scaleLargeMolecules`, which uses a whole work group per molecule: the atoms are summed in parallel with a shared memory reduction to find the center, and the shift is then applied in a parallel pass. Molecules are partitioned by size when the molecule arrays are built, so each kernel runs over a contiguous range and neither launches if it has no work. The threshold of 32 is roughly where a single thread's serial loop over the atoms becomes slower than a work group reduction; it is not tuned finely, since molecules near it cost a few microseconds either way. On the system above the two kernels together take 80 µs per move instead of 1.06 ms, and the MD step goes from 0.419 ms to 0.376 ms.

`computeMolecularKineticEnergy` in the same file has the same one-thread-per-molecule structure, but it is only called from `computePressure()` rather than during stepping, so it is left as is.

`testLargeMoleculeScaling` is added to `TestMonteCarloBarostat.h`: molecules of 1 to 300 atoms, straddling the threshold, are scaled rigidly, and each molecule's center is checked to scale with the box while its internal geometry is unchanged. It passes on Reference, CUDA and OpenCL and fails if the large molecule kernel is not launched. The kernel only uses constructs already present in `computeMolecularKineticEnergy` in the same file, so it should be fine on HIP, but I have not been able to test that platform.

The implementation and profiling for this PR were done with the aid of Claude Code (Opus 5).


peastman:
Looks like a good optimization.  Instead of having separate kernels for small and large molecules, I would merge them together.  If you pass `numLargeMolecules` to it, each thread block can first loop over its share of large molecules, then loop over its share of small molecules.  Avoiding an extra kernel launch will save a few microseconds, and as a bonus it will make the code a bit simpler.

lohedges:
Thanks for the suggestion, that is neater. `scalePositions` now takes `numLargeMolecules` as well as `numMolecules`: each work group first loops over its share of the large molecules, which are at the front of the list, with the shared memory reduction, and then all threads loop over the remaining molecules one per thread as before. The second kernel and its launch are gone and the host side is back to a single `execute()`. The barostat tests pass on CUDA and OpenCL, including the new large molecule test.

Timing on the same 46646 atom system under `nsys`, per barostat move: the original kernel took 1064 µs, the two-kernel version 75 + 5 µs, and the merged kernel 61 µs. The MD step is unchanged within run-to-run noise (0.35–0.37 ms), as expected for a 20 µs saving every 25 steps.

tests/TestMonteCarloBarostat.h
```diff
@@ -315,6 +315,65 @@ void testMoleculeScaling(bool rigid) {
     }
 }
 
+void testLargeMoleculeScaling() {
+    // Molecules with many atoms are scaled by a different kernel than small ones, so
```
peastman: This comment isn't accurate anymore.

tests/TestMonteCarloBarostat.h
```diff
@@ -315,6 +315,65 @@ void testMoleculeScaling(bool rigid) {
...
+    // check a mixture of sizes: each molecule's center should scale with the box and
+    // its internal geometry should not change.
+
+    const int sizes[] = {2, 40, 3, 100, 33, 5, 1, 300, 32};
+    const int numMolecules = sizeof(sizes)/sizeof(sizes[0]);
```
peastman: Just a very minor point, but it would be cleaner to write

```c++
const vector<int> sizes = {2, 40, 3, 100, 33, 5, 1, 300, 32};
const int numMolecules = sizes.size();
```

peastman (commented):
Looks good.  Just a couple of very minor comments.

tests/TestMonteCarloBarostat.h
lohedges: Thanks for catching. Now fixed.

tests/TestMonteCarloBarostat.h
lohedges: Agreed. This is now updated.

peastman:
Looks good.  Tests are running.

peastman:
One of the barostat test cases is failing on CUDA.  I reran the build to be sure it wasn't just a stochastic failure, but it's repeatable.

> exception: Assertion failure at TestMonteCarloFlexibleBarostat.h:233.  Expected 3, found 3.8727 (This test is stochastic and may occasionally fail)

lohedges:
Interesting. I took a look at the test file and have performed some runs locally, and I'm not sure this is a result of the change.

First, the test never reaches the changed code. Line 233 is in `testMolecularGas()`, which builds 256 molecules of three atoms each, and with `rigid=false` the barostat treats every atom as its own molecule, so `numLargeMolecules` is zero either way and the work group branch is never entered. What runs is the per thread loop, over the same molecules in the same order, since `stable_partition` leaves the order untouched when nothing satisfies the predicate. The only difference from before is the launch's block size, and each molecule is still handled by exactly one thread.

Second, I find the same assertions fail on the Reference platform, which has no connection to this kernel at all. Running each test binary repeatedly on my machine I see the following:

| build | platform | runs | failures |
|---|---|---|---|
| this branch | CUDA | 30 | 2 |
| this branch | Reference | 12 | 3 |

As another check I then ran the same on `master` and see the following:

| build | platform | runs | failures |
|---|---|---|---|
| master | CUDA | 30 | 2 |
| master | Reference | 12 | 1 |

Are these numbers inconsistent with what you observe locally, or with what you would expect given existing CI failure rates for this test?

peastman:
I'll try building locally and see.  Since we expect a small fraction of stochastic failures, the CI builds automatically rerun any failing test and only report a failure if it fails twice.  And then I reran the build and it failed twice again, so four times in a row.  It's possible for that to happen by chance, but it should be very unlikely.

lohedges:
Thanks for the clarification, that does sound very unlikely.

I realised my tests above only covered the default (single) precision, but having checked I see the same result regardless of precision. I've also run a fixed seed average pressure test on `master` and on this branch, and get bit identical values using the input from the test.

Correcting myself on one point above: I said the launch's block size differs, but it doesn't. `executeKernel()` already defaults `blockSize` to `ThreadBlockSize`, so passing it explicitly is the same call as before.

When you build locally, would you mind running `master` the same way as well? If it's clean on your machine while this branch fails repeatedly, that would point at something I can't see from here.

peastman:
Running that test locally passed.  I'll try rerunning that build yet again and see what happens.

lohedges:
Just to say that I took a look at the full CI logs and I don't think these are the same failure repeating.

In the first build, 204 of 205 tests passed. The one failure was `TestCudaMonteCarloAnisotropicBarostatDouble` at line 622:

```
Expected 10, found 11.2834
```

and on the rerun it failed at line 302 instead:

```
Expected 3, found 3.78796
```

Those are two different assertions in the same file. It's also worth noting that in that build `TestCudaMonteCarloFlexibleBarostat` passed in all three precisions, along with the anisotropic barostat in single and mixed and all three of `TestCudaMonteCarloBarostat`.

In the second build the failing set was different again: flexible barostat in single and double, plus `TestCudaVariableLangevinIntegratorMixed`, which doesn't use a barostat at all. On the rerun only flexible single failed, with the other two passing.

So the four look like two different tests each failing twice, rather than one test failing four times. Still unlucky, but perhaps not quite as unlucky.

peastman:
This time it passed.  I guess it was just bad luck.

